PR #3, opened 2016-07-27
fix test common setup for longhorn catalog and env VM enable

jimengliu:
@sangeethah @yasker 

fix test common setup for default longhorn catalog and env VM enable operations


sangeethah:
@jimengliu , we should wait for "VirtualMachine" stack to be active after vm setting is enabled for the environment to avoid the scenario of getting the tests to execute before the stack is fully activated.

sangeethah:
LGTM

PR #7, opened 2016-08-01
setup rancher and NFS GCE servers and register 3 Packet hosts

jimengliu:
@yasker @sangeethah

python scripts for setting up GCE VMs and Packet hosts: setupRancher.py and teardownRancher.py running in a container from docker image jimeng/longhorn-test:01

Jenkins jobs using the scripts: http://jenkins-poc.rancher.io:8080/job/test_longhorn_driver/
and also a quick cleanup job: http://jenkins-poc.rancher.io:8080/view/storage/job/test_longhorn_teardown/

### DIFF
--- a/environment-setup/Makefile
+++ b/environment-setup/Makefile
@@ -1,0 +1,7 @@
+target: flake8 package
+
+flake8:
+	flake8 *.py
+
+package:
+	docker build -t rancher/longhorn-env-setup .
--- a/environment-setup/Dockerfile
+++ b/environment-setup/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -y python openssh-client curl python-pip
+RUN pip install --upgrade pip google-api-python-client packet-python
+
+RUN curl -O https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-119.0.0-linux-x86_64.tar.gz
+RUN tar zxvf google-cloud-sdk-119.0.0-linux-x86_64.tar.gz
+ENV CLOUDSDK_CORE_DISABLE_PROMPTS="1"
+RUN ./google-cloud-sdk/install.sh
+RUN ./google-cloud-sdk/bin/gcloud components update
+
+ADD *.py ./
+
+ENTRYPOINT ["python"]
+
--- a/environment-setup/README.md
+++ b/environment-setup/README.md
@@ -1,0 +1,57 @@
+Steps to run the setupRancher.py and teardownRancher.py to setup and teardown machines for longhorn tests
+
+# first create a properties file to retrieve machine build results from container
+# it will contain CATTLE_TEST_URL and LONGHORN_BACKUP_SERVER_IP
+touch $PROPERTY_FILE_NAME
+
+# call setupRancher.py using environment variables to setup longhorn test machines
+docker run --rm -v $(pwd)/$PROPERTY_FILE_NAME:/$PROPERTY_FILE_NAME \
+    -e PROPERTY_FILE_NAME="$PROPERTY_FILE_NAME" \
+    -e GCE_SERVICE_KEY_JSON="$GCE_SERVICE_KEY_JSON" \
+    -e GCE_RANCHER_PROJECT_NAME="$GCE_RANCHER_PROJECT_NAME" \
+    -e GCE_RANCHER_PROJECT_ZONE="$GCE_RANCHER_PROJECT_ZONE" \
+    -e GCE_NFS_SERVER_NAME="$GCE_NFS_SERVER_NAME" \
+    -e GCE_RANCHER_SERVER_NAME="$GCE_RANCHER_SERVER_NAME" \
+    -e PACKET_HOST_NAMES="$PACKET_HOST_NAMES" \
+    -e PACKET_RANCHER_AUTH_TOKEN="$PACKET_RANCHER_AUTH_TOKEN" \
+    -e PACKET_RANCHER_PROJECT_ID="$PACKET_RANCHER_PROJECT_ID" \
+    -e PACKET_HOST_MACHINE_TYPE="$PACKET_HOST_MACHINE_TYPE" \
+    -e PACKET_HOST_OS_IMG="$PACKET_HOST_OS_IMG" \
+    -e PACKET_HOST_CLOUD_CONFIG="$PACKET_HOST_CLOUD_CONFIG" \
+    -e GCE_RANCHER_MACHINE_TYPE="$GCE_RANCHER_MACHINE_TYPE" \
+    -e GCE_RANCHER_OS_IMG="$GCE_RANCHER_OS_IMG" \
+    -e GCE_STARTUP_SCRIPT_RANCHER="$GCE_STARTUP_SCRIPT_RANCHER" \
+    -e GCE_STARTUP_SCRIPT_NFS="$GCE_STARTUP_SCRIPT_NFS" \
+    rancher/longhorn-env-setup setupRancher.py
+
+
+# call teardownRancher.py using environment variables to destroy longhorn test machines
+docker run --rm \
+    -e GCE_SERVICE_KEY_JSON="$GCE_SERVICE_KEY_JSON" \
+    -e GCE_RANCHER_PROJECT_NAME="$GCE_RANCHER_PROJECT_NAME" \
+    -e GCE_RANCHER_PROJECT_ZONE="$GCE_RANCHER_PROJECT_ZONE" \
+    -e GCE_NFS_SERVER_NAME="$GCE_NFS_SERVER_NAME" \
+    -e GCE_RANCHER_SERVER_NAME="$GCE_RANCHER_SERVER_NAME" \
+    -e PACKET_HOST_NAMES="$PACKET_HOST_NAMES" \
+    -e PACKET_RANCHER_AUTH_TOKEN="$PACKET_RANCHER_AUTH_TOKEN" \
+    -e PACKET_RANCHER_PROJECT_ID="$PACKET_RANCHER_PROJECT_ID" \
+    rancher/longhorn-env-setup teardownRancher.py 
+
+
+where environment variables are:
+PROPERTY_FILE_NAME:         result from running setupRancher.py, contains CATTLE_TEST_URL and LONGHORN_BACKUP_SERVER_IP
+GCE_SERVICE_KEY_JSON:       a GCE service account secrete key
+GCE_RANCHER_PROJECT_NAME:   rancher project name on GCE
+GCE_RANCHER_PROJECT_ZONE:   rancher project zone on GCE
+GCE_NFS_SERVER_NAME:        NFS server name on GCE to be created
+GCE_RANCHER_SERVER_NAME:    Rancher server name on GCE to be created
+GCE_RANCHER_MACHINE_TYPE:   Rancher server machine type, such as n1-standard-4
+GCE_RANCHER_OS_IMG:         Rancher server OS image name, such as ubuntu-1604-lts
+GCE_STARTUP_SCRIPT_RANCHER: Rancher server startup script run by GCE during boot
+GCE_STARTUP_SCRIPT_NFS:     NFS server startup script run by GCE during boot
+PACKET_HOST_NAMES:          a list of hosts to be created on Packet separated by commas, such as host1,host2,host3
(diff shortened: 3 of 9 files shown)


yasker:
Need a README on how to use.

common.py
```diff
@@ -0,0 +1,129 @@
...
+import json
+import time
+import errno
+
+
+logging.basicConfig(level=logging.INFO)
+log = logging.getLogger("common")
+
+packet_auth_token = os.environ["PACKET_AUTH_TOKEN"]
+packet_project_id = 'b4357aca-d706-4096-9e44-0a7c69d40ebe'
```
yasker: This should be a environment variable. And many things defined here should be able to override by a environment variable, e.g. machine type/os/zone/project etc.

yasker:
A general question: Can we run this as a Docker container? There are many environment setup things, which can be eliminated if it's packaged as a container.

jimengliu:
Jenkins job provides a way to change needed env variables at launch time, so I don't think it makes sense to run the script inside a container. Jenkins machine at least today has not had docker daemon running yet. I kind of limiting the env variable pollution in a way when test team does not care about configuring, I just hardcode these variable inside script, because we probably will never change them, such as project ID, zone, machine types etc

yasker:
It's a public repo...

yasker:
Also, have you run flake8 against the code yet? And many errors can be corrected by autopep8.

common.py
```diff
@@ -0,0 +1,129 @@
...
+            if 'error' in result:
+                raise Exception(result['error'])
+            return result
+
+        time.sleep(10)
+
+
+def install_python_client(python_client):
+    # install pip first, must be root
+    import getpass
```
yasker: I think this kind of import doesn't compliance with PEP8?

jimengliu:
didn't do flake8, but now fixing issues from flake8. thanks.

common.py
jimengliu: ran the flake8 and it does not complain import at function level. I already got rid of this case "import getpass" here anyway.

jimengliu:
addressed most of the comments.

yasker:
This batch of scripts should be in a separate directory of it own.

README.md
```diff
@@ -1,2 +1,4 @@
 # longhorn-tests
-Various test cases for longhorn
+A setup and a teardown script for setting up machines for various longhorn tests
```
yasker: This is still no instruction on how to use it. Need a README.md(and not in the root directory) which people can follow exactly then successful run the test.

common.py
```diff
@@ -0,0 +1,141 @@
...
+            project=gce_rancher_project_name,
+            zone=gce_rancher_project_zone,
+            operation=operation).execute()
+
+        if result['status'] == 'DONE':
+            if 'error' in result:
+                raise Exception(result['error'])
+            return result
+
+        time.sleep(10)
```
yasker: What's this for?

common.py
```diff
@@ -0,0 +1,141 @@
...
+    if retCode != 0:
+        raise Exception("failed to pip install --upgrade pip")
+    retCode = subprocess.call(
+        sudo +
+        "pip install --upgrade " +
+        python_client,
+        shell=True)
+    if retCode != 0:
+        raise Exception("failed to pip install python client")
+    time.sleep(10)
```
yasker: here too

common.py
```diff
@@ -0,0 +1,141 @@
...
+    install_gcloud(gcloud_sdk_url + gcloud_sdk_file_name, gcloud_sdk_file_name)
+
+    # install gcloud python client
+    log.info("install google-api-python-client ...")
+    install_python_client("google-api-python-client")
+
+    # before making any python client lib call, enable oauth
+    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = gce_key_file
+
+    from oauth2client.client import GoogleCredentials
```
yasker: Put lines below to a separate file handling gce? Don't like the import here.

README.md
jimengliu: I will move them to its own folder and add its own readme/usage. Talked to Bill, we need to run setup/teardown in a container, there is a docker daemon in that VM(I didn't use sudo). That could take me a day or two I guess.

common.py
jimengliu: sleep a while and query the status of GCE VM creation status.

common.py
jimengliu: this one, I think we maybe able to remove it now.

machine-setup/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:16.04
+
+ADD *.py ./
+RUN apt-get update
+RUN apt-get install -y openssh-client
```
yasker: Put in one line like `openssh-client python curl` etc, then Docker won't generate image for each time. In fact you can do `apt-get update && apt-get install -y xxxx`

machine-setup/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN apt-get install -y python
+RUN apt-get install -y curl
+RUN apt-get install -y python-pip
+RUN pip install --upgrade pip
```
yasker: `pip install` may able to follow `apt-get` example.

machine-setup/README.md
```diff
@@ -0,0 +1,11 @@
...
+# first create a properties file to retrieve machine build results from container
+# it will contain CATTLE_TEST_URL and LONGHORN_BACKUP_SERVER_IP
+touch $PROPERTY_FILE_NAME
+
+# call setupRancher.py using environment variables to setup longhorn test machines
+docker run --rm -v $(pwd)/$PROPERTY_FILE_NAME:/$PROPERTY_FILE_NAME -e PROPERTY_FILE_NAME="$PROPERTY_FILE_NAME" -e GCE_SERVICE_KEY_JSON="$GCE_SERVICE_KEY_JSON" -e GCE_RANCHER_PROJECT_NAME="$GCE_RANCHER_PROJECT_NAME" -e GCE_RANCHER_PROJECT_ZONE="$GCE_RANCHER_PROJECT_ZONE" -e GCE_NFS_SERVER_NAME="$GCE_NFS_SERVER_NAME" -e GCE_RANCHER_SERVER_NAME="$GCE_RANCHER_SERVER_NAME" -e PACKET_HOST_NAMES="$PACKET_HOST_NAMES" -e PACKET_RANCHER_AUTH_TOKEN="$PACKET_RANCHER_AUTH_TOKEN" -e PACKET_RANCHER_PROJECT_ID="$PACKET_RANCHER_PROJECT_ID" -e PACKET_HOST_MACHINE_TYPE="$PACKET_HOST_MACHINE_TYPE" -e PACKET_HOST_OS_IMG="$PACKET_HOST_OS_IMG" -e PACKET_HOST_CLOUD_CONFIG="$PACKET_HOST_CLOUD_CONFIG" -e GCE_RANCHER_MACHINE_TYPE="$GCE_RANCHER_MACHINE_TYPE" -e GCE_RANCHER_OS_IMG="$GCE_RANCHER_OS_IMG" -e GCE_STARTUP_SCRIPT_RANCHER="$GCE_STARTUP_SCRIPT_RANCHER" -e GCE_STARTUP_SCRIPT_NFS="$GCE_STARTUP_SCRIPT_NFS" jimeng/longhorn-test:001 setupRancher.py
```
yasker: 1. Format. Put `-e` in each line otherwise it's hard to read.
2. At least have a example for what each parameter means. I don't think anyone can understand only by the name here...

machine-setup/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+# call teardownRancher.py using environment variables to destroy longhorn test machines
+docker run --rm -e GCE_SERVICE_KEY_JSON="$GCE_SERVICE_KEY_JSON" -e GCE_RANCHER_PROJECT_NAME="$GCE_RANCHER_PROJECT_NAME" -e GCE_RANCHER_PROJECT_ZONE="$GCE_RANCHER_PROJECT_ZONE" -e GCE_NFS_SERVER_NAME="$GCE_NFS_SERVER_NAME" -e GCE_RANCHER_SERVER_NAME="$GCE_RANCHER_SERVER_NAME" -e PACKET_HOST_NAMES="$PACKET_HOST_NAMES" -e PACKET_RANCHER_AUTH_TOKEN="$PACKET_RANCHER_AUTH_TOKEN" -e PACKET_RANCHER_PROJECT_ID="$PACKET_RANCHER_PROJECT_ID" jimeng/longhorn-test:001 teardownRancher.py
```
yasker: Use `jimeng/longhorn-test` is better. Just tag the latest version as `latest` tag then it will be used.

yasker: I think it's better to have a Makefile. Just let it run `flake8` first then `docker build` to build the image. Then we can generate the image by just run `make`. `docker push` need to be done manually though.

machine-setup/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:16.04
+
+ADD *.py ./
```
yasker: Also put this line to the next to the last, otherwise everytime the script update will re-generated the whole image. By putting it in the end the change of script will only rebuild the script related part.

yasker:
`machine-setup` name is kind of weird. `environment-setup` sounds better.

machine-setup/Makefile
```diff
@@ -0,0 +1,7 @@
+target: flake8 docker-build
+
+flake8:
+	flake8 *.py
+
+docker-build:
```
yasker: Normally we call it `package`

machine-setup/Makefile
```diff
@@ -0,0 +1,7 @@
...
+	docker build -t jimeng/longhorn-test .
```
yasker: Don't point to your personal repo here. Just point to `rancher/longhorn-env-setup`. Upload can be done manually later.

Also I think `longhorn-env-setup` sounds better. Because it's not the only component in `longhorn-test`.

machine-setup/README.md
```diff
@@ -0,0 +1,57 @@
...
+    -e PACKET_RANCHER_AUTH_TOKEN="$PACKET_RANCHER_AUTH_TOKEN" \
+    -e PACKET_RANCHER_PROJECT_ID="$PACKET_RANCHER_PROJECT_ID" \
+    -e PACKET_HOST_MACHINE_TYPE="$PACKET_HOST_MACHINE_TYPE" \
+    -e PACKET_HOST_OS_IMG="$PACKET_HOST_OS_IMG" \
+    -e PACKET_HOST_CLOUD_CONFIG="$PACKET_HOST_CLOUD_CONFIG" \
+    -e GCE_RANCHER_MACHINE_TYPE="$GCE_RANCHER_MACHINE_TYPE" \
+    -e GCE_RANCHER_OS_IMG="$GCE_RANCHER_OS_IMG" \
+    -e GCE_STARTUP_SCRIPT_RANCHER="$GCE_STARTUP_SCRIPT_RANCHER" \
+    -e GCE_STARTUP_SCRIPT_NFS="$GCE_STARTUP_SCRIPT_NFS" \
+    jimeng/longhorn-test:001 setupRancher.py
```
yasker: Don't point to your own repo. Or with tag.

machine-setup/README.md
```diff
@@ -0,0 +1,57 @@
...
+docker run --rm \
+    -e GCE_SERVICE_KEY_JSON="$GCE_SERVICE_KEY_JSON" \
+    -e GCE_RANCHER_PROJECT_NAME="$GCE_RANCHER_PROJECT_NAME" \
+    -e GCE_RANCHER_PROJECT_ZONE="$GCE_RANCHER_PROJECT_ZONE" \
+    -e GCE_NFS_SERVER_NAME="$GCE_NFS_SERVER_NAME" \
+    -e GCE_RANCHER_SERVER_NAME="$GCE_RANCHER_SERVER_NAME" \
+    -e PACKET_HOST_NAMES="$PACKET_HOST_NAMES" \
+    -e PACKET_RANCHER_AUTH_TOKEN="$PACKET_RANCHER_AUTH_TOKEN" \
+    -e PACKET_RANCHER_PROJECT_ID="$PACKET_RANCHER_PROJECT_ID" \
+    jimeng/longhorn-test:001 teardownRancher.py
```
yasker: Same here.

yasker:
Otherwise LGTM. I can merge it after above comments are addressed. Thanks.

environment-setup/Makefile
```diff
@@ -0,0 +1,7 @@
+target: flake8 docker-build
```
yasker: package...

PR #8, opened 2016-08-01
Added code to wait for Storage pool and Virtual Machine stack to be available

sangeethah:
@yasker @jimengliu can you review the changes made to wait for Storage pool and Virtual Machine stack for maximum of 30 secs to prevent execution of test cases too quickly resulting in failure.


jimengliu:
logic looks good. If you can get rid of last list_funciton by creating a local variable before loop, then use it to do assertion, then that maybe better.

sangeethah:
@jimengliu, I have removed the redundant code that checks for storage pool and virtual Machine availability after the while loop.

Issue #13, opened 2017-01-20
verification test need to run inside container


yasker:
Tons of python library dependency need to be address in order for test to run.

yasker:
Fixed by #14

PR #14, opened 2017-03-01
fix test_container_replica_down

imikushin:
a fix for longhorn validation tests


imikushin:
I believe, I address #13 here as well

PR #15, opened 2017-11-30
Longhorn-test: commit the test code for volume driver

jimmy-peng:


manager/integration/tests/test_volume_driver.py
```diff
@@ -0,0 +1,82 @@
...
+    volumes = client.list_volume()
+    assert len(volumes) == 1
+    assert volumes[0]["name"] == volume["name"]
+    assert volumes[0]["size"] == str(volume_size)
+    assert volumes[0]["numberOfReplicas"] == int(volume["numberOfReplicas"])
+    assert volumes[0]["state"] == "healthy"
+    k8sclient.CoreV1Api().delete_namespaced_pod(
+        name=pod_name,
+        namespace='default', body=k8sclient.V1DeleteOptions())
+    wait_for_volume_delete(client, volume['name'])
```
yasker: Why volume is deleted here? It needs to be reused by others.

yasker:
Also need the test case for volume to be persistent across different pods. E.g., pod A write something to it, and then was deleted. Pod B can use the same volume and read from it.

jimmy-peng:
ok, this commitment has solved this problem.

PR #33, opened 2018-06-01
Add CSI test cases

niusmallnan:
Relate to: https://github.com/rancher/longhorn-manager/pull/68

How to run:
1. Deploy longhorn and [longhorn-csi](https://github.com/niusmallnan/longhorn-csi/tree/master/deploy/csi)
2. Run CSI cases: `py.test -v test_csi.py`

Results:
```
test_csi.py::test_volume_mount PASSED
test_csi.py::test_volume_io PASSED
```

### DIFF
--- a/manager/integration/tests/test_csi.py
+++ b/manager/integration/tests/test_csi.py
@@ -1,0 +1,216 @@
+import time
+
+from common import clients, csi_pvc_name  # NOQA
+from common import wait_for_volume_delete, wait_for_volume_state
+
+from kubernetes import client as k8sclient, config as k8sconfig
+from kubernetes.stream import stream
+from kubernetes.client import Configuration
+
+Gi = (1 * 1024 * 1024 * 1024)
+
+WAIT_POD_RETRY_COUNTS = 200
+WAIT_POD_RETRY_INTERVAL = 1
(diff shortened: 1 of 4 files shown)


manager/integration/tests/test_csi.py
```diff
@@ -0,0 +1,216 @@
...
+from kubernetes import client as k8sclient, config as k8sconfig
+from kubernetes.stream import stream
+from kubernetes.client import Configuration
+
+Gi = (1 * 1024 * 1024 * 1024)
+
+WAIT_POD_RETRY_COUNTS = 200
+WAIT_POD_RETRY_INTERVAL = 1
+
+DEFAULT_SC_NAME = "longhorn-csi"
```
yasker: It's ok to create a storage class here, but it's the default longhorn CSI storage class name here (as in your https://github.com/niusmallnan/longhorn-csi/blob/master/examples/storage-class.yaml ). You may get an object already exists error. Use a name like `longhorn-csi-test` to differentiate.

niusmallnan: @yasker Done

yasker:
Merged, thanks.

PR #41, opened 2018-06-12
manager: Pump in data for the integration test

zhoubofsy:
Issue: https://github.com/rancher/longhorn-tests/issues/21

Positive case :

```
ubuntu@ip-172-31-7-79:~/langs/golang/src/github.com/rancher/longhorn-tests/manager/integration$ kubectl logs -f longhorn-test
+ flake8 .
+ py.test -v -s -k test_basic .
platform linux2 -- Python 2.7.14, pytest-2.9.2, py-1.5.3, pluggy-0.3.1 -- /usr/local/bin/python
cachedir: .cache
rootdir: /integration/tests, inifile:
collecting ... collected 17 items

test_basic.py::test_hosts_and_settings PASSED
test_basic.py::test_volume_basic PASSED
test_basic.py::test_volume_iscsi_basic PASSED
test_basic.py::test_snapshot PASSED
test_basic.py::test_backup PASSED
test_basic.py::test_volume_multinode PASSED

==================== 11 tests deselected by '-ktest_basic' =====================
================== 6 passed, 11 deselected in 382.41 seconds ===================
```

Negative case:

```
ubuntu@ip-172-31-7-79:~/langs/golang/src/github.com/rancher/longhorn-tests/manager/integration$ kubectl logs -f longhorn-test
+ flake8 .
+ py.test -v -s -k test_basic .
platform linux2 -- Python 2.7.14, pytest-2.9.2, py-1.5.3, pluggy-0.3.1 -- /usr/local/bin/python
cachedir: .cache
rootdir: /integration/tests, inifile:
collecting ... collected 17 items

test_basic.py::test_hosts_and_settings PASSED
test_basic.py::test_volume_basic FAILED
test_basic.py::test_volume_iscsi_basic PASSED
test_basic.py::test_snapshot PASSED
test_basic.py::test_backup PASSED
test_basic.py::test_volume_multinode PASSED

=================================== FAILURES ===================================
______________________________ test_volume_basic _______________________________

clients = {'gke-bob-cluster-2-default-pool-bfd8924a-5480': <cattle.Client object at 0x7fef05f92a50>, 'gke-bob-cluster-2-default-...nt object at 0x7fef0b6da390>, 'gke-bob-cluster-2-default-pool-bfd8924a-vvph': <cattle.Client object at 0x7fef062bce90>}
volume_name = 'longhorn-testvol-pbqrqb'

    def test_volume_basic(clients, volume_name):  # NOQA
        # get a random client
        for host_id, client in clients.iteritems():
            break

        with pytest.raises(Exception):
            volume = client.create_volume(name="wrong_volume-name-1.0", size=SIZE,
                                          numberOfReplicas=2)
            volume = client.create_volume(name="wrong_volume-name", size=SIZE,
                                          numberOfReplicas=2)
            volume = client.create_volume(name="wrong_volume-name", size=SIZE,
                                          numberOfReplicas=2,
                                          frontend="invalid_frontend")

        volume = client.create_volume(name=volume_name, size=SIZE,
                                      numberOfReplicas=3)
        assert volume["name"] == volume_name
        assert volume["size"] == SIZE
        assert volume["numberOfReplicas"] == 3
        assert volume["frontend"] == "blockdev"

        volume = wait_for_volume_state(client, volume_name, "detached")
        assert len(volume["replicas"]) == 3

        assert volume["state"] == "detached"
        assert volume["created"] != ""

        assert len(volumes) == 1
        assert volumes[0]["name"] == volume["name"]
        assert volumes[0]["size"] == volume["size"]
        assert volumes[0]["numberOfReplicas"] == volume["numberOfReplicas"]
        assert volumes[0]["state"] == volume["state"]
        assert volumes[0]["created"] == volume["created"]

        volumeByName = client.by_id_volume(volume_name)
        assert volumeByName["name"] == volume["name"]
        assert volumeByName["size"] == volume["size"]
        assert volumeByName["numberOfReplicas"] == volume["numberOfReplicas"]
        assert volumeByName["state"] == volume["state"]
        assert volumeByName["created"] == volume["created"]

        lht_hostId = get_self_host_id()
        volume.attach(hostId=lht_hostId)
        volume = wait_for_volume_state(client, volume_name, "healthy")

        # soft anti-affinity should work, assume we have 3 nodes or more
        hosts = {}
        for replica in volume["replicas"]:
            id = replica["hostId"]
            assert id != ""
            assert id not in hosts
            hosts[id] = True
        assert len(hosts) == 3

        assert len(volumes) == 1
        assert volumes[0]["name"] == volume["name"]
        assert volumes[0]["size"] == volume["size"]
        assert volumes[0]["numberOfReplicas"] == volume["numberOfReplicas"]
        assert volumes[0]["state"] == volume["state"]
        assert volumes[0]["created"] == volume["created"]
        assert volumes[0]["endpoint"] == DEV_PATH + volume_name

        assert volume["endpoint"] == DEV_PATH + volume_name

        # RW test
>       assert volume_valid(volume["endpoint"])

test_basic.py:148:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

dev = '/dev/longhorn/longhorn-testvol-pbqrqb'

    def volume_valid(dev):
>       return stat.S_ISBLK(os.stat(dev).st_mode)
E       OSError: [Errno 2] No such file or directory: '/dev/longhorn/longhorn-testvol-pbqrqb'

common.py:232: OSError
==================== 11 tests deselected by '-ktest_basic' =====================
============= 1 failed, 5 passed, 11 deselected in 374.53 seconds ==============
```

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -43,8 +43,26 @@ spec:
 #           "--enable-recurring-job-test",
 #           ]
     imagePullPolicy: Always
+    securityContext:
+      privileged: true
     env:
     - name: LONGHORN_BACKUPSTORES
       value: "s3://backupbucket@us-east-1/backupstore$minio-secret, nfs://longhorn-test-nfs-svc.default:/opt/backupstore"
+    - name: NODE_NAME
+      valueFrom:
+        fieldRef:
+          fieldPath: spec.nodeName
+    volumeMounts:
+    - name: dev
+      mountPath: /dev
+    - name: proc
+      mountPath: /host/proc
+  volumes:
+  - name: dev
+    hostPath:
+      path: /dev/
+  - name: proc
+    hostPath:
+      path: /proc/
   restartPolicy: Never
   serviceAccountName: longhorn-test-service-account
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1,5 +1,6 @@
 import time
 import os
+import stat
 import random
 import string
 
@@ -12,7 +13,8 @@ from kubernetes import client as k8sclient, config as k8sconfig
 SIZE = str(16 * 1024 * 1024)
 VOLUME_NAME = "longhorn-testvol"
 DEV_PATH = "/dev/longhorn/"
-
+VOLUME_RWTEST_SIZE = 512
+VOLUME_INVALID_POS = -1
 PORT = ":9500"
 
 RETRY_COUNTS = 300
@@ -22,6 +24,8 @@ LONGHORN_NAMESPACE = "longhorn-system"
 
 COMPATIBILTY_TEST_IMAGE_PREFIX = "rancher/longhorn-test:version-test"
 UPGRADE_TEST_IMAGE_PREFIX = "rancher/longhorn-test:upgrade-test"
+
+ISCSI_DEV_PATH = "/dev/disk/by-path"
 
 
 @pytest.fixture
@@ -72,6 +76,11 @@ def get_mgr_ips():
     return mgr_ips
 
 
+def get_self_host_id():
+    envs = os.environ
+    return envs["NODE_NAME"]
+
+
 def get_backupstore_url():
     backupstore = os.environ['LONGHORN_BACKUPSTORES']
     backupstore = backupstore.replace(" ", "")
@@ -88,6 +97,18 @@ def get_clients(hosts):
         assert host["address"] is not None
         clients[host["uuid"]] = get_client(host["address"] + PORT)
     return clients
+
+
+def wait_for_device_login(dest_path, name):
+    dev = ""
+    for i in range(RETRY_COUNTS):
+        files = os.listdir(dest_path)
+        if name in files:
+            dev = name
+            break
+        time.sleep(RETRY_ITERVAL)
+    assert dev == name
+    return dev
 
 
 def wait_for_volume_state(client, name, state):
@@ -210,6 +231,117 @@ def get_compatibility_test_image(cli_v, cli_minv,
                                      cli_v, cli_minv,
                                      ctl_v, ctl_minv,
                                      data_v, data_minv)
+
+
+def generate_random_data(count):
+    return ''.join(random.choice(string.ascii_lowercase + string.digits)
+                   for _ in range(count))
+
+
+def volume_read(dev, start, count):
+    r_data = ""
+    fdev = open(dev, 'rb')
+    if fdev is not None:
+        fdev.seek(start)
+        r_data = fdev.read(count)
+        fdev.close()
+    return r_data
+
+
+def volume_write(dev, start, data):
+    w_length = 0
+    fdev = open(dev, 'rb+')
+    if fdev is not None:
+        fdev.seek(start)
+        fdev.write(data)
+        fdev.close()
+        w_length = len(data)
+    return w_length
+
+
+def volume_valid(dev):
+    return stat.S_ISBLK(os.stat(dev).st_mode)
+
+
+def parse_iscsi_endpoint(iscsi):
+    iscsi_endpoint = iscsi[8:]
+    return iscsi_endpoint.split('/')
+
+
+def get_iscsi_ip(iscsi):
+    iscsi_endpoint = parse_iscsi_endpoint(iscsi)
+    ip = iscsi_endpoint[0].split(':')
+    return ip[0]
+
+
+def get_iscsi_port(iscsi):
+    iscsi_endpoint = parse_iscsi_endpoint(iscsi)
+    ip = iscsi_endpoint[0].split(':')
+    return ip[1]
+
+
+def get_iscsi_target(iscsi):
+    iscsi_endpoint = parse_iscsi_endpoint(iscsi)
+    return iscsi_endpoint[1]
+
+
+def get_iscsi_lun(iscsi):
+    iscsi_endpoint = parse_iscsi_endpoint(iscsi)
+    return iscsi_endpoint[2]
+
+
+def exec_nsenter(cmd):
+    exec_cmd = "nsenter --mount=/host/proc/1/ns/mnt \
+               --net=/host/proc/1/ns/net bash -c \"" + cmd + "\""
+    fp = os.popen(exec_cmd)
+    ret = fp.read()
+    fp.close()
+    return ret
+
+
+def iscsi_login(iscsi_ep):
+    ip = get_iscsi_ip(iscsi_ep)
+    port = get_iscsi_port(iscsi_ep)
+    target = get_iscsi_target(iscsi_ep)
+    lun = get_iscsi_lun(iscsi_ep)
+    # discovery
+    cmd_discovery = "iscsiadm -m discovery -t st -p " + ip
+    exec_nsenter(cmd_discovery)
+    # login
+    cmd_login = "iscsiadm -m node -T " + target + " -p " + ip + " --login"
+    exec_nsenter(cmd_login)
+    blk_name = "ip-%s:%s-iscsi-%s-lun-%s" % (ip, port, target, lun)
+    wait_for_device_login(ISCSI_DEV_PATH, blk_name)
+    dev = os.path.realpath(ISCSI_DEV_PATH + "/" + blk_name)
+    return dev
+
+
+def iscsi_logout(iscsi_ep):
+    ip = get_iscsi_ip(iscsi_ep)
+    target = get_iscsi_target(iscsi_ep)
+    cmd_logout = "iscsiadm -m node -T " + target + " -p " + ip + " --logout"
+    exec_nsenter(cmd_logout)
+    cmd_rm_discovery = "iscsiadm -m discovery -p " + ip + " -o delete"
+    exec_nsenter(cmd_rm_discovery)
+
+
+def generate_random_pos(size, used={}):
+    for i in range(RETRY_COUNTS):
+        pos = 0
+        if int(SIZE) != size:
+            pos = random.randrange(0, int(SIZE)-size, 1)
+        collided = False
+        # it's [start, end) vs [pos, pos + size)
+        for start, end in used.items():
+            if pos + size <= start or pos >= end:
+                continue
+            collided = True
+            break
+        if not collided:
+            break
+    assert not collided
+    used[pos] = pos + size
+    return pos
 
 
 def get_upgrade_test_image(cli_v, cli_minv,
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -3,10 +3,16 @@ import time
 import common
 
 from common import clients, volume_name  # NOQA
-from common import SIZE, DEV_PATH
+from common import SIZE, DEV_PATH, VOLUME_RWTEST_SIZE
+from common import get_self_host_id
+from common import volume_read, volume_write
+from common import volume_valid
+from common import iscsi_login, iscsi_logout
 from common import wait_for_volume_state, wait_for_volume_delete
 from common import wait_for_snapshot_purge
 from common import generate_volume_name
+from common import generate_random_data
+from common import generate_random_pos
 
 
 def test_hosts_and_settings(clients):  # NOQA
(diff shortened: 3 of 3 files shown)


yasker:
I don't understand how you can call it `positive case` when there are failed cases and call traces.

manager/integration/tests/test_basic.py
```diff
@@ -59,6 +63,44 @@ def test_hosts_and_settings(clients):  # NOQA
...
+    fdev.seek(start)
+    fdev.write(data)
+    fdev.flush()
+
+
+def volume_valid(dev):
+    return stat.S_ISBLK(os.stat(dev).st_mode)
+
+
+def volume_rw_test(dev, count):
```
yasker: Just make the function that verifying the consistency of device at `loc` for `size`. Don't make caller compare check_count.

And put related functions in the `common.py`

manager/integration/tests/test_basic.py
```diff
@@ -59,6 +63,44 @@ def test_hosts_and_settings(clients):  # NOQA
...
+    check_count = 0
+    if not volume_valid(dev):
```
yasker: `assert volume_valid(dev)`

It's either a test bug or a manager bug if this happen.

niusmallnan:
@zhoubofsy Please update to the latest Longhorn image and test it again.

manager/integration/tests/test_basic.py
```diff
@@ -59,6 +63,44 @@ def test_hosts_and_settings(clients):  # NOQA
...
+        return check_count
+
+    with open(dev, "rb+") as fdev:
+        # save data
+        origin_data = volume_read(fdev, VOLUME_RWTEST_START_POSITION, count)
```
yasker: Don't use const for location. Pass in as parameter.

manager/integration/tests/test_basic.py
```diff
@@ -59,6 +63,44 @@ def test_hosts_and_settings(clients):  # NOQA
...
+
+        # rw test
+        w_data = b"\xfe"*(count)
```
yasker: Don't write the same character. Use random string, so if there is a difference between this write and previous write, it will be detected.

manager/integration/tests/test_basic.py
```diff
@@ -59,6 +63,44 @@ def test_hosts_and_settings(clients):  # NOQA
...
+        volume_write(fdev, VOLUME_RWTEST_START_POSITION, w_data)
+        r_data = volume_read(fdev, VOLUME_RWTEST_START_POSITION, count)
+
+        if r_data is not None and r_data == w_data:
```
yasker: `assert r_data == w_data`

manager/integration/tests/test_basic.py
yasker: And don't need backup origin data. It's a test volume.

manager/integration/tests/test_basic.py
```diff
@@ -101,7 +143,8 @@ def test_volume_basic(clients, volume_name):  # NOQA
     assert volumeByName["state"] == volume["state"]
     assert volumeByName["created"] == volume["created"]
 
-    volume.attach(hostId=host_id)
+    lht_hostId = get_node_name_by_pod("longhorn-test=test-job","longhorn-test")
```
yasker: It's easier to pass in an environment variable instead of using k8s API call.

manager/integration/tests/test_basic.py
zhoubofsy: I think , it's a manager bug.

manager/integration/deploy/test.yaml
```diff
@@ -43,11 +43,24 @@ spec:
...
     imagePullPolicy: Always
+    securityContext:
+      privileged: true
     env:
     - name: LONGHORN_BACKUPSTORE
       #value: "nfs://longhorn-test-nfs-svc.default:/opt/backupstore"
       value: "s3://backupbucket@us-east-1/backupstore"
     - name: LONGHORN_BACKUPSTORE_CREDENTIAL_SECRET
       value: "minio-secret"
+    - name: SELF_NODE_NAME
```
yasker: Just call it NODE_NAME, follow the example of Longhorn manager.

manager/integration/tests/common.py
```diff
@@ -12,6 +13,8 @@
 SIZE = str(16 * 1024 * 1024)
 VOLUME_NAME = "longhorn-testvol"
 DEV_PATH = "/dev/longhorn/"
+VOLUME_RWTEST_COUNT = 512
+VOLUME_RWTEST_START_POSITION = 0
```
yasker: Make location randomly as well. It's optimal if we can check the whole disk but that may take too long for a 20G disk. So I think it's OK for us to compromise for now using random data/location.

manager/integration/tests/test_basic.py
```diff
@@ -125,6 +144,10 @@ def test_volume_basic(clients, volume_name):  # NOQA
     volume = client.by_id_volume(volume_name)
     assert volume["endpoint"] == DEV_PATH + volume_name
 
+    # RW test
+    assert volume_valid(volume["endpoint"])
+    volume_rw_test(volume["endpoint"])
```
yasker: Put volume_valid() in volume_rw_test(). volume_rw_test() doesn't make sense if volume is not valid.

manager/integration/tests/common.py
```diff
@@ -12,6 +13,8 @@
 SIZE = str(16 * 1024 * 1024)
 VOLUME_NAME = "longhorn-testvol"
 DEV_PATH = "/dev/longhorn/"
+VOLUME_RWTEST_COUNT = 512
+VOLUME_RWTEST_START_POSITION = random.randrange(0, int(SIZE)-512, 1)
```
yasker: `-VOLUME_RWTEST_COUNT` instead of `-512`.

zhoubofsy:
Test test_basic:


```
+ flake8 .
+ py.test -v -s -k test_basic .
platform linux2 -- Python 2.7.14, pytest-2.9.2, py-1.5.3, pluggy-0.3.1 -- /usr/local/bin/python
cachedir: .cache
rootdir: /integration/tests, inifile:
collecting ... collected 17 items

test_basic.py::test_hosts_and_settings PASSED
test_basic.py::test_volume_basic PASSED
test_basic.py::test_volume_iscsi_basic PASSED
test_basic.py::test_snapshot PASSED
test_basic.py::test_backup PASSED
test_basic.py::test_volume_multinode PASSED

==================== 11 tests deselected by '-ktest_basic' =====================
================== 6 passed, 11 deselected in 363.13 seconds ===================
```

manager/integration/tests/test_basic.py
```diff
@@ -168,6 +193,10 @@ def test_volume_iscsi_basic(clients, volume_name):  # NOQA
     assert volumes[0]["frontend"] == "iscsi"
     assert volumes[0]["endpoint"].startswith("iscsi://")
 
+    dev = iscsi_login(volumes[0]["endpoint"])
```
yasker: Use `try...finally` in case there is an exception happened when iSCSI initiator connected.

manager/integration/tests/test_basic.py
```diff
@@ -125,6 +146,9 @@ def test_volume_basic(clients, volume_name):  # NOQA
     volume = client.by_id_volume(volume_name)
     assert volume["endpoint"] == DEV_PATH + volume_name
 
+    # RW test
```
yasker: It's very clear from code that it's for rw_test. The comment is redundant.

manager/integration/tests/test_basic.py
```diff
@@ -193,8 +222,10 @@ def test_snapshot(clients, volume_name):  # NOQA
     assert volume["numberOfReplicas"] == 2
     assert volume["state"] == "detached"
 
-    volume = volume.attach(hostId=host_id)
+    lht_hostId = get_self_host_id()
+    volume = volume.attach(hostId=lht_hostId)
     volume = wait_for_volume_state(client, volume_name, "healthy")
+    volume_rw_test(volume["endpoint"])
```
yasker: This doesn't make sense.

We don't need to test against the same type of scenario again and again.

The snapshot/backup test is to test if the data in the snapshot/backup is correct.

manager/integration/tests/test_basic.py
```diff
@@ -210,8 +236,13 @@ def test_snapshot(clients, volume_name):  # NOQA
 def snapshot_test(client, volname):
     volume = client.by_id_volume(volname)
 
+    start_pos = VOLUME_RWTEST_START_POSITION
     snap1 = volume.snapshotCreate()
+    snap2_wdata = generate_random_data(VOLUME_RWTEST_COUNT)
+    volume_write(volume["endpoint"], start_pos, snap2_wdata)
     snap2 = volume.snapshotCreate()
+    snap3_wdata = generate_random_data(VOLUME_RWTEST_COUNT)
+    volume_write(volume["endpoint"], start_pos, snap3_wdata)
```
yasker: After snap3 deleted, snap3_wdata should still exist. Deletion of a snapshot shouldn't affect the data of volume.

The data written for snap1/snap2/snap3 need to be in the different positions to verify that all the previous data still intact, even after purge, if no revert happened.

manager/integration/tests/common.py
```diff
@@ -212,6 +233,122 @@ def get_compatibility_test_image(cli_v, cli_minv,
...
+    free = {}
+
+    if used is None:
+        position = random.randrange(start, end - size, 1)
+    else:
+        for u_start in sorted(used.keys()):
+            u_end = used[u_start]
+            if u_start > start:
+                free[start] = u_start
+            start = u_end
```
yasker: This algorithm is correct, but it's unnecessarily complicated. We're writing a few bytes in a gigabyte-size disk. You can just generate a random number, calculate the start and the end and check if it collides with an existing position.

manager/integration/tests/common.py
```diff
@@ -212,6 +233,122 @@ def get_compatibility_test_image(cli_v, cli_minv,
...
 
+def generate_random_data(count):
+    return ''.join(random.choice(string.ascii_lowercase + string.digits)
+                   for _ in range(count))
+
+
+def volume_read(dev, start, count):
+    r_data = ""
+    fdev = open(dev, 'rb')
+    if fdev is not None:
```
yasker: Don't need to check None or not. If it's none, it will error out later. Or you can use `assert` to make sure it's not None.

manager/integration/tests/common.py
```diff
@@ -212,6 +233,122 @@ def get_compatibility_test_image(cli_v, cli_minv,
...
+        fdev.seek(start)
+        r_data = fdev.read(count)
+        fdev.close()
+    return r_data
+
+
+def volume_write(dev, start, data):
+    w_length = 0
+    fdev = open(dev, 'rb+')
+    if fdev is not None:
```
yasker: Same here.

manager/integration/tests/common.py
zhoubofsy: I previously considered using the algorithm of check collision and gave up.
Now the algorithm of check collision can solve the problem quickly and easily. Please imagine that if you have allocated a lot of scattered positions, you will need to produce the position frequently, because the possibility of collision increases. But my algorithm can't solve the fragmentation problem.
I think we should add an assert when the position cannot be allocated.

yasker: Do you mean hard to check collision using random location?

Currently the data we're going to write is a few hundred bytes. Total size of the volume is 30GB. The chance of collision is less than 1 in 30,000,000.

This algorithm works better if we have more constrain space to work with, though it still can't help with fragment if that need to be taken care of with constrain space.

The issue of this piece of code here is it's very easy to get it wrong and not easy to understand. Since there is an easy and straightforward alternative for our use case, I don't think it's necessary do it in this way.

And you're right. We need to make sure the position has been allocated.

manager/integration/tests/common.py
```diff
@@ -212,6 +233,122 @@ def get_compatibility_test_image(cli_v, cli_minv,
...
+
+    if not used:
+        position = random.randrange(0, int(SIZE)-size, 1)
+    else:
+        for i in range(RETRY_COUNTS):
+            pos = random.randrange(0, int(SIZE)-size, 1)
+            for start, end in used.items():
+                # check collision
+                if start <= pos and pos < end:
+                    continue
```
yasker: Looks like you basic wrote:
```
If the generated position collides with this used position, try the next used position.
Oh, this time it doesn't collide. Nice, I will forget about the first collision and pick this one.
```

Just use this function instead:
```
def generate_random_pos(size, used={}):
    for i in range(RETRY_COUNTS):
        pos = 0
        if int(SIZE) != size:
            pos = random.randrange(0, int(SIZE)-size, 1)
        collided = False
        # it's [start, end) vs [pos, pos + size)
        for start, end in used.items():
            if pos + size <= start or pos >= end:
                continue
            collided = True
            break
        if not collided:
            break
    assert not collided
    used[pos] = pos + size
    return pos
```

yasker:
LGTM. Merged, thanks.

PR #50, opened 2018-06-28
Add tests for base image volumes

LLParse:


manager/integration/tests/test_base_image.py
```diff
@@ -0,0 +1,203 @@
...
+import common
+
+from common import clients, volume_name  # NOQA
+from common import SIZE, DEV_PATH, BASE_IMAGE
+from common import wait_for_volume_state, wait_for_volume_delete
+from common import generate_volume_name
+from test_basic import snapshot_test, backup_test
+
+
+def test_volume_basic(clients, volume_name):  # NOQA
```
yasker: Can you refactor `test_basic.py` to avoid duplicated code? E.g. pass the BASE_IMAGE as a string parameter to the function and the comparison in the volume properties check would still work.

manager/integration/tests/test_basic.py
```diff
@@ -76,6 +76,9 @@ def volume_rw_test(dev):
 
 
 def test_volume_basic(clients, volume_name):  # NOQA
+    num_hosts = len(clients)
```
yasker: Put this into a separate commit since it doesn't require 3 nodes setup to work.

manager/integration/.gitignore
```diff
@@ -0,0 +1 @@
+bin/
```
yasker: Put changes other than the `tests` into one separate commit as well.

yasker:
And `test_ha`, `test_engine_upgrade` also needs update. We can give `test_recurring_job` a pass for now since the cronjob part shouldn't affect by the base image feature.

PR #52, opened 2018-07-03
Replace Host API in test

JacieChao:
Host API has been removed from manager, using Node API instead of it.

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -16,24 +16,24 @@ from common import generate_random_pos
 
 
 def test_hosts_and_settings(clients):  # NOQA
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -503,7 +503,7 @@ def test_replica_scheduler(clients, volume_name):  # NOQA
     for host_id, client in clients.iteritems():
         break
 
-    hosts = clients.itervalues().next().list_host()
+    hosts = clients.itervalues().next().list_node()
```
yasker: This won't be necessary now since we've removed the `list_host` API>

yasker:
LGTM. Merged, thanks.

PR #74, opened 2018-07-19
Add test for multiple disk scheduling

JacieChao:
1. multiple disk test
   - [x] Add/Delete multiple disks
   - [x] Update disk `AllowScheduling ` and `StorageReserved `
   - [x] Add/Delete multiple disks fail test
2. test scheduler
   - [x] Test no disk for scheduling
   - [x] Test volume too large to fill into a certain disk
   - [x] Test volume too large to fill any disks
   - [x] Test reduce `StorageReserved` that large volume could be scheduled to
3. test node controller
   - [x] test update `StorageScheduled ` after replica created
   - [x] test update `StorageAvailable ` after write certain size of file into disk.


manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,359 @@
...
+    disks = node["disks"]
+    for key, disk in disks.iteritems():
+        if disk["path"] == disk_path1 or disk["path"] == disk_path2:
+            assert not disk["allowScheduling"]
+            assert disk["storageReserved"] == Gi
+            assert disk["storageScheduled"] == 0
+
+    # delete disk
+    update_disk = generate_update_disk(disks)
+    delete_disk = []
```
yasker: Um, don't call it `delete_disk`. It sounds like the disk going to be deleted. It's more like the `remain_disk`.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,359 @@
...
+    small_disk_path = create_disk(client, "vol-small",
+                                  SIZE, lht_hostId)
+    small_disk = {"path": small_disk_path, "allowScheduling": True,
+                  "storageMaximum": SMALL_DISK_SIZE}
+    update_disks = generate_update_disk(node["disks"])
+    update_disks.append(small_disk)
+    node = node.diskUpdate(disks=update_disks)
+    assert len(node["disks"]) == len(update_disks)
+
+    unexpectd_disk = {}
```
yasker: Typo. `unexpected`.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,359 @@
...
+    lht_hostId = get_self_host_id()
+    nodes = client.list_node()
+
+    # create a volume and test update StorageScheduled of each node
+    vol_name = common.generate_volume_name()
+    volume = create_volume(client, vol_name, str(1*Gi), lht_hostId, len(nodes))
+    replicas = volume["replicas"]
+
+    # wait for node controller update
+    time.sleep(30)
```
yasker: Use a constant instead of `30`

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,359 @@
...
+        node = node.diskUpdate(disks=[])
+        assert len(node["disks"]) == 0
+
+    # test there's no disk fit for volume
+    vol_name = common.generate_volume_name()
+    volume = client.create_volume(name=vol_name,
+                                  size=SIZE, numberOfReplicas=len(nodes))
+    volume = common.wait_for_volume_status(client, vol_name,
+                                           common.VOLUME_FIELD_STATE, "")
+    assert volume["state"] == ""
```
yasker: This is not enough to determine if the volume is un-schedulable. When the volume was created its state can be "" as well. We need to figure out a way to report un-schedulable volumes.

yasker: I will add a new `Conditions` field in the volume to address this issue.

yasker: @JacieChao I've added Conditions to volumeStatus, and make scheduled as one condition.

See this commit for details on how it works: https://github.com/rancher/longhorn-manager/commit/e9ba26bad604dcb9c492c4790a02adbe02f3aa49

I've exposed in the API as volume.conditions, like this:
```
"conditions": {
"scheduled": {
"lastProbeTime": "2018-07-20T02:07:50Z",
"lastTransitionTime": "2018-07-20T02:07:37Z",
"message": "",
"reason": "ReplicaSchedulingFailure",
"status": "False",
"type": "scheduled",
},
},
```

`status: False` (using k8s words) means scheduling failed, as stated in the reason.

manager/integration/tests/common.py
```diff
@@ -807,3 +811,49 @@ def get_upgrade_test_image(cli_v, cli_minv,
...
+
+
+def cleanup_disk():
+    dir_list = os.listdir(DIRECTORY_PATH)
+    for dir in dir_list:
+        mount_path = generate_disk_directory(dir)
+        os.rmdir(mount_path)
+
+
+def umount_disk(vol_name):
```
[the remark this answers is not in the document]
yasker: And make sure you will unmount the disk even test failed or interrupted (as long as it doesn't receive SIGKILL)

manager/integration/deploy/test.yaml
```diff
@@ -57,12 +57,24 @@ spec:
       mountPath: /dev
     - name: proc
       mountPath: /host/proc
+    - name: disk-directory
+      mountPath: /tmp/longhorn-test
+      mountPropagation: Bidirectional
+    - name: longhorn
+      mountPath: /var/lib/rancher/longhorn/
+      mountPropagation: Bidirectional
```
yasker: Are we mounting anything to `/var/lib/rancher/longhorn`?  Not sure about bidirectional mount propagation here.

manager/integration/tests/common.py
```diff
@@ -807,3 +811,56 @@ def get_upgrade_test_image(cli_v, cli_minv,
...
+        subprocess.check_call(cmd)
+
+
+def umount_host_disk(vol_name):
+    mount_path = DIRECTORY_PATH + vol_name
+    cmd = ['umount', mount_path]
+    subprocess.check_call(cmd)
+
+
+def wait_for_volume_condition(client, name):
```
yasker: Don't quite make sense to just wait for a condition to show up, since there can be multiple conditions and you may not got the one you're waiting for.

Add conditionType ("scheduled") and conditionStatus("true/false") to the parameter as well.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,375 @@
...
+    assert volumeByName["state"] == volume["state"]
+    assert volumeByName["created"] == volume["created"]
+
+    volume.attach(hostId=node_id)
+    volume = common.wait_for_volume_healthy(client, vol_name)
+
+    return volume
+
+
+def clean_volume(client, vol_name):
```
yasker: `cleanup_volume`.

Also, you don't need to wait for the volume to detach before delete it.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,375 @@
...
+
+def clean_disk(client, *args):
+    # clean disk
+    for vol_name in args:
+        # umount disk
+        common.umount_host_disk(vol_name)
+        # clean volume
+        clean_volume(client, vol_name)
+
+    volumes = client.list_volume()
```
yasker: It's weird for the function to ask for the volume names, also check all the volumes at the same time. It can be either, but not both.

For here, I think it doesn't need to check if no volume has been left. `client` finalizer should able to do that.

manager/integration/tests/common.py
```diff
@@ -807,3 +811,56 @@ def get_upgrade_test_image(cli_v, cli_minv,
...
+
+def cleanup_host_disk_directory():
+    dir_list = os.listdir(DIRECTORY_PATH)
+    for dir in dir_list:
+        mount_path = DIRECTORY_PATH + dir
+        cmd = ['rm', '-r', mount_path]
+        subprocess.check_call(cmd)
+
+
+def umount_host_disk(vol_name):
```
yasker: You can merge `umount_host_disk` and `cleanup_host_disk_directory` into `cleanup_host_disk`, which will umount then remove the directory. Don't think either step will be useful by it own.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,375 @@
...
+        id = replica["hostId"]
+        assert id != ""
+        assert replica["running"]
+
+    # prepare the disk in the host filesystem
+    disk_path = common.prepare_host_disk(volume["endpoint"], volume["name"])
+    return disk_path
+
+
+def clean_disk(client, *args):
```
yasker: `cleanup_disks`

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,375 @@
...
+
+    # test add disk exception
+    with pytest.raises(Exception):
+        # add same disk by different mount path
+        disk = {"path": "/var/lib", "allowScheduling": True,
+                "storageMaximum": 5*Gi, "storageReserved": 2*Gi}
+        update_disk = get_update_disks(disks)
+        update_disk.append(disk)
+        node = node.diskUpdate(disks=update_disk)
+
```
yasker: You can check the for error message if an exception is expected. See https://github.com/rancher/longhorn-tests/blob/master/manager/integration/tests/test_engine_upgrade.py#L126

Also, only the line will cause exception in the `with`, otherwise if any other code caused exception it will result in a false negative result.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,375 @@
...
+    # test delete disk exception
+    with pytest.raises(Exception):
+        node.diskUpdate(disks=[])
```
yasker: Same here.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,375 @@
...
+    disks = node["disks"]
+    update_disk = get_update_disks(disks)
+    for disk in update_disk:
+        # keep default disk for other tests
+        if disk["path"] == disk_path1 or disk["path"] == disk_path2:
+            disk["allowScheduling"] = False
+            disk["storageReserved"] = Gi
+    node.diskUpdate(disks=update_disk)
+    # wait for node controller to update disk status
+    time.sleep(WAIT_FOR_NODE_STATUS)
```
yasker: Can we use polling (wait for the expected result) instead of sleep here (and below)? Interval can be longer but wait for 1 minutes everytime seems excessive.

manager/integration/tests/test_node.py
```diff
@@ -0,0 +1,375 @@
...
+                assert disk["storageScheduled"] == 1*Gi
+                break
+
+    # clean volumes
+    clean_volume(client, vol_name)
+
+    # write specified byte data into disk
+    if os.path.exists(TEST_FILE):
+        os.remove(TEST_FILE)
+    cmd = ['dd', 'if=/dev/zero', 'of=' + TEST_FILE, 'bs=1M', 'count=1024']
```
yasker: This test file is way too big. Should we able to detect e.g. 10M? That should be enough.

manager/integration/deploy/test.yaml
JacieChao: Yeah, I used this default disk to write some file into disk for testing node_controller. Or maybe create a new disk that use `/tmp/longhorn` directory is properly?

yasker: 1. I cannot find the hard disk we additionally mounted inside the container to the `/var/lib/rancher/longhorn`. If that's the case, we don't need mount propagation.
2. Yeah, as you said it's not a great idea to use this host directory for testing. It can potentially leave files in the host filesystem and would be hard to detect that as well. Let's create a new disk for that.

manager/integration/README.md
```diff
@@ -6,6 +6,8 @@ Requirement:
 1. A Kubernetes cluster with at least 3 nodes
 2. Longhorn system has already been successfully deployed in the cluster.
 3. No volume exists in the Longhorn system.
+4. Need kubernetes 1.10 or higher.
+5. Need to enable `feature-gates: MountPropagation=true` for kube-api and kubelet while using RKE
```
yasker: Don't need to mention `while using RKE`. The issue has been fixed on RKE. We can just say:

E.g. 
5. Make sure `MountPropagation` feature gate is enabled
5.1. For RKE before v0.1.9, you would need the extra parameter `feature-gates: MountPropagation=true` for `kube-api` and `kubelet` to enable the feature gate.

yasker:
Merged, thanks.

PR #87, opened 2018-08-02
Clear output of mkfs.ext4, mount and umount operation

JacieChao:
Fixed issue: https://github.com/rancher/longhorn/issues/139

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1066,26 +1066,28 @@ def get_upgrade_test_image(cli_v, cli_minv,
 
 
 def prepare_host_disk(dev, vol_name):
+    FNULL = open('py.stdout', 'w', 0)
     cmd = ['mkfs.ext4', dev]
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -1066,26 +1066,28 @@ def get_upgrade_test_image(cli_v, cli_minv,
 
 
 def prepare_host_disk(dev, vol_name):
+    FNULL = open('py.stdout', 'w', 0)
     cmd = ['mkfs.ext4', dev]
-    subprocess.check_call(cmd)
+    subprocess.check_call(cmd, stdout=FNULL, stderr=subprocess.STDOUT)
```
yasker: We just need to redirect the output to stdout. It will be useful for debugging.

yasker:
https://github.com/rancher/longhorn/issues/139 is invalid. I've specified `-s` in the command...

PR #88, opened 2018-08-02
Add cleanup disks after node test

JacieChao:
Add cleanup node disks after node test, delete all test disks from node.

Fixed issue: https://github.com/rancher/longhorn/issues/136 and https://github.com/rancher/longhorn/issues/137

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -639,6 +639,9 @@ def clients(request):
 
 
 def cleanup_client(client):
+    # cleanup test disks
+    cleanup_test_disks(client)
+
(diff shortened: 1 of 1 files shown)


yasker:
I don't think this fixes rancher/longhorn#136 ? This is only a better clean up. The user can still umount the disk manually on the host and cause it stuck in Longhorn.

yasker:
Need to deal with rancher/longhorn#136 separately. This part LGTM.

yasker:
Wait, no...

Github lost my review comments.

Need to deal with reserved space per disk too.

PR #89, opened 2018-08-02
FlexVolume tests

LLParse:


manager/test_containers/baseimage/generate.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash -e
+
+if [ ! -f parrot.qcow2 ]; then
+  qemu-img create -f qcow2 parrot.img 1G
```
yasker: 1G is too big. Can we do e.g. 32Mb? We need to run checksum for the whole disk.

manager/test_containers/baseimage/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM llparse/vm-base:930a055
```
yasker: I thought we're going to based on busybox?

manager/test_containers/baseimage/generate.sh
```diff
@@ -0,0 +1,27 @@
...
+  qemu-img convert -c -O qcow2 parrot.img parrot.qcow2
+  rm -f parrot.img
+fi
+
+docker build \
+  -t rancher/longhorn-test-baseimage:ext4 \
+  --build-arg BASE_IMAGE=parrot.qcow2 \
+    .
+
+docker push rancher/longhorn-test-baseimage:ext4
```
yasker: Use the `rancher/longhorn-test` repo like others.

manager/test_containers/baseimage/Dockerfile
LLParse: Yeah, that is literally just a pointer to busybox. I'll change it

PR #99, opened 2018-08-06
Conditionally run base image tests

LLParse:
Only run base image tests if all nodes support mount propagation


manager/integration/tests/conftest.py
```diff
@@ -50,3 +51,21 @@ def pytest_collection_modifyitems(config, items):
...
+    if not allNodesSupporMountPropagation:
+        skip_upgrade = pytest.mark.skip(reason="environment does not " +
+                                               "support base image")
+
+        for item in items:
+            if "baseimage" in item.keywords:
+                item.add_marker(skip_upgrade)
+
+
+def client_longhorn():
```
yasker: Use `get_longhorn_api_client`?

PR #107, opened 2018-08-10
test ready state of disk status in umount case

JacieChao:
Add tests for https://github.com/rancher/longhorn-manager/pull/164


manager/integration/tests/test_node.py
```diff
@@ -845,7 +848,13 @@ def test_node_delete_umount_disks(client):  # NOQA
             assert disk["storageAvailable"] == 0
             assert disk["storageReserved"] == 0
             assert disk["storageScheduled"] == SMALL_DISK_SIZE
+            conditions = disk["conditions"]
+            assert conditions[DISK_CONDITION_READY]["status"] == \
+                CONDITION_STATUS_FALSE
```
yasker: Need to test condition `schedulable` as well.

PR #108, opened 2018-08-13
Add test for disks and settings validation

JacieChao:
Add validation tests for https://github.com/rancher/longhorn-manager/pull/169

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -66,10 +66,50 @@ def test_hosts_and_settings(clients):  # NOQA
 
         old_value = setting["value"]
 
-        setting = client.update(setting, value="testvalue")
-        assert setting["value"] == "testvalue"
-        setting = client.by_id_setting(name)
-        assert setting["value"] == "testvalue"
+        if name == SETTING_STORAGE_OVER_PROVISIONING_PERCENTAGE:
+            with pytest.raises(Exception) as e:
+                client.update(setting, value="-100")
+            assert "with invalid StorageOverProvisioningPercentag" in \
+                   str(e.value)
+            with pytest.raises(Exception) as e:
+                client.update(setting, value="testvalue")
+            assert "with invalid StorageOverProvisioningPercentag" in \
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -66,10 +66,46 @@ def test_hosts_and_settings(clients):  # NOQA
...
+            with pytest.raises(Exception) as e:
+                client.update(setting, value="testvalue")
+            assert "with invalid StorageOverProvisioningPercentag" in \
+                   str(e.value)
+            setting = client.update(setting, value="200")
+            assert setting["value"] == "200"
+            setting = client.by_id_setting(name)
+            assert setting["value"] == "200"
+        elif name == SETTING_STORAGE_MINIMAL_AVAILABLE_PERCENTAGE:
+            with pytest.raises(Exception) as e:
```
yasker: Try negative value as well.

manager/integration/tests/test_node.py
```diff
@@ -113,7 +121,7 @@ def test_node_disk_update(client):  # NOQA
         # keep default disk for other tests
         if disk["path"] == disk_path1 or disk["path"] == disk_path2:
             disk["allowScheduling"] = False
-            disk["storageReserved"] = Gi
+            disk["storageReserved"] = 0
```
yasker: Yeah, I think you should keep the original test if possible, like I said in the manager PR.

JacieChao: This change is due to validation of disk storageReserved. My test disk size is 16MB, but when storageReserved is set to Gi, it couldn't pass this test.

yasker: Hmm, why?

JacieChao: Because we add a validation in node api, to validate `storageReserved` setting should no more than `storageMaximum`. If I keep the original test, it will return a validate exception, which has already added in disk update case.

yasker: Yeah, I've commented on that earlier…

PR #124, opened 2019-02-15
fix test_statefulset_restore: csi check

shuo-wu:


manager/integration/tests/common.py
```diff
@@ -1516,3 +1523,47 @@ def find_backup(client, vol_name, snap_name):
...
+                has_csi_plugin = True
+            elif item.metadata.labels.get('app', '') == 'longhorn-driver-deployer':
+                has_driver_deployer = True
+            elif item.metadata.labels.get('app', '') == 'longhorn-manager':
+                has_manager = True
+
+        if has_attacher and has_provisioner and has_engine_image_ei and \
+                has_csi_plugin and has_driver_deployer and has_manager:
+            csi = CSI_TRUE
+        elif not has_attacher and not has_provisioner and not has_engine_image_ei \
```
yasker: This part is wrong. Attacher, provisioner and CSI plugin are part of CSI. But driver deployer, engine image, and manager should exist for all the Longhorn setup. Flexvolume won't work without those as well.

You can deploy Longhorn using `--driver flexvolume` (see https://github.com/rancher/longhorn-manager/blob/master/deploy/install/02-components/04-driver.yaml#L33) and see what's going on with Flexvolume.

Though we can add an enhancement that fails test immediately if either CSI or Flexvolume driver or Longhorn deployment was found. It should able to be done in `conftest.py`.

manager/integration/tests/common.py
```diff
@@ -1516,3 +1523,47 @@ def find_backup(client, vol_name, snap_name):
...
+        longhorn_pod_list = core_api.list_namespaced_pod('longhorn-system', include_uninitialized=False)
+        for item in longhorn_pod_list.items:
+            if not item.metadata.labels:
+                pass
+            elif item.metadata.labels.get('app', '') == 'csi-attacher':
+                has_attacher = True
+            elif item.metadata.labels.get('app', '') == 'csi-provisioner':
+                has_provisioner = True
+            elif item.metadata.labels.get('longhorn', '') == 'engine-image':
+                has_engine_image_ei = True
```
yasker: `ei` is the prefix for `engine image` pods, so no need in the variable name.

yasker:
Also, put a more meaningful commit message, e.g. 

```
manager: fix test_statefulset_restore due to deployment change of CSI plugins
```

yasker:
Also, we should refactor `conftest.py` to use this logic as well. But put that change into a separate commit (in the same PR).

PR #127, opened 2019-03-16
manager: pid of dockerd can be detected dynamically when testing iscsi

shuo-wu:


yasker:
LGTM. Let’s wait for all the test passed, in case this one needs some extra touches.

yasker:
Add https://github.com/rancher/longhorn/issues/10 to the commit message.

Remove `bug fixed.` in the commit title.

Otherwise LGTM.

manager/integration/tests/test_node.py
```diff
@@ -959,6 +960,7 @@ def test_node_delete_umount_disks(client):  # NOQA
     subprocess.check_call(cmd)
 
 
+@pytest.mark.debug   # NOQA
```
yasker: what's this?

manager/integration/tests/test_node.py
```diff
@@ -1003,21 +1005,22 @@ def test_replica_cleanup(client):  # NOQA
...
         data_paths.append(replica["dataPath"])
 
+    dockerd_pid = find_dockerd_pid() or "1"
     # data path should exist now
     for data_path in data_paths:
-        assert subprocess.call(["nsenter",
-                                "--mount=/host/proc/1/ns/mnt",
-                                "--net=/host/proc/1/ns/net",
-                                "bash", "-c", "ls " + data_path]) == 0
+        assert subprocess.call(
```
yasker: Can we use `exec_nsenter` instead?

shuo-wu: No. Using exec_nsenter() will cause error.

manager/integration/tests/test_node.py
shuo-wu: sorry. I will delete this line

manager/integration/tests/test_node.py
yasker: why it would cause an error?

manager/integration/tests/test_ha.py
```diff
@@ -250,7 +250,7 @@ def test_ha_prohibit_deleting_last_replica(client, volume_name):  # NOQA
 
     with pytest.raises(Exception) as e:
         volume.replicaRemove(name=replica0["name"])
-    assert "no other healthy replica available" in str(e.value)
+    assert "no other healthy replica available" in str(e.value).lower()
```
yasker: this line has been fixed in the manager, so shouldn't need this one.

yasker:
LGTM.

PR #129, opened 2019-03-26
manager: add test for listing invalid backup volume

shuo-wu:
https://github.com/rancher/longhorn/issues/446

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -1,4 +1,6 @@
+import time
 import common
+import subprocess
 import pytest
 
 from common import clients, volume_name     # NOQA
@@ -13,6 +15,47 @@ from common import generate_volume_name
 from common import get_volume_endpoint, get_volume_engine
 from common import get_random_client
 from common import CONDITION_STATUS_FALSE, CONDITION_STATUS_TRUE
+from common import RETRY_COUNTS, RETRY_ITERVAL, RETRY_COMMAND_COUNT
+
+
+def create_volume(client, vol_name, num_of_replicas=2,
+                  size=SIZE, base_image="", frontend="blockdev"):
+    client.create_volume(name=vol_name, size=size,
+                         numberOfReplicas=num_of_replicas,
+                         frontend=frontend, baseImage=base_image)
+    volume = common.wait_for_volume_detached(client, vol_name)
+    assert volume["name"] == vol_name
+    assert volume["size"] == size
+    assert volume["numberOfReplicas"] == num_of_replicas
+    assert volume["state"] == "detached"
+    assert volume["baseImage"] == base_image
+    assert volume["frontend"] == frontend
+    assert volume["created"] != ""
+    return volume
+
+
+def create_backup(client, volname):
+    volume = client.by_id_volume(volname)
+    volume.snapshotCreate()
+    data = write_volume_random_data(volume)
+    snap = volume.snapshotCreate()
+    volume.snapshotCreate()
+    volume.snapshotBackup(name=snap["name"])
+
+    bv, b = common.find_backup(client, volname, snap["name"])
+
+    new_b = bv.backupGet(name=b["name"])
+    assert new_b["name"] == b["name"]
+    assert new_b["url"] == b["url"]
+    assert new_b["snapshotName"] == b["snapshotName"]
+    assert new_b["snapshotCreated"] == b["snapshotCreated"]
+    assert new_b["created"] == b["created"]
+    assert new_b["volumeName"] == b["volumeName"]
+    assert new_b["volumeSize"] == b["volumeSize"]
+    assert new_b["volumeCreated"] == b["volumeCreated"]
+
+    return bv, b, snap, data
+
 
 @pytest.mark.coretest   # NOQA
 def test_hosts(clients):  # NOQA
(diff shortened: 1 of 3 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -13,6 +15,47 @@
 from common import get_volume_endpoint, get_volume_engine
 from common import get_random_client
 from common import CONDITION_STATUS_FALSE, CONDITION_STATUS_TRUE
+from common import RETRY_COMMAND_COUNT
+
+
+def create_volume(client, vol_name, num_of_replicas=2,
```
yasker: Can this one be merged with `client.create_volume`?

manager/integration/tests/test_basic.py
```diff
@@ -13,6 +15,47 @@
...
+    assert new_b["name"] == b["name"]
+    assert new_b["url"] == b["url"]
+    assert new_b["snapshotName"] == b["snapshotName"]
+    assert new_b["snapshotCreated"] == b["snapshotCreated"]
+    assert new_b["created"] == b["created"]
+    assert new_b["volumeName"] == b["volumeName"]
+    assert new_b["volumeSize"] == b["volumeSize"]
+    assert new_b["volumeCreated"] == b["volumeCreated"]
+
+    return snap, data
```
yasker: If the function called `create_backup`, it should return the backup as well. Probably backup volume too.

manager/integration/tests/test_basic.py
```diff
@@ -469,26 +469,9 @@ def backup_test(clients, volume_name, size, base_image=""):  # NOQA
...
-    volume = client.by_id_volume(volname)
-    volume.snapshotCreate()
-    data = write_volume_random_data(volume)
-    snap2 = volume.snapshotCreate()
-    volume.snapshotCreate()
-
-    volume.snapshotBackup(name=snap2["name"])
-
+    snap2, data = create_backup(client, volname)
     bv, b = common.find_backup(client, volname, snap2["name"])
```
yasker: You won't need this if `create_backup` return `bv, b, snap, data`.

manager/integration/tests/test_basic.py
```diff
@@ -519,6 +502,115 @@ def backupstore_test(client, host_id, volname, size):
     assert not found
 
 
+@pytest.mark.coretest   # NOQA
+def test_listing_backup_volume(clients, base_image=""):   # NOQA
+    for host_id, client in clients.iteritems():
+        break
+    lht_hostId = get_self_host_id()
+
+    # create 2 volumes.
```
yasker: Create 3 with kind of random names, `volume-test-<random chars>`. We don't know the exact order which volume would be read first, which one second. But we want to make sure after reading an error volume, the read will continue.

Using random names ensures that at least in the multiple runs, the error volume will not always be the last one to be read.

manager/integration/tests/test_basic.py
```diff
@@ -519,6 +502,115 @@ def backupstore_test(client, host_id, volname, size):
...
+    volume1 = create_volume(client, volume1_name)
+    volume2 = create_volume(client, volume2_name)
+    volume1.attach(hostId=lht_hostId)
+    volume1 = common.wait_for_volume_healthy(client, volume1_name)
+    volume2.attach(hostId=lht_hostId)
+    volume2 = common.wait_for_volume_healthy(client, volume2_name)
+
+    setting = client.by_id_setting(common.SETTING_BACKUP_TARGET)
+    backupstores = common.get_backupstore_url()
+    for backupstore in backupstores:
```
yasker: Add a comment here, noted that we only test NFS, and the reason.

manager/integration/tests/test_basic.py
```diff
@@ -519,6 +502,115 @@ def backupstore_test(client, host_id, volname, size):
...
+        if common.is_backupTarget_nfs(backupstore):
+            nfs_url = backupstore.strip("nfs://")
+            setting = client.update(setting, value=backupstore)
+            assert setting["value"] == backupstore
+
```
yasker: Assert backupstore is `NFS` based. (you may need to retry and wait for the entry to update)

manager/integration/tests/test_basic.py
```diff
@@ -519,6 +502,115 @@ def backupstore_test(client, host_id, volname, size):
...
+    cmd = ["mv", volume2_backup_volume_cfg_path + ".tmp",
+           volume2_backup_volume_cfg_path]
+    subprocess.check_output(cmd)
+    subprocess.check_output(["sync"])
+    time.sleep(3)
+
+    bv1, b1 = common.find_backup(client, volume1_name, snap1["name"])
+    bv1.backupDelete(name=b1["name"])
+    backups1 = bv1.backupList()
```
yasker: The delay of removing backup may result in this check fail.

manager/integration/tests/test_basic.py
shuo-wu: no. client is imported from `cattle`

manager/integration/tests/test_basic.py
```diff
@@ -519,6 +501,150 @@ def backupstore_test(client, host_id, volname, size):
...
+    cmd = ["find", volume1_backup_volume_path, "-name", "volume.cfg"]
+    volume1_backup_volume_cfg_path = subprocess.check_output(cmd).strip()
+    cmd = ["mv", volume1_backup_volume_cfg_path,
+           volume1_backup_volume_cfg_path + ".tmp"]
+    subprocess.check_output(cmd)
+    subprocess.check_output(["sync"])
+    time.sleep(5)
+
+    found1 = found2 = found3 = False
+    for i in range(100):
```
yasker: We had `RETRY_COUNTS` and `RETRY_INTERVAL` defined.

manager/integration/tests/test_basic.py
```diff
@@ -519,6 +501,150 @@ def backupstore_test(client, host_id, volname, size):
...
+        if found1 & found2 & found3:
+            break
+        time.sleep(1)
+    assert found1 & found2 & found3
+
+    cmd = ["mv", volume1_backup_volume_cfg_path + ".tmp",
+           volume1_backup_volume_cfg_path]
+    subprocess.check_output(cmd)
+    subprocess.check_output(["sync"])
+    time.sleep(3)
```
yasker: Seems arbitrary.

yasker: In fact `sync` is a blocking operation. Why we need a `sleep` here?

PR #130, opened 2019-04-03
manager: add integration test for kubernetes_controller

shuo-wu:


manager/integration/tests/common.py
```diff
@@ -205,7 +209,7 @@ def delete_and_wait_statefulset(api, client, statefulset):
         found = False
         for item in ret.items:
             if item.metadata.name == \
-              statefulset['metadata']['name']:
+                    statefulset['metadata']['name']:
```
yasker: ?

manager/integration/tests/test_kubernetes_status.py
```diff
@@ -0,0 +1,225 @@
...
+    for i in range(RETRY_COMMAND_COUNT):
+        if k_status['PVStatus'] == 'Bound':
+            break
+        time.sleep(1)
+        volume = client.by_id_volume(volume_name)
+        k_status = volume["kubernetesStatus"]
+    assert k_status['PVStatus'] == 'Bound'
+    assert k_status['Namespace'] == 'default'
+    assert k_status['PVCName'] == p['pvc_name']
+    assert k_status['PodName'] == p['pod_name']
```
yasker: This should be after the loop.

manager/integration/tests/test_kubernetes_status.py
```diff
@@ -0,0 +1,225 @@
...
+    for i in range(RETRY_COMMAND_COUNT):
+        if k_status['PodStatus'] == 'Running':
+            break
+        time.sleep(1)
```
yasker: `RETRY_INTERVAL` (seems there is a typo made it `RETRY_ITERVAL`)

manager/integration/tests/test_kubernetes_status.py
```diff
@@ -0,0 +1,225 @@
...
+        volume = client.by_id_volume(volume_name)
+        k_status = volume["kubernetesStatus"]
+    assert k_status['PodStatus'] == 'Running'
+    assert not k_status['LastPVCRefAt']
+    assert not k_status['LastPodRefAt']
+
+    delete_and_wait_statefulset_only(core_api, statefulset)
+    volume = client.by_id_volume(volume_name)
+    k_status = volume["kubernetesStatus"]
+    assert k_status['PVName'] == p['pv_name']
```
yasker: I think you would need some wait for the status to be updated?

manager/integration/tests/test_kubernetes_status.py
```diff
@@ -0,0 +1,225 @@
...
+    assert k_status['PVStatus'] == 'Bound'
+    assert k_status['Namespace'] == 'default'
+    assert k_status['PVCName'] == p['pvc_name']
+    assert k_status['PodName'] == p['pod_name']
+    assert k_status['WorkloadName'] == statefulset_name
+    assert k_status['WorkloadType'] == 'StatefulSet'
+    for i in range(RETRY_COMMAND_COUNT):
+        if k_status['PodStatus'] == 'Running':
+            break
+        time.sleep(1)
```
yasker: `RETRY_INTERVAL`

manager/integration/tests/test_kubernetes_status.py
shuo-wu: Here I set `replicas=1` to speed up the test. Does 2 replicas for the statefulset matter?

yasker: How much is the difference in speed? It's better to have more than one replicas since it's more likely the real use case.

manager/integration/tests/test_kubernetes_status.py
```diff
@@ -0,0 +1,240 @@
...
+        assert k_status['PodStatus'] == 'Running'
+        assert not k_status['WorkloadName']
+        assert not k_status['WorkloadType']
+        assert not k_status['LastPVCRefAt']
+        assert not k_status['LastPodRefAt']
+
+        delete_and_wait_pod(core_api, p['pod_name'])
+
+    delete_and_wait_pvc(core_api, pod_info)
+    # Since reclaim policy is `Delete`, we don't need to
```
yasker: It seems the reclaim policy is `retain`?

yasker:
Please separate the commit into three commits:
1. Correct typo of RETRY_INTERVAL
2. Changes to existing frameworks.
3. New test case.

yasker:
> Also, please separate the change to the existing test cases from adding the new one.
> 
> Add test case for if a pod in the statefulset get deleted (and recreated automatically by Kubernetes).

I think you missed those comments.

PR #134, opened 2019-04-09
catched error caused by deleting a non-existing object

shuo-wu:
https://github.com/rancher/longhorn/issues/476

### DIFF
--- a/manager/integration/tests/longhorn.py
+++ b/manager/integration/tests/longhorn.py
@@ -172,12 +172,16 @@ class Schema(object):
 
 
 class ApiError(Exception):
-    def __init__(self, obj):
+    def __init__(self, obj, status_code):
+        if not obj:
+            obj = RestObject()
+            obj.message = ""
+        obj.code = status_code
         self.error = obj
         try:
             msg = '{} : {}\n{}'.format(obj.code, obj.message, obj)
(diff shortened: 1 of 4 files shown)


manager/integration/tests/requirements.txt
```diff
@@ -1,6 +1,9 @@
 flake8
 
-cattle==0.5.1
+argcomplete
```
yasker: Better to pin with version.

manager/integration/tests/longhorn.py
```diff
@@ -172,12 +172,11 @@ def __repr(self):
...
 class ApiError(Exception):
-    def __init__(self, obj):
-        self.error = obj
+    def __init__(self, response):
+        self.status = response.status_code
         try:
-            msg = '{} : {}\n{}'.format(obj.code, obj.message, obj)
-            super(ApiError, self).__init__(self, msg)
-        except:
+            super(ApiError, self).__init__(self, response.text)
```
yasker: It's better to preserve the format, since https://github.com/rancher/longhorn-manager/blob/master/api/router.go#L37 is still using it.

shuo-wu: you mean `self.error`?

PR #142, opened 2019-05-21
Use list_backupVolume to update last backup field in test_restore_inc()

shuo-wu:
https://github.com/rancher/longhorn/issues/558

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -650,6 +650,8 @@ def restore_inc_test(client, core_api, volume_name, pod):  # NOQA
     _, backup1, _, data1 = create_backup(
         client, volume_name,
         {'len': 2 * 1024, 'pos': 0, 'content': zero_string * 2 * 1024})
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -650,6 +650,8 @@ def restore_inc_test(client, core_api, volume_name, pod):  # NOQA
     _, backup1, _, data1 = create_backup(
         client, volume_name,
         {'len': 2 * 1024, 'pos': 0, 'content': zero_string * 2 * 1024})
+    # use this api to update field `last backup`
+    client.list_backupVolume()
```
yasker: This shouldn't be needed. 

https://github.com/rancher/longhorn-manager/blob/master/manager/engine.go#L173

shuo-wu: That one is only responsible for updating the original volume itself. `ListBackupVolume` will update `Last Backup` of all volumes.

PR #143, opened 2019-05-28
Fix issue 'test_replica_scheduler_too_large_volume_fit_any_disks failed sometime'

shuo-wu:
https://github.com/rancher/longhorn/issues/557

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -83,7 +83,7 @@ SETTING_STORAGE_OVER_PROVISIONING_PERCENTAGE = \
     "storage-over-provisioning-percentage"
 SETTING_STORAGE_MINIMAL_AVAILABLE_PERCENTAGE = \
     "storage-minimal-available-percentage"
-DEFAULT_DISK_PATH = "/var/lib/rancher/longhorn"
+DEFAULT_DISK_PATH = "/var/lib/rancher/longhorn/"
 DEFAULT_STORAGE_OVER_PROVISIONING_PERCENTAGE = "500"
 DEFAULT_STORAGE_MINIMAL_AVAILABLE_PERCENTAGE = "10"
 
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -9,7 +9,8 @@ from common import Gi, SIZE, CONDITION_STATUS_FALSE, \
     DISK_CONDITION_SCHEDULABLE, DISK_CONDITION_READY
 from common import get_self_host_id
 from common import SETTING_STORAGE_OVER_PROVISIONING_PERCENTAGE, \
-    SETTING_STORAGE_MINIMAL_AVAILABLE_PERCENTAGE
+    SETTING_STORAGE_MINIMAL_AVAILABLE_PERCENTAGE, \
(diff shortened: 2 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -83,7 +83,7 @@
     "storage-over-provisioning-percentage"
 SETTING_STORAGE_MINIMAL_AVAILABLE_PERCENTAGE = \
     "storage-minimal-available-percentage"
-DEFAULT_DISK_PATH = "/var/lib/rancher/longhorn"
+DEFAULT_DISK_PATH = "/var/lib/rancher/longhorn/"
```
yasker: Why is the ending slash needed?

manager/integration/tests/test_node.py
```diff
@@ -320,25 +320,34 @@ def test_replica_scheduler_too_large_volume_fit_any_disks(client):  # NOQA
...
+                         numberOfReplicas=len(nodes))
     volume = common.wait_for_volume_condition_scheduled(client, vol_name,
                                                         "status",
                                                         CONDITION_STATUS_FALSE)
 
-    # reduce StorageReserved of each default disk
+    # Reduce StorageReserved of each default disk so that each node can fit
+    # only one replica.
+    # If one node can fit multiple replicas and the following node update gets
+    # stuck for seconds,
```
yasker: Don't need the comments after the first two lines.

manager/integration/tests/test_node.py
```diff
@@ -320,25 +320,34 @@ def test_replica_scheduler_too_large_volume_fit_any_disks(client):  # NOQA
...
+    # there may be multiple replicas scheduled in the same node.
+    # To prevent this race condition in the test, `storageReserved` should be
+    # set with a appropriate value so that one node can fit one replicas only.
+    schedulable = int(volume_size * 1.5 / 5)
```
yasker: `5` is the value of OverProvisioningPercentage.

Also, name the value `neededForScheduling` instead.

manager/integration/tests/test_node.py
```diff
@@ -320,25 +320,34 @@ def test_replica_scheduler_too_large_volume_fit_any_disks(client):  # NOQA
...
         for disk in update_disks:
-            disk["storageReserved"] = 0
+            disk["storageReserved"] = disk["storageMaximum"] - schedulable
         node = node.diskUpdate(disks=update_disks)
         disks = node["disks"]
         for fsid, disk in disks.iteritems():
             wait_for_disk_status(client, node["name"],
-                                 fsid, "storageReserved", 0)
+                                 fsid, "storageReserved",
+                                 disk["storageMaximum"] - schedulable)
```
yasker: Do we need to change the `storageReserved` back?

shuo-wu: `cleanup_client()` will reset nodes

manager/integration/tests/common.py
shuo-wu: when we run one single test in `test_node`, we won't call `reset_disks_for_all_nodes()` and the default disk path contains the ending slash, then the check like here will fail: https://github.com/shuo-wu/longhorn-tests/blob/94c3a33b857d710f166a5a501462d06f5efb6323/manager/integration/tests/test_node.py#L313

PR #144, opened 2019-06-04
Add nightly integration test for Longhorn

meldafrawi:


manager/integration/deploy/test.yaml
```diff
@@ -40,7 +40,8 @@ metadata:
 spec:
   containers:
   - name: longhorn-test-pod
-    image: rancher/longhorn-manager-test:v0.5.0
+    image: rancher/longhorn-manager-test:fcd7a9e
+    args: [ "-s" , "--junitxml=${LONGHORN_JUNIT_REPORT_PATH}" ]
```
yasker: Add this to the args below.

manager/integration/deploy/test.yaml
```diff
@@ -40,7 +40,8 @@ metadata:
 spec:
   containers:
   - name: longhorn-test-pod
-    image: rancher/longhorn-manager-test:v0.5.0
+    image: rancher/longhorn-manager-test:fcd7a9e
```
yasker: I don't think the image is there?

PR #148, opened 2019-06-12
Added testcases for restoration of new restored volume

ywei88:
Test cases to check if restoration completed when restoration call ended

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -482,6 +482,31 @@ def backup_test(clients, volume_name, size, base_image=""):  # NOQA
     assert len(volumes) == 0
 
 
+# test normally created volume's "RestorationRequired" field
+def test_restoration_required_field(clients):
+    for host_id, client in clients.iteritems():
+        break
+
+    volname = generate_volume_name()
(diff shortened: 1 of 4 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -481,6 +481,28 @@ def backup_test(clients, volume_name, size, base_image=""):  # NOQA
     volumes = client.list_volume()
     assert len(volumes) == 0
 
+def test_restorationCompleted(clients):
+    for host_id, client in clients.iteritems():
+        break
+
```
shuo-wu: Add a comment to explain this test case

manager/integration/tests/test_basic.py
```diff
@@ -495,6 +517,8 @@ def backupstore_test(client, host_id, volname, size):
     assert volume["size"] == size
     assert volume["numberOfReplicas"] == 2
     assert volume["state"] == "detached"
+    assert volume["restorationCompleted"] is True
```
shuo-wu: Add a new wait function to check the intermediate status of restoring volumes

yasker:
LGTM.

PR #149, opened 2019-06-19
Test standby volume activation can still work when backup store polling disabled

ywei88:
Issue 573
https://github.com/longhorn/longhorn/issues/573

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -535,6 +535,69 @@ def test_restore_inc(clients, core_api, volume_name, pod):  # NOQA
         restore_inc_test(client, core_api, volume_name, pod)
 
 
+def test_standby_volume_activation(clients):
+    for _, client in clients.iteritems():
+        break
+
+    interval = client.by_id_setting("backupstore-poll-interval")
+    interval = client.update(interval, value="0")
+    assert interval["value"] == "0"
+
+    volname = generate_volume_name()
+    volume = client.create_volume(name=volname, size=SIZE, numberOfReplicas=3)
+    lht_host_id = get_self_host_id()
+    volume = common.wait_for_volume_detached(client, volname)
+    assert volume["name"] == volname
+    assert volume["size"] == SIZE
+    assert volume["numberOfReplicas"] == 3
+    assert volume["state"] == "detached"
+
+    volume.attach(hostId=lht_host_id)
+    volume = common.wait_for_volume_healthy(client, volname)
+    bv, b, snap, data = create_backup(client, volname)
+
+    sb_volume_name = "sb-" + volname
+    client.create_volume(name=sb_volume_name, size=SIZE,
+                         numberOfReplicas=2, fromBackup=b['url'],
+                         frontend="", standby=True)
+    sb_volume = common.wait_for_volume_detached(client, sb_volume_name)
+    sb_volume.attach(hostId=lht_host_id)
+    sb_volume = common.wait_for_volume_healthy(client, sb_volume_name)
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -549,6 +549,126 @@ def test_restore_inc(clients, core_api, volume_name, pod):  # NOQA
...
+    assert sb_volume3["lastBackup"] == b["name"]
+
+    activate_standby_volume(client, sb_volume_name)
+    sb_volume = client.by_id_volume(sb_volume_name)
+    sb_volume.attach(hostId=lht_host_id)
+    sb_volume = common.wait_for_volume_healthy(client, sb_volume_name)
+    check_volume_data(sb_volume, data, False)
+
+    # check if the standby volume data still remains the same with the first backup
+    # since backupstore polling has been disabled
```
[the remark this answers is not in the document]
ywei88: If the activation will call additional pulling, I also feel strange about why this can pass.....
I tested it, when we set polling interval to 0, and there is new backup that has not been polled, then we activate that DR volume, it will fail and say "incremental restore has not been finished so can't activate", and then pulling begins. Then we activate again, it will be successful. So I think setting interval to 0 doesn't affect the "activate" feature. And I will write test cases to make sure about that.

ywei88:
I removed the previous test case and added a new one, which is to see if activating a standby volume can still check for the latest backup can pull it when backup polling interval is set to 0

yasker:
@ywei88 Have you figured out why the test passed before?

ywei88:
I'm still trying to figure out...

manager/integration/tests/test_basic.py
```diff
@@ -535,6 +535,69 @@ def test_restore_inc(clients, core_api, volume_name, pod):  # NOQA
         restore_inc_test(client, core_api, volume_name, pod)
 
 
+def test_standby_volume_activation(clients):
```
yasker: Just use `client` if you're not going to use multiple nodes.

manager/integration/tests/test_basic.py
```diff
@@ -535,6 +535,69 @@ def test_restore_inc(clients, core_api, volume_name, pod):  # NOQA
...
+    for _, client in clients.iteritems():
+        break
+
+    interval = client.by_id_setting("backupstore-poll-interval")
```
yasker: Remember to restore the setting after the test completed.

manager/integration/tests/test_basic.py
```diff
@@ -535,6 +535,69 @@ def test_restore_inc(clients, core_api, volume_name, pod):  # NOQA
...
+        if sb_volume["lastBackup"] != b["name"]:
+            time.sleep(RETRY_INTERVAL)
+        else:
+            break
+    assert sb_volume["standby"] is True
+    assert sb_volume["lastBackup"] == b["name"]
+
+    _, b2, _, data2 = create_backup(client, volname)
+    time.sleep(10)
+    assert sb_volume["lastBackup"] == b["name"]
```
yasker: Don't use sleep. You don't know how long you should sleep for. If you want to do that, you need something like https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_basic.py#L576

I assume the field has been updated due to creating the backup. Since the backup polling is disabled, how can lastBackup automatic updated? It seems this doesn't test the pulling before the activate.

yasker:
I will close this PR. It doesn't look like the right way to test the standby volume activate. The only mystery is why previously the test case passed. Keep me updated on that. I will merge the manager PR.

PR #157, opened 2019-06-28
Run flake8 on manager build and configure drone setup

meldafrawi:


yasker:
This depends on the host flake8 version (if it's installed) instead of version going to be used in the longhorn-test container.

yasker:
Also, we need the ability to trigger drone build for this repo as well.

Dockerfile.dapper
```diff
@@ -0,0 +1,15 @@
...
+
+ENV DAPPER_SOURCE /go/src/github.com/longhorn/longhorn-tests
+
+WORKDIR ${DAPPER_SOURCE}
+
+ENV VIRTUAL_ENV=${DAPPER_SOURCE}/manager/integration/.venv
+
+RUN python2 -m virtualenv --python=/usr/bin/python2 ${VIRTUAL_ENV}
+
+CMD ["./manager/integration/scripts/flake8-check.sh"]
```
yasker:
```
/usr/bin/docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"./manager/integration/scripts/flake8-check.sh\": permission denied": unknown.
```

Seems flake8-check.sh is not executable.

meldafrawi: `flake8-check.sh` is now executable

manager/integration/scripts/flake8-check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
yasker: Trailing space

manager/integration/scripts/flake8-check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash 
+
+source manager/integration/.venv/bin/activate
```
yasker: Trailing space

manager/integration/scripts/flake8-check.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash 
+
+source manager/integration/.venv/bin/activate 
+
+pip install -r manager/integration/tests/requirements.txt
+
+flake8 manager/integration/tests
```
yasker: flake8 has no output, add `echo flake8 check succeed` at the end?

meldafrawi: Done

PR #161, opened 2019-07-03
manager: Update deployment to longhornio/longhorn-manager-test:master

meldafrawi:
I think it's better to use `master` tag to deploy `longhorn-manager-test` image, since it will be updated on every merge, and it will be ideal for test_framework to use.


yasker:
In fact, we cannot. Every time when we release Longhorn, all the deployment tag will be sync to the version tag. Also, we are going to keep this tag in sync with Longhorn manager tag. We need to make an adjustment in the test-framework part instead.

PR #170, opened 2019-07-16
Add testcase for backup volume deletion, and function to wait for backup completion

ywei88:
https://github.com/longhorn/longhorn/issues/598

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -194,6 +194,8 @@ def create_backup(client, volname, data={}, labels={}):
     # have added extra Labels (for things like BaseImage).
     for key, val in labels.iteritems():
         assert new_b["labels"].get(key) == val
+
+    volume = wait_for_backup_completion(client, volname, snap["name"])
     volume = wait_for_volume_status(client, volname,
                                     "lastBackup",
                                     b["name"])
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -752,6 +752,47 @@ def restore_inc_test(client, core_api, volume_name, pod):  # NOQA
...
+    volName = generate_volume_name()
+    volume = create_and_check_volume(client, volName)
+
+    volume.attach(hostId=lht_hostId)
+    volume = common.wait_for_volume_healthy(client, volName)
+
+    bv, _, snap1, _ = create_backup(client, volName)
+    _, _, snap2, _ = create_backup(client, volName)
+
+    found = False
```
yasker: Do we have a function to wait for the backup completion?

ywei88: I think no

yasker: Then we can add one.

manager/integration/tests/common.py
```diff
@@ -194,6 +194,8 @@ def create_backup(client, volname, data={}, labels={}):
     # have added extra Labels (for things like BaseImage).
     for key, val in labels.iteritems():
         assert new_b["labels"].get(key) == val
+
+    volume = wait_for_backup_completion(client, volname, snap["name"])
```
yasker: I think we should have code removed because of adding this function?

ywei88: I checked all the places where create_backup is called, and there is no redundant logic except the place we talked before.

manager/integration/tests/test_basic.py
```diff
@@ -784,6 +784,7 @@ def test_deleting_backup_volume(clients):  # NOQA
     bv = client.by_id_backupVolume(volName)
     client.delete(bv)
     common.wait_for_backup_volume_delete(client, volName)
+
```
yasker: extra blank line.

PR #177, opened 2019-07-19
Add offline node test scenario to manager integration test

meldafrawi:
longhorn/longhorn#635

### DIFF
--- a/manager/integration/tests/digitalocean.py
+++ b/manager/integration/tests/digitalocean.py
@@ -1,0 +1,102 @@
+import os
+import requests
+import signal
+import time
+
+from cloudprovider import cloudprovider
+
+
+class digitalocean(cloudprovider):
+    DO_API_TOKEN = os.getenv("ENV_DO_API_TOKEN")
+    DO_API_URL_BASE = "https://api.digitalocean.com/v2/"
+    DO_NODE_TAG = "longhorn-tests"
+    DO_REQ_HEADERS = {'Content-Type': 'application/json',
+                      'Authorization': 'Bearer {0}'.format(DO_API_TOKEN)}
+    DO_API_TIMEOUT_SEC = 120
+
+    def timeout_handler(self, signum, frame):
+        raise Exception("Err: Digitalocean API timed out")
+
+    def is_api_token_defined(self):
+        if self.DO_API_TOKEN is None or self.DO_API_TOKEN == "":
+            print "Err: ENV_DO_API_TOKEN is not defined"
+            return False
+        else:
+            return True
+
+    def node_id(self, node_name):
+        assert self.is_api_token_defined()
+
+        api_url = '{0}droplets'.format(self.DO_API_URL_BASE)
+
+        payload = {'tag_name': self.DO_NODE_TAG}
+
+        do_api_response = \
+            requests.get(api_url, params=payload, headers=self.DO_REQ_HEADERS)
+
+        do_nodes = do_api_response.json()
+
+        for do_droplet in do_nodes['droplets']:
+            if do_droplet['name'] == node_name:
+                return do_droplet['id']
+
+    def node_status(self, node_id):
+        assert self.is_api_token_defined()
+
+        api_url = '{0}droplets'.format(self.DO_API_URL_BASE)
+
+        payload = {'tag_name': self.DO_NODE_TAG}
+
+        do_api_response = requests.get(api_url,
+                                       params=payload,
+                                       headers=self.DO_REQ_HEADERS)
+
+        do_nodes = do_api_response.json()
+
+        for do_droplet in do_nodes['droplets']:
+            if do_droplet['id'] == node_id:
+                return do_droplet['status']
+
+    def droplet_exec_action(self,
+                            droplet_id,
+                            action,
+                            expected_status,
+                            desired_status):
+        assert self.is_api_token_defined()
+
+        api_url = '{0}droplets/{1}/actions'.format(self.DO_API_URL_BASE,
+                                                   droplet_id)
+
+        payload = {'type': action}
+
+        droplet_current_status = self.node_status(droplet_id)
+
+        if droplet_current_status == expected_status:
+            requests.post(api_url, headers=self.DO_REQ_HEADERS, json=payload)
+
+            signal.signal(signal.SIGALRM, self.timeout_handler)
+            signal.alarm(self.DO_API_TIMEOUT_SEC)
+
+            try:
+                while self.node_status(droplet_id) == expected_status:
+                    time.sleep(1)
+            except Exception, exc:
+                print exc
+                return False
+
+            return True
+
+        elif droplet_current_status == desired_status:
+            return True
+
+    def node_start(self, node_id):
+        self.droplet_exec_action(droplet_id=node_id,
+                                 action="power_on",
+                                 expected_status="off",
+                                 desired_status="active")
+
+    def node_shutdown(self, node_id):
+        self.droplet_exec_action(droplet_id=node_id,
+                                 action="shutdown",
+                                 expected_status="active",
+                                 desired_status="off")
--- a/manager/integration/tests/test_infra.py
+++ b/manager/integration/tests/test_infra.py
@@ -1,0 +1,161 @@
+import os
+import pytest
+import time
+
+from digitalocean import digitalocean
+from common import get_core_api_client, get_longhorn_api_client
+from common import RETRY_COUNTS, RETRY_INTERVAL
+
+
+def detect_cloudprovider():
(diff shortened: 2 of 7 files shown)


manager/integration/tests/test_node.py
```diff
@@ -1160,3 +1163,37 @@ def test_node_default_disk_labeled(client, core_api, random_disk_path,  reset_de
...
+                node_item.metadata.annotations['rke.cattle.io/external-ip']
+            break
+
+    droplet_id = digitalocean.get_droplet_id(node_external_ip)
+
+    digitalocean.droplet_exec_action(droplet_id,
+                                     action="shutdown",
+                                     expected_status="active",
+                                     desired_status="off")
+    time.sleep(30)
```
yasker: We need to query the API for the state instead of blindly waiting.

manager/integration/tests/digitalocean.py
```diff
@@ -0,0 +1,93 @@
...
+    if DO_API_TOKEN is None or DO_API_TOKEN == "":
+        print "Err: ENV_DO_API_TOKEN is not defined"
+        return False
+    else:
+        return True
+
+
+def get_droplet_id(node_ipaddr):
+    if is_api_token_defined() is False:
+        exit(1)
```
yasker: `assert is_api_token_defined()` or something can show us why it failed.

manager/integration/tests/test_node.py
```diff
@@ -1170,6 +1170,7 @@ def test_offline_node():
 
     k8s_api_client = get_core_api_client()
     longhorn_api_client = get_longhorn_api_client()
+    cloudprovider = digitalocean()
```
yasker: Feel we should able to implement it in a better way, so we can choose cloud provider using the environmental setting.

manager/integration/tests/test_node.py
meldafrawi: I've added 30 seconds wait here just to make sure the shutdown status gets reflected back on k8s, and hence longhorn API.

yasker: Yeah, that's what I meant. We need to query the API. We don't really know how long it would take for Kubernetes to change state under different circumstances.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,69 @@
...
+            if node_external_ip == longhorn_test_node_ip:
+                continue
+            else:
+                break
+
+    node = cloudprovider.node_id(node_external_ip)
+
+    cloudprovider.node_shutdown(node)
+
+    while True:
```
yasker: We cannot wait in an infinite loop, and we should have sleep between retrying otherwise it can be too many API calls. We should have a retry interval and maximum retry count. See https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/common.py#L1114 for examples.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,69 @@
...
+        for node_condition in node_status_k8s.status.conditions:
+            if node_condition.type == "Ready" and \
+               node_condition.status != "True":
+                break
+            else:
+                continue
+        break
+
+    # wait 30 seconds before querying node status in longhorn
+    time.sleep(30)
```
yasker: Same here.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,69 @@
...
+                    label_selector=pod_lable_selector).items:
+
+        if pod.metadata.name == "longhorn-test":
+            longhorn_test_node_ip = pod.status.host_ip
+
+    for node_item in k8s_api_client.list_node().items:
+        if node_worker_label in node_item.metadata.labels and \
+                node_item.metadata.labels[node_worker_label] == 'true':
+            node_name = node_item.metadata.name
+            node_external_ip = \
```
yasker: Why do we use IP as node ID? I thought the node name is a natural choice?

Also, this one used an RKE specific label. We may not want that since that limits what setup this test can run.

meldafrawi: - I rely on node IP to filter digital ocean nodes, and get node's unique ID. as node names can be redundant.

yasker: We can require the test environment to have different node names. It doesn't make sense to have the same node name in the production environment as well. It will make things easier.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,77 @@
...
+               node_condition.status != "True":
+                node_down = True
+                break
+
+        if node_down is True:
+            break
+        else:
+            time.sleep(RETRY_INTERVAL)
+            continue
+
```
yasker: missing `assert node_down` here.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,77 @@
...
+        if offline_node.conditions.Ready.status == 'False':
+            longhorn_node_down = True
+            break
+        else:
+            time.sleep(RETRY_INTERVAL)
+            continue
+
+    assert longhorn_node_down
+
+    cloudprovider.node_start(node)
```
yasker: Is it OK to run other test cases immediately after this? Should we wait for the node to be back before continuing?

It can also be problematic if something failed in this test which may result in other test failures. We can use `cloud_provider` as a fixture and add cleanup method in it to make sure everything is up even test failure happened.

test_framework/templates/3-nodes-k8s.yml
```diff
@@ -1,14 +1,18 @@
 kubernetes_version: v1.13.5-rancher1-2
 nodes:
   - address: K8S_CONTROLLER_IP
+    hostname_override: longhorn-tests-00
```
yasker: It's better for those to be replaced (as in a template) while running the script instead of hard coding. Otherwise, if we want more nodes or a different name defined in cloud-providers this will fail.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,129 @@
...
+            if node_name == longhorn_test_node_name:
+                continue
+            else:
+                break
+
+    node = cloudprovider.node_id(node_name)
+
+    cloudprovider.node_shutdown(node)
+
+    for i in range(RETRY_COUNTS):
```
yasker: We will need to use similar functions many times later to wait for the node to start/stop. We can write the `wait_for_node_down` function follow the pattern of e.g. https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/common.py#L1086 to make life easier.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,129 @@
...
+from common import RETRY_COUNTS, RETRY_INTERVAL
+
+
+def detect_cloudprovider():
+    cloudprovider_name = os.getenv("CLOUDPROVIDER")
+
+    if cloudprovider_name == "digitalocean":
+        cloudprovider = digitalocean()
+        return cloudprovider
+
```
yasker: `assert False` if we cannot find a suitable cloud provider.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,129 @@
...
+        else:
+            node_ready = False
+
+    return node_ready
+
+
+def is_node_ready_longhorn(node_name):
+    longhorn_api_client = get_longhorn_api_client()
+
+    for longhorn_node in longhorn_api_client.list_node():
```
yasker: We should able to get the node by name? Shouldn't need a list.

Also, if the node is not found, we should error out.

manager/integration/tests/test_infra.py
```diff
@@ -0,0 +1,129 @@
...
+        if node_worker_label in node_item.metadata.labels and \
+                node_item.metadata.labels[node_worker_label] == 'true':
+            node_name = node_item.metadata.name
+
+            if is_node_ready_k8s(node_name) is False:
+                node = cloudprovider.node_id(node_name)
+
+                cloudprovider.node_start(node)
+
+                for i in range(RETRY_COUNTS):
```
yasker: We can implement `wait_for_node_up()` here.

meldafrawi: done

manager/integration/tests/test_infra.py
meldafrawi: done

manager/integration/tests/test_infra.py
meldafrawi: done

manager/integration/tests/test_infra.py
yasker: Unresolved.

meldafrawi: done

PR #178, opened 2019-07-23
Test Attaching volume without frontend

shuo-wu:
https://github.com/longhorn/longhorn/issues/629


yasker:
https://github.com/longhorn/longhorn/issues/629

PR #179, opened 2019-07-24
Restore Progress and Snapshot revert

shreesha21:
The following code changes adds two things:
1. Adds a way to check the restoration progress percentage
2. Fixes the issue where snapshot revert was getting executed when the volume frontend is attached


manager/integration/tests/test_basic.py
```diff
@@ -584,6 +584,9 @@ def restore_inc_test(client, core_api, volume_name, pod):  # NOQA
     client.create_volume(name=sb_volume2_name, size=SIZE,
                          numberOfReplicas=2, fromBackup=backup0['url'],
                          frontend="", standby=True)
+    common.wait_for_volume_restoration_completed(client, sb_volume0_name)
+    common.wait_for_volume_restoration_completed(client, sb_volume0_name)
```
yasker: ? Same volume?

PR #187, opened 2019-08-03
Fixing unmount delete test failure

shreesha21:
After the longhorn refactor, this case was failing as Linux throws error
when the test case was trying to unmount the mount path when an active
volume has a replica running and storing meta files in this mount path.
The following code changes fixes it using -l flag to accomplish the
motive that the test case was trying to achieve.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1503,6 +1503,11 @@ def umount_disk(mount_path):
     subprocess.check_call(cmd)
 
 
+def lazy_umount_disk(mount_path):
+    cmd = ['umount', '-l', mount_path]
+    subprocess.check_call(cmd)
+
+
 def cleanup_host_disk(vol_name):
     mount_path = os.path.join(DIRECTORY_PATH, vol_name)
     umount_disk(mount_path)
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -878,7 +878,12 @@ def test_node_delete_umount_disks(client):  # NOQA
 
     # umount the disk
     mount_path = os.path.join(DIRECTORY_PATH, disk_volume_name)
-    common.umount_disk(mount_path)
+    # After longhorn refactor, umount_disk will fail with
+    # `target is busy` error from Linux as replica is using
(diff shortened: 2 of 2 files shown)


yasker:
If failed to umount due to Longhorn is actively using the disk, we still need to understand why it works before.

shreesha21:
This was working before because replica was running as a docker container, not as a regular linux process. As a result, there was an internal volume binding created for replica container by the docker as shown below:
```
"HostConfig": {
            "Binds": [
                "/tmp/longhorn-test/testdisk/replicas/test2-edbc5e08:/volume",
                 ...
            ]
}
"Mounts": [
            {
                "Type": "bind",
                "Source": "/tmp/longhorn-test/testdisk/replicas/test2-edbc5e08",
                "Destination": "/volume",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
            },
```

However, when I tried to do unmount docker command, I still see the same error:
```
root@mas1# docker exec 5f6002acacdb umount /volume
umount: /volume: target is busy
        (In some cases useful info about processes that
         use the device is found by lsof(8) or fuser(1).)
```
But when you unmount the actual path as a linux command (and not a docker command), I believe linux is agnostic of this path being re-mounted by the docker container and ends up un-mounting it.
I also think it relies on the fact that this mount path is not mentioned the ps output of the replica process:
```
v0.5.0:
Mount path: /tmp/longhorn-test/testdisk
root@mas1:/tmp/longhorn-test/testdisk/replicas/test2-edbc5e08# ps ax | grep replica
18359 ?        Ssl    0:03 longhorn replica --listen 0.0.0.0:9502 --size 21474836480 /volume
```

```
After Refactor:
Mount path: /tmp/longhorn-test/testdisk
 3771 ?        Sl     0:00 /usr/local/bin/longhorn replica /host/tmp/longhorn-test/testdisk/replicas/diskvol-36396a63 --size 2147483648 --listen 0.0.0.0:10015
```

yasker:
Make sense. Sounds like a bug for Linux before the refactoring, since `umount` should return the same result if the target was busy even the path is bind-mounted instead of directly mounted in the same namespace.

PR #190, opened 2019-08-21
Add stress test

meldafrawi:
**Updated**

longhorn/longhorn#679

Stress test actions:
- [x] Stress test job will, by default, create 10 test pods running in an infinite loop doing the following,
  1 - Create Longhorn volume, pv, pvc and a pod.
  2- Write random data with size **1 GB** to the volume, Validate the data and Make a snapshot (x2)
  3- Overwrite written data and Make a snapshot.
  4- Delete/purge snapshot (random)
  5- Choose 3 random actions from:
      - Write random data with size 1 GB
      - Delete data
      - Create a snapshot
      - Delete random snapshot / purge the snapshot randomly.
- [x]  Job keeps running till all pods are `completed`, in this case if test `failed`.

### DIFF
--- a/manager/integration/Dockerfile
+++ b/manager/integration/Dockerfile
@@ -1,7 +1,13 @@
 FROM python:2.7.14-stretch
 
+ARG KUBECTL_VERSION=v1.15.0
+
 RUN apt-get update && \
     apt-get install -y vim-tiny nfs-common
+
+RUN wget -q https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_VERSION/bin/linux/amd64/kubectl && \
+    mv kubectl /usr/local/bin/kubectl && \
+    chmod +x /usr/local/bin/kubectl
 
 ADD tests/requirements.txt .
 RUN pip install -r requirements.txt
--- a/manager/integration/tests/test_stress.py
+++ b/manager/integration/tests/test_stress.py
@@ -1,0 +1,271 @@
+import atexit
+import hashlib
+import os
+import pytest
+import subprocess
+import random
+import string
+
+from common import create_and_check_volume
+from common import create_and_wait_pod
+from common import create_pvc_for_volume
+from common import create_pv_for_volume
+from common import delete_and_wait_longhorn
+from common import delete_and_wait_pod
+from common import delete_and_wait_pv
+from common import delete_and_wait_pvc
+from common import delete_and_wait_volume_attachment
+from common import generate_pod_with_pvc_manifest
+from common import generate_random_data
+from common import get_core_api_client
+from common import get_longhorn_api_client
+from common import get_storage_api_client
+from common import Gi
+from common import read_volume_data
+from common import wait_for_snapshot_purge
+from common import wait_for_volume_detached
+from common import write_pod_volume_data
+from kubernetes.stream import stream
+from random import randrange
+
+
+NPODS = os.getenv("STRESS_TEST_NPODS")
+
+if NPODS is None:
+    NPODS = 1
+else:
+    NPODS = int(NPODS)
+
+count = [str(i) for i in range(NPODS)]
+
+
+STRESS_POD_NAME_PREFIX = "stress-test-pod-"
+STRESS_PVC_NAME_PREFIX = "stress-test-pvc-"
+STRESS_PV_NAME_PREFIX = "stress-test-pv-"
+STRESS_VOLUME_NAME_PREFIX = "stress-test-volume-"
+
+VOLUME_SIZE = str(2 * Gi)
+TEST_DATA_BYTES = 1 * Gi
+
+READ_MD5SUM_TIMEOUT = 60
+
+
+def get_random_suffix():
+    return ''.join(random.choice(string.ascii_lowercase + string.digits)
+                   for _ in range(6))
+
+
+def get_md5sum(file_path):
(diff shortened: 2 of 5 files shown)


manager/integration/tests/test_stress.py
```diff
@@ -0,0 +1,316 @@
...
+           "storageClassName": "longhorn",
+           "volumeMode": "Filesystem",
+           "volumeName": pv_name
+        }
+    }
+
+    return pvc_manifest
+
+
+def create_and_wait_pvc_pod(k8s_api_client, pod_manifest):
```
shuo-wu: Why not using `create_and_wait_pod()`?

manager/integration/tests/test_stress.py
```diff
@@ -0,0 +1,316 @@
...
+
+    wait_for_volume_detached(longhorn_api_client, volume_name)
+
+    pv_manifest = generate_pv_maifest_for_longhorn_volume(
+        pv_name,
+        volume_name,
+        VOLUME_SIZE
+    )
+
+    create_and_wait_pv(k8s_api_client, pv_manifest)
```
shuo-wu: Can we use `create_pv_for_volume()`? The storage class will be `longhorn-static` if you use this function. But I think it doesn't matter since you will manually clean up those components.

The workflow here is similar to `test_kubernetes_status()`, I think you can refer to that test and avoid redundant codes. Same below.

manager/integration/tests/test_stress.py
```diff
@@ -0,0 +1,316 @@
...
+    if trigger_purge == 1:
+        purge_random_snapshot(longhorn_api_client, volume_name, snapshot_name)
+
+
+def test_stress(generate_load):
+    pass
+
+
+@pytest.fixture(params=count)
+def generate_load(request):
```
shuo-wu: Will this fixture function be run sequentially? It means we are not operating multiple volumes simultaneously.

meldafrawi: No, This is actually a workaround to run the test in parallel.
I've added `pytest-xdist` to `requirements.txt` for this reason.
running the stress test in parallel will be done using this command: 
`py.test -n NUM -s test_stress.py`, where NUM is the number of threads that you want to start in parallel.

manager/integration/tests/test_stress.py
meldafrawi: fixed

manager/integration/tests/test_stress.py
meldafrawi: fixed

manager/integration/tests/test_stress.py
meldafrawi: - When you set `NPODS` to a number ,  count is created as a list of integers from `0` to `NPODS-1`.
- `generate_load` is a parameterized fixture with `params=count`, that means that this function will be executed for every value in count list.
- The value of count is passed to generate_load  in `request.param` which i use as an suffix for Volume/PV/PVC/POD names to ensure no duplicate names.
- with `pytest-xdist` we're able to run all these tests together to create a stress test.

manager/integration/tests/test_stress.py
```diff
@@ -0,0 +1,254 @@
...
+STRESS_POD_NAME_PREFIX = "stress-test-pod-"
+STRESS_PVC_NAME_PREFIX = "stress-test-pvc-"
+STRESS_PV_NAME_PREFIX = "stress-test-pv-"
+STRESS_VOLUME_NAME_PREFIX = "stress-test-volume-"
+
+VOLUME_SIZE = str(512 * Mi)
+TEST_DATA_BYTES = 102400
+
+
+def generate_pod_with_pvc_manifest(pod_name, pvc_name):
```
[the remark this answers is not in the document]
meldafrawi: done

manager/integration/Dockerfile
```diff
@@ -1,7 +1,12 @@
 FROM python:2.7.14-stretch
 
+ARG KUBECTL_VERSION=v1.15.0
+
 RUN apt-get update && \
-    apt-get install -y vim-tiny nfs-common
+    apt-get install -y vim-tiny nfs-common && \
+    wget -q https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_VERSION/bin/linux/amd64/kubectl && \
```
yasker: This doesn't belong to the `apt-get` lines. Add a new step instead.

manager/integration/tests/test_stress.py
```diff
@@ -0,0 +1,264 @@
...
+        elif action == 1:
+            delete_data(k8s_api_client, pod_name)
+        elif action == 2:
+            create_snapshot(longhorn_api_client, volume_name)
+        elif action == 3:
+            delete_random_snapshot(longhorn_api_client, volume_name)
+
+
+@pytest.mark.stress
+def test_reset_env():
```
yasker: It's a bit unsafe to depends on the order of the test running. Do we have other ways to do it?

manager/integration/Dockerfile
meldafrawi: fixed

manager/integration/tests/test_stress.py
meldafrawi: I've kept `reset_env` test in case stress environment cleanup is needed.

PR #196, opened 2019-08-30
test framework: add stress tests to nightly tests

meldafrawi:
depend on PR #190

### DIFF
--- a/manager/integration/tests/run.sh
+++ b/manager/integration/tests/run.sh
@@ -6,6 +6,10 @@ if [[ $@ =~ "--junitxml=" ]] ; then
   py.test -v . $@  >> /tmp/longhorn-pytest 2>&1
   
   cat ${LONGHORN_JUNIT_REPORT_PATH}
+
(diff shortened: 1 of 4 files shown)


manager/integration/tests/run.sh
```diff
@@ -6,6 +6,10 @@ if [[ $@ =~ "--junitxml=" ]] ; then
   py.test -v . $@  >> /tmp/longhorn-pytest 2>&1
   
   cat ${LONGHORN_JUNIT_REPORT_PATH}
+
+elif [[ $@ =~ "test_stress" ]] ; then
```
yasker: Is it possible to run the stress test with other tests? Not sure why we need a separate way to run it. It's for a small scale and should pass every night.

meldafrawi: We need to run stress test only in parallel with `pytest-xdist`, so we have to run the test in separate step.

yasker:
It seems the stress test's load is too much for the normal nightly test setup. We won't able to run nightly test with it for now.

PR #198, opened 2019-09-04
Add and use func delete_backup()

shuo-wu:
https://github.com/longhorn/longhorn/issues/709
https://github.com/longhorn/longhorn/issues/710

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -206,6 +206,28 @@ def create_backup(client, volname, data={}, labels={}):
     return bv, b, snap, data
 
 
+def delete_backup(backup_volume, backup_name):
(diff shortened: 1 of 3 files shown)


manager/integration/tests/common.py
```diff
@@ -206,6 +206,25 @@ def create_backup(client, volname, data={}, labels={}):
     return bv, b, snap, data
 
 
+def delete_backup(backup_volume, backup_name):
+    backup_volume.backupDelete(name=backup_name)
+    found = False
+    for i in range(RETRY_COMMAND_COUNT):
+        time.sleep(RETRY_INTERVAL)
```
yasker: Normally we sleep after the first round since it's quicker.

PR #200, opened 2019-09-06
add fromBackup for StorageClass & PV creation test

meldafrawi:
longhorn/longhorn#680


manager/integration/tests/test_basic.py
```diff
@@ -1136,3 +1149,137 @@ def test_attach_without_frontend(clients, volume_name):  # NOQA
...
+
+    volume_id = client.by_id_volume(volume_name)
+    snapshot = volume_id.snapshotCreate()
+
+    volume_id.snapshotBackup(name=snapshot['name'])
+
+    bv, b = find_backup(client, volume_name, snapshot['name'])
+
+    wait_for_backup_completion(client, volume_name, snapshot['name'])
+    status = wait_for_volume_status(client,
```
yasker: Seems we don't need to wait for this volume's `LastBackup` to update. Backup is usable once it's done. It may save us some time.

meldafrawi: fixed.

PR #203, opened 2019-09-09
fix fromBackup StorageClass & PV creation test

meldafrawi:
longhorn/longhorn#680

  - allow test pod to get pvc status
  - fix missing fixtures imports
  - remove delete_and_wait_pvc & delete_and_wait_pv
    since it will not be deleted before pods get deleted first

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -10,7 +10,7 @@ metadata:
   name: longhorn-test-role
 rules:
 - apiGroups: [""]
-  resources: ["nodes", "nodes/status", "pods", "pods/exec", "persistentvolumes", "persistentvolumeclaims", "secrets"]
+  resources: ["nodes", "nodes/status", "pods", "pods/exec", "persistentvolumes", "persistentvolumeclaims", "persistentvolumeclaims/status", "secrets"]
   verbs: ["*"]
 - apiGroups: ["storage.k8s.io"]
   resources: ["storageclasses"]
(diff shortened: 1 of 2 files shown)


manager/integration/deploy/test.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
   name: longhorn-test-role
 rules:
 - apiGroups: [""]
-  resources: ["nodes", "nodes/status", "pods", "pods/exec", "persistentvolumes", "persistentvolumeclaims", "secrets"]
+  resources: ["nodes", "nodes/status", "pods", "pods/exec", "persistentvolumes", "persistentvolumeclaims", "persistentvolumeclaims/status", "secrets"]
```
yasker: Why we need this?

meldafrawi: i use `read_namespaced_persistent_volume_claim_status` check if the new PVC status is `Bound`

https://github.com/longhorn/longhorn-tests/blob/b3ec4e9e64ce60608d90db638d6116a92232f258/manager/integration/tests/test_basic.py#L1240

manager/integration/tests/test_basic.py
```diff
@@ -1267,7 +1273,3 @@ def test_storage_class_from_backup(volume_name, pvc_name, storage_class, client,
     restored_data = read_volume_data(core_api, backup_pod_name)
 
     assert test_data == restored_data
-
-    delete_and_wait_pvc(core_api, pvc_name)
```
yasker: The reason for we don't need to wait for those?

meldafrawi: PVs & PVCs will not get deleted before the pods get removed first.
pod removal will happen after the test is finished, since we're using `pod_make` fixture.

PR #212, opened 2019-09-18
Check engine image state in toleration setting test

shuo-wu:
https://github.com/longhorn/longhorn/issues/734

### DIFF
--- a/manager/integration/tests/test_toleration_setting.py
+++ b/manager/integration/tests/test_toleration_setting.py
@@ -127,6 +127,14 @@ def wait_for_toleration_update(core_api, apps_api, count, set_tolerations):  # N
             time.sleep(RETRY_INTERVAL_LONG)
             continue
 
+        client = get_longhorn_api_client()
+        images = client.list_engine_image()
+        assert len(images) == 1
+        if images[0]["state"] != "ready":
+            updated = False
+            time.sleep(RETRY_INTERVAL_LONG)
+            continue
+
         break
 
     assert updated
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_toleration_setting.py
```diff
@@ -127,6 +127,14 @@ def wait_for_toleration_update(core_api, apps_api, count, set_tolerations):  # N
             time.sleep(RETRY_INTERVAL_LONG)
             continue
 
+        client = get_longhorn_api_client()
+        images = client.list_engine_image()
+        assert len(images) == 1
+        if images[0]["state"] != "ready":
+            updated = False
+            time.sleep(RETRY_INTERVAL_LONG)
+            continue
```
yasker: ??

yasker: It's confusing. Why add `time.sleep(RETRY_INTERVAL_LONG)` before every `continue` command? Is there a better way?

yasker: And there is a `continue` seems misaligned on line 116

shuo-wu: Puting `time.sleep(RETRY_INTERVAL_LONG)` at the top of the loop can avoid the case. But it looks weird.

yasker: I think it's OK to put it on the top. One extra second is not a big deal in the test.

PR #218, opened 2019-09-20
fix wait_for_snapshot_purge

meldafrawi:

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1358,16 +1358,21 @@ def wait_for_snapshot_purge(client, volume_name, *snaps):
     assert completed == len(purge_status)
 
     # Now that the purge has been reported to be completed, the Snapshots
-    # should be gone.
+    # should should be removed or "marked as removed" in the case of
+    # the latest snapshot.
     found = False
     snapshots = v.snapshotList(volume=volume_name)
-    snap_list = []
-    for snap in snapshots:
-        snap_list.append(snap.name)
+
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -1217,16 +1217,17 @@ def wait_for_snapshot_purge(client, volume_name, *snaps):
     assert completed == len(purge_status)
 
     # Now that the purge has been reported to be completed, the Snapshots
-    # should be gone.
+    # should be marked as removed or exist as last snapshot.
```
yasker: `should be removed or "marked as removed" in the cast of the latest snapshot`

manager/integration/tests/common.py
```diff
@@ -1217,16 +1217,17 @@ def wait_for_snapshot_purge(client, volume_name, *snaps):
...
-    for snap in snapshots:
-        snap_list.append(snap["name"])
+
     for snap in snaps:
-        if snap in snap_list:
-            found = True
-            break
+        for vs in snapshots.data:
+            if snap == vs["name"]:
+                if vs["removed"] is False:
```
yasker: The logic here is wrong. The latest snapshot (with `volume-head` as a child) needs to be marked as `removed`. All other snapshot should be removed already.

We have another case that a snapshot cannot be immediately removed, but that involves a snapshot tree.

yasker:
LGTM.

meldafrawi:
Changes already merged in #256

PR #219, opened 2019-09-21
Disable SETTING_CREATE_DEFAULT_DISK_LABELED_NODES before cleanup

shuo-wu:
https://github.com/longhorn/longhorn/issues/747


manager/integration/tests/test_node.py
```diff
@@ -54,9 +54,9 @@ def reset_default_disk_label():
 def reset_disk_settings():
     yield
     api = get_longhorn_api_client()
-    setting = api.by_id_setting(CREATE_DEFAULT_DISK_SETTING)
+    setting = api.by_id_setting(SETTING_DEFAULT_DATA_PATH)
```
yasker: Wrong setting.

PR #221, opened 2019-09-25
add wait to create snapshot function

meldafrawi:

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -179,13 +179,13 @@ def cleanup_volume(client, volume):
 
 def create_backup(client, volname, data={}, labels={}):
     volume = client.by_id_volume(volname)
-    volume.snapshotCreate()
+    create_snapshot(client, volname)
     if not data:
         data = write_volume_random_data(volume)
     else:
         data = write_volume_data(volume, data)
-    snap = volume.snapshotCreate()
-    volume.snapshotCreate()
+    snap = create_snapshot(client, volname)
+    create_snapshot(client, volname)
     volume.snapshotBackup(name=snap["name"], labels=labels)
 
     bv, b = find_backup(client, volname, snap["name"])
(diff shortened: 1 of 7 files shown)


manager/integration/tests/common.py
```diff
@@ -2433,3 +2433,24 @@ def wait_for_engine_image_deletion(client, core_api, engine_image_name):
...
     assert deleted
+
+
+def create_snapshot(longhorn_api_client, volume_name):
+    volume = longhorn_api_client.by_id_volume(volume_name)
+
+    snapshots = volume.snapshotList(volume=volume_name)
+    snapshots_count = len(snapshots.data)
+
+    volume.snapshotCreate()
```
yasker: snapshotCreate() should return the name of the snapshot as a part of response, which can be used to determine if the snapshot has been created or not.

manager/integration/tests/common.py
```diff
@@ -2433,3 +2433,24 @@ def wait_for_engine_image_deletion(client, core_api, engine_image_name):
...
+    for i in range(RETRY_COUNTS):
+        snapshots = volume.snapshotList(volume=volume_name)
+
+        if len(snapshots) == snapshots_count + 1:
+            snapshot_created = True
+            break
+        time.sleep(RETRY_INTERVAL)
+
+    assert snapshot_created
+    return snapshots
```
yasker: Doesn't make sense for the `create_snapshot` to return a list of snapshots. Make more sense to only return the newly created snapshot's name.

yasker:
We should also replace the existing `volume.snapshotCreate()` call with this call.

manager/integration/tests/common.py
meldafrawi: fixed

manager/integration/tests/common.py
meldafrawi: fixed

yasker:
LGTM. Need to fix the conflict.

meldafrawi:
conflict resolved.

PR #225, opened 2019-10-03
Avoid checking endpoint for disaster recovery volumes

shuo-wu:
There is no assertion after waiting for the endpoint of dr volumes hence the test won't error out. But we need to fix it to avoid meaningless waiting.


yasker:
Should we add assert in `wait_for_volume_endpoint`?

PR #228, opened 2019-10-04
Add process crash case for ha_salvage_test()

shuo-wu:
longhorn/longhorn#789

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1350,6 +1350,48 @@ def delete_replica_processes(client, api, volname):
             assert deleted
 
 
+def crash_replica_processes(client, api, volname):
+    volume = client.by_id_volume(volname)
+    for r in volume["replicas"]:
+        kill_command = [
+            '/bin/sh', '-c',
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -1350,6 +1350,52 @@ def delete_replica_processes(client, api, volname):
...
+    replica_managers = []
+
+    volume = client.by_id_volume(volname)
+    for r in volume["replicas"]:
+        replica_managers.append(r["instanceManagerName"])
+
+    kill_command = [
+        '/bin/sh', '-c',
+        "kill `ps aux | " +
+        "grep 'longhorn replica' | grep -v grep | awk '{print $2}'`"
```
yasker: This would kill all the replicas in the instance manager?

shuo-wu: In this case, there should be only one `attached` volume and each replica manager contains only one replica process

yasker: This will break the stress test since we're running the test for multiple volumes at once.

PR #231, opened 2019-10-16
Upgrade integration tests to python 3

meldafrawi:
https://github.com/longhorn/longhorn/issues/641

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -173,9 +173,9 @@ def cleanup_volume(client, volume):
     :param volume: The volume to clean up.
     """
     volume.detach()
-    volume = wait_for_volume_detached(client, volume["name"])
+    volume = wait_for_volume_detached(client, volume.name)
     client.delete(volume)
-    wait_for_volume_delete(client, volume["name"])
+    wait_for_volume_delete(client, volume.name)
     volumes = client.list_volume()
     assert len(volumes) == 0
 
@@ -189,29 +189,29 @@ def create_backup(client, volname, data={}, labels={}):
         data = write_volume_data(volume, data)
     snap = create_snapshot(client, volname)
     create_snapshot(client, volname)
-    volume.snapshotBackup(name=snap["name"], labels=labels)
+    volume.snapshotBackup(name=snap.name, labels=labels)
 
-    bv, b = find_backup(client, volname, snap["name"])
+    bv, b = find_backup(client, volname, snap.name)
 
-    new_b = bv.backupGet(name=b["name"])
-    assert new_b["name"] == b["name"]
-    assert new_b["url"] == b["url"]
-    assert new_b["snapshotName"] == b["snapshotName"]
-    assert new_b["snapshotCreated"] == b["snapshotCreated"]
-    assert new_b["created"] == b["created"]
-    assert new_b["volumeName"] == b["volumeName"]
-    assert new_b["volumeSize"] == b["volumeSize"]
-    assert new_b["volumeCreated"] == b["volumeCreated"]
+    new_b = bv.backupGet(name=b.name)
+    assert new_b.name == b.name
+    assert new_b.url == b.url
+    assert new_b.snapshotName == b.snapshotName
+    assert new_b.snapshotCreated == b.snapshotCreated
+    assert new_b.created == b.created
+    assert new_b.volumeName == b.volumeName
+    assert new_b.volumeSize == b.volumeSize
+    assert new_b.volumeCreated == b.volumeCreated
     # Don't directly compare the Label dictionaries, since the server could
     # have added extra Labels (for things like BaseImage).
-    for key, val in labels.iteritems():
-        assert new_b["labels"].get(key) == val
+    for key, val in iter(labels.items()):
+        assert new_b.labels.get(key) == val
 
-    volume = wait_for_backup_completion(client, volname, snap["name"])
+    volume = wait_for_backup_completion(client, volname, snap.name)
     volume = wait_for_volume_status(client, volname,
                                     "lastBackup",
-                                    b["name"])
-    assert volume["lastBackupAt"] != ""
+                                    b.name)
+    assert volume.lastBackupAt != ""
 
     return bv, b, snap, data
 
@@ -228,7 +228,7 @@ def delete_backup(backup_volume, backup_name):
             continue
 
         for b in backups:
-            if b["name"] == backup_name:
+            if b.name == backup_name:
                 found = True
                 break
         if not found:
@@ -257,13 +257,13 @@ def create_and_check_volume(client, volume_name, num_of_replicas=3, size=SIZE,
                          numberOfReplicas=num_of_replicas,
                          baseImage=base_image, frontend=frontend)
     volume = wait_for_volume_detached(client, volume_name)
-    assert volume["name"] == volume_name
-    assert volume["size"] == size
-    assert volume["numberOfReplicas"] == num_of_replicas
-    assert volume["state"] == "detached"
-    assert volume["baseImage"] == base_image
-    assert volume["frontend"] == frontend
-    assert volume["created"] != ""
+    assert volume.name == volume_name
+    assert volume.size == size
+    assert volume.numberOfReplicas == num_of_replicas
+    assert volume.state == "detached"
+    assert volume.baseImage == base_image
+    assert volume.frontend == frontend
+    assert volume.created != ""
     return volume
 
 
@@ -568,12 +568,12 @@ def check_volume_replicas(volume, spec, tag_mapping):
     """
     found_hosts = {}
     # Make sure that all the Tags the Volume requested were fulfilled.
-    for replica in volume["replicas"]:
-        found_hosts[replica["hostId"]] = {}
+    for replica in volume.replicas:
+        found_hosts[replica.hostId] = {}
         assert not len(set(spec["disk"]) -
-                       set(tag_mapping[replica["hostId"]]["disk"]))
+                       set(tag_mapping[replica.hostId]["disk"]))
         assert not len(set(spec["node"]) -
-                       set(tag_mapping[replica["hostId"]]["node"]))
+                       set(tag_mapping[replica.hostId]["node"]))
 
     # The Volume should have replicas on as many nodes as matched
     # the requirements (specified by "expected" in the spec variable).
@@ -591,7 +591,7 @@ def set_node_tags(client, node, tags=[]):  # NOQA
     :param tags: The tags to set on the node.
     :return: The updated Node.
     """
-    return client.update(node, allowScheduling=node["allowScheduling"],
+    return client.update(node, allowScheduling=node.allowScheduling,
                          tags=tags)
 
 
@@ -665,7 +665,7 @@ def pod_make(request):
             try:
                 delete_and_wait_pod(api, pod_manifest['metadata']['name'])
             except Exception as e:
-                print "Exception when waiting for pod deletion", e
+                print("Exception when waiting for pod deletion", e)
                 return
 
         request.addfinalizer(finalizer)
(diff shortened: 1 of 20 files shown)


manager/integration/tests/common.py
```diff
@@ -173,11 +173,11 @@ def cleanup_volume(client, volume):
...
     """
     volume.detach()
-    volume = wait_for_volume_detached(client, volume["name"])
+    volume = wait_for_volume_detached(client, volume.name)
     client.delete(volume)
-    wait_for_volume_delete(client, volume["name"])
+    wait_for_volume_delete(client, volume.name)
     volumes = client.list_volume()
-    assert len(volumes) == 0
+    assert volumes.__len__() == 0
```
yasker: Can you check if this PR fixes the inconsistency of __len__()? https://github.com/longhorn/longhorn-tests/pull/244

manager/integration/tests/longhorn.py
```diff
@@ -10,6 +10,8 @@
 import json
 import time
 import operator
+from functools import reduce
```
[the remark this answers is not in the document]
meldafrawi: yes, in:
https://github.com/longhorn/longhorn-tests/blob/4b74107615fc96df10b5b9f68668691530985d2b/manager/integration/tests/longhorn.py#L613

manager/integration/tests/longhorn.py
```diff
@@ -193,8 +195,8 @@ class GdapiClient(object):
     def __init__(self, access_key=None, secret_key=None, url=None, cache=False,
                  cache_time=86400, strict=False, headers=HEADERS, **kw):
         self._headers = headers
-        self._access_key = access_key
-        self._secret_key = secret_key
+        self._access_key = access_key.__str__()
```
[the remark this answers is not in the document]
meldafrawi: This is used to fix the following warnings:

```
- DeprecationWarning: Non-string usernames will no longer be supported in Requests 3.0.0.
Please convert the object you've passed in (None) to a string or bytes object in the near future to avoid problem
    
- DeprecationWarning: Non-string passwords will no longer be supported in Requests 3.0.0.
Please convert the object you've passed in (None) to a string or bytes object in the near future to avoid problems.
```
And  `access_key` & `secret_key` is set by default to `None`

manager/integration/tests/common.py
meldafrawi: ok, will check.

manager/integration/tests/longhorn.py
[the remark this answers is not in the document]
yasker: Can we set them to "" by default then? as well as the URL.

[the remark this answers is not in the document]
meldafrawi: fixed.

manager/integration/tests/common.py
meldafrawi: PR works.

manager/integration/tests/common.py
```diff
@@ -995,7 +995,7 @@ def client(request):
     # Make sure nodes and managers are all online.
     ips = get_mgr_ips()
     client = get_client(ips[0] + PORT)
-    hosts = client.list_node()
+    hosts = client.list_node().data
```
yasker: no longer needed?

manager/integration/tests/common.py
```diff
@@ -1010,17 +1010,17 @@ def clients(request):
     k8sconfig.load_incluster_config()
     ips = get_mgr_ips()
     client = get_client(ips[0] + PORT)
-    hosts = client.list_node()
+    hosts = client.list_node().data
```
yasker: no longer needed?

manager/integration/tests/common.py
```diff
@@ -1434,16 +1434,16 @@ def wait_for_replica_failed(client, volname, replica_name):
         time.sleep(RETRY_INTERVAL)
         failed = True
         volume = client.by_id_volume(volname)
-        for r in volume["replicas"]:
-            if r['name'] != replica_name:
+        for r1 in volume.replicas:
```
yasker: Why renamed to `r1`?

manager/integration/tests/common.py
```diff
@@ -1506,7 +1506,7 @@ def write_volume_random_data(volume, position={}):
 
 def check_device_data(dev, data, check_checksum=True):
     r_data = dev_read(dev, data['pos'], data['len'])
-    assert r_data == data['content']
+    assert r_data == bytes(data['content'], encoding='utf8')
```
yasker: Do we need `encoding`? Random data is obviously not utf8.

manager/integration/tests/common.py
```diff
@@ -1571,6 +1571,7 @@ def volume_write(v, start, data):
 
 
 def dev_write(dev, start, data):
+    data = bytes(data, encoding='utf-8')
```
yasker: Same here.

manager/integration/tests/common.py
```diff
@@ -2348,10 +2393,13 @@ def wait_volume_kubernetes_status(client, volume_name, expect_ks):
     for i in range(RETRY_COUNTS):
         expected = True
         volume = client.by_id_volume(volume_name)
-        ks = volume["kubernetesStatus"]
+        ks = volume.kubernetesStatus
+        ks = json.loads(json.dumps(ks, default=lambda o: o.__dict__))
```
yasker: Is this still necessary?

manager/integration/tests/test_basic.py
```diff
@@ -886,15 +888,15 @@ def test_listing_backup_volume(clients, base_image=""):   # NOQA
     for i in range(RETRY_COUNTS):
         bvs = client.list_backupVolume()
         for bv in bvs:
-            if bv["name"] == volume1_name:
-                if "error" in bv.messages:
-                    assert "volume.cfg" in bv.messages["error"].lower()
+            if bv.name == volume1_name:
+                if "error" in bv.messages.__dict__:
```
yasker: Why we need `__dict__` here?

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -282,79 +282,79 @@ def engine_live_upgrade_test(client, core_api, volume_name, base_image=""):  # N
...
-    wait_for_engine_image_deletion(client, core_api, new_img['name'])
+    wait_for_engine_image_deletion(client, core_api, new_img.name)
 
 
 def test_engine_image_incompatible(client, core_api, volume_name):  # NOQA
     images = client.list_engine_image()
     assert len(images) == 1
-    assert images[0]["default"]
-    assert images[0]["state"] == "ready"
+    assert images.data[0].default
```
yasker: Is this still necessary?

manager/integration/tests/test_toleration_setting.py
```diff
@@ -138,7 +138,7 @@ def wait_for_toleration_update(core_api, apps_api, count, set_tolerations):  # N
         client = get_longhorn_api_client()
         images = client.list_engine_image()
         assert len(images) == 1
-        if images[0]["state"] != "ready":
+        if images.data[0].state != "ready":
```
yasker: Is this still necessary?

manager/integration/tests/common.py
meldafrawi: In `fdev.write(data)` function expects the `data` to be in a bytes-like object or this error occurs:

> TypeError: a bytes-like object is required, not 'str'

manager/integration/tests/common.py
meldafrawi: Type of `data['contents']` is an object of string class, and type of `r_data` is an object of bytes class
encoding here is needed to convert string to bytes and compare.

manager/integration/tests/test_basic.py
meldafrawi: `bv.messages` is of type `longhorn.RestObject` and is not iteratable.

>*** TypeError: argument of type 'RestObject' is not iterable

manager/integration/tests/common.py
meldafrawi: Yes, `workloadsStatus` object in `volume.kubernetesStatus` is of type `longhorn.RestObject`, and need to be converted to dict to be compared with `expected_ks` value, which of type `dict`

manager/integration/tests/common.py
meldafrawi: fixed

manager/integration/tests/common.py
meldafrawi: fixed

manager/integration/tests/common.py
meldafrawi: Sorry, debugging typo
fixed

manager/integration/tests/test_engine_upgrade.py
meldafrawi: fixed

manager/integration/tests/test_toleration_setting.py
meldafrawi: fixed

manager/integration/tests/common.py
yasker: Do we need to specify encoding `utf8`? By reading this code, one might think the data is encoded in utf8. But it's just raw data in alphabets. The encoding shouldn't matter so it's confusing.

manager/integration/tests/common.py
yasker: But now longhorn.RestObject has __getitem__ for any keys, so `ks[k]` should work directly?

manager/integration/tests/common.py
meldafrawi: When encoding is not used, assert is always `False` even if the data matched.

for example if `data['content']` is a string like this `"ABC....."`
`r_data` is `b'ABC.....'`

manager/integration/tests/test_basic.py
yasker: Can you try the following?
```
diff --git a/manager/integration/tests/longhorn.py b/manager/integration/tests/longhorn.py
index 1620567..faeaa35 100644
--- a/manager/integration/tests/longhorn.py
+++ b/manager/integration/tests/longhorn.py
@@ -134,6 +134,7 @@ class RestObject:
     def __iter__(self):
         if self._is_list():
             return iter(self.data)
+        return iter(self.__dict__)
```

yasker: Per https://docs.python.org/3/reference/expressions.html#membership-test-operations

```
For user-defined classes which define the __contains__() method, x in y returns True if y.__contains__(x) returns a true value, and False otherwise.

For user-defined classes which do not define __contains__() but do define __iter__(), x in y is True if some value z, for which the expression x is z or x == z is true, is produced while iterating over y. If an exception is raised during the iteration, it is as if in raised that exception.

Lastly, the old-style iteration protocol is tried: if a class defines __getitem__(), x in y is True if and only if there is a non-negative integer index i such that x is y[i] or x == y[i], and no lower integer index raises the IndexError exception. (If any other exception is raised, it is as if in raised that exception).
```

We do defined `__getitem__()` and `__iter__()` in the object, why `in` doesn't work?

manager/integration/tests/common.py
meldafrawi: it didn't work in this case, the output of `ks[k]` is of type  `longhorn.RestObject` not a `dict`

yasker:
We shouldn't need to use any built-in functions like `__dict__` or `__str__` if `longhorn.py` has been implemented correctly. Something was wrong.

manager/integration/tests/common.py
yasker: Looks like Python 3 enforced choosing the encoding for string:

>If it is a string, you must also give the encoding (and optionally, errors) parameters; bytearray() then converts the string to bytes using str.encode().

manager/integration/tests/test_basic.py
meldafrawi: Adding `return iter(self.__dict__)` to `__iter__` definition works.

PR #233, opened 2019-10-22
Ignore the error caused by listing the deleting backups

shuo-wu:
https://github.com/longhorn/longhorn/issues/826

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -549,7 +549,7 @@ def flexvolume(request):
     flexvolume_manifest = {
         'name': generate_volume_name(),
         'flexVolume': {
-            'driver': 'rancher.io/longhorn',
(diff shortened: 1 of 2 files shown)


yasker:
Can we fix the error instead?

shuo-wu:
1. It's an error caused by the intermediate status of backup volumes. I don't think there is a big impact on the Longhorn system.
2. We cannot find a simple way to fix the error that caused by listing and deleting the backups of the same volume simultaneously.

yasker:
Longhorn system itself is OK but it's an unexpected error to the API and user.

I think we should put this on hold and think more about it.

shuo-wu:
We can fix the error caused by listing backup volumes only and deleting the backups simultaneously, which leads to this test failure. But the case I mentioned above is unavoidable.

PR #234, opened 2019-10-28
test framework: add optional upgrade step

meldafrawi:
longhorn/longhorn#746


yasker:
We also need to have a test case which can simulate the steps in https://github.com/longhorn/longhorn-tests/wiki/Manual-Test-Plan#longhorn-upgrade-test , to make sure our test is closer to the real environment.

meldafrawi:
Closing this PR in favor of longhorn/longhorn-tests#243

PR #237, opened 2019-11-05
add node offline test

meldafrawi:
longhorn/longhorn#635

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -77,6 +77,10 @@ spec:
     - name: longhorn
       mountPath: /var/lib/longhorn/
       mountPropagation: Bidirectional
+  dnsConfig:
+    nameservers:
+      - 8.8.8.8
+      - 1.1.1.1
   volumes:
   - name: dev
     hostPath:
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -76,6 +76,10 @@ DEFAULT_POD_TIMEOUT = 180
 DEFAULT_STATEFULSET_INTERVAL = 5
 DEFAULT_STATEFULSET_TIMEOUT = 180
 
+DEFAULT_DEPLOYMENT_INTERVAL = 1
+DEFAULT_DEPLOYMENT_TIMEOUT = 120
+
+
 DEFAULT_VOLUME_SIZE = 3  # In Gi
 EXPANDED_VOLUME_SIZE = 4  # In Gi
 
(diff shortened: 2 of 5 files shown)


manager/integration/tests/test_infra.py
```diff
@@ -159,3 +181,181 @@ def test_offline_node(reset_cluster_ready_status):
...
+    apps_api = get_apps_api_client()
+    longhorn_api_client = get_longhorn_api_client()
+    cloudprovider = detect_cloudprovider()
+
+    volume_name = generate_volume_name()
+    pv_name = volume_name + "-pv"
+    pvc_name = volume_name + "-pvc"
+    deployment_name = volume_name + "-dep"
+
+    atexit.register(delete_and_wait_longhorn, longhorn_api_client, volume_name)
```
meldafrawi: use fixture instead of atexit function

manager/integration/tests/test_infra.py
```diff
@@ -159,3 +181,181 @@ def test_offline_node(reset_cluster_ready_status):
...
+    atexit.register(delete_and_wait_pv, core_api, pv_name)
+    atexit.register(delete_and_wait_pvc, core_api, pvc_name)
+    atexit.register(delete_and_wait_deployment, apps_api, deployment_name)
+
+    for pod in core_api.list_namespaced_pod(
```
meldafrawi: use get_self_hostid to get the node name

manager/integration/tests/test_infra.py
```diff
@@ -159,3 +181,181 @@ def test_offline_node(reset_cluster_ready_status):
...
+                          core_api,
+                          longhorn_volume,
+                          pvc_name)
+
+    create_and_wait_deployment(apps_api, deployment_manifest)
+
+    deployment_label_selector =\
+        "name=" + deployment_manifest["metadata"]["labels"]["name"]
+
+    deployment_pod_list =\
```
meldafrawi: use get_namespaced_deployment

manager/integration/tests/test_infra.py
```diff
@@ -159,3 +181,181 @@ def test_offline_node(reset_cluster_ready_status):
...
+
+    assert k8s_node_down
+
+    longhorn_node_down = wait_for_node_down_longhorn(node_name,
+                                                     longhorn_api_client)
+    assert longhorn_node_down
+
+    time.sleep(toleration_seconds + 5)
+
+    deployment_label_selector =\
```
meldafrawi: - remove it, already defined above
- don't use numbers in timeout

manager/integration/tests/test_infra.py
```diff
@@ -159,3 +181,181 @@ def test_offline_node(reset_cluster_ready_status):
...
+            core_api.list_namespaced_pod(
+                namespace="default",
+                label_selector=deployment_label_selector
+            )
+
+        assert deployment_pod_list.items.__len__() == 1
+
+        new_pod_name = deployment_pod_list.items[0].metadata.name
+
+        wait_deployment_replica_ready(apps_api, deployment_name, 1)
```
meldafrawi: check if new pod is running state instead of deployment replicas is ready.

manager/integration/tests/test_infra.py
```diff
@@ -159,3 +181,181 @@ def test_offline_node(reset_cluster_ready_status):
...
+            if pod.metadata.__getattribute__("deletion_timestamp") is not None:
+                terminating_pod_name = pod.metadata.name
+                break
+
+        if terminating_pod_name is not None:
+            break
+        else:
+            time.sleep(1)
+
+    if terminating_pod_name is not None:
```
meldafrawi: use `assert terminating_pod_name ` instead

manager/integration/tests/common.py
```diff
@@ -1466,7 +1466,18 @@ def wait_for_replica_failed(client, volname, replica_name):
 
 @pytest.fixture
 def volume_name(request):
-    return generate_volume_name()
+    volume_name = generate_volume_name()
```
yasker: Is it safe to change this? It seems lots of assumption is made here and this function has been widely used in the test cases.

manager/integration/deploy/test.yaml
```diff
@@ -77,6 +77,10 @@ spec:
     - name: longhorn
       mountPath: /var/lib/rancher/longhorn/
       mountPropagation: Bidirectional
+  dnsConfig:
```
yasker: Do we still need this after coreDNS scales to 3?

meldafrawi:
@yasker

- The readiness probe is used solve `client_cleanup` teardown issue,  the manager pod will be marked ready when port 9500 is open, and ready to accept connections longhorn/longhorn#909

- For port checking:
  Previously when `client` object is created, it used first IP of the list of manager IP addresses 
https://github.com/longhorn/longhorn-tests/blob/bd3281bb43a00830e53faab22a51c3e2817214e3/manager/integration/tests/common.py#L1008 

  In the case of the offline node test, if by chance that first manager IP address used to create  `client` object belongs to the manager pod on the node that will be made offline, then `client` object will not be able to do any operations since it's already trying to connect to a manager on an offline node, and test fails.

PR #240, opened 2019-11-08
Add test for CSI block volume

shuo-wu:
longhorn/longhorn#678

### DIFF
--- a/manager/integration/tests/test_csi.py
+++ b/manager/integration/tests/test_csi.py
@@ -1,12 +1,18 @@
 #!/usr/sbin/python
 import pytest
-
+import random
 import common
-from common import client, core_api, csi_pv, pod_make, pvc  # NOQA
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_csi.py
```diff
@@ -195,3 +201,58 @@ def backupstore_test(client, core_api, csi_pv, pvc, pod_make, pod_name, base_ima
...
+
+    create_storage_class(storage_class)
+    create_pvc(pvc)
+    pv_name = wait_and_get_pv_for_pvc(core_api, pvc_name).metadata.name
+    create_and_wait_pod(core_api, pod_manifest)
+
+    test_data = generate_random_data(VOLUME_RWTEST_SIZE)
+    write_pod_block_volume_data(
+        core_api, pod_name, test_data, device_path)
+    md5_sum = get_pod_block_volume_data_md5sum(
```
yasker: Need to check the raw data once before calculation md5sum. The stored data can be wrong.

PR #243, opened 2019-11-21
Add integration test case for upgrade

meldafrawi:
longhorn/longhorn#746

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -10,7 +10,8 @@ metadata:
   name: longhorn-test-role
 rules:
 - apiGroups: [""]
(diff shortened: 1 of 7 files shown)


manager/integration/deploy/test.yaml
```diff
@@ -18,6 +18,9 @@ rules:
 - apiGroups: ["apps"]
   resources: ["statefulsets", "deployments", "daemonsets"]
   verbs: ["*"]
+- apiGroups: ["*"]
```
meldafrawi: customize api/resource group needed

meldafrawi: fixed

meldafrawi:
**Usage:**
`pytest --include-upgrade-test -svvl test_upgrade.py::test_upgrade --upgrade-image-tag=IMAGE_TAG`

by default the test will upgrade from the latest stable Longhorn version, currently `v0.7.0` to `master`
unless `--upgrade-image-tag=` options is used to specify different image tag.

PR #252, opened 2020-01-14
fix cleanup_client call in clients fixture finalizer

meldafrawi:

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1057,12 +1057,10 @@ def clients(request):
     clis = get_clients(hosts)
 
     def finalizer():
-        client = next(iter(clis.values()))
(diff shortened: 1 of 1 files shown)


manager/integration/tests/common.py
```diff
@@ -1057,8 +1057,7 @@ def clients(request):
     clis = get_clients(hosts)
 
     def finalizer():
-        client = next(iter(clis.values()))
-        cleanup_client(client)
+        cleanup_client()
 
     request.addfinalizer(finalizer)
```
yasker: Do we still need the `client = next(iter(clis.values()))` below?

meldafrawi: No, I've removed it.

PR #257, opened 2020-01-27
Remove Flexvolume tests

meldafrawi:
longhorn/longhorn#1032


manager/integration/tests/conftest.py
```diff
@@ -58,14 +58,7 @@ def pytest_collection_modifyitems(config, items):
...
     using_csi = check_csi(core_api)
-    if using_csi:
-        skip_upgrade = pytest.mark.skip(reason="environment is not using " +
-                                               "flexvolume")
-        for item in items:
-            if "flexvolume" in item.keywords:
-                item.add_marker(skip_upgrade)
-
-    else:
+    if not using_csi:
```
yasker: Error out if it's not CSI since it's the only driver we support now.

manager/integration/tests/test_statefulset.py
```diff
@@ -347,19 +347,6 @@ def test_statefulset_restore(client, core_api, storage_class,  # NOQA
             },
             'volumeHandle': ''
         }
-    else:
```
yasker: Change `if csi` to `assert csi`.

manager/integration/tests/test_statefulset.py
```diff
@@ -380,9 +367,6 @@ def test_statefulset_restore(client, core_api, storage_class,  # NOQA
             wait_for_volume_detached(client, pod['pvc_name'])
 
             pv['spec']['csi']['volumeHandle'] = pod['pvc_name']
-        else:
```
yasker: Change `if csi:` to `assert csi`.

manager/integration/tests/test_statefulset.py
meldafrawi: fixed

manager/integration/tests/test_statefulset.py
meldafrawi: fixed

manager/integration/tests/conftest.py
meldafrawi: removed the check, since we're only support csi driver now.

PR #258, opened 2020-02-07
add k8s zone tag integration test

meldafrawi:
longhorn/longhorn#929

### DIFF
--- a/manager/integration/tests/test_zone.py
+++ b/manager/integration/tests/test_zone.py
@@ -1,0 +1,220 @@
+import pytest
+import time
+
+from common import client # NOQA
+from common import core_api  # NOQA
+from common import create_and_check_volume
+from common import get_self_host_id
+from common import RETRY_COUNTS
+from common import RETRY_INTERVAL
+from common import volume_name # NOQA
+from common import wait_for_volume_degraded
+from common import wait_for_volume_healthy
+from common import wait_for_volume_replica_count
+from random import randrange
+from test_scheduling import wait_new_replica_ready
+from common import get_version_api_client
+
+
+# label deprecated for k8s >= v1.17
+DEPRECATED_K8S_ZONE_LABEL = "failure-domain.beta.kubernetes.io/zone"
+
+K8S_ZONE_LABEL = "topology.kubernetes.io/zone"
+
+ZONE1 = "lh-zone1"
+ZONE2 = "lh-zone2"
+
+
+def get_k8s_zone_label():
+    ver_api = get_version_api_client()
+    k8s_ver_data = ver_api.get_code()
+
+    k8s_ver_major = k8s_ver_data.major
+    assert k8s_ver_major == '1'
+
+    k8s_ver_minor = k8s_ver_data.minor
+
+    if int(k8s_ver_minor) >= 17:
+        k8s_zone_label = K8S_ZONE_LABEL
+    else:
+        k8s_zone_label = DEPRECATED_K8S_ZONE_LABEL
+
+    return k8s_zone_label
+
+
+@pytest.fixture
+def k8s_node_zone_tags(client, core_api):  # NOQA
+
+    k8s_zone_label = get_k8s_zone_label()
+    lh_nodes = client.list_node()
+
+    node_index = 0
+    for node in lh_nodes:
+        node_name = node.name
+
+        if node_index % 2 == 0:
+            zone = ZONE1
+        else:
+            zone = ZONE2
+
+        payload = {
+            "metadata": {
+                "labels": {
+                    k8s_zone_label: zone}
+            }
+        }
+
+        core_api.patch_node(node_name, body=payload)
+        node_index += 1
+
+    yield
+
+    lh_nodes = client.list_node()
+
+    node_index = 0
+    for node in lh_nodes:
+        node_name = node.name
+
+        payload = {
+            "metadata": {
+                "labels": {
+                    k8s_zone_label: None}
+            }
+        }
+
+        core_api.patch_node(node_name, body=payload)
+
+
+def wait_longhorn_node_zone_updated(client): # NOQA
+
+    lh_nodes = client.list_node()
+    node_names = map(lambda node: node.name, lh_nodes)
+
+    for node_name in node_names:
+        for j in range(RETRY_COUNTS):
+            lh_node = client.by_id_node(node_name)
+            if lh_node.zone != '':
+                break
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,190 @@
...
+            "metadata": {
+                "labels": {
+                    K8s_ZONE_LABEL: zone}
+            }
+        }
+
+        core_api.patch_node(node_name, body=payload)
+        node_index += 1
+
+    yield
```
shuo-wu: This fixture/generator is not invoked in the test. Will the following (cleanup) part be executed after the test complete?

manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,190 @@
...
+
+        if node_index % 2 == 0:
+            zone = ZONE1
+        else:
+            zone = ZONE2
+
+        payload = {
+            "metadata": {
+                "labels": {
+                    K8s_ZONE_LABEL: zone}
```
shuo-wu: This label is used for k8s version v1.17 or higher. Need to check the cluster version first then use `failure-domain.beta.kubernetes.io/zone` for the lower version cluster. check [here](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone) for details.

manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,190 @@
...
+        if lh_node.zone != '':
+            break
+        time.sleep(RETRY_INTERVAL)
+
+    assert lh_node.zone != ''
+
+
+def test_zone_tags(client, core_api, volume_name, k8s_node_zone_tags):  # NOQA
+
+    for lh_node in client.list_node():
```
shuo-wu: Why not waiting for **all** longhorn nodes update in the func `wait_longhorn_node_zone_updated()`?

manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,190 @@
...
+
+    host_id = get_self_host_id()
+
+    volume.attach(hostId=host_id)
+
+    volume = wait_for_volume_healthy(client, volume_name)
+
+    volume = client.by_id_volume(volume_name)
+
+    zone1_replica_count = 0
```
shuo-wu: The lines 91-102 are redundant. The logic is the same as that of the lines 125-146.

manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,190 @@
...
+    wait_for_volume_healthy(client, volume_name)
+
+    volume = client.by_id_volume(volume_name)
+
+    lh_node_names = list(map(lambda node: node.name, client.list_node()))
+
+    for replica in volume.replicas:
+        lh_node_names.remove(replica.hostId)
+
+    assert lh_node_names == []
```
shuo-wu: Does this imply that there are 3 nodes at most in the cluster? I think the logic is like:
if len(nodes) >= len(replicas):
  each replica is in a different node

shuo-wu: And the logic here is similar to the following part. (line 183-190)

manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,190 @@
...
+        wait_for_volume_replica_count(client, volume_name, replica_count)
+
+        volume = client.by_id_volume(volume_name)
+
+        lh_node_names = list(map(lambda node: node.name, client.list_node()))
+
+        for replica in volume.replicas:
+            lh_node_names.remove(replica.hostId)
+
+        assert lh_node_names == []
```
shuo-wu: Add one more part or one more separate test case: scale the replicas to test the soft anti-affinity feature.

manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,190 @@
+import pytest
+import time
+
+from common import client # NOQA
```
shuo-wu: Sort out this import part. I think it's better to follow the schema of the engine integration test.

meldafrawi: fixed

manager/integration/tests/test_zone.py
meldafrawi: yes, `By using a yield statement instead of return, all the code after the yield statement serves as the teardown code:`
https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

manager/integration/tests/test_zone.py
meldafrawi: This checks if only one replica is scheduled per node.

manager/integration/tests/test_zone.py
meldafrawi: fixed

manager/integration/tests/test_zone.py
meldafrawi: fixed

manager/integration/tests/test_zone.py
meldafrawi: fixed

manager/integration/tests/test_zone.py
meldafrawi: fixed

manager/integration/tests/test_zone.py
```diff
@@ -0,0 +1,215 @@
...
+    ver_api = get_version_api_client()
+    k8s_ver_data = ver_api.get_code()
+
+    k8s_ver_major = k8s_ver_data.major
+    assert k8s_ver_major == '1'
+
+    k8s_ver_minor = k8s_ver_data.minor
+
+    if int(k8s_ver_minor) >= 17:
+        k8s_zone_label = "topology.kubernetes.io/zone"
```
shuo-wu: It's better to continue using constants for the label keys.

manager/integration/tests/test_zone.py
shuo-wu: I mean:
```
from common import (
......
)

from test_scheduling import ...
```

meldafrawi: * According to python style guide
  > https://www.python.org/dev/peps/pep-0008/#imports

  it's recommended to imports to be on separate lines.

* I'll create a separate PR to move `wait_new_replica_ready()` to common.

manager/integration/tests/test_zone.py
meldafrawi: fixed

shuo-wu:
LGTM

PR #265, opened 2020-03-04
run all integration tests after upgrade test

meldafrawi:

### DIFF
--- a/test_framework/scripts/longhorn-setup.sh
+++ b/test_framework/scripts/longhorn-setup.sh
@@ -24,7 +24,6 @@ check_longhorn_status() {
 
 }
 
-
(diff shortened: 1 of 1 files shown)


test_framework/scripts/longhorn-setup.sh
```diff
@@ -75,4 +96,5 @@ while [[ -z "`kubectl get pods longhorn-test --no-headers=true | awk '{print $3}
   sleep 5m
 done
 
+# get integration test junit report
 kubectl logs longhorn-test  > "${TF_VAR_tf_workspace}/longhorn-test-junit-report.xml"
```
yasker: The report will be overwritten.

meldafrawi: Fixed, Thanks @yasker

PR #266, opened 2020-03-04
move backupstore creation after longhorn setup

meldafrawi:


yasker:
Reason? What if we want to run backup cases for upgrade test in the future?

yasker:
OK, just saw the reason. We can create the namespace beforehand then.

meldafrawi:
minio backupstore is deployed to `longhorn-system` namespace.

meldafrawi:
@yasker updated.

PR #273, opened 2020-03-06
add wait for remount check for auto salvage tests

meldafrawi:

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2900,23 +2900,6 @@ def disable_auto_salvage(client):
     assert setting.value == "true"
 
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -2930,3 +2930,24 @@ def get_liveness_probe_spec(initial_delay=5, period=5):
...
+
+
+def wait_for_pod_remount(core_api, pod_name):
+    check_command = [
+        '/bin/sh',
+        '-c',
+        'ls /data/'
+    ]
+
+    for i in range(10):
```
yasker: 1. If we have this, we don't need `wait_for_pod_restart` since that's not the real case we want to check.
2. Retry time is not enough. Assuming the pod will restart multiple times during the period.

PR #275, opened 2020-03-18
Add test case for default disks and node tags setup

shuo-wu:
longhorn/longhorn#991
longhorn/longhorn#1053


yasker:
Need one more test case to make sure the wrong annotation wouldn't block the node controller from working.

PR #276, opened 2020-03-23
Add tests for no annotation and invalid node annotations

shuo-wu:
longhorn/longhorn#1130

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -2,6 +2,7 @@ import common
 import pytest
 import os
 import subprocess
+import time
 
 from random import choice
 from string import ascii_lowercase, digits
@@ -21,7 +22,8 @@ from common import SETTING_STORAGE_OVER_PROVISIONING_PERCENTAGE, \
 from common import get_volume_endpoint
 from common import get_update_disks
 from common import wait_for_disk_status, wait_for_disk_update, \
-    wait_for_disk_conditions, wait_for_node_tag_update
+    wait_for_disk_conditions, wait_for_node_tag_update, \
+    cleanup_node_disks
 from common import exec_nsenter
 from common import wait_for_replica_directory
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_node.py
```diff
@@ -1308,3 +1310,150 @@ def test_node_config_annotations(client, core_api,  # NOQA
...
+            "annotations": {
+                DEFAULT_DISK_CONFIG_ANNOTATION:
+                    '[{"path":"/invalid-path","allowScheduling":false,' +
+                    '"storageReserved":1024,"tags":["ssd","fast"]}]',
+                DEFAULT_NODE_TAG_ANNOTATION: '[",.*invalid-tag"]',
+            }
+        }
+    })
+    # Make sure Longhorn has tried to apply the annotation.
+    time.sleep(2)
```
yasker: Magic number `2`. Use a constant.

manager/integration/tests/test_node.py
```diff
@@ -1308,3 +1310,150 @@ def test_node_config_annotations(client, core_api,  # NOQA
...
+    # then the correct annotation should be applied.
+    node = client.by_id_node(node_name)
+    disks = node.disks
+    for _, disk in iter(disks.items()):
+        disk.allowScheduling = False
+    update_disks = get_update_disks(disks)
+    node = client.by_id_node(node_name)
+    node.diskUpdate(disks=update_disks)
+    node.diskUpdate(disks=[])
+    time.sleep(2)
```
yasker: There is a wait below, why we need to use `time.sleep` here?

manager/integration/tests/test_node.py
```diff
@@ -1308,3 +1310,150 @@ def test_node_config_annotations(client, core_api,  # NOQA
...
+        }
+    })
+    time.sleep(2)
+    node = client.by_id_node(node_name)
+    assert set(node.tags) == {"tag3", "tag4"}
+
+    # Clean up all node tags
+    # then the correct annotation should be applied.
+    client.update(node, tags=[])
+    time.sleep(2)
```
yasker: there is a wait below.

manager/integration/tests/test_node.py
shuo-wu: Before the above disk cleanup, there is one existing disk on the node.
After the disk cleanup, node controller will re-create a new disk immediately according to the annotation.
Hence the below wait cannot guarantee that the cleanup and recreation are complete or not.

Similar below

yasker: Then why sleeping 2 seconds can guarantee that?

shuo-wu: OK. It's better to use `wait_for_disk_status()` rather than sleep and `wait_for_disk_update`

PR #278, opened 2020-03-24
Add test case test_replica_rebuild_with_intensive_data() 

shuo-wu:
longhorn/longhorn#1103

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -555,9 +555,9 @@ def read_pod_block_volume_data(api, pod_name, data_size, offset, device_path):
             tty=False)
 
 
-def get_pod_block_volume_data_md5sum(api, pod_name, device_path):
+def get_pod_data_md5sum(api, pod_name, path):
     md5sum_command = [
-        '/bin/sh', '-c', 'md5sum ' + device_path + " | awk '{print $1}'"
+        '/bin/sh', '-c', 'md5sum ' + path + " | awk '{print $1}'"
     ]
     with timeout(seconds=STREAM_EXEC_TIMEOUT * 3,
                  error_message='Timeout on executing stream md5sum'):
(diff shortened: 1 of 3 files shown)


yasker:
Can this reproduce https://github.com/longhorn/longhorn/issues/1103 ?

shuo-wu:
No. I cannot find a way to reproduce file sending timeout error... 
This test tries to guarantee that the rebuild feature works fine even if there is intensive data in the volume.

yasker:
Can we introduce an error in the rebuild process?

shuo-wu:
Now the test can trigger rebuild failure. The rebuild failure is triggered by the sender replica crash rather than the file sending timeout, but the rebuilt/receiver replica will still wrongly mark the rebuild state as `complete` then be cleaned up by the engine immediately. 

As I commented in [issue 1103](https://github.com/longhorn/longhorn/issues/1103), the intermediate status (receiver replica's rebuild state becoming `complete`) can not be caught without using the image `shuowu/longhorn-engine:rebuild-test` and manually checking the log. Hence there is no related assertion in the test.

yasker:
I am OK with it as long as it can trigger the original issue and failed the test.

Need a minor rebase. Otherwise LGTM.

manager/integration/tests/test_ha.py
```diff
@@ -504,3 +509,57 @@ def test_salvage_auto_crash_replicas_long_wait(client, core_api, volume_name, po
...
+
+    data_path_1 = "/data/test1"
+    write_pod_volume_random_data(core_api, pod_name, data_path_1, 300)
+    original_md5sum_1 = get_pod_data_md5sum(core_api, pod_name, data_path_1)
+    create_snapshot(client, volume_name)
+    data_path_2 = "/data/test2"
+    write_pod_volume_random_data(core_api, pod_name, data_path_2, 300)
+    original_md5sum_2 = get_pod_data_md5sum(core_api, pod_name, data_path_2)
+
+    for i in range(3):
```
yasker: Magic number

manager/integration/tests/test_ha.py
```diff
@@ -504,3 +509,57 @@ def test_salvage_auto_crash_replicas_long_wait(client, core_api, volume_name, po
...
+    volume = create_and_check_volume(client, volume_name,
+                                     num_of_replicas=3, size=str(1 * Gi))
+    assert len(volume.replicas) == 3
+    create_pv_for_volume(client, core_api, volume, pv_name)
+    create_pvc_for_volume(client, core_api, volume, pvc_name)
+    pod['spec']['volumes'] = [create_pvc_spec(pvc_name)]
+    create_and_wait_pod(core_api, pod)
+
+    data_path_1 = "/data/test1"
+    write_pod_volume_random_data(core_api, pod_name, data_path_1, 300)
```
yasker: Magic number `300`

manager/integration/tests/common.py
```diff
@@ -567,6 +567,19 @@ def get_pod_block_volume_data_md5sum(api, pod_name, device_path):
             tty=False)
 
 
+def write_pod_volume_random_data(api, pod_name, path, size):
```
yasker: Use `size_mb` or `size_in_mb`. It's not obvious.

PR #279, opened 2020-03-25
Integration test for PV with XFS  filesystem

meldafrawi:
longhorn/longhorn#994


yasker:
What if we start a pod with a blank PV set to `fstype: xfs`? Will Kubernetes or Longhorn driver format it correctly?

meldafrawi:
I've update the test start a pod with a blank PV set to `fstype: xfs`

meldafrawi:
@yasker I've added a test case for  existing XFS volume.

PR #281, opened 2020-03-27
Add test case test_engine_image_daemonset_restart()

shuo-wu:
longhorn/longhorn#1116

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -4,7 +4,7 @@ import subprocess
 import pytest
 
 from common import clients, random_labels, volume_name  # NOQA
-from common import core_api, pod   # NOQA
+from common import core_api, apps_api, pod   # NOQA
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -1543,3 +1543,31 @@ def test_restore_inc_with_expansion(clients, core_api, volume_name, pod):  # NOQ
...
+    apps_api.delete_namespaced_daemon_set(ds_name, common.LONGHORN_NAMESPACE)
+
+    check_volume_data(volume, snap1_data)
+
+    # Wait for the restart complete
+    common.wait_for_engine_image_state(client, default_img.name, "ready")
+
+    # Longhorn is still able to use the corresponding engine binary to
+    # operate snapshot
+    snap2_data = write_volume_random_data(volume)
```
yasker: Better to check volume data again before writing more random data.

PR #287, opened 2020-04-13
Add a new test test_csi_expansion_canceling()

shuo-wu:
longhorn/longhorn#978

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -45,6 +45,8 @@ from common import expand_attached_volume
 from common import wait_for_dr_volume_expansion
 from common import check_block_device_size
 from common import wait_for_rebuild_complete
+from common import wait_for_volume_expansion
+from common import fail_replica_expansion, wait_for_expansion_failure
 
 
 @pytest.mark.coretest   # NOQA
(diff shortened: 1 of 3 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -1794,3 +1796,92 @@ def test_engine_image_daemonset_restart(clients, apps_api, volume_name):  # NOQA
...
+    4. Delete the pod and wait for volume detachment
+    5. Try to expand the volume using Longhorn API
+    6. Wait for expansion failure then use Longhorn API to cancel it
+    7. Create a new pod and validate the volume content,
+       then re-write random data to the pod
+    8. Delete the pod and wait for volume detachment
+    9. Retry expansion then verify the expansion done using Longhorn API
+    10. Create a new pod
+    11. Validate the volume content, then check if data writing looks fine
+    12. Cleam up pod, PVC, and PV
```
yasker, line 1819: Typo.

manager/integration/tests/test_basic.py
```diff
@@ -1794,3 +1796,92 @@ def test_engine_image_daemonset_restart(clients, apps_api, volume_name):  # NOQA
     snap2_data = write_volume_random_data(volume)
     create_snapshot(client, volume_name)
     check_volume_data(volume, snap2_data)
+
+
+@pytest.mark.coretest  # NOQA
+def test_expansion_canceling(clients, core_api, volume_name, pod):  # NOQA
+    """
+    Test expansion canceling feature
```
yasker, line 1804: Remove `feature`. It's a part of expansion feature.

manager/integration/tests/test_ha.py
```diff
@@ -708,3 +712,82 @@ def test_rebuild_failure_with_intensive_data(client, core_api, volume_name, pod_
...
+    2. Generate `test_data` and write to the pod
+    3. Create an empty directory with expansion snapshot tmp meta file path
+       for one replica so that the replica expansion will fail
+    4. Delete the pod and wait for volume detachment
+    5. Expand the volume using Longhorn API
+    6. Check expansion result using Longhorn API. There will be expansion error
+       caused by the failed replica but overall the expansion should succeed.
+    7. Create a new pod and
+       check if the volume will rebuild for the failed replica
+    8 .Validate the volume content, then check if data writing looks fine
```
yasker, line 732: Wrong position for `.`

shuo-wu:
1. As long as the PVC spec.size is expanded, we cannot shrink it anymore. Hence the expansion canceling cannot be triggered via PVC/CSI.
2. If users choose PVC to expand the volume then try to use Longhorn API to cancel the expansion, it would lead to the inconsistency between PVC and Longhorn volume. 

Hence expansion canceling feature is mainly for the users using the Longhorn API to expand volumes. And we need to add a tip for the API on the UI page.

yasker:
@shuo-wu How about one replica failure when expansion? The expansion should succeed, right? That's something we should test with CSI as well.

shuo-wu:
Yes. That's pretty similar to the 2nd test in this PR. Do we still need to retain the 2nd test after adding the CSI test you mentioned?

yasker:
It's better to do it through the CSI if we only want to keep one case.

PR #297, opened 2020-04-20
Add test for the corner case that the from-replica and the rebuilding replica are on the same node

shuo-wu:
longhorn/longhorn#1167
longhorn/longhorn#1220

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -32,6 +32,7 @@ from common import wait_for_rebuild_start
 from kubernetes.stream import stream
 
 RANDOM_DATA_SIZE = 300
+RANDOM_DATA_SIZE2 = 800
 
 
 @pytest.mark.coretest   # NOQA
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -635,3 +636,83 @@ def test_rebuild_failure_with_intensive_data(client, core_api, volume_name, pod_
...
+    original_replicas = volume.replicas
+    assert len(original_replicas) == 3
+    # Trigger rebuild
+    crash_replica_processes(client, core_api, volume_name,
+                            [original_replicas[0]])
+    wait_for_volume_degraded(client, volume_name)
+    from_replica_name, rebuilding_replica_name = \
+        wait_for_rebuild_start(client, volume_name)
+    assert from_replica_name != rebuilding_replica_name
+    found_from_replica = False
```
yasker, line 696: `verified_from_replica`. `found` just meant it's found.

manager/integration/tests/test_ha.py
```diff
@@ -635,3 +636,83 @@ def test_rebuild_failure_with_intensive_data(client, core_api, volume_name, pod_
     delete_and_wait_pod(core_api, pod_name)
     delete_and_wait_pvc(core_api, pvc_name)
     delete_and_wait_pv(core_api, pv_name)
+
+
+def test_rebuild_duplicate_replica(client, core_api, volume_name, pod_make):  # NOQA
```
yasker, line 641: `duplicate_replica` part in the name is confusing.

Be more specific, e.g. `test_rebuild_from_replica_field`.

Also, do we need a separate test case for the field? We can add it in the existing replica rebuild test.

shuo-wu: This test is not only for the field `fromReplica, but also for the corner case that the from-replica and the rebuilding replica are on the same node.

yasker: Make sense. Rename the test case to show that should be good enough.

PR #298, opened 2020-04-20
fix infra test condition check

meldafrawi:

### DIFF
--- a/test_framework/scripts/longhorn-setup.sh
+++ b/test_framework/scripts/longhorn-setup.sh
@@ -83,7 +83,7 @@ else
 fi
 
 
-if [[ ${LONGHORN_INFRA_TEST} ]] ; then
+if [[ "${LONGHORN_INFRA_TEST}" =~ [Tt]rue ]] ; then
   ## enable infrastructure tests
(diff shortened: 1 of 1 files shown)


test_framework/scripts/longhorn-setup.sh
```diff
@@ -83,7 +83,7 @@ else
 fi
 
 
-if [[ ${LONGHORN_INFRA_TEST} ]] ; then
+if ${LONGHORN_INFRA_TEST} ; then
```
yasker, line 86: Need a better way to ensure:
1. $LONGHORN_INFRA_TEST is true/True, `if` executes
2. Otherwise, `if` shouldn't

Now if the variable is an invalid value Or `True`, the script will error out.

meldafrawi: @yasker fixed

test_framework/scripts/longhorn-setup.sh
```diff
@@ -83,7 +83,7 @@ else
 fi
 
 
-if [[ ${LONGHORN_INFRA_TEST} ]] ; then
+if [[ ${LONGHORN_INFRA_TEST} =~ [Tt]rue ]] ; then
```
yasker, line 86: Better add quotes for ${LONGHORN_INFRA_TEST} to ensure it's treated as a string.

meldafrawi: fixed

PR #302, opened 2020-04-27
Update the endpoint value check

shuo-wu:
https://github.com/longhorn/longhorn/issues/1248


manager/integration/deploy/test.yaml
```diff
@@ -51,12 +51,12 @@ spec:
   containers:
   - name: longhorn-test-pod
     image: longhornio/longhorn-manager-test:master
-#    args: [
+    args: [
```
yasker, line 54: ??

shuo-wu: Sorry for pushing the wrong commit...

PR #304, opened 2020-04-28
Update test-minio-backupstore to use https with custom cert

joshimoo:


yasker:
Need to add issue URL next time.

PR #306, opened 2020-05-01
Add backup status test for unavailable replicas

joshimoo:
This test is used to make sure that we do not try to get the backup 
status from unavailable replicas. This could lead to the engine backup 
status retrieval process exceeding the maximum allowed time, after which
the manager would terminate that process. Since we wouldn't be able to
retrieve updated backup statuses we could no longer show new backups in
the frontend.

longhorn/longhorn#1230

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -468,6 +468,128 @@ def snapshot_test(clients, volume_name, base_image):  # NOQA
     cleanup_volume(client, volume)
 
 
+def test_backup_status_for_unavailable_replicas(clients, volume_name):    # NOQA
+    """
+    Test backup status for unavailable replicas
+
+    Context:
+
+    We want to make sure that we do not try to retrieve the backup status
+    of no longer valid replicas (offline, deleted, etc). The reason for
+    this is that trying to establish a tcp connection with an old replica
+    address `(tcp://ip:port)` could block the engine retrieval process,
+    since we will wait upto 1 minute for each individual backup status.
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -467,6 +467,116 @@ def snapshot_test(clients, volume_name, base_image):  # NOQA
...
+    2. Run the test for all the available backupstores.
+
+    Steps:
+
+    1. Create a backup of volume
+    2. Find the replica for that backup
+    3. Disable scheduling on the node of that replica
+    4. Delete the replica
+    5. Wait for volume backup status state to go to error
+    6. Verify backup status error contains `unknown replica`
```
yasker, line 487: Explain a little more on why this will result in `unknown replicas`, e.g. why we need to disable the scheduling etc.

manager/integration/tests/test_basic.py
```diff
@@ -467,6 +467,116 @@ def snapshot_test(clients, volume_name, base_image):  # NOQA
 
     cleanup_volume(client, volume)
 
+@pytest.mark.coretest   # NOQA
```
yasker, line 470: Not a core functional test.

manager/integration/tests/test_basic.py
joshimoo: Doesn't really fit under the Steps heading, should I add an explanation/context Heading?

yasker: Yeah, that's better

PR #310, opened 2020-05-07
fix offline node with attached volume and pod test

meldafrawi:
test failed on restoring default cluster status (all nodes running)
because the very short time between node shutdown and node start again
increasing node unreachable/not-ready toleration_seconds will fix the
issue.

Longhorn/Longhorn#1303


yasker:
As we discussed, if DO reports the command success, then the following operation should be fine. Need to check why that's not the case here.

PR #312, opened 2020-05-12
Add test skeletons for restore failures

shuo-wu:
longhorn/longhorn#1260

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -888,3 +888,36 @@ def test_restore_volume_with_invalid_backupstore():
     12. Verify this faulted volume can be deleted.
     """
     pass
+
+
+@pytest.mark.skip(reason="TODO")
+def test_all_replica_restore_failure():
+    """
+    [HA] Test if all replica restore failure will lead to the restore volume
+    becoming Faulted, and if the auto salvage feature is disabled for
+    the faulted restore volume.
+
+    1. Enable auto-salvage.
+    2. Set the a random backupstore.
+    3. Do cleanup for the backupstore.
+    4. Create a pod with a volume and wait for pod to start.
+    5. Write data to the pod volume and get the md5sum.
+    6. Create a backup for the volume.
+    7. Randomly delete some data blocks of the backup, which will lead to
+       all replica restore failures later.
+    8. Restore a volume from the backup.
+    9. Wait for the volume restore in progress by checking if:
+       9.1. `volume.restoreStatus` shows the related restore info.
+       9.2. `volume.conditions[restore].status == True &&
+            volume.conditions[restore].reason == "RestoreInProgress"`.
+       9.3. `volume.ready == false`.
+    10. Wait for the restore volume Faulted.
+    11. Check if `volume.conditions[restore].status == False &&
+        volume.conditions[restore].reason == "RestoreFailure"`.
+    12. Check if `volume.ready == false`.
+    13. Make sure auto-salvage is not triggered even the feature is enabled.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -836,3 +836,107 @@ def test_rebuild_with_inc_restoration(
...
+    4. Invalidate the target URL.
+       (e.g.: s3://backupbucket-invalid@us-east-1/backupstore-invalid)
+    5. Restore a volume from the backup.
+       (The fromBackup fields of the volume create API should consist of
+       the invalid target URL and the valid backup volume info)
+    6. Wait for the restore volume Faulted.
+    7. Check if the volume condition "restore":
+       `volume.conditions[restore].status == False &&
+       volume.conditions[restore].reason == "RestoreFailure".`
+    8. Check if the auto-salvage is disabled even if the setting is enabled.
```
sowmyav27, line 861: Why would `auto-salvage` be disabled even if we enable the setting?

manager/integration/tests/test_ha.py
```diff
@@ -836,3 +836,107 @@ def test_rebuild_with_inc_restoration(
...
+    9. Check if `volume.ready` is false
```
sowmyav27, line 862: Can we also add the following:

10. Verify if PV/PVC cannot be created from Longhorn.
11. Verify the faulted volume cannot be attached to a node.
12. Verify this faulted volume can be deleted

manager/integration/tests/test_ha.py
```diff
@@ -836,3 +836,107 @@ def test_rebuild_with_inc_restoration(
...
+    13. Check if
+        13.1. there is no rebuilt replica for the restored volume
+              before reattachment.
+        13.2. `volume.ready == false`
+        13.3. `volume.conditions[restore].status == False &&
+              volume.conditions[restore].reason == ""`.
+    14. Create PV/PVC/Pod for the restored volume and wait for the pod start.
+    15. Check if the restored volume is state `degraded`
+        after the attachment.
+    16. Wait for the rebuild complete and the volume becoming healthy.
```
sowmyav27, line 938: We can add an additional step here

Verify `len(restore_volume.replicas) == 3`.

manager/integration/tests/test_ha.py
shuo-wu: I mean, if the faulted volume is caused by the restore failure, Longhorn won't do auto salvage and the volume will keep state`Faulted` even if the setting is enabled

manager/integration/tests/test_ha.py
shuo-wu: Sure, but I am not sure if I need to put this kind of implementation details in the skeleton.
The reason I mention some detailed checks in the skeleton is that some fields are newly introduced after the fix and some fields are critical for the state check.

manager/integration/tests/test_ha.py
shuo-wu: YES, I will add it

manager/integration/tests/test_ha.py
yasker: @shuo-wu Probably you should put it this way:

`Make sure auto-salvage is not triggered even if the feature is enabled.`

Disabled is a state.

manager/integration/tests/test_ha.py
yasker: If volume suppose to have three replicas and it's healthy, that means all three replicas should be healthy. So in this case, there is no need to mention it since we will make sure the volume is healthy.

meldafrawi:
LGTM

PR #314, opened 2020-05-12
test: Add test case description for "Disable Scheduling On Cordoned Node" setting

boknowswiki:
This test covers the replica scheduler should be able to schedule replica on cordoned node when this setting is set to false.
And the replica scheduler should not schedule replica on cordoned node when this setting is set to true.

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -373,6 +373,25 @@ def test_replica_zone_anti_affinity(client):  # NOQA
     pass
 
 
+@pytest.mark.skip(reason="TODO")
+@pytest.mark.node  # NOQA
+def test_disable_scheduling_on_cordoned_node(client):  # NOQA
+    """
+    Test replica scheduler: schedule replica based on
+    `Disable Scheduling On Cordoned Node` setting
+
+    1. Set `Disable Scheduling On Cordoned Node` to true.
+    2. Set `Replica Soft Anti-Affinity` to false.
+    3. Set cordon on one node.
+    4. Create a volume with 3 replicas.
+    5. Verify the scheduler should fail due to cordoned node.
+    6. Set `Disable Scheduling On Cordoned Node` to false.
+    7. Automatically the scheduler should creates three replicas
+    from step 5 failure.
+    """
+    pass
+
+
 @pytest.mark.node  # NOQA
 @pytest.mark.mountdisk # NOQA
 def test_replica_scheduler_large_volume_fit_small_disk(client):  # NOQA


manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,22 @@ def test_replica_scheduler_no_disks(client):  # NOQA
...
+@pytest.mark.node  # NOQA
+def test_disable_scheduling_on_cordoned_node(client):  # NOQA
+    """
+    Test replica scheduler: schedule replica based on
+    `Disable Scheduling On Cordoned Node` setting
+
+    1. Set `Disable Scheduling On Cordoned Node` to true.
+    2. Set `Replica Soft Anti-Affinity` to false.
+    3. Set cordon on one node.
+    4. Create a volume and verify the scheduler should fail due to cordoned node.
```
yasker, line 362: `Create a volume with three replicas`

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,22 @@ def test_replica_scheduler_no_disks(client):  # NOQA
...
+    5. Remove step 4 volume.
```
yasker, line 363: Don't need to remove the volume on step 4. Once the scheduling resumed, the volume should be scheduled correctly.

yasker:
Also, `flake8` failed.

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,22 @@ def test_replica_scheduler_no_disks(client):  # NOQA
     common.wait_for_volume_delete(client, vol_name)
 
 
+@pytest.mark.node  # NOQA
+def test_disable_scheduling_on_cordoned_node(client):  # NOQA
+    """
+    Test replica scheduler: schedule replica based on
+    `Disable Scheduling On Cordoned Node` setting
+
+    1. Set `Disable Scheduling On Cordoned Node` to true.
```
sowmyav27, line 359: Is this a new setting which will be introduced for the users on the Longhorn UI?

manager/integration/tests/test_node.py
sowmyav27, line 362: Why would we expect the scheduler to fail when one node is cordoned? It should be able to schedule the replicas on other nodes which are not cordoned.

manager/integration/tests/test_node.py
boknowswiki: Yes, this is a new node condition we introduced. I will update the condition name from "Cordoned" to "Schedulable".

manager/integration/tests/test_node.py
boknowswiki: This is because we disable the soft anti-affinity, we shouldn't schedule multiple replicas on the same node.

sowmyav27: @boknowswiki But Scheduler will be able to schedule replicas on the node which is not cordoned isnt it?

boknowswiki: @sowmyav27 , the case here is we have 3 nodes, 1 is cordoned, and we need 3 replicas, and we have disabled the soft on node anti-affinity. So we can schedule 2 replicas on the 2 good nodes, but for the cordoned node, we should not schedule replica on it. Hence the scheduler failed to schedule 3 replicas. Please let me know what you think.

Thanks.

meldafrawi:
LGTM

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,23 @@ def test_replica_scheduler_no_disks(client):  # NOQA
     common.wait_for_volume_delete(client, vol_name)
 
 
+@pytest.mark.node  # NOQA
+def test_disable_scheduling_on_cordoned_node(client):  # NOQA
```
yasker, line 354: Add `@pytest.mark.skip(reason="TODO")`, so we know which tests need to be implemented.

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,23 @@ def test_replica_scheduler_no_disks(client):  # NOQA
...
+    1. Set `Disable Scheduling On Cordoned Node` to true.
+    2. Set `Replica Soft Anti-Affinity` to false.
+    3. Set cordon on one node.
+    4. Create a volume with 3 replicas.
+    5. Verify the scheduler should fail due to cordoned node.
+    6. Set `Disable Scheduling On Cordoned Node` to false.
+    7. Automatically the scheduler should creates three replicas
+    from step 5 failure.
+    """
+
```
yasker, line 368: Add `pass` at the end of the function.

manager/integration/tests/test_node.py
boknowswiki: Thanks for the input.

yasker:
LGTM. Need rebase.

boknowswiki:
@meldafrawi , I will start to work on this test case today. Just make sure we don't duplicate our work. Please let me know if you have worked on this issue already.

PR #316, opened 2020-05-13
Add test_single_replica_failed_during_engine_start() skeleton

shuo-wu:
longhorn/longhorn#1253

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -836,3 +836,27 @@ def test_rebuild_with_inc_restoration(
     delete_and_wait_pod(core_api, dr_pod_name)
     delete_and_wait_pvc(core_api, dr_pvc_name)
     delete_and_wait_pv(core_api, dr_pv_name)
+
+
+@pytest.mark.skip
+def test_single_replica_failed_during_engine_start():
+    """
+    Test if the volume still works fine when there is
+    an invalid replica/backend in the engine starting phase.
+
+    1. Create a pod using Longhorn volume.
+    2. Write some data to the volume then get the md5sum.
+    3. Create a snapshot.
+    4. Repeat step2 and step3 for 3 times then there should be 3 snapshots.
+    5. Randomly pick up a replica and
+       manually messing up the snapshot meta file.
+    6. Delete the pod and wait for the volume detached.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -836,3 +836,27 @@ def test_rebuild_with_inc_restoration(
...
+    Test if the volume still works fine when there is
+    an invalid replica/backend in the engine starting phase.
+
+    1. Create a pod using Longhorn volume.
+    2. Write some data to the volume then get the md5sum.
+    3. Create a snapshot.
+    4. Repeat step2 and step3 for 3 times then there should be 3 snapshots.
+    5. Randomly pick up a replica and
+       manually messing up the snapshot meta file.
+    6. Delete the pod and wait for the volume detached.
```
sowmyav27, line 853: Why are we deleting the pod here after corrupting the data? For a replica rebuild?

khushboo-rancher:
Test looks good, just curious if we find a way to figure out the replica is bad based on snapshot data not the meta data https://github.com/longhorn/longhorn/issues/1298, then we will have to add another scenario to mess with snapshot data to make replica go bad.

manager/integration/tests/test_ha.py
shuo-wu: Deleting the pod will trigger volume detach. The purpose of this step is to detach the volume.

shuo-wu:
> longhorn/longhorn#1298

Unfortunately, Longhorn may not be able to detect if one replica snapshot data is messed up if the replica r/w itself doesn't error out. @yasker Do we need to consider this part? I remembered that we had discusses if we need to add the replica consistency check before attaching and the conclusion was No since the check may take a long time.

yasker:
> > [longhorn/longhorn#1298](https://github.com/longhorn/longhorn/issues/1298)
> 
> Unfortunately, Longhorn may not be able to detect if one replica snapshot data is messed up if the replica r/w itself doesn't error out. @yasker Do we need to consider this part? I remembered that we had discusses if we need to add the replica consistency check before attaching and the conclusion was No since the check may take a long time.

Snapshot data check is out of scope. And currently, we don't have a way to detect the data corruption inside the replica automatically.

PR #317, opened 2020-05-13
scheduler: Test case for zone anti-affinity

boknowswiki:
This test case is to cover zone anti-affinity.

zone hard anti-affinity means the scheduler should not schedule replica on the same zone
zone soft anti-affinity means the scheduler can schedule replica on the same zone

https://github.com/longhorn/longhorn/issues/1052

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -350,6 +350,29 @@ def test_replica_scheduler_no_disks(client):  # NOQA
     common.wait_for_volume_delete(client, vol_name)
 
 
+@pytest.mark.skip(reason="TODO")
+@pytest.mark.node  # NOQA
+def test_replica_zone_anti_affinity(client):  # NOQA
+    """
+    Test replica scheduler with zone anti-affinity
+
+    1. Set zone anti-affinity to hard.
+    2. Label nodes 1 & 2 with same zone label "zone1".
+    Label node 3 with zone label "zone2".
+    3. Create a volume with 3 replicas.
+    4. Wait for volume condition `scheduled` to be false.
+    5. Label node 2 with zone label "zone3".
+    6. Wait for volume condition `scheduled` to be success.
+    7. Clear the volume.
+    8. Set zone anti-affinity to soft.
+    9. Change the zone labels on node 1 & 2 & 3 to "zone1".
+    10. Create a volume.
+    11. Wait for volume condition `scheduled` to be success.
+    12. Clean up the replica count, the zone labels and the volume.
+    """
+    pass
+
+
 @pytest.mark.node  # NOQA
 @pytest.mark.mountdisk # NOQA
 def test_replica_scheduler_large_volume_fit_small_disk(client):  # NOQA


yasker:
@boknowswiki remember to replace the link to the issue.

Flake8 failed.

boknowswiki:
@yasker , thanks for the notice.

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,27 @@ def test_replica_scheduler_no_disks(client):  # NOQA
     common.wait_for_volume_delete(client, vol_name)
 
 
+@pytest.mark.node  # NOQA
+def test_replica_zone_anti_affinity(client):  # NOQA
+    """
+    Test replica scheduler with zone anti-affinity
+
+    1. Set zone anti-affinity to hard, and replica count to 2.
```
meldafrawi, line 358: - we don't need to set default replica count to `2`, instead, we can create a volume with `2` replicas in step `3`

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,27 @@ def test_replica_scheduler_no_disks(client):  # NOQA
...
+    2. Create 2 nodes and add the same zone label to them.
```
meldafrawi, line 359: - we automatically deploy Longhorn on 3 node cluster for integration tests, so may be we can do like this:
  1- Label nodes 1 & 2 with  same zone label  e.g `zone1` 
  2- Label nodes 3 with another zone label  e.g `zone2` 
  3- Disabled scheduling on node 3

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,27 @@ def test_replica_scheduler_no_disks(client):  # NOQA
...
+    3. Create a volume.
+    4. Wait for volume condition `scheduled` to be false.
+    5. Change the zone label to be different on one node.
+    6. Create a volume.
+    7. Wait for volume condition `scheduled` to be success.
+    8. Clear the volume.
+    9. Set zone anti-affinity to soft.
```
meldafrawi, line 366: After we set `zone anti-affinity` to soft, we can increase the number of volume replicas to e.g 5 and check replicas scheduled successfully and in healthy state.

manager/integration/tests/test_node.py
boknowswiki: I will move this to step 3, since we have 3 nodes we will need 3 replicas.

manager/integration/tests/test_node.py
boknowswiki: Thanks for the input. I will update this part, but for "Disabled scheduling on node 3", we don't need to change the scheduling part of the node, we only need to change the node's zone label.

manager/integration/tests/test_node.py
boknowswiki: I think 3 is good enough. We can set the replicas to 5, but in this case we need to make sure we have the node soft anti-affinity is set as well.

boknowswiki:
@meldafrawi , please take a look at the new PR, and let me know what you think.

Thanks.

manager/integration/tests/test_node.py
```diff
@@ -350,6 +350,27 @@ def test_replica_scheduler_no_disks(client):  # NOQA
     common.wait_for_volume_delete(client, vol_name)
 
 
+@pytest.mark.node  # NOQA
```
yasker, line 353: Add skip mark here and `pass` at the end.

meldafrawi:
LGTM

PR #319, opened 2020-05-14
add test_single_replica_failed_during_engine_start

meldafrawi:
longhorn/longhorn#1253

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -2,8 +2,10 @@ import pytest
 import common
 import time
 import random
+import subprocess
 
 from common import client, core_api, volume_name  # NOQA
+from common import DATA_SIZE_IN_MB_1, DATA_SIZE_IN_MB_2, DATA_SIZE_IN_MB_3
 from common import check_volume_data, cleanup_volume, create_and_check_volume
 from common import delete_replica_processes, crash_replica_processes
 from common import get_self_host_id, check_volume_endpoint
@@ -13,7 +15,7 @@ from common import expand_attached_volume, check_block_device_size
 from common import write_volume_data, generate_random_data
 from common import wait_for_rebuild_complete
 from common import disable_auto_salvage # NOQA
-from common import pod_make  # NOQA
+from common import pod_make, pod  # NOQA
 from common import create_pv_for_volume
 from common import create_pvc_for_volume
 from common import create_pvc_spec
@@ -36,7 +38,6 @@ from common import check_volume_last_backup
 from common import activate_standby_volume
 
 from common import SIZE, VOLUME_RWTEST_SIZE, EXPAND_SIZE, Gi
-from common import DATA_SIZE_IN_MB_2, DATA_SIZE_IN_MB_3
 from common import RETRY_COUNTS, RETRY_INTERVAL
 from common import SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY
 
@@ -849,8 +850,8 @@ def test_rebuild_with_inc_restoration(
     delete_and_wait_pv(core_api, dr_pv_name)
 
 
-@pytest.mark.skip(reason="TODO")
-def test_single_replica_failed_during_engine_start():
+@pytest.mark.coretest
+def test_single_replica_failed_during_engine_start(client, core_api, volume_name, pod): # NOQA
     """
     Test if the volume still works fine when there is
     an invalid replica/backend in the engine starting phase.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -859,4 +860,75 @@ def test_single_replica_failed_during_engine_start():
...
+    volume = client.by_id_volume(volume_name)
+    for repl in volume.replicas:
+        if repl.name == replica_name:
+            break
+
+    assert repl.running is False
+    assert repl.failedAt != ''
+
+    wait_for_volume_healthy(client, volume_name)
+    vol_data = read_volume_data(core_api, pod_name)
```
shuo-wu, line 926: I prefer to use these functions to write and check data:
```
write_pod_volume_random_data(core_api, pod_name,
                                 data_path, DATA_SIZE_IN_MB_1)
get_pod_data_md5sum(core_api, pod_name, data_path)
```

Then we can check all data written to the volume rather than the last written data.

manager/integration/tests/test_ha.py
```diff
@@ -859,4 +860,75 @@ def test_single_replica_failed_during_engine_start():
...
+
+    assert vol_data == test_data3
+    test_data4 = generate_random_data(VOLUME_RWTEST_SIZE)
+    write_pod_volume_data(core_api, pod_name, test_data4)
+
+    vol_data = read_volume_data(core_api, pod_name)
+
+    assert vol_data == test_data4
```
shuo-wu, line 934: Try to create one more snapshot and delete some snapshots to see if the snapshot chain is still fine.

manager/integration/tests/test_ha.py
```diff
@@ -859,4 +860,75 @@ def test_single_replica_failed_during_engine_start():
...
+    pod['metadata']['name'] = pod_name
+    pod['spec']['volumes'] = [{
+        'name': pod['spec']['containers'][0]['volumeMounts'][0]['name'],
+        'persistentVolumeClaim': {
+            'claimName': pvc_name,
+        },
+    }]
+
+    create_and_wait_pod(core_api, pod)
+    wait_for_volume_healthy(client, volume_name)
```
shuo-wu, line 880: You can use `prepare_pod_with_data_in_mb()` in the beginning. This modification/suggestion is optional for you.

manager/integration/tests/test_ha.py
```diff
@@ -859,7 +860,137 @@ def test_single_replica_failed_during_engine_start():
...
+    res_data_md5sum2 = get_pod_data_md5sum(core_api, pod_name, data_path2)
+    assert data_md5sum2 == res_data_md5sum2
+
+    res_data_md5sum3 = get_pod_data_md5sum(core_api, pod_name, data_path3)
+    assert data_md5sum3 == res_data_md5sum3
+
+    data_path4 = "/data/file4"
+    write_pod_volume_random_data(core_api, pod_name,
+                                 data_path4, DATA_SIZE_IN_MB_1)
+    data_md5sum1 = get_pod_data_md5sum(core_api, pod_name, data_path4)
```
shuo-wu, line 949: Should it be named as `data_md5sum4` rather than `data_md5sum1`? And it seems that this returned variable is not used in the test.

manager/integration/tests/test_ha.py
```diff
@@ -859,7 +860,137 @@ def test_single_replica_failed_during_engine_start():
...
+
+    replica_data_path = replica.dataPath
+    replica_name = replica.name
+    snap2_meta_file = replica_data_path + \
+        "/volume-snap-" + \
+        snap2.name + ".img.meta"
+
+    command = ["dd", "if=/dev/zero",
+               "of="+snap2_meta_file,
+               "count=" + str(DATA_SIZE_IN_MB_1),
```
yasker, line 917: You don't need to write a 100MB file to replace the metadata file. The metadata file is only a few bytes.

meldafrawi: fixed

manager/integration/tests/test_ha.py
meldafrawi: fixed

manager/integration/tests/test_ha.py
```diff
@@ -870,7 +871,141 @@ def test_single_replica_failed_during_engine_start():
...
+
+    replica_data_path = replica.dataPath
+    replica_name = replica.name
+    snap2_meta_file = replica_data_path + \
+        "/volume-snap-" + \
+        snap2.name + ".img.meta"
+
+    command = ["dd", "if=/dev/zero",
+               "of="+snap2_meta_file,
+               "count=" + str(100),
```
yasker, line 928: Writing 100M of 0 into the metadata file? The file is only a few bytes.

meldafrawi: Sorry, I should have removed `bs=1M` part

meldafrawi: fixed

PR #320, opened 2020-05-15
manager: Add rebuild concurrent limit test skeleton

yasker:

Signed-off-by: Sheng Yang <sheng.yang@rancher.com>


meldafrawi:
LGTM

PR #321, opened 2020-05-15
controller: Test case skeleton for adding default disk back when extra disk is unmounted

boknowswiki:
This test case is to cover adding default disk back when the extra disk is unmounted
Since the extra disk will show the same filesystem ID as the default disk
when it is unmounted.

https://github.com/longhorn/longhorn/issues/1291

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -979,6 +979,28 @@ def test_node_controller_sync_disk_state(client):  # NOQA
                                      CONDITION_STATUS_TRUE)
 
 
+@pytest.mark.skip(reason="TODO")
+@pytest.mark.node  # NOQA
+@pytest.mark.mountdisk # NOQA
+def test_node_default_disk_added_back_with_extra_disk_unmounted(client):  # NOQA
+    """
+    [Node] Test adding default disk back with extra disk is unmounted
+    on the node
+
+    1. Create a host disk `extra_disk` using Longhorn volume and
+    attach it to node 1.
+    2. Delete the default disk on node 1.
+    3. Unmount the extra disk on node 1.
+    And wait for it becoming "Unschedulable".
+    4. Create and add the default disk back on node 1.
+    5. Wait and verify the default disk should become "Schedulable".
+    6. Mount extra disk back on node 1.
+    7. Wait and verify this extra disk should become "Schedulable".
+    8. Delete the host disk `extra_disk`.
+    """
+    pass
+
+
 @pytest.mark.node  # NOQA
 @pytest.mark.mountdisk # NOQA
 def test_node_umount_disk(client):  # NOQA


manager/integration/tests/test_node.py
```diff
@@ -983,6 +983,29 @@ def test_node_controller_sync_disk_state(client):  # NOQA
...
+    2. Create an schedulable extra Longhorn disk on node 1 with
+    this extra new disk.
+    3. Delete the default disk on node 1.
+    4. Unmount the extra disk on node 1.
+    And wait for it becoming "Unschedulable".
+    5. Create and add the default disk back on node 1.
+    6. Wait and verify the default disk should become "Schedulable".
+    7. Mount extra disk back on node 1.
+    8. Wait and verify this extra disk should become "Schedulable".
+    9. Clean the extra disk on node 1.
```
sowmyav27, line 1004: Does "clean" mean deleting the disk?

boknowswiki: Yes, sorry about the confusion. I will update it. Thanks.

meldafrawi:
LGTM

manager/integration/tests/test_node.py
```diff
@@ -1002,6 +1002,30 @@ def test_node_controller_sync_disk_state(client):  # NOQA
...
+@pytest.mark.skip(reason="TODO")
+@pytest.mark.node  # NOQA
+@pytest.mark.mountdisk # NOQA
+def test_node_default_disk_addedback_with_extra_disk_unmounted(client):  # NOQA
+    """
+    [Node] Test adding default disk back with extra disk is unmounted
+    on the node
+
+    1. Create and attach a new (Linux) disk to node 1.
+    2. Create an schedulable extra Longhorn disk on node 1 with
```
yasker, line 1014: The first two steps can be simple as:
> 1. Create a host disk `extra_disk` using Longhorn volume and attach it to node 1.

`host disk` refers to Longhorn's disk created using Longhorn volume in our test cases.

manager/integration/tests/test_node.py
```diff
@@ -1002,6 +1002,30 @@ def test_node_controller_sync_disk_state(client):  # NOQA
...
+    this extra new disk.
+    3. Delete the default disk on node 1.
+    4. Unmount the extra disk on node 1.
+    And wait for it becoming "Unschedulable".
+    5. Create and add the default disk back on node 1.
+    6. Wait and verify the default disk should become "Schedulable".
+    7. Mount extra disk back on node 1.
+    8. Wait and verify this extra disk should become "Schedulable".
+    9. Delete the extra longhorn disk and delete the extra (Linux) disk
```
yasker, line 1023: `Delete the host disk extra_disk` will be enough.

manager/integration/tests/test_node.py
```diff
@@ -1002,6 +1002,28 @@ def test_node_controller_sync_disk_state(client):  # NOQA
...
 
+@pytest.mark.skip(reason="TODO")
+@pytest.mark.node  # NOQA
+@pytest.mark.mountdisk # NOQA
+def test_node_default_disk_addedback_with_extra_disk_unmounted(client):  # NOQA
+    """
+    [Node] Test adding default disk back with extra disk is unmounted
+    on the node
+
+    1. Create a host disk extra_disk using Longhorn volume and
```
yasker, line 1013: Add backquote (`) around the extra_disk so it's treated as a name. Same below for the last step.

manager/integration/tests/test_node.py
```diff
@@ -1002,6 +1002,28 @@ def test_node_controller_sync_disk_state(client):  # NOQA
                                      CONDITION_STATUS_TRUE)
 
 
+@pytest.mark.skip(reason="TODO")
+@pytest.mark.node  # NOQA
+@pytest.mark.mountdisk # NOQA
+def test_node_default_disk_addedback_with_extra_disk_unmounted(client):  # NOQA
```
yasker, line 1008: `test_node_default_disk_add_back_with_extra_disk_unmounted`

`addedback` is not one word. Should be `add_back` or `added_back`

PR #324, opened 2020-05-18
Add test skeleton for test_setting_guaranteed_engine_cpu 

yasker:
https://github.com/longhorn/longhorn/issues/1318

Also remove test skeleton for test_replica_rebuild_concurrent_limit.

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -198,3 +198,34 @@ def check_tolerations_set(current_toleration_list, set_tolerations):
                  "operator": t.operator, "effect": t.effect}
 
     return current_tolerations == set_tolerations
+
+
+@pytest.skip(reason="TODO")
+def test_setting_guaranteed_engine_cpu():
+    """
+    Test setting Guaranteed Engine CPU
+
+    Notice any change of the setting will result in all the instance manager
+    recreated, no matter if there is any volume attached or not.
+
+    1. Change the setting to `xxx`. Update setting should fail.
+    2. Change the setting to `0.1`
+    3. Wait for all the IM pods to recreated and become running
+    4. Verify every IM pod has guaranteed CPU set correctly to 100m
+    5. Change the setting to `0`
+    6. Wait for all the IM pods to recreated and become running
+    7. Verify every IM pod no longer has guaranteed CPU set
+    8. Change the setting to `200m`
+    9. Wait for all the IM pods to recreated and become running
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_settings.py
```diff
@@ -198,3 +198,29 @@ def check_tolerations_set(current_toleration_list, set_tolerations):
...
 
     return current_tolerations == set_tolerations
+
+
+@pytest.skip(reason="TODO")
+def test_setting_guaranteed_engine_cpu():
+    """
+    Test setting Guaranteed Engine CPU
+
+    1. Change the setting to `0.1`
```
sowmyav27, line 208: Does a change in the "Guaranteed Engine CPU" setting value, result in instance manager pods to be recreated ?

manager/integration/tests/test_settings.py
```diff
@@ -198,3 +198,29 @@ def check_tolerations_set(current_toleration_list, set_tolerations):
...
+    2. Wait for all the instance manager pods to recreated and become running
+    3. Verify every pod has guaranteed CPU set correctly to 100m
+    4. Change the setting to `0`
+    5. Wait for all the instance manager pods to recreated and become running
+    6. Verify every pod no longer has guaranteed CPU set
+    7. Change the setting to `200m`
```
sowmyav27, line 214: Is this an invalid value check ?

yasker: `200m` is a valid value. It's the same as `0.2`. See https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/

yasker: Thinking about it, yes, we should have an invalid value check.

manager/integration/tests/test_settings.py
yasker: Yes.

manager/integration/tests/test_settings.py
```diff
@@ -198,3 +198,29 @@ def check_tolerations_set(current_toleration_list, set_tolerations):
...
+    8. Wait for all the instance manager pods to recreated and become running
+    9. Verify every pod no longer has guaranteed CPU set to 200m
```
sowmyav27, line 216: if 200m is same as 0.2, why is it expected here that we "no longer have guaranteed CPU set to 200m" ?

yasker: Good catch. Typo.

PR #325, opened 2020-05-18
fix test_replica_datapath_cleanup

meldafrawi:


yasker:
I merged it too quick..

We should reset soft antiaffinity in https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/common.py#L2220 so it will be done automatically when the function exits.

meldafrawi:
@yasker 
Submitted https://github.com/longhorn/longhorn-tests/pull/326 to fix that.

PR #326, opened 2020-05-18
add replica node and replica zone anti-affinity settings to reset_setting

meldafrawi:


manager/integration/tests/test_node.py
```diff
@@ -1328,11 +1328,6 @@ def test_replica_datapath_cleanup(client):  # NOQA
 
     cleanup_host_disk(client, 'extra-disk')
 
-    # reset soft antiaffinity setting to false
-    node_soft_anti_affinity_setting = \
-        client.by_id_setting(SETTING_REPLICA_SOFT_ANTI_AFFINITY)
```
yasker, line 1333: We should also change the import to `common` instead of `test_scheduling` for all the references of `SETTING_REPLICA_SOFT_ANTI_AFFINITY`. We shouldn't define a constant twice.

meldafrawi: fixed

manager/integration/tests/test_scheduling.py
```diff
@@ -7,11 +7,9 @@
     wait_for_volume_detached, wait_for_volume_degraded, \
     wait_for_volume_healthy, wait_scheduling_failure, \
     write_volume_random_data, wait_for_rebuild_complete
-
+from common import SETTING_REPLICA_SOFT_ANTI_AFFINITY
```
yasker, line 10: This should be in the previous commit since the constant was moved then.

This commit should deal with SETTING_ZONE_SOFT_ANTI_AFFINITY only.

Also, now it shows that the naming of two constants is inconsistent. We call them either:
`SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY` and `SETTING_REPLICA_ZONE_SOFT_ANTI_AFFINITY`
Or:
`SETTING_NODE_SOFT_ANTI_AFFINITY` and `SETTING_ZONE_SOFT_ANTI_AFFINITY`

But `SETTING_REPLICA_SOFT_ANTI_AFFINITY` and `SETTING_ZONE_SOFT_ANTI_AFFINITY` doesn't work.

meldafrawi: fixed

PR #330, opened 2020-05-19
test: Add test case for "Disable Scheduling On Cordoned Node" setting

boknowswiki:
This test covers the replica scheduler should be able to schedule replica on
cordoned node when this setting is set to false.

And the replica scheduler should not schedule replica on cordoned node when
this setting is set to true.

https://github.com/longhorn/longhorn/issues/1287

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -10,21 +10,23 @@ from string import ascii_lowercase, digits
 from common import core_api, client  # NOQA
 from common import Gi, SIZE, CONDITION_STATUS_FALSE, \
     CONDITION_STATUS_TRUE, DEFAULT_DISK_PATH, DIRECTORY_PATH, \
-    DISK_CONDITION_SCHEDULABLE, DISK_CONDITION_READY
+    DISK_CONDITION_SCHEDULABLE, DISK_CONDITION_READY, \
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_node.py
```diff
@@ -368,7 +382,56 @@ def test_disable_scheduling_on_cordoned_node(client):  # NOQA
...
+        get_longhorn_api_client(), node.name)
+    assert node.conditions[NODE_CONDITION_SCHEDULABLE]["status"] == "True"
+
+    # Created volume schedulable condition change to true
+    common.wait_for_volume_condition_scheduled(client, vol_name,
+                                               "status",
+                                               CONDITION_STATUS_TRUE)
+
+    # Disk storage available test on this node
+    node_controller_sync_storage_available_test(client, get_self_host_id())
```
yasker, line 434: You can test the volume instead of the disk, e.g. use `create_and_check_volume`. The disk should be fine.

boknowswiki: @yasker , I realize I have create a volume, is it ok to use that volume to attach it to a node? Also do we need to use `write_volume_random_data` to write data on this volume as well?

Thanks

yasker: Yes, we need to attach the volume to the current node before testing it.

PR #332, opened 2020-05-19
controller: Test case for adding default disk back when extra disk is unmounted

boknowswiki:
This test case is to cover adding default disk back when the extra disk is unmounted
Since the extra disk will show the same filesystem ID as the default disk
when it is unmounted.

https://github.com/longhorn/longhorn/issues/1291

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -981,7 +981,6 @@ def test_node_controller_sync_disk_state(client):  # NOQA
                                      CONDITION_STATUS_TRUE)
 
 
-@pytest.mark.skip(reason="TODO")
 @pytest.mark.node  # NOQA
 @pytest.mark.mountdisk # NOQA
 def test_node_default_disk_added_back_with_extra_disk_unmounted(client):  # NOQA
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_node.py
```diff
@@ -1000,7 +999,106 @@ def test_node_default_disk_added_back_with_extra_disk_unmounted(client):  # NOQA
...
+
+    update_disk = get_update_disks(node.disks)
+    update_disk["extra-disk"] = extra_disk
+    node = node.diskUpdate(disks=update_disk)
+    node = common.wait_for_disk_update(client, lht_hostId,
+                                       len(update_disk))
+    assert len(node.disks) == len(update_disk)
+
+    # Make sure all disks are schedulable
+    for fsid, disk in iter(node.disks.items()):
```
boknowswiki, line 1030: replace fsid with name

boknowswiki: doesn't need to use iter here.

manager/integration/tests/test_node.py
```diff
@@ -1000,7 +999,106 @@ def test_node_default_disk_added_back_with_extra_disk_unmounted(client):  # NOQA
...
+
+    # Add default disk back and wait for schedulable condition
+    default_disk = {"path": DEFAULT_DISK_PATH, "allowScheduling": True,
+                    "storageReserved": SMALL_DISK_SIZE}
+    update_disk = get_update_disks(node.disks)
+    update_disk["default-disk"] = default_disk
+    node = node.diskUpdate(disks=update_disk)
+    node = common.wait_for_disk_update(client, lht_hostId,
+                                       len(update_disk))
+    for fsid, disk in iter(node.disks.items()):
```
boknowswiki, line 1065: fix fsid and iter issue, and check for all of them.

PR #334, opened 2020-05-20
Add and update test for the scheduling failed volume

shuo-wu:
longhorn/longhorn#1360
longhorn/longhorn#1373

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -2053,3 +2053,36 @@ def test_expansion_canceling(clients, core_api, volume_name, pod):  # NOQA
     delete_and_wait_pod(core_api, pod_name)
     delete_and_wait_pvc(core_api, expansion_pvc_name)
     delete_and_wait_pv(core_api, expansion_pv_name)
+
+
+@pytest.mark.coretest  # NOQA
+@pytest.mark.skip(reason="TODO")
+def test_running_volume_with_scheduling_failure():
+    """
+    Test if the running volume still work fine
+    when there is a scheduling failed replica
+
+    Prerequisite:
+    Setting "soft anti-affinity" is false.
+
+    1. Create a volume, then create the corresponding PV, PVC and Pod.
+    2. Wait for the pod running and the volume healthy.
+    3. Write data to the pod volume and get the md5sum.
+    4. Disable the scheduling for a node contains a running replica.
+    5. Crash the replica on the scheduling disabled node for the volume.
+    6. Wait for the new replica created.
+    7. Verify the volume is Degraded and fails to scheduled the replica.
+    8. Verify:
+      8.1. `volume.ready == True`.
+      8.2. `volume.conditions[scheduled].status == False`
+    9. Delete the pod and wait for the volume detached.
+    10. Verify the failed replica is removed and
+        the volume contains healthy replicas only.
+    11. Verify:
+      11.1. `volume.ready == True`.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -2051,3 +2053,30 @@ def test_expansion_canceling(clients, core_api, volume_name, pod):  # NOQA
...
+def test_running_volume_with_scheduling_failure():
+    """
+    Test if the running volume still work fine
+    when there is a scheduling failed replica
+
+    1. Disable the soft anti-affinity setting.
+    2. Create a volume, then create the corresponding PV, PVC and Pod.
+    3. Wait for the pod running and the volume healthy.
+    4. Write data to the pod volume and get the md5sum.
+    5. Disable a node scheduling.
```
sowmyav27, line 2069: Is. this same as "Disable scheduling on a node where a replica is running" ?

manager/integration/tests/test_basic.py
```diff
@@ -2051,3 +2053,30 @@ def test_expansion_canceling(clients, core_api, volume_name, pod):  # NOQA
...
+    6. Crash the replica on the scheduling disabled node for the volume.
+    7. Wait for the new replica created.
+    8. Verify the volume is Degraded and fails to scheduled the replica.
+    9. Verify the `volume.ready == True`.
+    10. Delete the pod and wait for the volume detached.
+    11. Verify the failed replica is removed and
+        the volume contains healthy replicas only.
+    12. Verify the `volume.ready == True`.
```
sowmyav27, line 2077: The volume will be in this state isnt it error - `Scheduling Failure Replica Scheduling Failure` and here volume read = No ?

manager/integration/tests/test_basic.py
shuo-wu: Yes.

manager/integration/tests/test_basic.py
shuo-wu: No. 
`Scheduling Failure` is caused by scheduling failed replica. But this failed replica will be cleaned up during the volume detachment (step 10) then the error will disappear. 
Otherwise, the error will prevent the volume from being reattached/reused by the workloads and it would become a bug.

shuo-wu:
Added one more test skeleton for #1373

manager/integration/tests/test_basic.py
```diff
@@ -2053,3 +2053,30 @@ def test_expansion_canceling(clients, core_api, volume_name, pod):  # NOQA
...
+    4. Write data to the pod volume and get the md5sum.
+    5. Disable the scheduling for a node contains a running replica.
+    6. Crash the replica on the scheduling disabled node for the volume.
+    7. Wait for the new replica created.
+    8. Verify the volume is Degraded and fails to scheduled the replica.
+    9. Verify the `volume.ready == True`.
+    10. Delete the pod and wait for the volume detached.
+    11. Verify the failed replica is removed and
+        the volume contains healthy replicas only.
+    12. Verify the `volume.ready == True`.
```
yasker, line 2077: Also verify the volume schedulable false condition was removed.

manager/integration/tests/test_basic.py
```diff
@@ -2080,3 +2080,38 @@ def test_running_volume_with_scheduling_failure():
...
```
yasker, line 2104: Also verify the volume schedulable false condition was removed.

PR #335, opened 2020-05-20
add integration tests for State transitions for a failure in backup restore

meldafrawi:
https://github.com/longhorn/longhorn/issues/1260

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2293,8 +2293,8 @@ def reset_settings(client):
                       value="true")
     except Exception as e:
         print("Exception when update "
-              "Replica Node Level Soft Anti-Affinity setting",
-              replica_node_soft_anti_affinity_setting, e)
+              "Replica Zone Level Soft Anti-Affinity setting",
+              replica_zone_soft_anti_affinity_setting, e)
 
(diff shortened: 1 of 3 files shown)


yasker:
LGTM.

manager/integration/tests/common.py
```diff
@@ -3343,3 +3343,52 @@ def settings_reset():
...
+            backup_target_setting = client.update(backup_target_setting,
+                                                  value=backupsettings[0])
+            assert backup_target_setting.value == backupsettings[0]
+
+            backup_target_credential_setting = client.by_id_setting(
+                SETTING_BACKUP_TARGET_CREDENTIAL_SECRET)
+            backup_target_credential_setting = \
+                client.update(backup_target_credential_setting,
+                              value=backupsettings[1])
+            assert backup_target_credential_setting.value == backupsettings[1]
```
yasker, line 3388: We should `break` after found the s3 backup target. Same below.

meldafrawi:
Yes, `add wait replica fail for crash_replica_processes` commit should be the second to last.

manager/integration/tests/common.py
meldafrawi: fixed

PR #336, opened 2020-05-21
Add test_dr_volume_with_last_backup_deletion() skeleton

shuo-wu:
longhorn/longhorn#1380

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2751,7 +2751,8 @@ def activate_standby_volume(client, volume_name,
         volume = client.by_id_volume(volume_name)
         engines = volume.controllers
         if len(engines) != 1 or \
-                engines[0].lastRestoredBackup != volume.lastBackup:
+                (volume.lastBackup != "" and
+                 engines[0].lastRestoredBackup != volume.lastBackup):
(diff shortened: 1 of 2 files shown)


yasker:
> Do we also need to add a test case for the second scenario mentioned in the original issue?

We only have one cluster in the test framework, so we won't able to address the Scenario #2.

sowmyav27:
@yasker The second scenario does not need 2 clusters.

yasker:
@sowmyav27 you're right, I misread it.

yasker:
@sowmyav27 In fact, the scenario #2 is the same as scenario #1. create b2 and delete it before DR volume can restore it. From Longhorn’s perspective, the last backup is empty and no extra backup in the backupstore, so it’s the same.

shuo-wu:
@sowmyav27 
I tried to write an integration test for this scenario https://github.com/longhorn/longhorn/issues/1380/#issuecomment-632147058.
but the situation of step 9 was impossible for the single cluster environment: 
Once the 2nd backup creation complete, the field `lastBackup` of both volumes will be updated immediately.

yasker:
LGTM. @sowmyav27 ?

PR #338, opened 2020-05-23
Improve upgrade test to run multiple pods

meldafrawi:
Longhorn/Longhorn#1391

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -3298,15 +3298,20 @@ def expand_attached_volume(client, volume_name):
 
 def prepare_pod_with_data_in_mb(
         client, core_api, pod_make, volume_name, volume_size=str(1*Gi),
-        data_path="/data/test", data_size_in_mb=DATA_SIZE_IN_MB_1):  # NOQA:
+        data_path="/data/test", data_size_in_mb=DATA_SIZE_IN_MB_1,
+        add_liveness_prope=True):  # NOQA:
     pod_name = volume_name + "-pod"
     pv_name = volume_name + "-pv"
     pvc_name = volume_name + "-pvc"
 
     pod = pod_make(name=pod_name)
-    pod_liveness_probe_spec = get_liveness_probe_spec(initial_delay=1,
-                                                      period=1)
-    pod['spec']['containers'][0]['livenessProbe'] = pod_liveness_probe_spec
+
+    if add_liveness_prope is True:
+        pod_liveness_probe_spec = \
+            get_liveness_probe_spec(initial_delay=1,
+                                    period=1)
+        pod['spec']['containers'][0]['livenessProbe'] = \
+            pod_liveness_probe_spec
 
     volume = create_and_check_volume(client, volume_name,
                                      num_of_replicas=3, size=volume_size)
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -3324,3 +3324,11 @@ def prepare_pod_with_data_in_mb(
            stderr=True, stdin=False, stdout=True, tty=False)
 
     return pod_name, pv_name, pvc_name, md5sum
+
+
+@pytest.fixture
+def settings_reset():
+    yield
+
+    client = get_longhorn_api_client()
+    reset_settings(client)
--- a/manager/integration/tests/test_upgrade.py
+++ b/manager/integration/tests/test_upgrade.py
@@ -1,5 +1,6 @@
 import pytest
 import subprocess
+import time
 
 from common import SIZE
 from common import volume_name  # NOQA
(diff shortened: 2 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -3287,15 +3287,20 @@ def expand_attached_volume(client, volume_name):
 
 def prepare_pod_with_data_in_mb(
         client, core_api, pod_make, volume_name, volume_size=str(1*Gi),
-        data_path="/data/test", data_size_in_mb=DATA_SIZE_IN_MB_1):  # NOQA:
+        data_path="/data/test", data_size_in_mb=DATA_SIZE_IN_MB_1,
+        add_liveness_prope=True):  # NOQA:
```
yasker, line 3291: The title of the commit is wrong. This part is not a fix for a bug.

manager/integration/tests/test_upgrade.py
```diff
@@ -31,7 +48,7 @@ def longhorn_upgrade(image_tag):
 
 
 @pytest.mark.upgrade
-def test_upgrade(upgrade_image_tag, volume_name):  # NOQA
+def test_upgrade(upgrade_image_tag, volume_name, pod_make, statefulset, storage_class):  # NOQAA
```
yasker, line 51: "NOQAA"? How did this pass flake8...

manager/integration/tests/test_upgrade.py
```diff
@@ -31,7 +48,7 @@ def longhorn_upgrade(image_tag):
...
     """
     Test Longhorn upgrade
```
yasker, line 54: Add prerequisite. 

As I mentioned, `auto-salvage` should be disabled here.

manager/integration/tests/test_upgrade.py
```diff
@@ -45,39 +62,111 @@ def test_upgrade(upgrade_image_tag, volume_name):  # NOQA
...
+        name=statefulset_name,
+        namespace='default',
+        body={
+            'spec': {
+                'replicas': replicas
+            }
+        })
 
-    volume = client.by_id_volume(volume_name)
+    delete_and_wait_pod(core_api, pod_name)
```
yasker, line 125: Have we checked the pod's data?

manager/integration/tests/test_upgrade.py
```diff
@@ -45,39 +62,111 @@ def test_upgrade(upgrade_image_tag, volume_name):  # NOQA
...
 
-    data = write_volume_random_data(volume)
+    for sspod_info in statefulset_pod_info:
+        sspod = core_api.read_namespaced_pod(name=sspod_info['pod_name'],
+                                             namespace='default')
+        assert \
+            sspod.status.container_statuses[0].restart_count == 0
 
-    assert longhorn_upgrade(upgrade_image_tag)
+    statefulset['spec']['replicas'] = replicas = 0
```
yasker, line 113: Have we checked the data of statefulset?

manager/integration/tests/test_upgrade.py
```diff
@@ -45,39 +62,111 @@ def test_upgrade(upgrade_image_tag, volume_name):  # NOQA
...
+        body={
+            'spec': {
+                'replicas': replicas
+            }
+        })
 
-    volume = client.by_id_volume(volume_name)
+    delete_and_wait_pod(core_api, pod_name)
 
+    volume = client.by_id_volume(volume_name)
```
yasker, line 127: Check the volume data?

manager/integration/tests/test_upgrade.py
yasker: Also, update the skeleton.

manager/integration/tests/test_upgrade.py
```diff
@@ -45,39 +62,111 @@ def test_upgrade(upgrade_image_tag, volume_name):  # NOQA
...
+    apps_api.patch_namespaced_stateful_set(
+        name=statefulset_name,
+        namespace='default',
+        body={
+            'spec': {
+                'replicas': replicas
+            }
+        })
+
+    wait_statefulset(statefulset)
```
yasker, line 159: We need to check the statefulset data. (do we have a way to do it?)

manager/integration/tests/test_upgrade.py
yasker: We need to make sure data is correct after the manager upgrade but before the engine upgrade.

After the engine upgrade, write more data to the volume, then upgrade the engine.

yasker:
The titles of two commits need to be updated as well. It doesn't indicate what's the changes and why we made them.

manager/integration/tests/common.py
meldafrawi: fixed

manager/integration/tests/test_upgrade.py
meldafrawi: fixed

manager/integration/tests/test_upgrade.py
meldafrawi: Updated

manager/integration/tests/test_upgrade.py
meldafrawi: Updated it to check data before/after engine upgrade

manager/integration/tests/test_upgrade.py
meldafrawi: fixed

manager/integration/tests/test_upgrade.py
meldafrawi: yes, fixed

manager/integration/tests/test_upgrade.py
```diff
@@ -31,53 +53,177 @@ def longhorn_upgrade(image_tag):
 
 
 @pytest.mark.upgrade
-def test_upgrade(upgrade_image_tag, volume_name):  # NOQA
+def test_upgrade(upgrade_image_tag, volume_name, pod_make, statefulset, storage_class):  # NOQA
     """
     Test Longhorn upgrade
 
+    Prerequisite:
+      - Disable Auto Salvage Setting
```
yasker, line 61: Does this test case has a side-effect of disabling the auto salvage setting?

meldafrawi: added `reset_settings` to the end of the test.

yasker: That won't work if the test failed.

PR #339, opened 2020-05-26
Fix the nightly test failures

shuo-wu:
longhorn/longhorn#1402

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -1069,38 +1069,45 @@ def test_node_default_disk_added_back_with_extra_disk_unmounted(client):  # NOQA
     [Node] Test adding default disk back with extra disk is unmounted
     on the node
 
-    1. Create a host disk `extra_disk` using Longhorn volume and
-    attach it to node 1.
-    2. Delete the default disk on node 1.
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_node.py
```diff
@@ -1069,38 +1069,45 @@ def test_node_default_disk_added_back_with_extra_disk_unmounted(client):  # NOQA
...
-    And wait for it becoming "Unschedulable".
-    4. Create and add the default disk back on node 1.
-    5. Wait and verify the default disk should become "Schedulable".
-    6. Mount extra disk back on node 1.
-    7. Wait and verify this extra disk should become "Schedulable".
-    8. Delete the host disk `extra_disk`.
+    1. Clean up all disks on node 1.
+    2. Recreate the default disk with "allowScheduling" disabled for
+       node 1.
+    3. Create a Longhorn volume and attach it to node 1.
```
yasker, line 1075: We have three nodes, with hard anti-affinity set to true and one disk set to scheduling disabled. How did this operation succeed?

yasker: Oh, `create_host_disk` only creates the volume with one replica.

PR #342, opened 2020-05-27
setting: Test case for Guaranteed Engine CPU setting

boknowswiki:
This test case covers guarantted engine cpu setting with both valid and
invalid settings, also verifies the requests cpu value.

https://github.com/longhorn/longhorn/issues/1318

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -137,6 +137,8 @@ SETTING_AUTO_SALVAGE = "auto-salvage"
 SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY = "replica-soft-anti-affinity"
 SETTING_REPLICA_ZONE_SOFT_ANTI_AFFINITY = "replica-zone-soft-anti-affinity"
 
+SETTING_GUARANTEED_ENGINE_CPU = "guaranteed-engine-cpu"
+
 CSI_UNKNOWN = 0
 CSI_TRUE = 1
 CSI_FALSE = 2
@@ -2336,6 +2338,23 @@ def reset_settings(client):
         print("Exception when update Auto Salvage setting",
               auto_salvage_setting, e)
 
+    guaranteed_engine_cpu_setting = \
+        client.by_id_setting(SETTING_GUARANTEED_ENGINE_CPU)
+    try:
+        client.update(guaranteed_engine_cpu_setting,
+                      value="0.25")
+    except Exception as e:
+        print("Exception when update "
+              "Guaranteed Engine CPU setting",
+              guaranteed_engine_cpu_setting, e)
+
+    instance_managers = client.list_instance_manager()
+    core_api = get_core_api_client()
+    # Wait for the current instance manager running
+    for im in instance_managers:
+        wait_for_instance_manager_desire_state(client, core_api,
+                                               im.name, "Running", True)
+
 
 def reset_engine_image(client):
     core_api = get_core_api_client()
@@ -3552,3 +3571,34 @@ def wait_for_pod_restart(core_api, pod_name, namespace="default"):
 
         time.sleep(RETRY_INTERVAL)
     assert pod_restarted
+
+
+def wait_for_instance_manager_desire_state(client, core_api, im_name,
+                                           state, desire=True):
+    for i in range(RETRY_COUNTS):
+        im = client.by_id_instance_manager(im_name)
+        try:
+            pod = core_api.read_namespaced_pod(name=im_name,
+                                               namespace=LONGHORN_NAMESPACE)
+        except Exception as e:
+            # Continue with pod restarted case
+            if e.reason == "Not Found":
+                time.sleep(RETRY_INTERVAL)
+                continue
+            # Report any other error
+            else:
+                assert(not e)
+        if desire:
+            if im.currentState == state.lower() and pod.status.phase == state:
+                break
+        else:
+            if im.currentState != state.lower() and pod.status.phase != state:
+                break
+        time.sleep(RETRY_INTERVAL)
+    if desire:
+        assert im.currentState == state.lower()
+        assert pod.status.phase == state
+    else:
+        assert im.currentState != state.lower()
+        assert pod.status.phase != state
+    return im
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -9,9 +9,14 @@ from common import (  # NOQA
     write_volume_random_data, check_volume_data,
     get_default_engine_image,
     wait_for_engine_image_state,
+    wait_for_instance_manager_desire_state,
+    generate_volume_name,
+    wait_for_volume_condition_scheduled,
+    client, core_api, settings_reset,
 
     LONGHORN_NAMESPACE, SETTING_TAINT_TOLERATION,
-    RETRY_COUNTS, RETRY_INTERVAL_LONG,
+    RETRY_COUNTS, RETRY_INTERVAL_LONG, SETTING_GUARANTEED_ENGINE_CPU,
+    SIZE, RETRY_INTERVAL
 )
 
 from test_infra import wait_for_node_up_longhorn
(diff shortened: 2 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -3337,3 +3339,36 @@ def settings_reset():
 
     client = get_longhorn_api_client()
     reset_settings(client)
+
+
+def wait_for_instance_manager_desire_state(client, core_api, im_name,
+                                           state, desire=True):
+    time.sleep(2 * RETRY_INTERVAL)
```
yasker, line 3346: This is kind of random. Why do we need this?

manager/integration/tests/common.py
```diff
@@ -3337,3 +3339,36 @@ def settings_reset():
...
+            pod = core_api.read_namespaced_pod(name=im_name,
+                                               namespace=LONGHORN_NAMESPACE)
+        except Exception as e:
+            # Continue with pod restarted case
+            if e.reason == "Not Found":
+                time.sleep(RETRY_INTERVAL)
+                continue
+            # Report any other error
+            else:
+                print("Failed to get pod based on name: %s, namespace: %s"
```
yasker, line 3359: Do we expect any other exception to be raised? Sounds like we should `assert` then fail the test.

manager/integration/tests/test_settings.py
```diff
@@ -200,7 +204,6 @@ def check_tolerations_set(current_toleration_list, set_tolerations):
     return current_tolerations == set_tolerations
 
 
-@pytest.mark.skip(reason="TODO")
 def test_setting_guaranteed_engine_cpu():
```
yasker, line 207: Why don't use the `client` and `core_api` fixture? Otherwise, we cannot guarantee it's cleaned up correctly.

yasker: Also, in the `reset_setting` fixture, we should reset the setting back to the default value, and wait for the IM to be recreated correctly.

manager/integration/tests/test_settings.py
```diff
@@ -228,4 +231,98 @@ def test_setting_guaranteed_engine_cpu():
...
+    org_setting = setting
+
+    # Get current running instance managers
+    instance_managers = client.list_instance_manager()
+
+    # Set an invalid value, and it should return error
+    try:
+        client.update(setting, value="xxx")
+    except Exception as e:
+        print("Exception when update "
```
yasker, line 248: We should expect an exception. See e.g. https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_basic.py#L109

manager/integration/tests/test_settings.py
```diff
@@ -228,4 +231,98 @@ def test_setting_guaranteed_engine_cpu():
...
+                                        "100m", instance_managers,
+                                        "Running", True)
+
+    # Update guaranteed engine cpu setting to 0
+    guaranteed_engine_cpu_setting_check(client, core_api, setting, "0",
+                                        None, instance_managers,
+                                        "Running", True)
+
+    # Update guaranteed engine cpu setting to 0.2
+    guaranteed_engine_cpu_setting_check(client, core_api, setting, "0.2",
```
yasker, line 263: The input should be `200m` instead of `0.2`. We need to test different types of inputs.

manager/integration/tests/test_settings.py
```diff
@@ -228,4 +231,98 @@ def test_setting_guaranteed_engine_cpu():
...
+        print("Exception when update "
+              "Guaranteed Engine CPU to", org_setting.value, e)
+        assert(not e)
+
+
+def guaranteed_engine_cpu_setting_check(client, core_api, setting, val,
+                                        cpu_val, instance_managers,
+                                        state, desire):
+    """
+    We chech if intance managers are in the desired state with
```
yasker, line 303: Typos. `check`, `instance`.

And the comment doesn't explain what is `desire` variable mean, It's not obvious from the function signature.

manager/integration/tests/test_settings.py
```diff
@@ -228,4 +231,98 @@ def test_setting_guaranteed_engine_cpu():
...
+    correct setting
+    """
+    # Update guaranteed engine cpu setting
+    try:
```
yasker, line 307: Don't need the `try`. Those updates shouldn't fail. If those failed, then the test should fail.

manager/integration/tests/test_settings.py
```diff
@@ -228,4 +231,98 @@ def test_setting_guaranteed_engine_cpu():
...
+        client.update(setting, value=val)
+    except Exception as e:
+        print("Exception when update "
+              "Guaranteed Engine CPU to ", val, e)
+        assert(not e)
+
+    # Wait for the current instance manager running
```
yasker, line 314: Remove this comment. It's obvious from the function name. Also, it's not waiting for it to be `running`.

manager/integration/tests/test_settings.py
```diff
@@ -228,4 +231,98 @@ def test_setting_guaranteed_engine_cpu():
...
+    volume = wait_for_volume_healthy(client, vol_name)
+    assert len(volume.replicas) == 3
+
+    data = write_volume_random_data(volume)
+    check_volume_data(volume, data)
+
+    # Cleanup volume
+    cleanup_volume(client, volume)
+
+    # restore original setting
```
yasker, line 290: Cleanup should be done in the fixture.

manager/integration/tests/common.py
boknowswiki: No, we don't expect any other exception here. I will add the assert there to fail the test.

manager/integration/tests/test_settings.py
boknowswiki: Sure, let me add the fixture.

manager/integration/tests/test_settings.py
boknowswiki: Oh, i see. I will update it.

manager/integration/tests/test_settings.py
boknowswiki: will fix them.

manager/integration/tests/common.py
boknowswiki: Giving sometime before we check the status. Say if we check the status immediately after we update the setting, I am not sure if the status were still in the previous states.

yasker: For this function, if the caller is not sure the state is good to check for the `wait_for`, the caller should take steps to ensure the state is good to check before calling the function. In this case, at least move the `sleep` into the caller.

Normally we shouldn't have any random `sleep` call since you won't know how long is enough given a certain environment. That's what introduced race condition. In this case, we can check the previous pods' UUID before and after, to make sure the change is happening.

manager/integration/tests/test_settings.py
boknowswiki: This is the extra volume I created during the test, I think I need to clean it up, right?

boknowswiki:
Verified:
`$ klogs -f longhorn-test
+ flake8 .
+ py.test -v . -x -s -k test_setting_guaranteed_engine_cpu
platform linux -- Python 3.7.4, pytest-5.3.1, py-1.8.1, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
collecting ... collected 118 items / 117 deselected / 1 selected

test_settings.py::test_setting_guaranteed_engine_cpu PASSED

================ 1 passed, 117 deselected in 116.06s (0:01:56) =================
`

manager/integration/tests/test_settings.py
yasker: The volume should be cleaned up by the `client` fixture automatically.

The setting below should be updated back to the default setting as well.

So those two cleanups won't be necessary.

manager/integration/tests/common.py
yasker: Don't need this print line. `assert` will pinpoint the line as well as show up the content of `e`.

manager/integration/tests/test_settings.py
```diff
@@ -39,9 +44,9 @@ def test_setting_toleration():
     14. Attach the volume and validate `data2`.
     15. Generate and write `data3` to the volume.
     """
-    client = get_longhorn_api_client()
+    client = get_longhorn_api_client()  # NOQA
```
yasker, line 47: Why do we need `# NOQA` here and lines below?

boknowswiki: pytest will complain about redefined client, since at the beginning we imported it from common.

PR #347, opened 2020-06-01
fix test_single_replica_restore_failure

meldafrawi:
Longhorn/Longhorn#1432

sometimes, replica instanceManagerName field is empty, causing crash replica function to fail.
we need to wait for volume state to be healthy, before crashing replicas

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -47,6 +47,7 @@ from common import delete_random_backup_block
 from common import wait_for_volume_condition_restore
 from common import wait_for_pod_restart
 from common import crash_engine_process_with_sigkill
+from common import wait_for_volume_healthy_no_frontend
 
 from common import SIZE, VOLUME_RWTEST_SIZE, EXPAND_SIZE, Gi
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -1242,6 +1243,8 @@ def test_single_replica_restore_failure(client, core_api, volume_name, pod_make)
     res_volume = client.by_id_volume(res_name)
     assert res_volume.ready is False
 
+    res_volume = wait_for_volume_healthy_no_frontend(client, res_name)
+
     crash_replica_processes(client, core_api, res_name,
```
yasker, line 1248: We better add an `assert` in `crash_replica_processes` and similar functions that the instance manager must exist before continue. It can be a separate commit.

meldafrawi: fixed in a separate commit.

yasker: hmm? which one?

yasker:
LGTM. Can you add the fix explanation to the commit?

meldafrawi:
commit message updated.

PR #348, opened 2020-06-04
Add skeleton for backup block deletion test

joshimoo:
We want to make sure that we only delete non referenced backup blocks,
we also don't want to delete blocks while there other backups in progress.
The reason for this is that we don't yet know which blocks are required by
the in progress backup, so blocks deletion could lead to a faulty backup.

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -600,6 +600,48 @@ def backup_status_for_unavailable_replicas_test(clients, volume_name, # NOQA
     cleanup_volume(client, volume)
 
 
+@pytest.mark.skip(reason="TODO")
+def test_backup_block_deletion():  # NOQA
+    """
+    Test backup block deletion
+
+    Context:
+
+    We want to make sure that we only delete non referenced backup blocks,
+    we also don't want to delete blocks while there other backups in progress.
+    The reason for this is that we don't yet know which blocks are required by
+    the in progress backup, so blocks deletion could lead to a faulty backup.
+
+    Setup:
+
+    1. Setup NFS backupstore since we can manipulate the content easily
+
+    Steps:
+
+    1.  Create a volume and attach to the current node
+    2.  Write 4 MB to the beginning of the volume (2 x 2MB backup blocks)
+    3.  Create backup(1) of the volume
+    4.  Overwrite the first of the backup blocks of data on the volume
+    5.  Create backup(2) of the volume
+    6.  Overwrite the first of the backup blocks of data on the volume
+    7.  Create backup(3) of the volume
+    8.  Verify backup block count == 4
+        assert volume["DataStored"] == str(BLOCK_SIZE * expected_count)
+        assert count of *.blk files for that volume == expected_count
+    9.  Create an artificial in progress backup.cfg file
+        json.dumps({"Name": name, "VolumeName": volume, "CreatedTime": ""})
+    10. Delete backup(2)
+    11. Verify backup block count == 4 (because of the in progress backup)
+    12. Delete the artificial in progress backup.cfg file
+    13. Delete backup(1)
+    14. Verify backup block count == 2
+    15. Delete backup(3)
+    16. Verify backup block count == 0
+    17. Delete the backup volume
+    18. Cleanup the volume
+    """
+    pass
+
 @pytest.mark.coretest   # NOQA
 def test_backup(clients, volume_name):  # NOQA
     """


manager/integration/tests/test_basic.py
```diff
@@ -600,6 +600,45 @@ def backup_failure_predicate(b):
...
+    the in progress backup, so blocks deletion could lead to a faulty backup.
+
+    Setup:
+
+    1. Create a volume and attach to the current node
+    2. Setup NFS backupstore since we can manipulate the content easily
+
+    Steps:
+
+    1.  Write 2 backup blocks of data on to the volume
```
yasker, line 622: Be a little more specific, e.g. 
1. Create and attach the volume on the current node.
2. Write 4 MB to the beginning of the volume (2 x 2MB backup blocks)
3. Create backup1 etc
…
7. Get the backup volume. Check to make sure the volume has 4 blocks.

manager/integration/tests/test_basic.py
```diff
@@ -600,6 +600,45 @@ def backup_failure_predicate(b):
...
+    2.  Create backup(1) of the volume
+    3.  Overwrite the first of the backup blocks of data on the volume
+    4.  Create backup(2) of the volume
+    5.  Overwrite the first of the backup blocks of data on the volume
+    6.  Create backup(3) of the volume
+    7.  Verify backup block count == 4
+    8.  Create an artificial in progress backup.cfg file
```
yasker, line 629: How to create the artificial backup.cfg?

manager/integration/tests/test_basic.py
```diff
@@ -600,6 +600,45 @@ def backup_failure_predicate(b):
...
+
+    Steps:
+
+    1.  Write 2 backup blocks of data on to the volume
+    2.  Create backup(1) of the volume
+    3.  Overwrite the first of the backup blocks of data on the volume
+    4.  Create backup(2) of the volume
+    5.  Overwrite the first of the backup blocks of data on the volume
+    6.  Create backup(3) of the volume
+    7.  Verify backup block count == 4
```
yasker, line 628: Also check the backup directory, search for `.blk` make sure it matched the block count.

manager/integration/tests/test_basic.py
joshimoo: I can make these changes, but to clarify my understanding of test steps description.
Is the objective of the test steps to literally describe the implementation of the python code?

For the block count check one wants to check the volume.cfg as well as the blocks on disk, like I do in the engine integration test. Should I include these details in the descriptions? since they are implementation details?

manager/integration/tests/test_basic.py
joshimoo: One can place an in progress backup.cfg into the backups folder for that volume via the nfs mount. An example can be found in the engine integration test, except for the nfs part.
An example of the nfs operations can be found in [here](https://github.com/longhorn/longhorn-tests/blob/099d46e9182e081dd76bb45011966d49968d0989/manager/integration/tests/test_basic.py#L1153)

manager/integration/tests/test_basic.py
joshimoo: For the block count check one wants to check the volume.cfg as well as the blocks on disk, like I do in the engine integration test. Should I include these details in the descriptions? since they are implementation details?

manager/integration/tests/test_basic.py
yasker: Yes, we should describe the implementation as much as possible, to make sure it's implemented according to the skeleton.

Yes, we can refer to those detail when encountered in the skeleton for the first time, then later the same check operation should follow the same established pattern.

manager/integration/tests/test_basic.py
yasker: Don't need details on how to modify NFS backupstore, but need to specify what to be put in the file (e.g. a blank file? a file with some content?).

manager/integration/tests/test_basic.py
yasker: Yes, include them when encountered for the first time.

yasker:
LGTM.

Ping @khushboo-rancher

joshimoo:
@khushboo-rancher for writing custom backup data have a look [here](https://github.com/longhorn/longhorn-tests/blob/f138fe9829d0168a282f54702d7cc27d6cff4220/manager/integration/tests/test_basic.py#L1798-L1801)

PR #350, opened 2020-06-05
Cleanup  PV/PVC after test completion - test_ha.py

khushboo-rancher:
Added deletion of PV/PVC in pod_make fixture.
https://github.com/longhorn/longhorn/issues/1413

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -727,6 +727,18 @@ def pod_make(request):
             except Exception as e:
                 print("Exception when waiting for pod deletion", e)
                 return
+            try:
+                volume_details = pod_manifest['spec']['volumes'][0]
+                pvc_name = volume_details['persistentVolumeClaim']['claimName']
+                delete_and_wait_pvc(api, pvc_name)
+            except Exception as e:
+                print("Exception when waiting for PVC deletion", e)
+            try:
+                pv = wait_and_get_pv_for_pvc(api, pvc_name)
+                pv_name = pv.metadata.name
+                delete_and_wait_pv(api, pv_name)
+            except Exception as e:
+                print("Exception when waiting for PV deletion", e)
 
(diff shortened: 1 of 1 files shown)


manager/integration/tests/common.py
```diff
@@ -727,6 +727,18 @@ def finalizer():
             except Exception as e:
                 print("Exception when waiting for pod deletion", e)
                 return
+            try:
```
yasker, line 730: We have fixtures for `pv` and `pvc`, why doesn't work as expected? Want to understand the reason for failure before fixing it.

khushboo-rancher: There are some tests which are not using fixtures for pv/pvc but they are getting created in the definition of tests. 
```
test_salvage_auto_crash_all_replicas(client, core_api, volume_name, pod_make):
pod_name, pv_name, pvc_name, md5sum = \
        prepare_pod_with_data_in_mb(client, core_api, pod_make, volume_name)
```
These tests are using ```pod_make``` fixture to get pod manifest and deleting pod later. So, instead of changing tests to make them use pv/pvc fixture, a try block is added to delete pv/pvc here. This way we don't have to disturb existing tests and pv/pvc can be deleted in pod_make fixture.

yasker: Noticed `prepare_pod_with_data_in_mb`, `create_pvc_for_volume` as well as `create_pv_for_volume` are in fact bad practise. We should always try to use the fixture first. Not sure if all the references of the above functions are really needed.

Can you file an issue to replace `create_pvc_for_volume` and `create_pv_for_volume` with `pv/pvc` fixtures in order to properly clean the objects up? We can have this PR merged first but we will still want to fix the problem from the root.

khushboo-rancher: https://github.com/longhorn/longhorn/issues/1462

PR #358, opened 2020-06-24
Replace create_pv_for_volume and create_pvc_for_volume with fixture

khushboo-rancher:
Replaced the create_pv_for_volume and create_pvc_for_volume in method 'prepare_pod_with_data_in_mb'.
Identified few more cases where 'create_pv_for_volume' and 'create_pvc_for_volume' are used(not straight though). Will change them too if this approach is right.

https://github.com/longhorn/longhorn/issues/1462

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -15,7 +15,7 @@ from common import expand_attached_volume, check_block_device_size
 from common import write_volume_data, generate_random_data
 from common import wait_for_rebuild_complete
 from common import disable_auto_salvage # NOQA
-from common import pod_make, pod  # NOQA
+from common import pod_make, pod, csi_pv, pvc  # NOQA
 from common import create_pv_for_volume
 from common import create_pvc_for_volume
 from common import create_pvc_spec
@@ -25,7 +25,6 @@ from common import wait_for_volume_healthy, wait_for_volume_degraded
 from common import get_pod_data_md5sum
 from common import wait_for_pod_remount
 from common import delete_and_wait_pod
-from common import delete_and_wait_pvc, delete_and_wait_pv
 from common import wait_for_rebuild_start
 from common import prepare_pod_with_data_in_mb
 from common import wait_for_replica_running
@@ -53,6 +52,7 @@ from common import SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY
 
 from backupstore import backupstore_cleanup
 from backupstore import backupstore_delete_random_backup_block
+
 
 @pytest.mark.coretest   # NOQA
 def test_ha_simple_recovery(client, volume_name):  # NOQA
@@ -426,8 +426,8 @@ def test_ha_recovery_with_expansion(client, volume_name):   # NOQA
 
 
 def wait_pod_for_auto_salvage(
-        client, core_api, volume_name, pod_name, pv_name, pvc_name,   # NOQA
-        original_md5sum, data_path="/data/test"):
+        client, core_api, volume_name, pod_name, original_md5sum,  # NOQA
+        data_path="/data/test"):
     # this line may fail if the recovery is too quick
     common.wait_for_volume_faulted(client, volume_name)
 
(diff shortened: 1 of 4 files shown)


yasker:
The general approach looks good. Btw, flake8 failed.

khushboo-rancher:
Removed the pv/pvc cleanup from ```pod_make``` fixture as pv/pvc is getting cleaned up in their respective fixtures.
There are few test cases which deals with multiple pv/pvc creation, only in those case deletion is taken care explicitly in test case itself because if we change the fixtures to become dynamically called, it will have impact on the existing ones.
@shuo-wu Updated the volume size as 1 Gi.
@yasker Fixed the flake8 issue.

shuo-wu:
In order to deal with the test cases containing multiple pv/pvc creation, can we implement an extra fixture that will remove all pods, pvc, and pv during the teardown?

khushboo-rancher:
I've updated the code. In the case of multiple PVC/PV creation, the pod_make fixture will delete the PVC/PV which will not get deleted by csi_pv and pvc fixtures.
Earlier I removed the deletion of PV/PVC from pod_make fixture which is added back now.

manager/integration/tests/test_ha.py
```diff
@@ -755,16 +751,10 @@ def test_rebuild_with_restoration(
 
     # cleanup
     backupstore_cleanup(client)
-    delete_and_wait_pod(core_api, original_pod_name)
-    delete_and_wait_pvc(core_api, original_pvc_name)
-    delete_and_wait_pv(core_api, original_pv_name)
-    delete_and_wait_pod(core_api, restore_pod_name)
-    delete_and_wait_pvc(core_api, restore_pvc_name)
```
shuo-wu, line 762: The PV and PVC of the restore volume are created by `create_pv_for_volume` and `create_pvc_for_volume`. Need to check if this cleanup can be removed.

manager/integration/tests/test_ha.py
```diff
@@ -855,16 +845,11 @@ def test_rebuild_with_inc_restoration(
 
     # cleanup
     backupstore_cleanup(client)
-    delete_and_wait_pod(core_api, std_pod_name)
-    delete_and_wait_pvc(core_api, std_pvc_name)
-    delete_and_wait_pv(core_api, std_pv_name)
-    delete_and_wait_pod(core_api, dr_pod_name)
-    delete_and_wait_pvc(core_api, dr_pvc_name)
```
shuo-wu, line 862: Same above.

shuo-wu:
IIRC, we need to use `create_pv_for_volume` and `create_pvc_for_volume` to test if our Longhorn APIs work fine. For those tests, we cannot use the fixtures `csi_pv` and `pvc` then remove the manual cleanup. As we discussed previously, can we add a finalizer in `client` to find & remove all Longhorn related PVCs & PVs?

shuo-wu:
BTW, you can create a separate commit for the data size modification, and mention issue 1609 in this PR and the commit.

manager/integration/tests/test_ha.py
khushboo-rancher: In this case, the pod is getting created with pod_make fixture and as we have PV/PVC cleanup there. These delete methods are not required.

manager/integration/tests/test_ha.py
khushboo-rancher: In this case, the pod is getting created with pod_make fixture and as we have PV/PVC cleanup there. These delete methods are not required.

khushboo-rancher:
I've removed the data size related changes from here. 
Reg creating finalizer in ```client``` fixture, I analyzed it more and found that those PV/PVC are getting deleted in ```pod_make``` fixture because in the test the second pod is getting created by changing name in ```pod_make```. As we have pv/pvc cleanup in ```pod_make``` fixture, we don't need to add extra cleanup in ```client``` fixture.
```
restore_pod_name = restore_volume_name + "-pod"
restore_pv_name = restore_volume_name + "-pv"
restore_pvc_name = restore_volume_name + "-pvc"
restore_pod = pod_make(name=restore_pod_name)
create_pv_for_volume(client, core_api, restore_volume, restore_pv_name)
create_pvc_for_volume(client, core_api, restore_volume, restore_pvc_name)
restore_pod['spec']['volumes'] = [create_pvc_spec(restore_pvc_name)]
create_and_wait_pod(core_api, restore_pod)
```

PR #359, opened 2020-06-25
Fix a BUG related to the PV cleanup in pod_make()

shuo-wu:
The PV may be already deleted before calling the finalizer, hence
we cannot use the wait function to find out the PV.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -738,9 +738,16 @@ def pod_make(request):
             except Exception as e:
                 print("Exception when waiting for PVC deletion", e)
             try:
-                pv = wait_and_get_pv_for_pvc(api, pvc_name)
-                pv_name = pv.metadata.name
-                delete_and_wait_pv(api, pv_name)
+                found = False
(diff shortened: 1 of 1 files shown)


khushboo-rancher:
I've deleted the clean up code of pv/pvc in https://github.com/longhorn/longhorn-tests/pull/358 because once that will get merged the pv/pvc will be deleted in their respective fixture. I think this will be no longer required.

shuo-wu:
> I've deleted the clean up code of pv/pvc in #358 because once that will get merged the pv/pvc will be deleted in their respective fixture. I think this will be no longer required.

But we cannot assume that the pod/pvc/pv always exists when the finalizer is invoked and everything works as expected. We need to make sure this finalizer won't get stuck there if somehow the pod/pvc/pv has been removed in advance.

shuo-wu:
@khushboo-rancher Just in case, do we still need this PR?

khushboo-rancher:
@shuo-wu Yes, we'll need this.

khushboo-rancher (approved):
LGTM

PR #364, opened 2020-07-09
backupstore: Test case skeleton for validation creating DRV option

boknowswiki:
This test case is to cover creating DRV option is valid
when the latest backup of a volume is deleted

https://github.com/longhorn/longhorn/issues/1381

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -1038,6 +1038,24 @@ def backup_labels_test(client, random_labels, volume_name, size=SIZE, base_image
     cleanup_volume(client, volume)
 
 
+@pytest.mark.skip(reason="TODO")
+def test_create_DRV_option_validation(client, core_api, volume_name, pod):  # NOQA
+    """
+    Test create disaster recovery volume option
+
+    1. Create a volume and attach it to the current node
+    2. Generate `data0`, write to the volume, make a backup `backup0`
+    3. Verify Backup Volume's `lastBackup` is `backup0`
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -1038,6 +1038,22 @@ def backup_labels_test(client, random_labels, volume_name, size=SIZE, base_image
...
 
 
+@pytest.mark.skip(reason="TODO")
+def test_create_DRV_option_validation(client, core_api, volume_name, pod):  # NOQA
+    """
+    Test create disaster recovery volume option
+
+    1. Create a volume and attach it to the current node
+    2. Generate `data0`, write to the volume, make a backup `backup0`
+    3. Verify DR volume's `lastBackup` is `backup0`
```
yasker, line 1048: Not DR volume. Backup Volume's `lastBackup`

manager/integration/tests/test_basic.py
```diff
@@ -1038,6 +1038,22 @@ def backup_labels_test(client, random_labels, volume_name, size=SIZE, base_image
...
+    4. Generate `data1`, write to the volume, make a backup `backup1`
+    5. Verify DR volume's `lastBackup` is `backup1`
+    6. Delete backup `backup1` and verify DR volume's `lastBackup` is
+    `backup0`
+    """
```
yasker, line 1053: also add creating a DR volume, activate it and verify the data to ensure the data is indeed from `backup0`.

PR #365, opened 2020-07-10
Add skeleton for backup locks test

joshimoo:
Longhorn #612

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -736,6 +736,47 @@ def test_backup_volume_list():  # NOQA
     pass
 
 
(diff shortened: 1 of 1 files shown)


yasker:
I think we should focus on more high-level operations for this test case, e.g. execute the create/delete, restore/delete etc at the same time.

joshimoo:
Good point I agree

> I think we should focus on more high-level operations for this test case, e.g. execute the create/delete, restore/delete etc at the same time.
> 
> The locking mechanism has already been covered in engine test and it's not user-facing change.

Good point, I agree will update the test descriptions.

joshimoo:
superseded by #longhorn/longhorn-tests/424

PR #366, opened 2020-07-10
tests: add test_backup_volume_list to make sure that listing volume does not fail even there are bad names in backups folder

PhanLe1010:
test_backup_volume_list() make sure that instead of erroring the backup list should ignore the badly named files in backups folder.

longhorn/longhorn#1410

### DIFF
--- a/manager/integration/tests/backupstore.py
+++ b/manager/integration/tests/backupstore.py
@@ -1,7 +1,7 @@
 import os
 import base64
 import hashlib
-
+import json
 
 from minio import Minio
 from minio.error import ResponseError
@@ -14,6 +14,9 @@ from common import is_backupTarget_s3
 from common import is_backupTarget_nfs
 from common import get_longhorn_api_client
 from common import wait_for_backup_volume_delete
+
+
+TEMP_FILE_PATH = "/tmp/temp_file"
 
 
 def backupstore_cleanup(client):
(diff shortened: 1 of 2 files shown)


manager/integration/tests/backupstore.py
```diff
@@ -213,6 +255,36 @@ def minio_write_backup_cfg_file(client, core_api, volume_name, backup_name, back
...
+    backup_target_setting = client.by_id_setting(SETTING_BACKUP_TARGET)
+    backupstore = backup_target_setting.value
+
+    if is_backupTarget_s3(backupstore):
+        return mino_delete_file_in_backupstore(client,
+                                               core_api,
+                                               prefix,
+                                               file_name)
+
+    elif is_backupTarget_nfs(backupstore):
```
shuo-wu, line 268: `else:`
Why not be consistent with `backupstore_create_file`?

manager/integration/tests/backupstore.py
```diff
@@ -170,6 +170,48 @@ def nfs_get_backup_blocks_dir():
     raise NotImplementedError
 
 
+def backupstore_create_file(client, core_api, prefix, file_name, data={}):
```
shuo-wu, line 173: The input args look a little bit weird at first glance. Can we use `(client, core_api, file_path, data={})` instead?
Same below.

manager/integration/tests/backupstore.py
```diff
@@ -170,6 +170,48 @@ def nfs_get_backup_blocks_dir():
...
+
+    secret_name = backup_target_credential_setting.value
+
+    minio_api = minio_get_api_client(client, core_api, secret_name)
+    bucket_name = minio_get_backupstore_bucket_name(client)
+
+    if len(data) == 0:
+        data = {"testkey": "test data from mino_create_file_in_backupstore()"}
+
+    temp_file_path = "/tmp/temp_file.cfg"
```
shuo-wu, line 199: It's better not to hard code `temp_file_path` in the util function.

manager/integration/tests/test_basic.py
```diff
@@ -733,7 +729,71 @@ def test_backup_volume_list():  # NOQA
...
+                            "backup_1234@failure.cfg")
+
+    verify_no_err()
+
+    backupstore_delete_file(client,
+                            core_api,
+                            prefix,
+                            "backup_1234@failure.cfg")
+
+    bv1 = client.by_id_backupVolume(volume1_name)
```
shuo-wu, line 788: duplicate cleanup since the test will call `backupstore_cleanup` at the end.
Same below.

shuo-wu:
And remember to correct the commit name.

manager/integration/tests/backupstore.py
PhanLe1010: Fixed

manager/integration/tests/backupstore.py
PhanLe1010: fixed

manager/integration/tests/backupstore.py
PhanLe1010: fixed

manager/integration/tests/test_basic.py
PhanLe1010: fixed

PhanLe1010:
corrected the commit name

manager/integration/tests/common.py
```diff
@@ -29,6 +29,7 @@
 EXPAND_SIZE = str(32 * Mi)
 VOLUME_NAME = "longhorn-testvol"
 DEV_PATH = "/dev/longhorn/"
+TEMP_FILE_PATH = "/tmp/temp_file.cfg"
```
shuo-wu, line 32: 1. This tmp file is used in `backupstore_create_file` only, right? I think you can put this definition in the file `backupstore.py`
2. The function `backupstore_create_file` itself is a common function and it's not only for creating invalid backup config files, hence we can remove the surffix `cfg` from the tmp file name used in the function.

PhanLe1010: Great point! Fixed. Thank you

PR #372, opened 2020-07-17
Refactor test to always use backup_delete / backup_volume_delete

joshimoo:
Since both deletion operations are async we need to always wait for
completion of the requested operation before proceeding with our tests.

longhorn/longhorn#710
longhorn/longhorn#1603
longhorn/longhorn#1604


yasker:
Flake8 check failed.

PR #376, opened 2020-07-27
Test skeleton for rwx volumes

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/1606

I've added a new file test_rwx.py and added a few test skeletons for rwx volumes.

### DIFF
--- a/manager/integration/tests/test_rwx.py
+++ b/manager/integration/tests/test_rwx.py
@@ -1,0 +1,178 @@
+import pytest
+'''
+Fixture required for this module to clean up longhorn-nfs-provisioner related
+resources.
+'''
+
+
+@pytest.mark.skip(reason="TODO")
+def test_rwx_with_statefulset_multi_pods():
+    """
+    Test writing of data in same volume from 2 pods
+
+    1. Create PodSecurityPolicy for longhorn-nfs-provisioner as sample yaml
+       https://raw.githubusercontent.com/longhorn/longhorn/master/examples
+       /rwx/01-security.yaml
+
+    2. Create serviceAccount, service, longhorn-nfs-provisioner deployment,
+       PVC and StorageClass as sample yaml https://raw.githubusercontent.com/
+       longhorn/longhorn/master/examples/rwx/02-longhorn-nfs-provisioner.yaml
+
+    3. Wait for longhorn-nfs-provisioner deployment to come up healthy.
+    4. Verify the volume named as the PV attached to longhorn-nfs-provisioner
+       `volume.ready == True`
+    5. Create a StatefulSet of 2 pods with VolumeClaimTemplate and
+       longhorn-nfs class in default namespace.
+    6. Wait for both pods to come up running.
+    7. Verify two folders get created with PV names in longhorn-nfs-provisioner
+       volume.
+    8. Write data in both pods and compute md5sum.
+    9. Compare md5sum of the data in longhorn-nfs-provisioner volume
+    """
+    pass
+
+
+@pytest.mark.skip(reason="TODO")
+def test_rwx_multi_statefulset_with_same_pvc():
+    """
+    Test writing of data with multiple pods using same PVC
+
+    1. Create PodSecurityPolicy for longhorn-nfs-provisioner as sample yaml
+       https://raw.githubusercontent.com/longhorn/longhorn/master/examples
+       /rwx/01-security.yaml
+
+    2. Create serviceAccount, service, longhorn-nfs-provisioner deployment,
+       PVC and StorageClass as sample yaml https://raw.githubusercontent.com/
+       longhorn/longhorn/master/examples/rwx/02-longhorn-nfs-provisioner.yaml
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_rwx.py
```diff
@@ -0,0 +1,152 @@
...
+    3. Wait for longhorn-nfs-provisioner deployment to come up healthy.
+    4. Verify the volume named as the PV attached to longhorn-nfs-provisioner
+       `volume.ready == True`
+    5. Create a StatefulSet of 1 pod with VolumeClaimTemplate and
+       longhorn-nfs class in default namespace.
+    6. Wait for StatefulSet to come up healthy.
+    7. Create another statefulset with same pvc which got created with first
+       statefulset.
+    8. Wait for statefulset to come up healthy.
+    9. Start writing 800 MB data in first stateful set and start writing 500 MB
```
joshimoo, line 87: The data should be written to two different files.

manager/integration/tests/test_rwx.py
```diff
@@ -0,0 +1,152 @@
...
+
+    3. Wait for longhorn-nfs-provisioner deployment to come up healthy.
+    4. Verify the volume named as the PV attached to longhorn-nfs-provisioner
+       `volume.ready == True`
+    5. Create a StatefulSet of 2 pods with VolumeClaimTemplate and
+       longhorn-nfs class in default namespace.
+    6. Wait for both pods to come up running.
+    7. Verify two folders get created with PV names in longhorn-nfs-provisioner
+       volume.
+    8. Write data in both pods and compute md5sum.
```
joshimoo, line 29: How do we do this?
Maybe map the export folder from the nfs provisioner?

manager/integration/tests/test_rwx.py
```diff
@@ -0,0 +1,152 @@
...
+       longhorn/longhorn/master/examples/rwx/02-longhorn-nfs-provisioner.yaml
+
+    3. Wait for longhorn-nfs-provisioner deployment to come up healthy.
+    4. Verify the volume named as the PV attached to longhorn-nfs-provisioner
+       `volume.ready == True`
+    5. Create a StatefulSet of 2 pods with VolumeClaimTemplate and
+       longhorn-nfs class in default namespace.
+    6. Wait for StatefulSet pods to come up healthy.
+    7. Write data and compute md5sum in both pods
+    8. Force delete pods.
```
joshimoo, line 116: What does force delete refer to in this case?

manager/integration/tests/test_rwx.py
```diff
@@ -0,0 +1,152 @@
...
+
+    3. Wait for longhorn-nfs-provisioner deployment to come up healthy.
+    4. Verify the volume named as the PV attached to longhorn-nfs-provisioner
+       `volume.ready == True`
+    5. Create a StatefulSet of 1 pod with VolumeClaimTemplate and
+       longhorn-nfs class in default namespace.
+    6. Wait for StatefulSet to come up healthy.
+    7. Write data and compute md5sum.
+    8. Shutdown the node where longhorn-nfs-provisioner is running.
+    9. Force delete the deployment in unknown state
```
joshimoo, line 146: Why force delete the deployment, you should be able to add a liveness probe which would restart the pod.
That should be sufficient to be able to use the newly brought up nfs-provisioner.

joshimoo:
@khushboo-rancher we should probably add a test with a regular scaled deployment.
If the pods come up on different nodes while they refer to the same PVC that should be sufficient evidence.

manager/integration/tests/test_rwx.py
khushboo-rancher: Having pods name and their data path, we can write data to pods.

manager/integration/tests/test_rwx.py
khushboo-rancher: Done

manager/integration/tests/test_rwx.py
khushboo-rancher: Actually force delete is not required, a delete will do. I wrote it to make sure this delete pod in any case.

manager/integration/tests/test_rwx.py
khushboo-rancher: Make sense. Modified it.

khushboo-rancher:
@joshimoo Added a test case with a regular scaled deployment

manager/integration/tests/test_rwx.py
```diff
@@ -0,0 +1,178 @@
...
+    11. Write more data to it and compute md5sum.
+    12. Check the data md5sum in longhorn-nfs-provisioner volume
+    """
+    pass
+
+
+@pytest.mark.skip(reason="TODO")
+def test_rwx_deployment_with_multi_pods():
+    """
+    Test moving of longhorn-nfs-provisioner pod from one node to another.
```
joshimoo, line 159: Title refers to the previous test.

PR #380, opened 2020-07-31
Increase the writing data size in few tests to minimize the failure

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/1614
https://github.com/longhorn/longhorn/issues/1609

Some tests fail while checking the run time rebuild status or while trying removing replica when the restoration is in progress because rebuilding/restoration for small data get completed quickly and sometimes doesn't get detected by code.
By increasing the data size a little, we will increase the time taken by the rebuilding/restoration process so that we can capture real-time status.

We can observe the integration test results for a few days to make sure these changes minimize the failure occurrence.


manager/integration/tests/test_ha.py
```diff
@@ -1429,8 +1431,7 @@ def test_volume_reattach_after_engine_sigkill(
                                  data_path2, DATA_SIZE_IN_MB_1)
     md5sum2 = get_pod_data_md5sum(core_api, pod_name, data_path2)
 
-    res_md5sum2 = get_pod_data_md5sum(core_api, pod_name, data_path2)
-    assert md5sum2 == res_md5sum2
+    assert md5sum2 is not None
```
yasker, line 1434: ? How can this get passed before?

khushboo-rancher: I saw this code snippet, I think asserting ```md5sum2 == res_md5sum2``` was not making sense, both are same computed value so I changed it. This is not related to the failure of the test.

shuo-wu: It's better to write the specific data then read and verify it. e.g., using `echo` to write some strings and then using `cat` to read the file and verify the strings.

khushboo-rancher: Updated this with write and reading a string using ```echo``` and ```cat```.

PR #387, opened 2020-08-05
Update and add new tests for the refactored restore feature

shuo-wu:
longhorn/longhorn#1279


shuo-wu:
@yasker Need to update the image due to the engine version bump and API modification.

yasker:
Since it's blocking the nightly test now, I am going to merge this PR for now. Let's track any issue found by the QA in a separate issue later.

PR #392, opened 2020-08-12
run nightly tests using v1.0.x branch instead of master

meldafrawi:
longhorn/longhorn/issues/1672


yasker:
Note: we will change it back and add a more flexible solution for multi-branches after v1.0.2 released.

PR #397, opened 2020-08-19
Set a longer backup wait interval for backup wait function

shuo-wu:


yasker:
Reason for the change?

shuo-wu:
Some of the recent nightly test failures is caused by the short backup wait interval.

yasker:
Why do we need a longer backup completion wait time now? Is our backup speed slowed down compared to previous versions?

c3y1huang:
This seems already resolve in https://github.com/longhorn/longhorn-tests/commit/b8fbb2d67bfa5714bbda85095ec1a40aab49ea2b .

yasker:
The original issue has been fixed in https://github.com/longhorn/longhorn-tests/pull/409

PR #398, opened 2020-08-19
Add test skeleton for node eviction

boknowswiki:
https://github.com/longhorn/longhorn/issues/298

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -2184,3 +2184,52 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
     node = common.wait_for_disk_update(client, lht_hostId,
                                        len(update_disks))
     cleanup_host_disk(client, 'vol-small')
+
+
+@pytest.mark.skip(reason="TODO")
+def test_node_eviction():
+    """
+    Test node eviction (assuming this is a 3 nodes cluster)
+
+    1. Disable scheduling on node 3.
+    2. Create volume 1 with 2 replicas.
+    3. Attach volume 1 to node 1 and write some data to it and get
+    checksum 1.
+    4. Disable scheduling and set 'Eviction Requested' to 'true' on node 1.
+    5. Volume 1 should be failed to schedule new replica.
+    6. Set 'Eviction Requested' to 'false' to cancel node 1 eviction and
+    check there should be 1 replica on node 1 and node 2. And remove the
+    unscheduled replica.
+    7. Set 'Eviction Requested' to 'true' on node 1.
+    8. Set 'Replica Node Level Soft Anti-Affinity' to 'true'.
+    9. The eviction should be success, and no replica on node 1, 2 replicas
+    on node 2.
+    10. Enable scheduling on node 3, and set 'Eviction Requested' to
+    'false', enable scheduling on node 1.
+    11. Set 'Replica Node Level Soft Anti-Affinity' to 'false'.
+    12. Disable scheduling and set 'Eviction Requested' to 'true' on
+    node 2. And make sure the volume is in healthy state during the
+    eviction.
+    13. The eviction should be success and no replica on node 2. And 1
+    replica on node 1 and node 3. And verify the data with checksum 1.
+    14. Set 'Eviction Requested' to 'false' and enable scheduling on node 2.
+    15. Remove the replica on node 1 to make volume 1 in 'Degraded'
+    State. And set 'Eviction Requested' to 'true' and disable scheduling
+    on node 3.
+    16. Once volume 1 is back at 'Healthy' state, and the eviction is
+    done, there should be 1 replica on node 1 and node 2. And verify the
+    data with checksum 1.
+    17. Disable scheduling on node 1.
+    18. Create volume 2 with 2 replicas.
+    19. Attach volume 2 to node 2 and write some data to it and get
+    checksum 2. (volume 1 has replicas on node 1&2, volume 2 has replicas
+    on node 2&3)
+    20. Enable scheduling on node 1. And set 'Eviction Requested' to 'true'
+    and disable scheduling on node 2.
+    21. After the eviction is success, volume 1 should has replicas on node
+    1&3 and volume 2 should has replicas on node 1&3.
+    22. Verify the data on volume 1 and volume 2, the checksum should be
+    the same as checksum 1 and checksum 2.
+    23. Set 'Eviction Requested' to 'false' and enable scheduling on node 2.
+    """
+    pass


yasker:
Also, we need to test it with more than one volume.

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,28 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    5. Set 'Replica Node Level Soft Anti-Affinity' to 'true'.
+    6. The eviction should be success, and no replica on node 1, 2 replicas
+    on node 2.
+    7. Enable scheduling on node 3, and set 'Eviction Requested' to 'false'
+    , enable scheduling on node 1.
+    8. Set 'Replica Node Level Soft Anti-Affinity' to 'false'.
+    9. set 'Eviction Requested' to 'true' and disable scheduling on node 2.
+    10. The eviction should be success and no replica on node 2. And 1
+    replica on node 1 and node 3.
+    11. Verify the data on the volume, the checksum should be the same.
```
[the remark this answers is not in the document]
boknowswiki: Sure, I will update it.

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,28 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+def test_node_eviction():
+    """
+    Test node eviction (assuming this is a 3 nodes cluster)
+
+    1. Disable scheduling on node 3.
+    1. Create a volume with 2 replicas.
+    2. Attach the volume to node 1 and write some data to it and get
+    checksum.
+    3. Set 'Eviction Requested' to 'true' and disable scheduling on node 1.
+    4. The volume should be failed to schedule new replica.
```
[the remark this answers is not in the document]
boknowswiki: Then everything stays the same. Let me add this step. Thanks.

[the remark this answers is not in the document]
yasker: Unscheduled replica should be removed automatically?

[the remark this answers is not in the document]
boknowswiki: Sorry, it won't be removed automatically. We need to remove it manually for now. Let me update the step. Thanks.

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,49 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+@pytest.mark.skip(reason="TODO")
+def test_node_eviction():
+    """
+    Test node eviction (assuming this is a 3 nodes cluster)
+
+    1. Disable scheduling on node 3.
+    2. Create volume 1 with 2 replicas.
+    3. Attach volume 1 to node 1 and write some data to it and get
+    checksum 1.
+    4. Set 'Eviction Requested' to 'true' and disable scheduling on node 1.
```
khushboo-rancher, line 2198: Shouldn't be the disabling of scheduling set first and then eviction requested should be set to ```True``` ?

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,49 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    5. Volume 1 should be failed to schedule new replica.
+    6. Set 'Eviction Requested' to 'false' to cancel node 1 eviction and
+    check there should be 1 replica on node 1 and node 2. And remove the
+    unscheduled replica.
+    7. Set 'Eviction Requested' to 'true' on node 1.
+    8. Set 'Replica Node Level Soft Anti-Affinity' to 'true'.
```
khushboo-rancher, line 2204: Should we set 'Replica Node Level Soft Anti-Affinity' to ```True``` before 'Eviction Requested' to ```True``` to make sure replica scheduling doesn't fail on same node.?

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,49 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    9. The eviction should be success, and no replica on node 1, 2 replicas
+    on node 2.
+    10. Enable scheduling on node 3, and set 'Eviction Requested' to
+    'false', enable scheduling on node 1.
+    11. Set 'Replica Node Level Soft Anti-Affinity' to 'false'.
+    12. Set 'Eviction Requested' to 'true' and disable scheduling on node 2.
```
khushboo-rancher, line 2210: Same as https://github.com/longhorn/longhorn-tests/pull/398/files#r473483283

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,49 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    13. The eviction should be success and no replica on node 2. And 1
+    replica on node 1 and node 3.
```
khushboo-rancher, line 2212: Before proceeding to next step we should ```Assert``` the checksum of data in new replica(on node 3).

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,49 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    14. Set 'Eviction Requested' to 'false' and enable scheduling on node 2.
+    15. Remove the replica on node 1 to make volume 1 in 'Degraded'
+    State. And set 'Eviction Requested' to 'true' and disable scheduling
+    on node 3.
+    16. Once volume 1 is back at 'Healthy' state, and the eviction is
+    done, there should be 1 replica on node 1 and node 2.
```
khushboo-rancher, line 2218: And we should Assert the checksum of the data in two replicas.

khushboo-rancher:
@boknowswiki @yasker Instead of having a single test with multiple test steps, I think we should bifurcate it into smaller tests like below. I can help writing it.
``` 
1. test_node_eviction_with_node_available_to_schedule
2. test_node_eviction_with_no_node_available_to_schedule
3. test_node_eviction_with_soft_anti_afinity
4. test_node_eviction_with_degraded_volume
5. test_node_eviction_with_multiple_volumes
6. test_node_eviction_with_detach_volume
7. test_node_eviction_cancel
```
One more thing, I want to ask is if there are multiple volume replicas, does the eviction happen one by one or in parallel?

manager/integration/tests/test_node.py
boknowswiki: I was thinking to set them at the same time. Let me put the disabling scheduling before the eviction to solve the confusion.

manager/integration/tests/test_node.py
boknowswiki: One reason we set 'Eviction Requested' to 'True' before that is we can see the eviction got suspend and continues.

manager/integration/tests/test_node.py
boknowswiki: I will update it.

manager/integration/tests/test_node.py
boknowswiki: I think we can check the data by the end. But it doesn't hurt we check it here as well. Let me update it.

manager/integration/tests/test_node.py
boknowswiki: Sure, we can do that.

boknowswiki:
> @boknowswiki @yasker Instead of having a single test with multiple test steps, I think we should bifurcate it into smaller tests like below. I can help writing it.
> 
> ```
> 1. test_node_eviction_with_node_available_to_schedule
> 2. test_node_eviction_with_no_node_available_to_schedule
> 3. test_node_eviction_with_soft_anti_afinity
> 4. test_node_eviction_with_degraded_volume
> 5. test_node_eviction_with_multiple_volumes
> 6. test_node_eviction_with_detach_volume
> 7. test_node_eviction_cancel
> ```
> 
> One more thing, I want to ask is if there are multiple volume replicas, does the eviction happen one by one or in parallel?

Like we discussed, one reason we put things together is we want to see the eviction suspend and continues, and how it handles difference failure cases. And for different volumes the eviction is parallel.

manager/integration/tests/test_node.py
meldafrawi: if `Eviction Requested` is set to `True`  before setting `Replica Node Level Soft Anti-Affinity` to `True` 
we should get the same result as step `5.  Volume 1 should be failed to schedule new replica.` and volume should end up with 1 healthy replica one node 2 and one suspended replica since scheduling is disabled on nodes 1 & 3. 
Then after setting `Replica Node Level Soft Anti-Affinity` to `True` replica should scheduled to node 2

@boknowswiki  is that what you mean ?

meldafrawi:
LGTM

manager/integration/tests/test_node.py
boknowswiki: @meldafrawi , step 7, the volume should have 2 healthy replicas, 1 on node 1, 1 on node 2. And 1 suspended replica. Since we only remove the replica when there is healthy replica count > replica spec number, which means we rebuild one first, until it success, we remove the extra one. And after set 'Replica Node Level Soft Anti-Affinity' to 'True', there will be 2 replicas on node 2 (step 9).

manager/integration/tests/test_node.py
```diff
@@ -2184,3 +2184,50 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    unscheduled replica.
+    7. Set 'Eviction Requested' to 'true' on node 1.
+    8. Set 'Replica Node Level Soft Anti-Affinity' to 'true'.
+    9. The eviction should be success, and no replica on node 1, 2 replicas
+    on node 2.
+    10. Enable scheduling on node 3, and set 'Eviction Requested' to
+    'false', enable scheduling on node 1.
+    11. Set 'Replica Node Level Soft Anti-Affinity' to 'false'.
+    12. Disable scheduling and set 'Eviction Requested' to 'true' on node 2.
+    13. The eviction should be success and no replica on node 2. And 1
```
shuo-wu, line 2211: Verify the volume keeps `Healthy` during the eviction

shuo-wu (commented):
In general, LGTM

manager/integration/tests/test_node.py
boknowswiki: Do you mean to check the volume state after each step?

shuo-wu: No. I just want to verify the volume HA is not affected during the eviction. Hence checking the volume robustness during the eviction is enough.

boknowswiki: Sure, I just updated step 12 to make sure volume is in healthy state during the eviction.

PR #404, opened 2020-08-27
manager/integration/tests: add test skeleton to test data locality feature

PhanLe1010:
longhorn/longhorn#1045

### DIFF
--- a/manager/integration/tests/test_scheduling.py
+++ b/manager/integration/tests/test_scheduling.py
@@ -377,3 +377,83 @@ def test_replica_rebuild_per_volume_limit():
 
     """
     pass
+
+
+@pytest.mark.skip(reason="TODO")
+def test_data_locality_basic():
+    """
+    Test data locality basic feature
+
+    Context:
+
+    Data Locality feature allows users to have an option to keep a local
+    replica on the same node as the consuming pod.
+    Longhorn is currently supporting 2 modes:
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_scheduling.py
```diff
@@ -377,3 +377,50 @@ def test_replica_rebuild_per_volume_limit():
...
+       a replica on the new-engine-node and remove the replica on
+       old-engine-node
+    10. Update dataLocality to disabled for volume(1)
+    11. detach the volume(1) and attach it to a different node.
+       Let's call the new node is new-engine-node and the old
+       node is old-engine-node
+    12. Wait for volume(1) to finish detaching
+    13. Use a retry loop to verify that Longhorn does not create
+        a replica on the new-engine-node
+    14. clean up
```
yasker, line 424: What about the case that replica cannot be created on the attached node?

Also, what if the volume detached (workload shutdown) during the time of detaching? Need to think more about negative cases.

PhanLe1010: Thank you! I will fix this

PhanLe1010: @yasker I added the 2 negative test cases and also check for replica deletion preference.

meldafrawi (approved):

meldafrawi:
LGTM

PR #405, opened 2020-08-27
Add test skeleton for extra replica cleanup

boknowswiki:
https://github.com/longhorn/longhorn/issues/1679

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -1919,3 +1919,23 @@ def test_rebuild_after_replica_file_crash(client, volume_name): # NOQA
     assert crashed_replica["failedAt"] is not None
 
     check_volume_data(volume, data)
+
+
+@pytest.mark.skip(reason="TODO")
+def test_extra_replica_cleanup():
+    """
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -1919,3 +1919,22 @@ def test_rebuild_after_replica_file_crash(client, volume_name): # NOQA
...
 
     check_volume_data(volume, data)
+
+
+@pytest.mark.skip(reason="TODO")
+def test_extra_replica_cleanup():
+    """
+    Test extra failed to scheduled replica cleanup
+    when no eviction requested
+
```
shuo-wu, line 1929: Mention `Disable soft anti-affinity` explicitly

shuo-wu (changes requested):

manager/integration/tests/test_ha.py
boknowswiki: Sure, will add it at the beginning.

shuo-wu (approved):

meldafrawi (approved):

PR #411, opened 2020-09-14
add data locality integration test

meldafrawi:

longhorn/longhorn#1045

### DIFF
--- a/manager/integration/tests/test_scheduling.py
+++ b/manager/integration/tests/test_scheduling.py
@@ -1,13 +1,28 @@
 import pytest
+import time
 
 from common import RETRY_COUNTS, RETRY_INTERVAL
-from common import client, volume_name  # NOQA
+from common import client, core_api, volume_name, pod  # NOQA
 from common import check_volume_data, cleanup_volume, \
     create_and_check_volume, get_longhorn_api_client, get_self_host_id, \
     wait_for_volume_detached, wait_for_volume_degraded, \
     wait_for_volume_healthy, wait_scheduling_failure, \
     write_volume_random_data, wait_for_rebuild_complete
 from common import SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY
+from common import SETTING_DEFAULT_DATA_LOCALITY
+from common import create_pv_for_volume
+from common import create_pvc_for_volume
+from common import write_pod_volume_random_data
+from common import wait_for_volume_replica_count
+
+from common import Mi, Gi, DATA_SIZE_IN_MB_2, DATA_SIZE_IN_MB_4
+from common import create_and_wait_pod
+from common import settings_reset # NOQA
+from common import wait_for_rebuild_start
+from common import delete_and_wait_pod
+from common import wait_for_replica_failed
+from common import crash_engine_process_with_sigkill
+
 from time import sleep
 
 
@@ -379,8 +394,7 @@ def test_replica_rebuild_per_volume_limit():
     pass
 
 
-@pytest.mark.skip(reason="TODO")
-def test_data_locality_basic():
+def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset): # NOQA
     """
     Test data locality basic feature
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_scheduling.py
```diff
@@ -454,6 +468,424 @@ def test_data_locality_basic():
...
+    for replica in volume3.replicas:
+        if replica["hostId"] == node1.name:
+            err_replica = replica
+            break
+
+    assert err_replica is not None
+    wait_for_replica_failed(client, volume3_name, err_replica['name'])
+
+    delete_and_wait_pod(core_api, pod3_name)
+    volume3 = wait_for_volume_detached(client, volume3_name)
```
meldafrawi, line 755: Test still fails on case 3 here

Error: 
```
MountVolume.SetUp failed for volume "longhorn-testvol-skp0d3-3-pv" : rpc error: code = InvalidArgument desc = There is no block device frontend for volume longhorn-testvol-skp0d3-3
```

manager/integration/tests/test_scheduling.py
```diff
@@ -454,6 +468,424 @@ def test_data_locality_basic():
...
+        {"kubernetes.io/hostname": volume1_attached_node}
+    create_and_wait_pod(core_api, pod1)
+
+    write_pod_volume_random_data(core_api, pod1_name,
+                                 volume1_data_path, DATA_SIZE_IN_MB_2)
+
+    for i in range(10):
+        volume1 = client.by_id_volume(volume1_name)
+        assert len(volume1.replicas) == 1
+        assert volume1.replicas[0]['hostId'] == volume1_replica_node
```
PhanLe1010, line 550: I guess that we are imeplementing step #5  here. Should we do like `assert volume1.replicas[0]['hostId'] != volume1_attached_node` so it is a bit clearer that the replica is not on the same node as the engine. It is totally fine not to change though. 🙂

manager/integration/tests/test_scheduling.py
```diff
@@ -454,6 +468,424 @@ def test_data_locality_basic():
...
+
+    volume1 = wait_for_volume_replica_count(client, volume1_name, 2)
+    volume1 = client.by_id_volume(volume1_name)
+    assert len(volume1.replicas) == 2
+
+    wait_for_rebuild_complete(client, volume1_name)
+
+    volume1 = client.by_id_volume(volume1_name)
+
+    volume1 = wait_for_volume_replica_count(client, volume1_name, 1)
```
PhanLe1010, line 566: should we wait until the `replica_count==1` before get the updated volume object? e.g. should it be like this?:
```
volume1 = wait_for_volume_replica_count(client, volume1_name, 1)
volume1 = client.by_id_volume(volume1_name)
```

manager/integration/tests/test_scheduling.py
```diff
@@ -454,6 +468,424 @@ def test_data_locality_basic():
...
+
+    wait_for_rebuild_start(client, volume1_name)
+
+    volume1 = client.by_id_volume(volume1_name)
+    assert len(volume1.replicas) == 2
+
+    wait_for_rebuild_complete(client, volume1_name)
+
+    volume1 = client.by_id_volume(volume1_name)
+    volume1 = wait_for_volume_replica_count(client, volume1_name, 1)
```
PhanLe1010, line 595: Same as https://github.com/longhorn/longhorn-tests/pull/411/files#r488416332

manager/integration/tests/test_scheduling.py
```diff
@@ -454,6 +468,424 @@ def test_data_locality_basic():
...
+        "persistentVolumeClaim": {
+            "claimName": pvc2_name
+        }
+    }]
+
+    pod2['spec']['nodeSelector'] = {"kubernetes.io/hostname": node3.name}
+    create_and_wait_pod(core_api, pod2)
+
+    volume2 = wait_for_volume_healthy(client, volume2_name)
+
```
PhanLe1010, line 650: Can we check to make sure that there are 3 replicas at this moment?
` assert len(volume1.replicas) == 3`

manager/integration/tests/test_scheduling.py
PhanLe1010: hmm. Is it because we detach the volume manually at https://github.com/longhorn/longhorn-tests/blob/c66584c2b280967cb3dba7c54e6fab9e3b349990/manager/integration/tests/test_scheduling.py#L740
so the engine of the volume is killed at that moment and the pod can not find the block device frontend (which is the engine)?
If it is the case, can we try this?
```
    wait_for_rebuild_start(client, volume3_name)
     volume3 = client.by_id_volume(volume3_name)
     delete_and_wait_pod(core_api, pod3_name)
     volume3 = client.by_id_volume(volume3_name)
     assert len(volume3.replicas) == 2
     volume3 = client.by_id_volume(volume3_name)

      err_replica = None
     for replica in volume3.replicas:
         if replica["hostId"] == node1.name:
             err_replica = replica
             break

      assert err_replica is not None
     wait_for_replica_failed(client, volume3_name, err_replica['name'])

  
     volume3 = wait_for_volume_detached(client, volume3_name)
```

PhanLe1010:
This is a lot of work! 🚀 Thank you so much Mohamed!

manager/integration/tests/test_scheduling.py
meldafrawi: @PhanLe1010 
Time waiting for pod deletion may exceed replica rebuilding time, causing the test to fail
I've updated the test to work around the issue by killing volume engine to simulate rebuild interruption

manager/integration/tests/test_scheduling.py
meldafrawi: fixed

manager/integration/tests/test_scheduling.py
meldafrawi: fixed

manager/integration/tests/test_scheduling.py
meldafrawi: fixed

khushboo-rancher (approved):
LGTM

PhanLe1010 (approved):
It LGTM

PR #412, opened 2020-09-15
Cleanup backup related test code

joshimoo:
There where issues in the test code
- we where not waiting for the backup deletion during the backup_list test
- we where trying to find the backup before waiting for the backup completion in the create_backup function
- there is a couple of spots where we call find_backup but where not waiting for backup completion, since the backup.cfg only shows up once the backup is completed this would not work correctly all the time and introduce timing dependent flaky test behavior.

Since backup operations needs to be thought of as async, we always need to wait for completion of the backup operation. We do this in the utility functions delete_backup, create_backup, etc.

This is only tangentially related to the backup locks longhorn/longhorn#612

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -257,7 +257,12 @@ def create_backup(client, volname, data={}, labels={}):
         data = write_volume_data(volume, data)
     snap = create_snapshot(client, volname)
     create_snapshot(client, volname)
+
+    # after backup request we need to wait for completion of the backup
+    # since the backup.cfg will only be available once the backup operation
+    # has been completed
     volume.snapshotBackup(name=snap.name, labels=labels)
+    wait_for_backup_completion(client, volname, snap.name)
 
     verified = False
     for i in range(RETRY_COMMAND_COUNT):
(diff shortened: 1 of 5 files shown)


manager/integration/tests/common.py
```diff
@@ -2499,31 +2523,30 @@ def is_backupTarget_nfs(s):
...
-            break
-        time.sleep(1)
-    assert found
-
-    return bv, b
+                return bv
+        return None
+
+    for i in range(120):
+        bv = find_backup_volume()
```
khushboo-rancher, line 2541: Can we set ```bv``` to ```None``` initially and find backup only if ```bv``` is ```None```? That way we can avoid calling the method ```find_backup_volume``` after we have the value of ```bv``` already.

khushboo-rancher (dismissed):
LGTM

manager/integration/tests/common.py
joshimoo: We can but initially the volume.cfg will not be present, since the BV existence is based on the existence of a folder.
This shouldn't have an impact since calling backupList() on a non existent backupVolume should just return an empty list.

joshimoo: Will change it.

meldafrawi:
LGTM

meldafrawi (dismissed):

joshimoo:
@khushboo-rancher @meldafrawi I added 3 more commits, can you rereview the last 3 commits?

meldafrawi:
Reviewed the last 3 commits, LGTM

meldafrawi (approved):

khushboo-rancher (approved):
LGTM

PR #414, opened 2020-09-18
Wait for lock expiration in the restore crash tests

joshimoo:
since we crash the replica that initiated the restore it's backupstore
lock will still be present, so we need to wait till the lock is expired,
before we can delete the backups.

longhorn/longhorn#612

### DIFF
--- a/manager/integration/tests/backupstore.py
+++ b/manager/integration/tests/backupstore.py
@@ -1,4 +1,5 @@
 import os
+import time
 import base64
 import hashlib
 import json
@@ -18,6 +19,7 @@ from common import get_longhorn_api_client
 from common import delete_backup_volume
 
 BACKUPSTORE_BV_PREFIX = "/backupstore/volumes/"
+BACKUPSTORE_LOCK_DURATION = 150
 
 
 TEMP_FILE_PATH = "/tmp/temp_file"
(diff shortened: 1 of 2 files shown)


meldafrawi (dismissed):
LGTM

khushboo-rancher:
Just to understand, now we have a lock mechanism and while restoring a backup it can't be deleted. 
Currently, the expiration time is 150 secs for a lock but going forward we'll delete the locks which are not required because of the crashing of replica etc.
Will this deletion of locks be part of the feature?

joshimoo:
> Just to understand, now we have a lock mechanism and while restoring a backup it can't be deleted.

The lock mechanism works like this, once a lock is acquired there is a refresh routine that refreshes the lock every 60s, a locks duration is 150s (2 minutes 30s) which means that if one refresh fails we still have another try.

The tradeoffs are 
- The longer the lock duration the longer a crashed operation is still presumed to be active.
- The lower the lock duration the more writes a needed to keep a lock active.

So we decided for a refresh interval of 1 minute, to have a second chance in case a refresh fails we used a lock duration of 2 minutes and 30s.

> Currently, the expiration time is 150 secs for a lock but going forward we'll delete the locks which are not required because of the crashing of replica etc.
> Will this deletion of locks be part of the feature?

We will not delete a lock of a failed operation, it will remain on the backupstore indefinitely till the volume gets deleted.
In the future we could add the ability to cleanup expired locks after a certain time, we could do this as part of a backup delete operation. Please create a separate issue for this.

For the longhorn-tests code you can refactor the wait for lock expiration code to deleting the locks from the backupstore.
Since we know that the lock is no longer in use and the reason why it has not been cleaned up.

khushboo-rancher (approved):
LGTM

PR #418, opened 2020-09-24
Add auto salvage with revision counter disabled test skeleton

boknowswiki:
https://github.com/longhorn/longhorn/issues/508

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -139,6 +139,12 @@ def ha_rebuild_replica_test(client, volname):   # NOQA
 def test_ha_salvage(client, core_api, volume_name, disable_auto_salvage):  # NOQA
     """
     [HA] Test salvage when volume faulted
+    TODO
+    The test cases should cover the following four cases:
+    1. Manual salvage with revision counter enabled.
+    2. Manual salvage with revision counter disabled.
+    3. Auto salvage with revision counter enabled.
+    4. Auto salvage with revision counter enabled.
 
     Setting: Disable auto salvage
 
@@ -159,6 +165,39 @@ def test_ha_salvage(client, core_api, volume_name, disable_auto_salvage):  # NOQ
     Case 2: Crash all replica processes
 
     Same steps as Case 1 except on step 3, use SIGTERM to crash the processes
+
+    Setting: Enabled auto salvage.
+
+    Case 3: Revision counter disabled.
+
+    1. Set 'Automatic salvage' to true.
+    2. Set 'Disable Revision Counter' to true.
+    3. Create a volume with 3 replicas.
+    4. Attach the volume to a node and write some data to it and save the
+    checksum.
+    5. Delete all replica processes using instance manager or
+    crash all replica processes using SIGTERM.
+    6. Wait for volume to `faulted`, then `healthy`.
+    7. Verify there are 5 replicas, 3 good replicas
+        (one is from the first three replicas), and 2 failed replicas
+        (both are from the first three replicas).
+    8. Check the data in the volume and make sure it's the same as the
+    checksum saved on step 5.
+
+    Case 4: Revision counter enabled.
+
+    1. Set 'Automatic salvage' to true.
+    2. Set 'Disable Revision Counter' to false.
+    4. Create a volume with 3 replicas.
+    5. Attach the volume to a node and write some data to it and save the
+    checksum.
+    6. Delete all replica processes using instance manager or
+    crash all replica processes using SIGTERM.
+    7. Wait for volume to `faulted`, then `healthy`.
+    8. Verify there are 3 replicas, they are all from previous replicas.
+    9. Check the data in the volume and make sure it's the same as the
+    checksum saved on step 5.
+
     """
     ha_salvage_test(client, core_api, volume_name)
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -1941,3 +1941,24 @@ def test_extra_replica_cleanup():
...
     """
     pass
+
+
+@pytest.mark.skip(reason="TODO")
+def test_auto_salvage_with_revision_counter_disabled():
+    """
+    Test auto salvage when revision counter is disabled
+
+    1. Set 'Diable Revision Counter' to true.
```
yasker, line 1951: `Disable`

yasker (changes requested):
We also need a test case on manual salvage.

shuo-wu:
There is already a test `ha_salvage_test` similar to this test skeleton. Can we directly modify it instead? At least you can rename the test to `test_ha_salvage_with_revision_counter_disabled` then indicate let QAs/developers to use function `ha_salvage_test`

boknowswiki:
> There is already a test `test_ha_salvage` similar to this test skeleton. Can we directly modify it instead? At least you can rename the test to `test_ha_salvage_with_revision_counter_disabled` then indicate let QAs/developers to use function `ha_salvage_test`

The behaviors are different when revision counter is disabed for auto salvage. The original one is for revision counter enabled case. This is the reason I create a new one. Thanks.

yasker:
@boknowswiki Is there any difference from the user's perspective? What I can think of is the replica number might be different in the end, but the process should be largely the same?

boknowswiki:
> @boknowswiki Is there any difference from the user's perspective? What I can think of is the replica number might be different in the end, but the process should be largely the same?

I think so, only the replica number are different in the end.

yasker:
@boknowswiki in that case, we should be able to create a function out of the existing code and reuse it for the both case.

boknowswiki:
@shuo-wu and @yasker , I have updated the PR, please take a look and let me know if my understand is wrong. Thanks.

manager/integration/tests/test_ha.py
```diff
@@ -233,6 +233,42 @@ def ha_salvage_test(client, core_api, # NOQA
...
+    # test auto salvage when revision counter enabled ordisabled.
+    # One for revision counter enabled case.
+    test_ha_salvage_autosalvage()
+
+    # One for revision counter disabled case.
+    test_ha_salvage_autosalvage()
+
+
+@pytest.mark.skip(reason="TODO")
+def test_ha_salvage_autosalvage():
```
yasker, line 245: if it's a common function, it's better to not start with `test`. Otherwise, it will be picked up by pytest.

Also, you don't need to add the function above. You can just describe once the test is written, what's the expected result. For example
```
The test cases should cover the following four cases:
1. Manual salvage with revision counter enabled.
2. Manual salvage with revision counter disabled.
3. Auto salvage with revision counter enabled.
4. Auto salvage with revision counter enabled.
```

yasker (changes requested):

manager/integration/tests/test_ha.py
```diff
@@ -233,6 +233,42 @@ def ha_salvage_test(client, core_api, # NOQA
 
     cleanup_volume(client, volume)
 
+    # test auto salvage when revision counter enabled ordisabled.
+    # One for revision counter enabled case.
+    test_ha_salvage_autosalvage()
```
shuo-wu, line 238: `ha_salvage_test` is a common function. You need to call this common function in the test functions rather than in a reverse way.

manager/integration/tests/test_ha.py
shuo-wu, line 245: And you can rename the salvage related tests to `test_ha_auto_salvage_with_revision_counter`, `test_ha_auto_salvage_without_revision_counter`, `test_ha_manual_salvage_with_revision_counter`, and `test_ha_manual_salvage_without_revision_counter`.

shuo-wu (changes requested):

boknowswiki:
@shuo-wu and @yasker , since there was a `test_ha_salvage`, I prefer to use that func as the `main` function, which will cover all the cases. I have updated the test skeleton for that function, please let me know what you think. Thanks.

yasker (approved):

yasker:
@boknowswiki can you add another test case regarding live upgrade and revision counter disabled? I realized it's the other place that might use the revision counter.

manager/integration/tests/test_ha.py
```diff
@@ -139,6 +139,12 @@ def ha_rebuild_replica_test(client, volname):   # NOQA
 def test_ha_salvage(client, core_api, volume_name, disable_auto_salvage):  # NOQA
     """
     [HA] Test salvage when volume faulted
+    TODO
+    The test cases should cover the following four cases:
+    1. Manual salvage with revision counter enabled.
+    2. Manual salvage with revision counter disabled.
```
shuo-wu, line 145: The test step for this case is missed.

shuo-wu (changes requested):

manager/integration/tests/test_ha.py
boknowswiki: @shuo-wu , that's the current test implementation. I was expecting whoever works on this test case will restructure it. Thanks.

boknowswiki:
> @boknowswiki can you add another test case regarding live upgrade and revision counter disabled? I realized it's the other place that might use the revision counter.

@yasker , updated `test_upgrade` skeleton, please take a look. Thanks.

shuo-wu (approved):

yasker (approved):

PR #419, opened 2020-09-25
doc: Add uninstallation test for v1.1.0

shuo-wu:
longhorn/longhorn#1820


shuo-wu:
Will test if this manual test can block the v1.0.2 uninstallation before the merge.

yasker:
The script should be included in the PR.

shuo-wu:
The script is added to the test as an example.

yasker (approved):

PR #426, opened 2020-10-07
Add tests for backup-lock

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/612

Added tests for backup locks and fixed a few flake8 errors in the test_basic.py file.

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -63,6 +63,11 @@ from common import delete_backup
 from common import delete_backup_volume
 from common import BACKUP_BLOCK_SIZE
 from common import assert_backup_state
+from common import wait_for_volume_condition_restore, wait_for_backup_delete
+from common import VOLUME_FIELD_ROBUSTNESS, VOLUME_ROBUSTNESS_HEALTHY
+from common import VOLUME_ROBUSTNESS_FAULTED
+from common import DATA_SIZE_IN_MB_2, DATA_SIZE_IN_MB_3
+from common import wait_for_backup_to_start
 
 from backupstore import backupstore_corrupt_backup_cfg_file
 from backupstore import backupstore_delete_volume_cfg_file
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -2842,11 +2989,11 @@ def test_backup_lock_restoration_during_deletion():  # NOQA
...
     6. Write more data to the volume and take another backup.
     7. Wait for the 2nd backup to be completed.
-    6. Delete the 1st backup.
-    7. Without waiting for the backup deletion completion, restore the 2nd
+    8. Delete the 1st backup.
+    9. Without waiting for the backup deletion completion, restore the 2nd
        backup from the backup store.
-    8. Verify the restored volume become faulted.
-    9. Wait for the 1st backup deletion and assert the count of the backups
+    10. Verify the restored volume become faulted.
```
joshimoo, line 2995: Not sure if this should become faulted, since the restore should be retried as long as the fromBackup flag is still set in the volume. not sure if and when we clean this up.

khushboo-rancher: Restore should be retried if the ```fromBackup``` flag has a non-existent backup. In this test, backup b2 deletion is in progress while restore is being initiated from backup b1 resulting in the restoring volume becoming faulted. 

However, to repro this scenario, we need a backup of at least 1.5 Gi of data and this test takes around 22 mins to run. Do we want to include this test or should we skip this?
@yasker What's your opinion on this?

joshimoo: I would say skip the test for scheduled testing. But I would also have assumed that the restore of B1 on the new volume should be retried periodically :)

khushboo-rancher:
@meldafrawi This test should have covered test cases for https://github.com/longhorn/longhorn/issues/1393 as well.

manager/integration/tests/test_basic.py
yasker: We can skip this part and add them to the manual test instead.

khushboo-rancher: Skipped the test.

joshimoo (approved):
LGTM

PR #427, opened 2020-10-08
Fix recurring job integration tests

PhanLe1010:
We changed the recurring jobs' logic: recurring jobs only take backup/snapshot when the volume has new data since the last backup/snapshot. Therefore, we need to update recurring job integration tests.

Now we can control precisely when recurring snapshot/backup will happen . As the result, this PR fixes longhorn/longhorn#1354 as a side affect

Related issues:
longhorn/longhorn#1858
longhorn/longhorn#1354

### DIFF
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -2,6 +2,8 @@ import pytest
 import time
 import json
 
+from datetime import datetime
+
 import common
 from common import client, clients, core_api, random_labels, volume_name  # NOQA
 from common import storage_class, statefulset  # NOQA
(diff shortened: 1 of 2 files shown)


yasker (approved):
LGTM except for the flake8 error.

Once the flake8 error is fixed we can merge this in.

PhanLe1010:
Fixed the flake8 error

manager/integration/tests/test_recurring_job.py
```diff
@@ -50,24 +61,34 @@ def test_recurring_job(clients, volume_name):  # NOQA
...
+    10. Make sure there are exactly 4 completed backups.
+        1. old backup job completed 2 backups
+        2. new backup job completed 2 backups
+    11. Make sure we have no backup in progress
     """
-    for host_id, client in iter(clients.items()):  # NOQA
-        break
+
+    '''
+    The timeline looks like this:
```
joshimoo, line 83: I like the addition of the timeline 👍

yasker (approved):

PR #433, opened 2020-10-15
Add step to deploy NFS provisioner setup dependencies in README

c3y1huang:
In case forget, this seems required by test_rwx_*.


joshimoo (approved):
LGTM, we will refactor the NFS-provisioner tests once 1.1 rwx is implemented then a provisioner deployment is no longer required.

But currently it is required, so I am merging this as is :)

PR #434, opened 2020-10-16
Disable replica rebuild test skeleton

boknowswiki:

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -2037,3 +2037,35 @@ def test_extra_replica_cleanup(client, volume_name, settings_reset): # NOQA
     assert volume.robustness == "healthy"
 
     check_volume_data(volume, data)
+
+
+@pytest.mark.skip(reason="TODO") # NOQA
+def test_disable_replica_rebuild():
+    """
+    Test disable replica rebuild
+
+    1. Disable node scheduling on node-2 and node-3. To make sure
+    replica scheduled on node-1.
+    2. Set 'Disable Replica Rebuild' to true.
+    3. Create a volume with 1 replica and attach it to node-1.
+    4. Enable scheduling on node-2 and node-3. Set node-1 scheduling to
+    'Disable' and 'Enable' eviction on node-1.
+    5. Wait for 30 seconds, and check no eviction happen.
+    6. 'Enable' node-1 scheduling and 'Disable' node-1 eviction.
+    7. Detach the volume and update data locality to 'best-effort'.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -2037,3 +2037,27 @@ def test_extra_replica_cleanup(client, volume_name, settings_reset): # NOQA
...
+    3. Create a volume with 1 replica and attach it to node-1.
+    4. Enable scheduling on node-2 and node-3. Set node-1 scheduling to
+    'Disable' and 'Enable' eviction on node-1.
+    5. Wait for 30 seconds, and check no eviction happen.
+    6. 'Enable' node-1 scheduling and 'Disable' node-1 eviction.
+    7. Detach the volume and update data locality to 'best-effort'.
+    8. Attach the volume to node-2, and wait for 30 seconds, and check
+    no data locality happen.
+    9. Detach the volume and update data locality to 'disable'.
+    10. Attach the volume to node-2 and reboot node-1.
```
yasker, line 2059: We don't have the ability to reboot a node in the integration test now.

boknowswiki: ok, I was thinking to increase the replica number instead of reboot the node, this should be easier and get the same test. Please let me know if you have different idea. Thanks.

yasker (approved):

yasker:
LGTM. @khushboo-rancher or @meldafrawi ?

khushboo-rancher:
LGTM in general.
Can we add one more case of simply deleting any replica should not trigger rebuild?

Example:
```
Step 13 - Set 'Disable Replica Rebuild' to True, delete one replica.
Step 14 - Wait for 30 secs, no rebuild should get triggered. The volume should stay in degraded state with 1 replica.
Step 15 - Set 'Disable Replica Rebuild' to false, and wait for replica rebuild and volume becomes 'healthy' state with 2 replicas.
```

boknowswiki:
> LGTM in general.
> Can we add one more case of simply deleting any replica should not trigger rebuild?
> 
> Example:
> 
> ```
> Step 13 - Set 'Disable Replica Rebuild' to True, delete one replica.
> Step 14 - Wait for 30 secs, no rebuild should get triggered. The volume should stay in degraded state with 1 replica.
> Step 15 - Set 'Disable Replica Rebuild' to false, and wait for replica rebuild and volume becomes 'healthy' state with 2 replicas.
> ```

Sure, will update it. Thanks.

khushboo-rancher (approved):
LGTM

yasker (approved):

PR #435, opened 2020-10-16
doc: Add a manual test for the node cleanup feature

shuo-wu:
longhorn/longhorn#1232

### DIFF
--- a/docs/content/manual/pre-release/node/node-deletion.md
+++ b/docs/content/manual/pre-release/node/node-deletion.md
@@ -1,0 +1,29 @@
+---
+title: Test node deletion
+---
+### Case 1: Delete multiple kinds of nodes:
+1. Deploy Longhorn.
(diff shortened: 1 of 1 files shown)


yasker (approved):

khushboo-rancher:
LGTM.

However, I think we can have more test scenarios like 
1. Multiple nodes removal.
2. Removing the node when in down/disable state and make sure the corresponding Kubernetes node object is deleted.

yasker (changes requested):
Let's move this to the `pre-release` test instead of `v1.1.0` test.

And we should add the test steps from https://github.com/longhorn/longhorn/issues/1232 too.

shuo-wu:
The test doc is updated:
1. Scale the existing test case to multiple-node scenarios
2. Add the reproduce step of longhorn/longhorn#1232 as the 2nd test case.

yasker (approved):

PR #436, opened 2020-10-16
Fix the test failure for backup lock

khushboo-rancher:
This is a fix for https://ci.longhorn.io/job/public/job/longhorn-tests/127/testReport/tests/test_basic/test_backup_lock_deletion_during_restoration/

It looks like in the nightly test environment the restore from a back got completed quickly even before the deletion of the same backup could initiate. So, increased the data backup size from 100 MB to 300 MB and deleted extra wait.

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -63,7 +63,7 @@ from common import delete_backup
 from common import delete_backup_volume
 from common import BACKUP_BLOCK_SIZE
 from common import assert_backup_state
-from common import wait_for_volume_condition_restore, wait_for_backup_delete
+from common import wait_for_backup_delete
 from common import VOLUME_FIELD_ROBUSTNESS, VOLUME_ROBUSTNESS_HEALTHY
 from common import VOLUME_ROBUSTNESS_FAULTED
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -2785,10 +2786,6 @@ def test_backup_lock_deletion_during_restoration(client, core_api, volume_name,
 
     _, b = common.find_backup(client, std_volume_name, snap1.name)
     client.create_volume(name=restore_volume_name, fromBackup=b.url)
-    wait_for_volume_condition_restore(client, restore_volume_name,
```
yasker, line 2788: We don't need to wait for the RestoreInProgress condition?

yasker (changes requested):
Also, flake8 test failed.

manager/integration/tests/test_basic.py
khushboo-rancher: We have ```wait_for_volume_restoration_start``` check at the next line, so this was not required.

khushboo-rancher:
Flake8 failure is fixed now.

yasker (approved):

PR #439, opened 2020-10-20
test plan: Allow volume creation with degraded availability

yasker:


khushboo-rancher (dismissed):
LGTM

Serial numbers of the test steps in the integration tests can be serialized. (Can be ignored, not important)

yasker:
Serial numbers updated in the integration test.

PR #443, opened 2020-10-21
Enhance and fix unstable test result from test_node_eviction

c3y1huang:
* Split test case into different scenarios.
* Removed intermediate state check. 
* Fix test failure when k8s decided to relocate replicas one time only during anti-affinity node eviction.
* Fix test fails when relocated to the wrong node due to a missing node eviction.

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -37,9 +37,7 @@ from common import Mi, DATA_SIZE_IN_MB_2
 from common import create_and_wait_pod
 from common import get_pod_data_md5sum
 from common import wait_for_volume_healthy
-from common import wait_for_volume_replica_count, \
-    wait_for_volume_replica_count_not_equal
-from common import wait_for_volume_degraded
+from common import wait_for_volume_replica_count
 from common import delete_and_wait_pod
 from common import wait_for_volume_detached
 
@@ -2214,10 +2212,8 @@ def test_node_eviction(client, core_api, csi_pv, pvc, pod_make, volume_name): # 
     3. Write some data and get the checksum.
     4. Set 'Eviction Requested' to 'false' and enable scheduling on node 2.
     5. Set 'Eviction Requested' to 'true' and disable scheduling on node 3.
-    6. Remove replica on node 1 to make volume degraded.
-    7. Wait for volume 'healthy' and eviction completed.
-    8. Check replicas running on node 1 and 2.
-    9. Check volume data checksum.
+    6. Check volume 'healthy' and wait for replicas running on node 1 and 2.
+    7. Check volume data checksum.
     """
     nodes = client.list_node()
     node1 = nodes[0]
@@ -2234,11 +2230,8 @@ def test_node_eviction(client, core_api, csi_pv, pvc, pod_make, volume_name): # 
                                            num_of_replicas=2,
                                            data_path=data_path)
 
-    volume = wait_for_volume_healthy(client, volume_name)
-    assert len(volume.replicas) == 2
-    for r in volume.replicas:
-        assert r.mode == "RW"
-        assert r.running is True
+    common.wait_for_replica_scheduled(client, volume_name,
+                                      to_nodes=[node1.name, node3.name])
 
     # replica now running on node 1, 3
     # enable node 2
@@ -2246,30 +2239,8 @@ def test_node_eviction(client, core_api, csi_pv, pvc, pod_make, volume_name): # 
     # disable node 3 to have replica schedule to node 1, 2
     client.update(node3, allowScheduling=False, evictionRequested=True)
 
-    volume = client.by_id_volume(volume_name)
-    for r in volume.replicas:
-        if r.hostId == node1.name:
-            volume.replicaRemove(name=r.name)
-            break
-    wait_for_volume_degraded(client, volume_name)
-    wait_for_volume_healthy(client, volume_name)
-
-    try:
-        wait_for_volume_replica_count(client, volume_name, 3)
-    except AssertionError as e:
-        print("\nException when asserting replica count, "
-              "could have missed this intermediate state for ",
-              volume_name)
-        print(e)
-        pass
-    wait_for_volume_replica_count(client, volume_name, 2)
-
-    volume = client.by_id_volume(volume_name)
-    for r in volume.replicas:
-        assert r.hostId == node1.name or \
-               r.hostId == node2.name
-        assert r.running is True
-        assert r.mode == "RW"
+    common.wait_for_replica_scheduled(client, volume_name,
+                                      to_nodes=[node1.name, node2.name])
 
     expect_md5sum = get_pod_data_md5sum(core_api, pod_name, data_path)
     assert expect_md5sum == created_md5sum
@@ -2304,13 +2275,12 @@ def test_node_eviction_no_schedulable_node(client, core_api, csi_pv, pvc, pod_ma
 
     volume = wait_for_volume_healthy(client, volume_name)
     assert len(volume.replicas) == 2
-    replica_names = []
-    replica_hostIDs = []
+
+    created_replicas = {}
     for r in volume.replicas:
         assert r.mode == "RW"
         assert r.running is True
-        replica_names.append(r.name)
-        replica_hostIDs.append(r["hostId"])
+        created_replicas[r.name] = r["hostId"]
 
     client.update(node1, allowScheduling=False, evictionRequested=True)
     wait_for_volume_replica_count(client, volume_name, 3)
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_node.py
```diff
@@ -2388,93 +2401,97 @@ def test_node_eviction(client, core_api, volume_name, pod, settings_reset): # NO
...
-    wait_for_volume_replica_count(client, volume1_name, 3)
+    # relocation from the only replica node of anti-affinity could
+    # spawn single or mutiple replica, hense we do not know the
+    # expected replica size during replication
+    wait_for_volume_replica_count_not_equal(client, volume1_name, 2)
     wait_for_volume_replica_count(client, volume1_name, 2)
 
-    volume1 = client.by_id_volume(volume1_name)
-    assert volume1.robustness == "healthy"
+    # k8s may or may not run an extra relocation
```
yasker, line 2414: What does this mean?

manager/integration/tests/test_node.py
```diff
@@ -2388,93 +2401,97 @@ def test_node_eviction(client, core_api, volume_name, pod, settings_reset): # NO
...
-    assert volume1.robustness == "healthy"
-
-    wait_for_volume_replica_count(client, volume1_name, 3)
-    wait_for_volume_replica_count(client, volume1_name, 2)
-
-    volume1 = client.by_id_volume(volume1_name)
-    assert volume1.robustness == "healthy"
-
-    wait_for_volume_replica_count(client, volume1_name, 3)
+    # relocation from the only replica node of anti-affinity could
```
yasker, line 2408: Do we need to check the volume is healthy across the process?

manager/integration/tests/test_node.py
```diff
@@ -2388,93 +2401,97 @@ def test_node_eviction(client, core_api, volume_name, pod, settings_reset): # NO
...
+    # spawn single or mutiple replica, hense we do not know the
+    # expected replica size during replication
+    wait_for_volume_replica_count_not_equal(client, volume1_name, 2)
```
yasker, line 2411: It's an intermediate state, not sure we can catch it. Same below.

manager/integration/tests/test_node.py
```diff
@@ -2388,93 +2401,97 @@ def test_node_eviction(client, core_api, volume_name, pod, settings_reset): # NO
...
+        wait_for_volume_replica_count_not_equal(client, volume1_name, 2)
+        wait_for_volume_replica_count(client, volume1_name, 2)
+    except AssertionError as e:
+        print("\nException when asserting replica count, "
+              "should already correctly allocated",
+              volume1_name)
+        print(e)
+        pass
 
+    volume1 = wait_for_volume_healthy(client, volume1_name)
```
yasker, line 2425: Volume should be always healthy during the process?

yasker (changes requested):
I don't think we should depend on the intermediate state.

Also, this test case is too long and hard to read. It's better if we can split it into a few different scenarios instead of a big test case.

manager/integration/tests/test_node.py
c3y1huang: By observation, K8s could replicate to single or multi replicas for relocation.
Sometimes the extra check is not required when multiple replicas are replicated correctly on the first try of relocation. But IMO this is still a necessary step in case K8s decided to have relocation done multiple times.

I've corrected the comment.

manager/integration/tests/test_node.py
c3y1huang: IMO this still an important check to ensure there is enough time for the replication. I've wrapped in try-except so there is no dependency on this.

manager/integration/tests/test_node.py
c3y1huang: Addressed.

manager/integration/tests/test_node.py
c3y1huang: Addressed.

yasker (changes requested):
LGTM in general. Commented a few minor issues.

manager/integration/tests/test_node.py
```diff
@@ -2211,367 +2204,406 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
-    27. Verify the data on volume 1 and volume 2, the checksum should be
-    the same as checksum 1 and checksum 2.
-    28. Set 'Eviction Requested' to 'false' and enable scheduling on node 2.
+    Case: node 1, 3 to node 1, 2 eviction
+    1. Disable scheduling on node 2.
+    2. Create pv, pvc, pod with volume of 2 replicas.
+    3. Write some data and get the checksum.
+    4. Set 'Eviction Requested' to 'false' and enable scheduling on node 2.
+    5. Set 'Eviction Requested' to 'true' and disable scheduling on node 3.
+    6. Remove replica on node 1 to make volume degraded.
```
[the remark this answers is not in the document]
c3y1huang: Addressed in https://github.com/longhorn/longhorn-tests/pull/443/commits/65671fa95883bd661ae683a6823949d135949790.

manager/integration/tests/test_node.py
```diff
@@ -2211,367 +2204,406 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+
+    volume = wait_for_volume_healthy(client, volume_name)
+    assert len(volume.replicas) == 2
+    replica_names = []
+    replica_hostIDs = []
+    for r in volume.replicas:
+        assert r.mode == "RW"
+        assert r.running is True
+        replica_names.append(r.name)
+        replica_hostIDs.append(r["hostId"])
```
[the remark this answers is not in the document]
c3y1huang: Addressed inhttps://github.com/longhorn/longhorn-tests/pull/443/commits/65671fa95883bd661ae683a6823949d135949790.

manager/integration/tests/test_node.py
```diff
@@ -2211,367 +2204,406 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    2. Create pv, pvc, pod with volume of 2 replicas.
+    3. Write some data and get the checksum.
+    7. Set 'Eviction Requested' to 'true' and disable scheduling on node 1.
+    8. Set 'Replica Node Level Soft Anti-Affinity' to 'true'.
+    9. The eviction should succeed, and all replicas running on node 2.
+    Case #2: node 2 to node 1, 3 eviction
+    10. Enable scheduling on node 1 and 3.
+    11. Set 'Replica Node Level Soft Anti-Affinity' to 'false'.
+    12. Set 'Eviction Requested' to 'true' and disable scheduling on node 2.
+    13. Check volume is still `healty`.
```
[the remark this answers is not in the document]
c3y1huang: Addressed in https://github.com/longhorn/longhorn-tests/pull/443/commits/65671fa95883bd661ae683a6823949d135949790.

manager/integration/tests/test_node.py
```diff
@@ -2211,367 +2204,406 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
-
-    for replica in volume1.replicas:
-        if replica.hostId == node1.name:
-            break
+    # By observation, when K8s try to relocate from the single node with
+    # anti-affinity sometimes would replicate to single or mutiple replicas
+    # for the relocation, hense we do not know the expected replica size
+    # here.
+    try:
+        wait_for_volume_replica_count_not_equal(client, volume_name, 2)
```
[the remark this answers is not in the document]
c3y1huang: Removed all intermediate replica count checks and instead wait for the end result. Address in https://github.com/longhorn/longhorn-tests/pull/443/commits/65671fa95883bd661ae683a6823949d135949790. 

And thanks for the correction.

manager/integration/tests/test_node.py
```diff
@@ -2211,367 +2204,406 @@ def test_replica_scheduler_rebuild_restore_is_too_big(client):  # NOQA
...
+    3. Write some data to volume 1 and get the checksum.
+    2. Create pv, pvc, pod with volume 2 of 2 replicas.
+    3. Write some data to volume 2 and get the checksum.
+    4. Set 'Eviction Requested' to 'true' and disable scheduling on node 2.
+    5. Set 'Eviction Requested' to 'false' and enable scheduling on node 1.
+    6. Check all replicas running on node 1 and 3.
+    7. delete pods to detach volume 1 and 2.
+    8. Set 'Eviction Requested' to 'false' and enable scheduling on node 2.
+    9. Set 'Eviction Requested' to 'true' and disable scheduling on node 1.
+    10. Create pod 1 and pod 2 to automatically attach to volume 1 and 2.
```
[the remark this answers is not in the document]
c3y1huang: Addressed in https://github.com/longhorn/longhorn-tests/pull/443/commits/65671fa95883bd661ae683a6823949d135949790.

yasker (approved):

yasker:
Merged, thanks @c3y1huang !

PR #446, opened 2020-10-22
Enhance test_engine_live_upgrade_with_intensive_data_writing

c3y1huang:
* Add checks for volume replica rebuild.
* Fix exception while cleaning up engine image.

### DIFF
--- a/manager/integration/tests/test_engine_upgrade.py
+++ b/manager/integration/tests/test_engine_upgrade.py
@@ -593,7 +593,7 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
         Then write new data to the upgraded volume and get the md5sum.
     11. Delete the pod and wait for the volume detached.
         Then check engine and replicas's engine image again.
-    12. Recreate the pod and wait for the volume attached.
+    12. Recreate the pod and wait for the volume replicas rebuild.
     13. Check if the attached volume is state `healthy`
         rather than `degraded`.
     14. Check the data.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_engine_upgrade.py
```diff
@@ -683,10 +683,11 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
     for replica in volume.replicas:
         assert replica.engineImage == engine_upgrade_image
 
-    # The reattached volume should be state `healthy` rather than`degraded`.
     create_and_wait_pod(core_api, pod)
-    volume = client.by_id_volume(volume_name)
-    assert volume.robustness == "healthy"
+    common.wait_for_volume_degraded(client, volume_name)
```
yasker, line 687: There is a problem with this approach. The degraded/rebuild start/rebuild complete might happen quickly, so it's hard to detect those intermediate states. And if we miss them, then the test case will fail.

We can wait for the volume to become healthy instead.

c3y1huang: Since the replica rebuilds sequentially and this one has 800MiB data and 100Mib ongoing, and this sometimes fails for me when the test failed by timed out during the rebuilding of the last replica.  To include degraded/rebuild start/rebuild complete, the test result can have more info about which stage was causing the issues. 

And since some combinations implemented in other tests and so I assumed this shouldn't be a problem. 
https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_ha.py#L1225-L1234
https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_ha.py#L1966-L1967

yasker: With enough data, yes this makes sense.

yasker (approved):

yasker:
The test case failed at `wait_for_volume_degraded` https://ci.longhorn.io/job/public/job/longhorn-tests/141/testReport/junit/tests/test_engine_upgrade/test_engine_live_upgrade_with_intensive_data_writing/

c3y1huang:
> https://ci.longhorn.io/job/public/job/longhorn-tests/141/testReport/junit/tests/test_engine_upgrade/test_engine_live_upgrade_with_intensive_data_writing/

Addressed in https://github.com/longhorn/longhorn-tests/pull/455.

yasker:
@c3y1huang Thanks!

PR #453, opened 2020-10-27
Add test skeleton for volume 'TooManySnapshots' condition

boknowswiki:
https://github.com/longhorn/longhorn/issues/496

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -3155,3 +3155,19 @@ def test_cleanup_system_generated_snapshots(client, core_api, volume_name, csi_p
 
     read_md5sum1 = get_pod_data_md5sum(core_api, pod_name, "/data/test")
     assert md5sum1 == read_md5sum1
+
+
+@pytest.mark.skip(reason="TODO")
+def test_volume_toomanysnapshots_condition(): # NOQA
+    """
+    Test Volume TooManySnapshots Condition
+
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -3155,3 +3155,18 @@ def test_cleanup_system_generated_snapshots(client, core_api, volume_name, csi_p
...
+
+@pytest.mark.skip(reason="TODO")
+def test_volume_toomanysnapshots_condition(): # NOQA
+    """
+    Test Volume TooManySnapshots Condition
+
+    1. Create a volume and attach it to a node.
+    2. Check the 'TooManySnapshots' condition is False.
+    3. Writing data to this volume and meanwhile taking 100 snapshots.
+    4. Check the 'TooManySnapshots' condition is True.
```
khushboo-rancher, line 3168: Should we add one more snapshot after ```TooManySnapshots``` condition becomes ```True```, just to make sure this condition is just a warning and will not prevent the user to take more snapshots?

khushboo-rancher (approved):
LGTM, left one comment.

manager/integration/tests/test_basic.py
boknowswiki: Sure, let me update it. Thanks.

khushboo-rancher (approved):

PR #454, opened 2020-10-28
Update test skeleton to check backup cronjob that runs on node with taint

PhanLe1010:
longhorn/longhorn#1903

### DIFF
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -483,3 +483,43 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
     17. cleanup
     """
     pass
+
+
+@pytest.mark.skip(reason="TODO")
+def test_recurring_jobs_on_nodes_with_taints():  # NOQA
+    """
+    Test recurring jobs on nodes with taints
+
+    Context:
+
+    Test the prevention of creation of multiple pods due to
+    recurring job's pod being rejected by Taint controller
+    on nodes with taints
+
+    Steps:
+
+    1. Set taint toleration for Longhorn components
+       `persistence=true:NoExecute`
+    2. Taint `node-1` with `persistence=true:NoExecute`
+    3. Create a volume, vol-1.
+       Attach vol-1 to node-1
+       Write some data to vol-1
+    4. Create a recurring backup job which:
+       Has retain count 10
+       Runs every minute
+    5. Wait for 3 minutes.
+       Verify that the there is 1 backup created
+       Verify that the total number of pod in longhorn-system namespace < 50
+       Verify that the number of pods of the cronjob is <= 2
+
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_recurring_job.py
```diff
@@ -483,3 +483,39 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+
+    Steps:
+
+    1. Set taint toleration for Longhorn components
+       `persistence=true:NoExecute`
+    1. Taint `node-1` with `persistence=true:NoExecute`
+    1. Create a volume, vol-1.
+       Attach vol-1 to node-1
+       Write some data to vol-1
+    1. Create a recurring backup job which runs every minute
```
khushboo-rancher, line 505: What's the retain count of the recurring job here, more than 1?

manager/integration/tests/test_recurring_job.py
```diff
@@ -483,3 +483,39 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+    1. Write some data to vol-1
+    1. Wait for 3 minutes.
+       Verify that the there are 2 backups created in total
+       Verify that the total number of pod in longhorn-system namespace < 50
+       Verify that the number of pods of the cronjob is <= 2
+
+    1. Remove `persistence=true:NoExecute` from all nodes and Longhorn setting
+       Clean up backups, volumes
+    """
+    pass
```
khushboo-rancher, line 521: I think there is a new line required here or you can remove the ```pass``` as the test is already skipped.

manager/integration/tests/test_recurring_job.py
PhanLe1010: Thank you! Fixed

manager/integration/tests/test_recurring_job.py
PhanLe1010: Thank you! Fixed

manager/integration/tests/test_recurring_job.py
```diff
@@ -483,3 +483,41 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+
+
+@pytest.mark.skip(reason="TODO")
+def test_recurring_jobs_on_nodes_with_taints():  # NOQA
+    """
+    Test recurring jobs on nodes with taints
+
+    Context:
+
+    longhorn/longhorn#1903
```
khushboo-rancher, line 495: Can we write something like 'Test the prevention of the creation of multiple pods due to recurring job failure' instead of the bug id?

khushboo-rancher:
LGTM, Serial number should be serialized.

manager/integration/tests/test_recurring_job.py
PhanLe1010: Thank you! Fixed

khushboo-rancher (approved):

PR #464, opened 2020-11-02
Fix test_salvage_auto_crash_all_replicas test case

PhanLe1010:
It is hard to capture the pending state of the workload pod. Moreover, we only need to make sure that the pod is in `running` state and `wait_for_pod_remount()` returns true. That is enough to conclude that the pod was restarted and the volume was remount. Therefore, we removed the `wait_for_pod_phase(core_api, pod_name, pod_phase="Pending")`

After this PR is merged, this test case sometimes still fails because of the issue https://github.com/longhorn/longhorn/issues/1942

longhorn/longhorn#1914

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -488,7 +488,6 @@ def wait_pod_for_auto_salvage(
 
     wait_for_volume_healthy(client, volume_name)
 
-    common.wait_for_pod_phase(core_api, pod_name, pod_phase="Pending")
     common.wait_for_pod_phase(core_api, pod_name, pod_phase="Running")
 
     wait_for_pod_remount(core_api, pod_name)


PhanLe1010:
cc @c3y1huang

yasker (approved):

c3y1huang:
@PhanLe1010 sorry about the late review, I'm a little concerned about this change since removing this means there is no checking for the intermediate state of pod restart. By the look, it takes at least 5 seconds until K8s get the pod termination request and I think the test could potentially miss the restart, checked for the pod before terminate and pass. I suggest to try-except this instead.

PhanLe1010:
@c3y1huang Thank you for your review! I am sorry that I missed your comment.

> By the look, it takes at least 5 seconds until K8s get the pod termination request and I think the test could potentially miss the restart, checked for the pod before terminate and pass

I think that is impossible that the test passed but the pod has never restarted. In fact, I think that we can skip both checking for the pod to be in `Running` state as well as `Pending` state. 

Following the follow of the e2e test, we can see that the e2e test fist crash all replicas:
https://github.com/longhorn/longhorn-tests/blob/f8d1cebba1102a2bab365b784cfe8b5ba43b0e04/manager/integration/tests/test_ha.py#L531

This makes the volume faulty and Longhorn does auto reattachment. The e2e test then check this 
https://github.com/longhorn/longhorn-tests/blob/f8d1cebba1102a2bab365b784cfe8b5ba43b0e04/manager/integration/tests/test_ha.py#L497

Because we removed all the auto-remount logic, the only way to pass this check is that the workload pod has to restart (either manually by user or automatically by Longhorn).

However, adding the  `Running` state as well as `Pending` state checking for the workload pod is nicer because it makes the e2e test clearer, easier to read  and follow. Therefore, I agree that we should try-except for checking the  `Pending` state.

Please correct me if I miss something :)

PR #465, opened 2020-11-03
Try except the pod restart intermediate phase

c3y1huang:
Without checking for the intermediate phase could potentially miss the pod restart, try-except to allow enough time for K8s to handle the pod deletion.

https://github.com/longhorn/longhorn-tests/pull/464

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -488,6 +488,11 @@ def wait_pod_for_auto_salvage(
 
     wait_for_volume_healthy(client, volume_name)
 
+    try:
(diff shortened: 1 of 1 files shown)


yasker (approved):

yasker:
It's the practical way of dealing with it right now. But it bothers me we don't have a better way to track those intermediate states. I am thinking is there any other better ways to deal with it? Can we rely on `events`?

PhanLe1010:
I have a thought that is related to this PR:
https://github.com/longhorn/longhorn-tests/pull/464#issuecomment-722682107

PR #466, opened 2020-11-03
Fix test_engine_live_upgrade_with_intensive_data_writing

c3y1huang:
~Refresh volume object and clean before detached by pod.~
Remove cleanups since already handled by the finalizer.

https://ci.longhorn.io/job/public/job/longhorn-tests/148/testReport/junit/tests/test_engine_upgrade/test_engine_live_upgrade_with_intensive_data_writing/

### DIFF
--- a/manager/integration/tests/test_engine_upgrade.py
+++ b/manager/integration/tests/test_engine_upgrade.py
@@ -15,7 +15,6 @@ from common import pod_make  # NOQA
 from common import create_pv_for_volume, create_pvc_for_volume
 from common import create_pvc_spec
 from common import create_and_check_volume, create_and_wait_pod
-from common import delete_and_wait_pvc, delete_and_wait_pv
 from common import delete_and_wait_pod
 from common import write_pod_volume_random_data, get_pod_data_md5sum
 from common import copy_pod_volume_data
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_engine_upgrade.py
```diff
@@ -693,8 +693,9 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
         core_api, pod_name, data_path2)
     assert volume_file_md5sum2 == original_md5sum2
 
+    volume = client.by_id_volume(volume_name)
+    common.cleanup_volume(client, volume)
```
yasker, line 697: It's not a good idea to clean up the volume before cleanup the Pod/PV/PVC.

The error happens because the volume might be gone already.

We can remove those cleanup logic and leave it to the finalizer part of `pod_make` https://github.com/longhorn/longhorn-tests/blob/065257bc50a2aae0e0acb16484ed452313ee9b91/manager/integration/tests/common.py#L737

c3y1huang: Addressed.

yasker (approved):

PR #468, opened 2020-11-04
Fix data_locality_test failure

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/1951

Signed-off-by: khushboo <khushboo@rancher.com>

### DIFF
--- a/manager/integration/tests/test_scheduling.py
+++ b/manager/integration/tests/test_scheduling.py
@@ -22,6 +22,7 @@ from common import wait_for_rebuild_start
 from common import delete_and_wait_pod
 from common import wait_for_replica_failed
 from common import crash_engine_process_with_sigkill
+from common import wait_for_replica_running
 
 from time import sleep
 
@@ -394,7 +395,7 @@ def test_replica_rebuild_per_volume_limit():
     pass
 
 
-def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset): # NOQA
+def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset):  # NOQA
     """
     Test data locality basic feature
 
@@ -735,7 +736,6 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
     create_and_wait_pod(core_api, pod3)
 
     wait_for_rebuild_start(client, volume3_name)
-    volume3 = client.by_id_volume(volume3_name)
     crash_engine_process_with_sigkill(client, core_api, volume3_name)
     volume3 = client.by_id_volume(volume3_name)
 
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_scheduling.py
```diff
@@ -881,6 +880,7 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
     v4_node3_replica_count = 0
 
     for replica in volume4.replicas:
+        wait_for_replica_running(client, volume4_name, replica["name"])
```
khushboo-rancher, line 883: Additional wait to check replica is in running state is introduced before checking the ```hostId``` of the replica because immediately after ```wait_for_volume_replica_count``` method completes, we don't have the ```hostId``` and other details available in the replica.
```
[{'address': '', 'currentImage': '', 'dataPath': '', 'diskID': '', 'engineImage': 'longhornio/longhorn-engine:master', 'failedAt': '', 'hostId': '', 'instanceManagerName': '', 'mode': '', 'name': 'longhorn-testvol-l2gpw4-4-r-3caf1504', 'running': False}, {'address': 'xx.xx.xx.xx', 'currentImage': 'longhornio/longhorn-engine:master', 'dataPath': '/var/lib/longhorn/replicas/longhorn-testvol-l2gpw4-4-a776fd90', 'diskID': 'default-disk-3e72533a17f25ad4', 'engineImage': 'longhornio/longhorn-engine:master', 'failedAt': '', 'hostId': 'node-wk2', 'instanceManagerName': 'instance-manager-r-3cc712b8', 'mode': 'RW', 'name': 'longhorn-testvol-l2gpw4-4-r-c67fb79d', 'running': True}]
```

manager/integration/tests/test_scheduling.py
```diff
@@ -754,16 +754,15 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
...
     create_and_wait_pod(core_api, pod3)
     wait_for_rebuild_start(client, volume3_name)
     volume3 = client.by_id_volume(volume3_name)
     assert len(volume3.replicas) == 2
     wait_for_rebuild_complete(client, volume3_name)
 
-    volume3 = client.by_id_volume(volume3_name)
-    assert len(volume3.replicas) == 1
+    # Wait for deletion of extra replica
+    volume3 = wait_for_volume_replica_count(client, volume3_name, 1)
```
khushboo-rancher, line 765: In case of the rebuilding of a replica due to data locality, we need to wait for deletion of the extra replica before counting the number of replicas

manager/integration/tests/test_scheduling.py
yasker: Does that mean we need to wait for replica running for line 765 as well?

khushboo-rancher: This is a case where the replica count has just increased and ```wait_for_volume_replica_count``` returns the correct count but the ```hostId``` and other value are yet to be assigned to the replica. 
Whereas at line 765 the replica is being deleted and once  ```wait_for_volume_replica_count``` confirms that we'll not need ```wait_for_replica_running```.

yasker (approved):

PR #471, opened 2020-11-05
Update tests for disk migration

shuo-wu:
longhorn/longhorn#1269


yasker (approved):
LGTM. Wait for the manager part to be merged in.

PR #481, opened 2020-11-12
Add manual test plan to verify that CSI components' images are updated

PhanLe1010:
longhorn/longhorn#1642

### DIFF
--- a/docs/content/manual/pre-release/upgrade/update_csi_components_when_images_change.md
+++ b/docs/content/manual/pre-release/upgrade/update_csi_components_when_images_change.md
@@ -1,0 +1,10 @@
+---
+title: Re-deploy CSI components when their images change
+---
+
+1. Install Longhorn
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Will add negative test case

PhanLe1010:
Added negative test cases

yasker:
Let's put this into the pre-release test since deploying the same image shouldn't trigger the redeploy even Longhorn is upgraded.

PhanLe1010:
I moved it to `pre-release`

yasker (approved):
LGTM. Waiting for the manager part to be checked in.

yasker:
Not sure why this test case was missed. cc @PhanLe1010

PhanLe1010:
I am not sure neither. May be we were waiting and forgot 
> LGTM. Waiting for the manager part to be checked in.

PR #483, opened 2020-11-13
Implement test_allow_volume_creation_with_degraded_availability_restore

c3y1huang:
Split original into two scenarios.

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -3099,8 +3099,8 @@ def test_allow_volume_creation_with_degraded_availability(client, volume_name): 
     common.wait_for_replica_scheduled(client, volume_name,
                                       to_nodes=[node1.name],
                                       expect_success=1, expect_fail=2,
-                                      is_vol_healthy=False,
-                                      is_replica_running=False)
+                                      chk_vol_healthy=False,
+                                      chk_replica_running=False)
 
     # enable node 2 to schedule to node 1 and 2
     client.update(node2, allowScheduling=True)
@@ -3109,8 +3109,8 @@ def test_allow_volume_creation_with_degraded_availability(client, volume_name): 
     common.wait_for_replica_scheduled(client, volume_name,
                                       to_nodes=[node1.name, node2.name],
                                       expect_success=2, expect_fail=1,
-                                      is_vol_healthy=False,
-                                      is_replica_running=False)
+                                      chk_vol_healthy=False,
+                                      chk_replica_running=False)
 
     volume = client.by_id_volume(volume_name)
     assert volume.conditions[VOLUME_CONDITION_SCHEDULED]['status'] == "True"
@@ -3208,8 +3208,8 @@ def test_allow_volume_creation_with_degraded_availability_error(
     common.wait_for_replica_scheduled(client, volume_name,
                                       to_nodes=[node1.name],
                                       expect_success=1, expect_fail=2,
-                                      is_vol_healthy=False,
-                                      is_replica_running=False)
+                                      chk_vol_healthy=False,
+                                      chk_replica_running=False)
     volume = common.wait_for_volume_status(client, volume_name,
                                            VOLUME_FIELD_READY, True)
     assert volume.conditions[VOLUME_CONDITION_SCHEDULED]['status'] == "True"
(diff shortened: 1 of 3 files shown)


c3y1huang:
Currently, the two tests both failed.
The `_restore` failed from 2.2, by observation only 1 replica is found where expecting a total of 3 replicas(1 running, 2 stopped). 
The `_dr` failed from 2.1 where the schedule condition stayed false.

shuo-wu:
The scheduling failed replicas somehow are marked as `ERROR` during the restore, which leads to unexpected test failures. I will check it later.

c3y1huang:
Rebased.

yasker:
@shuo-wu Can you check the issue?

shuo-wu:
I just re-checked the test step. The failure of the tests is caused by the transient state. For restore/DR volumes, Longhorn will automatically and immediately attach the volume once it's created. The volume schedule condition will be false once the volume is attached/attaching. And the scheduling failed replicas will be cleaned up once Longhorn automatically detaches the volume after restoration complete. In brief, you need to:
1. Check the volume current state before verifying other fields.
2. Avoid checking the transient state

c3y1huang:
@shuo-wu thanks for the review and explanation. 

I've removed the checks for failed replicas to bypass the transient state. However, tests are still failing.
* No additional replica got scheduled after enable the second node. Observe that the restoration of additional replica starts when the volume gets attached near the end of the tests.
* During restore, test `_restore` schedule condition stayed true, but test `_dr` stayed false.

shuo-wu:
- Sometimes the replica scheduling of the restoring volume may take several seconds after enabling the 2nd node. And you may need to be careful of the race condition: the restore is complete before the scheduling success. But based on my manual test with Longhorn v1.1.0, it does work fine. BTW, you can find the restore status besides the rebuilding state in the engine for the newly scheduled replica since the rebuilding of a restore/DR volume is actually full restoration.
- The scheduling condition of both DR volumes and restore volume is False during the restore. I am sorry for not pointing out that this wrong in the skeleton. Remember, the scheduling condition of a volume is always False if the node scheduling is not enabled and there is an unscheduled replica in the volume.

shuo-wu:
Here is the modified test skeleton for the restore volume part:
```
1. Disable scheduling for node 2 and 3.
2. Restore a volume with 3 replicas.
    1. Volume should be attached automatically and `Scheduled` is true.
    2. Only 1 replica becomes scheduled.
3. During the restore, enable scheduling for node 2. Then check the followings if the volume restore is still in progress:
    1. 1 additional replica of the volume will become scheduled.
    2. The restore status and the rebuild status for this newly scheduled replica. 
    3. The scheduled condition is still false.
4. Wait for the restore to complete and volume detach automatically. Then check the followings:
    1. After the volume detached, the only unscheduled replica will be removed.
    2. The scheduled condition will become true.
5. Attach and wait for the volume. 
    1. A new replica will be created with scheduling failure.
    2. The scheduled condition becomes false.
```

As for DR volumes, the main difference is that the scheduling condition will become true as long as all node scheduling is enabled and all replicas become running before activation.

yasker:
Any update @c3y1huang ?

c3y1huang:
> Any update @c3y1huang ?

I've almost forgotten about this one, will come back to this later.

c3y1huang:
> * Sometimes the replica scheduling of the restoring volume may take several seconds after enabling the 2nd node. And you may need to be careful of the race condition: the restore is complete before the scheduling success. But based on my manual test with Longhorn v1.1.0, it does work fine. BTW, you can find the restore status besides the rebuilding state in the engine for the newly scheduled replica since the rebuilding of a restore/DR volume is actually full restoration.

@shuo-wu I've tried again and observed the restore gets completed very quickly (~1 second), and the 2nd node replica never got scheduled, nor is the rebuilding. I guess this is the race condition you've talked about. I've tried to increase the backup to 1600MB and seeing the same result. Is this expected?

shuo-wu:
Hmmm... I just retied the test. Restoring 800MB~900MB data from the test NFS backupstore took 5~15s, which does not be able to guarantee the test always successful. Maybe we can remove the healthy/replica count check for the restore volume. As long as the data of the restored volume is correct, I will consider the test is passed.

c3y1huang:
> Hmmm... I just retied the test. Restoring 800MB~900MB data from the test NFS backupstore took 5~15s, which does not be able to guarantee the test always successful. Maybe we can remove the healthy/replica count check for the restore volume. As long as the data of the restored volume is correct, I will consider the test is passed.

Seem the intermediate checks are hard to catch (couldn't in my case), I've modified tests to mostly only verify the expecting result.

manager/integration/tests/test_basic.py
```diff
@@ -3252,45 +3252,233 @@ def test_multiple_volumes_creation_with_degraded_availability():
...
-        3. Scheduled condition is still true
-    4. Wait for the restore to complete.
-    5. Enable the scheduling for node 3.
-        1. DR volume should automatically rebuild the third replica.
-    6. Activate the volume and verify the data.
+    1. Disable scheduling for node 2 and 3.
+    2. Create a DR volume from backup with 3 replicas.
+        1. The scheduled condition is false.
+        2. Only node 1 replica become scheduled.
+    3. Enable scheduling for node 2 and 3.
```
shuo-wu, line 3385: We can wait for all replica scheduling then restore/rebuild success here since DR volume won't be detached automatically if it's not activated

shuo-wu (changes requested):

manager/integration/tests/test_basic.py
c3y1huang: Addressed.

shuo-wu (approved):

PR #487, opened 2020-11-20
Implement test_replica_rebuild_per_volume_limit

c3y1huang:

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -67,6 +67,8 @@ VOLUME_ROBUSTNESS_UNKNOWN = "unknown"
 
 VOLUME_FIELD_RESTOREREQUIRED = "restoreRequired"
 
+VOLUME_REPLICA_WO_LIMIT = 1
+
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -1442,26 +1444,34 @@ def wait_for_volume_replica_count(client, name, count):
...
+
+        count = 0
+        wo_count = 0
         for r in replicas:
             if r.mode == mode:
                 count += 1
-        if count == len(replicas):
+            if r.mode == 'WO':
+                wo_count += 1
+            assert wo_count <= VOLUME_REPLICA_WO_LIMIT
```
yasker, line 1471: Should put this outside of the loop.

yasker (changes requested):

manager/integration/tests/common.py
c3y1huang: Addressed.

yasker (approved):

PR #490, opened 2020-11-26
implement test_recurring_jobs_for_detached_volume

c3y1huang:
https://github.com/longhorn/longhorn/issues/1509

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -120,6 +120,8 @@ SETTING_DEGRADED_AVAILABILITY = \
     "allow-volume-creation-with-degraded-availability"
 SETTING_REPLICA_REPLENISHMENT_WAIT_INTERVAL = \
     "replica-replenishment-wait-interval"
+SETTING_RECURRING_JOB_WHILE_VOLUME_DETACHED = \
+    "allow-recurring-job-while-volume-detached"
 
 DEFAULT_DISK_PATH = "/var/lib/longhorn/"
 DEFAULT_STORAGE_OVER_PROVISIONING_PERCENTAGE = "500"
@@ -1334,6 +1336,12 @@ def wait_for_volume_endpoint(client, name):
     return v
 
 
+def wait_for_volume_attached(client, name):
+    return wait_for_volume_status(client, name,
+                                  VOLUME_FIELD_STATE,
+                                  VOLUME_STATE_ATTACHED)
+
+
 def wait_for_volume_detached(client, name):
     return wait_for_volume_status(client, name,
                                   VOLUME_FIELD_STATE,
@@ -1479,6 +1487,16 @@ def wait_for_volume_replicas_mode(client, volname, mode,
 
     assert verified
     return volume
+
+
+def wait_for_volume_frontend_disabled(client, volume_name, state=True):
+    for _ in range(RETRY_COUNTS):
+        vol = client.by_id_volume(volume_name)
+        try:
+            assert vol.disableFrontend is state
+            break
+        except AssertionError:
+            time.sleep(RETRY_INTERVAL)
 
 
 def wait_for_snapshot_purge(client, volume_name, *snaps):
@@ -2214,13 +2232,15 @@ def get_volume_attached_nodes(v):
     return nodes
 
 
-def wait_for_backup_completion(client, volume_name, snapshot_name,
+def wait_for_backup_completion(client, volume_name, snapshot_name=None,
                                retry_count=RETRY_BACKUP_COUNTS):
     completed = False
-    for i in range(retry_count):
+    for _ in range(retry_count):
         v = client.by_id_volume(volume_name)
         for b in v.backupStatus:
-            if b.snapshot == snapshot_name and b.state == "complete":
+            if snapshot_name is not None and b.snapshot != snapshot_name:
+                continue
+            if b.state == "complete":
                 assert b.progress == 100
                 assert b.error == ""
                 completed = True
@@ -2232,13 +2252,44 @@ def wait_for_backup_completion(client, volume_name, snapshot_name,
     return v
 
 
-def wait_for_backup_to_start(client, volume_name, snapshot_name,
+def wait_pod_auto_attach_after_first_backup_completion(
+        client, core_api, volume_name, label_name):
+    completed = False
+    for _ in range(RETRY_BACKUP_COUNTS):
+        vol = client.by_id_volume(volume_name)
+        for b in vol.backupStatus:
+            if b.state == 'complete':
+                assert b.progress == 100
+                assert b.error == ''
+                completed = True
+                break
+        if completed:
+            wait_for_volume_detached(client, vol.name)
+            wait_for_volume_frontend_disabled(client, vol.name, False)
+            wait_for_volume_attached(client, vol.name)
+            break
+
+        label_selector = "name=" + label_name
+        pods = core_api.list_namespaced_pod(namespace="default",
+                                            label_selector=label_selector)
+        for pod in pods.items:
+            assert pod.status.phase != 'Running'
+        assert vol.disableFrontend is True
+
+        time.sleep(RETRY_BACKUP_INTERVAL)
+    assert completed is True
+    return vol
+
+
+def wait_for_backup_to_start(client, volume_name, snapshot_name=None,
                              retry_count=RETRY_BACKUP_COUNTS):
     in_progress = False
-    for i in range(retry_count):
+    for _ in range(retry_count):
         v = client.by_id_volume(volume_name)
         for b in v.backupStatus:
-            if b.snapshot == snapshot_name and b.state == "in_progress":
+            if snapshot_name is not None and b.snapshot != snapshot_name:
+                continue
+            if b.state == "in_progress":
                 assert b.progress > 0
                 assert b.error == ""
                 in_progress = True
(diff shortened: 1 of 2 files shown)


c3y1huang:
This currently failed at step 15.

c3y1huang:
This currently failed at step 15.

yasker:
@PhanLe1010 Can you help to review this?

PhanLe1010:
Sure

manager/integration/tests/test_recurring_job.py
```diff
@@ -476,13 +479,96 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+        {
+            "name": RECURRING_JOB_NAME,
+            "cron": "*/2 * * * *",
+            "task": "backup",
+            "retain": 1
+        }
+    ]
+    vol.recurringUpdate(jobs=jobs)
+
+    time.sleep(120)
```
PhanLe1010, line 556: Can we skip this sleep?

manager/integration/tests/common.py
```diff
@@ -2231,13 +2235,43 @@ def wait_for_backup_completion(client, volume_name, snapshot_name,
     return v
 
 
-def wait_for_backup_to_start(client, volume_name, snapshot_name,
+def wait_for_recurring_backup_completion(client, core_api,
```
PhanLe1010, line 2238: This function is waiting for the first backup completion. Also, if the volume is currently attaching to a running pod, the pod phase will be `running` while the backup is happening. 
Should we name it `wait_for_first_autoattach_recurring_backup_completion` to avoid miss-using this function in the future?

PhanLe1010 (dismissed):
In general, LGTM

manager/integration/tests/test_recurring_job.py
```diff
@@ -476,13 +479,96 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
     13. Verify that during the recurring backup, the volume's frontend is
         disabled, and pod cannot start.
     14. Wait for the recurring backup finishes.
-        Delete the recurring backup
-    15. In a 6-mintute retry loop, verify that the pod can eventually start.
-
-    16. Change the setting allow-recurring-job-while-volume-detached to false
-    17. cleanup
+        Delete the recurring backup.
+    15. In a 10-minutes retry loop, verify that the pod can eventually start.
```
PhanLe1010, line 483: Wow. I takes 10 mins. Seems like a significant downtime.

PhanLe1010:
Oops. I missed this comment 
> This currently failed at step 15.

I will retest this feature.

manager/integration/tests/common.py
c3y1huang: You are right the function name should speak more for itself. The `wait_for_first_autoattach_recurring_backup_completion` is a little hard to read imo, how about `wait_pod_auto_attach_after_first_backup_completion`?

manager/integration/tests/test_recurring_job.py
c3y1huang: The intention is to show the to wait for the cronjob to start, but sure we could skip this since it's already covered in wait_for_backup_to_start.

Addressed.

manager/integration/tests/common.py
PhanLe1010: Yeah, good idea!

manager/integration/tests/test_recurring_job.py
PhanLe1010: Thanks

manager/integration/tests/test_recurring_job.py
```diff
@@ -476,13 +479,96 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
-        Delete the recurring backup
-    15. In a 6-mintute retry loop, verify that the pod can eventually start.
-
-    16. Change the setting allow-recurring-job-while-volume-detached to false
-    17. cleanup
+        Delete the recurring backup.
+    15. In a 6-minutes retry loop, verify that the pod can eventually start.
+    16. Change the setting allow-recurring-job-while-volume-detached to false.
+    17. Cleanup.
     """
```
PhanLe1010, line 486: Can we set Backupstore here
```
set_random_backupstore(client)
```
Without this, the test will fail when it is run separately since it cannot find a backup store.

manager/integration/tests/test_recurring_job.py
c3y1huang: Changed back 6 minutes in the original test plan.

manager/integration/tests/test_recurring_job.py
```diff
@@ -476,13 +479,96 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+    client.update(recurring_job_setting, value="true")
+
+    vol = common.create_and_check_volume(client, volume_name, size=str(1 * Gi))
+
+    lht_hostId = get_self_host_id()
+    vol.attach(hostId=lht_hostId)
+    vol = common.wait_for_volume_healthy(client, vol.name)
+
+    data = {
+        'pos': int(SIZE),
```
PhanLe1010, line 498: I am not sure why we choose `pos=SIZE`. I see this on some other places as well. Should `pos` be 0 because it is the location in the block device that we start writing to?

manager/integration/tests/test_recurring_job.py
```diff
@@ -476,13 +479,96 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+        'content': common.generate_random_data(50 * Mi),
+    }
+    common.write_volume_data(vol, data)
+
```
PhanLe1010, line 502: Before detaching, should we add 
```
time.sleep(15)
```
so that the volume calm down and data are flushed to disk after a big write?

manager/integration/tests/test_recurring_job.py
```diff
@@ -476,13 +479,96 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+            "retain": 1
+        }
+    ]
+    vol.recurringUpdate(jobs=jobs)
+    common.wait_for_backup_completion(client, vol.name)
+    for _ in range(4):
+        bv = client.by_id_backupVolume(vol.name)
+        backups = bv.backupList().data
+        assert len(backups) == 1
+        time.sleep(60)
```
PhanLe1010, line 520: I think 2 mins is enough, e.g. `time.sleep(30)`,  because the recurring backup job run every minute and we only need to make sure that it doesn't make a new backup in the next run. How do you think?

manager/integration/tests/test_recurring_job.py
```diff
@@ -476,13 +479,96 @@ def test_recurring_jobs_for_detached_volume():  # NOQA
...
+
+    deployment['spec']['replicas'] = 1
+    apps_api.patch_namespaced_deployment(body=deployment,
+                                         namespace='default',
+                                         name=deployment["metadata"]["name"])
+
+    deployment_label_name = deployment["metadata"]["labels"]["name"]
+    common.wait_pod_auto_attach_after_first_backup_completion(
+        client, core_api, vol.name, deployment_label_name)
+
```
PhanLe1010, line 566: Can we put 
```
time.sleep(30)
```
here? After the recurring job finish, we need to give it sometime so that it can detach the volume properly. Deleting it immediately will make it not be able to detach the volume and the consuming workload can never use the volume which makes the test case fail.

PhanLe1010 (changes requested):
Thank you for supper quick response. I have left some comments. The test passed when I implemented those comments.

manager/integration/tests/test_recurring_job.py
PhanLe1010: Oops, It is good. I think 10 mins is good! It actually take longer than I thought.

manager/integration/tests/test_recurring_job.py
c3y1huang: Addressed.

manager/integration/tests/test_recurring_job.py
c3y1huang: Addressed.

manager/integration/tests/test_recurring_job.py
c3y1huang: Addressed.

manager/integration/tests/test_recurring_job.py
c3y1huang: Addressed. The wait time looks random though, I wonder if there are ways to check when recurring job can be deleted?

PhanLe1010: Good idea! We can wait for the volume to be in (detached state or frontend is enabled). It means that the recurring job has finished and thus we can delete it

manager/integration/tests/test_recurring_job.py
c3y1huang: Addressed.

PhanLe1010 (dismissed):
LGTM

PhanLe1010 (commented):
LGTM except that we need to change the waittime from 6 to 10 mins https://github.com/longhorn/longhorn-tests/pull/490#discussion_r538154819

manager/integration/tests/test_recurring_job.py
yasker: @c3y1huang Does it require 10 minutes? or 6 minutes is good enough. If 6-minute is good enough, we don't need to update it to 10 minutes.

manager/integration/tests/test_recurring_job.py
c3y1huang: Addressed. I've removed the sleep and check for volume detach/attach and volume frontend instead. Thanks :)

manager/integration/tests/test_recurring_job.py
c3y1huang: @yasker I've tried with 6 minutes but occasionally will fail on time-out. Wait for 10 minutes gives more stable results and it's I see it's also the default time wait for other tests.

manager/integration/tests/test_recurring_job.py
PhanLe1010: Thank you!

PhanLe1010 (approved):
LGTM

PR #493, opened 2020-11-30
Implement test_setting_toleration_extra

c3y1huang:
https://github.com/longhorn/longhorn/issues/1977

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -102,11 +102,20 @@ def test_setting_toleration():
     assert 'invalid effect' in str(e.value)
 
     setting_value_str = "key1=value1:NoSchedule; key2:NoExecute"
-    setting_value_dict = \
-        {"key1": {"key": "key1", "value": "value1",
-                  "operator": "Equal", "effect": "NoSchedule"},
-         "key2": {"key": "key2", "value": None,
-                  "operator": "Exists", "effect": "NoExecute"}, }
+    setting_value_dicts = [
+        {
+            "key": "key1",
+            "value": "value1",
+            "operator": "Equal",
+            "effect": "NoSchedule"
+        },
+        {
+            "key": "key2",
+            "value": None,
+            "operator": "Exists",
+            "effect": "NoExecute"
+        },
+    ]
 
     volume_name = "test-toleration-vol"  # NOQA
     volume = create_and_check_volume(client, volume_name)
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_settings.py
```diff
@@ -183,26 +191,87 @@ def test_setting_toleration_extra():
...
+        },
+        {
+            "value": "",
+            "expect": [],
+        },
+    ]
+
+    chk_removed_tolerations = []
+    for setting in settings:
+        client = get_longhorn_api_client()  # NOQA
```
PhanLe1010, line 242: Why do you put this one inside the `for` loop?

manager/integration/tests/test_settings.py
```diff
@@ -183,26 +191,87 @@ def test_setting_toleration_extra():
...
+        taint_toleration = client.by_id_setting(SETTING_TAINT_TOLERATION)
+        updated = client.update(taint_toleration,
+                                value=setting["value"])
+        assert updated.value == setting["value"]
+
+        core_api = get_core_api_client()  # NOQA
```
PhanLe1010, line 248: Why do you put this one inside the `for` loop?

manager/integration/tests/test_settings.py
```diff
@@ -183,26 +191,87 @@ def test_setting_toleration_extra():
...
         for p in pod_list:
-            if p.status.phase != "Running" or \
-                    not check_tolerations_set(p.spec.tolerations,
-                                              set_tolerations):
-                updated = False
+            if p.status.phase == "Running" \
+                and check_tolerations_set(p.spec.tolerations,
+                                          expected_tolerations,
+                                          chk_removed_tolerations):
+                updated = True
```
PhanLe1010, line 274: I think there is an issue here. The expectation is that all pods in `longhorn-system` namespace must be updated. However, the new code is reporting ok if there is a pod that was updated. How do you think?

manager/integration/tests/test_settings.py
c3y1huang: You are right, it's not required. Addressed.

manager/integration/tests/test_settings.py
c3y1huang: You are correct. Addressed.

manager/integration/tests/test_settings.py
c3y1huang: Seems the update triggers pods to restart and have a new cluster IP assignment. Hence makes the previous client invalid.

c3y1huang:
@PhanLe1010 thanks for your review, all comments are addressed.

manager/integration/tests/test_settings.py
PhanLe1010: Thank you for the explanation!

PhanLe1010 (approved):
LGTM

PR #494, opened 2020-12-02
Implement test_allow_volume_creation_with_degraded_availability_csi

c3y1huang:

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -3508,6 +3508,20 @@ def create_and_wait_deployment(apps_api, deployment_manifest):
     )
 
 
+def wait_and_get_any_deployment_pod(core_api, deployment_name,
+                                    is_phase="Running"):
+    for _ in range(DEFAULT_DEPLOYMENT_TIMEOUT):
+        label_selector = "name=" + deployment_name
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -3493,19 +3493,34 @@ def wait_deployment_replica_ready(apps_api, deployment_name,
     assert replicas_ready
 
 
-def create_and_wait_deployment(apps_api, deployment_manifest):
+def create_and_wait_deployment(apps_api, deployment_manifest, wait=True):
```
shuo-wu, line 3496: Since this function name is "create_**and_wait**_deployment", it's weird to add this argument. You can create another function to create deployment only.
```
def create_and_wait_deployment(apps_api, deployment_manifest):
    create_deployment()
    wait_deployment_replica_ready()
    ......


def create_deployment(apps_api, deployment_manifest):
```

shuo-wu (changes requested):
Others LGTM

manager/integration/tests/common.py
c3y1huang: You are right, this does look weird. Addressed.

shuo-wu (approved):

PR #495, opened 2020-12-02
Remove the special toleration that tolerates everything from the test case test_setting_toleration_extra

PhanLe1010:
The special toleration `:` makes Longhorn deploy manager pods on control/etcd nodes and when the test clears the toleration the control/etcd nodes become "down" in the Longhorn system. The result is that Longhorn has 4 nodes with 1 failed node. This breaks other test cases.

We will re-enable this special toleration once we implement the logic for deleting failed nodes

longhorn/longhorn#2024

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -220,17 +220,23 @@ def test_setting_toleration_extra(core_api, apps_api):  # NOQA
                 },
             ],
         },
-        {
-            "value": ":",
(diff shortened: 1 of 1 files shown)


PhanLe1010:
cc @c3y1huang

yasker (approved):

yasker:
Related issue: https://github.com/longhorn/longhorn/issues/2026

PR #497, opened 2020-12-08
Ensure priority class setting gets reset irregardless of test failure

joshimoo:
longhorn/longhorn#2055
The api returns 500 when trying to set the priority class setting, while there are volumes that are non detached.
Therefore I only clear the priority setting when it's not empty already, to prevent misleading error messages in the log.
There seems to be some timing issues here, that there are some volumes still not detached (might be in detaching) state during cleanup.


khushboo-rancher (approved):
LGTM

PR #502, opened 2020-12-17
Fix: test_data_locality_basic sometimes fails

PhanLe1010:
Fixed some timing issues. More explanation in comments below

### DIFF
--- a/manager/integration/tests/test_scheduling.py
+++ b/manager/integration/tests/test_scheduling.py
@@ -13,6 +13,7 @@ from common import check_volume_data, cleanup_volume, \
     write_volume_random_data, wait_for_rebuild_complete
 from common import SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY
 from common import SETTING_DEFAULT_DATA_LOCALITY
+from common import VOLUME_FIELD_ROBUSTNESS, VOLUME_ROBUSTNESS_HEALTHY
 from common import create_pv_for_volume
 from common import create_pvc_for_volume
 from common import write_pod_volume_random_data
@@ -23,7 +24,6 @@ from common import create_and_wait_pod
 from common import settings_reset # NOQA
 from common import wait_for_rebuild_start
 from common import delete_and_wait_pod
-from common import wait_for_replica_failed
 from common import crash_engine_process_with_sigkill
 from common import wait_for_replica_running
 
@@ -439,6 +439,8 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
 
     Steps:
 
+    Case 1: Test that Longhorn builds a local replica on the engine node
+
     1. Create a volume(1) with 1 replica and dataLocality set to disabled
     2. Find node where the replica is located on.
        Let's call the node is replica-node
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_scheduling.py
```diff
@@ -611,18 +611,12 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
...
         {"kubernetes.io/hostname": volume1_attached_node}
     create_and_wait_pod(core_api, pod1)
-
-    wait_for_rebuild_start(client, volume1_name)
-
-    volume1 = client.by_id_volume(volume1_name)
-    assert len(volume1.replicas) == 2
-
-    wait_for_rebuild_complete(client, volume1_name)
-
```
PhanLe1010, line 621: We can miss the rebuilding event if the pod takes a bit longer to come up. Moreover, making sure that there is exactly a local replica in the below check is enough. Therefore, skip this check.

manager/integration/tests/test_scheduling.py
```diff
@@ -611,18 +611,12 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
...
     volume1 = wait_for_volume_replica_count(client, volume1_name, 1)
     volume1 = client.by_id_volume(volume1_name)
     assert len(volume1.replicas) == 1
     assert volume1.replicas[0]['hostId'] == volume1_attached_node
+    delete_and_wait_pod(core_api, pod1_name)
+    wait_for_volume_detached(client, volume1_name)
```
PhanLe1010, line 619: Make sure to clean up the pod after we done with it. Otherwise, PVC cannot be removed at the in the finalized at the end of the test.

manager/integration/tests/test_scheduling.py
```diff
@@ -695,6 +689,8 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
     volume2_rep1 = volume2_healthy_replicas[0]
     volume2_rep2 = volume2_healthy_replicas[1]
     assert volume2_rep1["hostId"] != volume2_rep2["hostId"]
+    delete_and_wait_pod(core_api, pod2_name)
+    wait_for_volume_detached(client, volume2_name)
```
PhanLe1010, line 693: Make sure to clean up the pod after we done with it. Otherwise, PVC cannot be removed at the in the finalized at the end of the test.

manager/integration/tests/test_scheduling.py
```diff
@@ -765,20 +761,17 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
...
         if replica["hostId"] == node1.name:
             err_replica = replica
             break
 
     assert err_replica is not None
-
     wait_for_replica_failed(client, volume3_name, err_replica["name"])
 
+    delete_and_wait_pod(core_api, pod3_name)
+    wait_for_volume_detached(client, volume3_name)
```
PhanLe1010, line 774: Make sure to clean up the pod after we done with it. Otherwise, PVC cannot be removed at the in the finalized at the end of the test.

manager/integration/tests/test_scheduling.py
```diff
@@ -794,6 +787,9 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
     assert volume3.replicas[0]["mode"] == "RW"
     assert volume3.replicas[0]["running"] is True
 
+    delete_and_wait_pod(core_api, pod3_name)
+    wait_for_volume_detached(client, volume3_name)
+
```
PhanLe1010, line 792: Make sure to clean up the pod after we done with it. Otherwise, PVC cannot be removed at the in the finalized at the end of the test.

manager/integration/tests/test_scheduling.py
```diff
@@ -765,20 +761,17 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
...
-
     err_replica = None
     for replica in volume3.replicas:
         if replica["hostId"] == node1.name:
             err_replica = replica
             break
 
     assert err_replica is not None
-
     wait_for_replica_failed(client, volume3_name, err_replica["name"])
```
PhanLe1010, line 771: There is an ordering problem here. When the volume is detached, the failed replica is already gone. So, move the `wait_for_replica_failed ` to before the detach command.

manager/integration/tests/test_scheduling.py
```diff
@@ -908,13 +904,15 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
 
     for replica in volume4.replicas:
         wait_for_replica_running(client, volume4_name, replica["name"])
+
+    volume4 = client.by_id_volume(volume4_name)
+    for replica in volume4.replicas:
```
PhanLe1010, line 909: Get the updated list of replica before checking its host. Otherwise, the `hostId` might be empty

manager/integration/tests/test_scheduling.py
PhanLe1010, line 771: There is an ordering problem here. When the volume is detached, the failed replica is already gone. So, move the wait_for_replica_failed to before the detach command.

manager/integration/tests/test_scheduling.py
```diff
@@ -901,6 +897,7 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
     volume4 = volume4.updateReplicaCount(replicaCount=2)
     wait_for_volume_healthy(client, volume4_name)
     volume4 = wait_for_volume_replica_count(client, volume4_name, 2)
+    volume4 = wait_for_volume_healthy(client, volume4_name)
```
PhanLe1010, line 900: Must wait for the volume to be healthy because if we don't, the volume might be in the state with 1 failed replica and 1 running replica.

manager/integration/tests/test_scheduling.py
```diff
@@ -899,8 +895,10 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
 
     volume4 = volume4.updateDataLocality(dataLocality="disabled")
     volume4 = volume4.updateReplicaCount(replicaCount=2)
-    wait_for_volume_healthy(client, volume4_name)
+    wait_for_rebuild_start(client, volume1_name)
+    wait_for_rebuild_complete(client, volume1_name)
     volume4 = wait_for_volume_replica_count(client, volume4_name, 2)
+    volume4 = wait_for_volume_healthy(client, volume4_name)
```
PhanLe1010, line 901: Must wait for the volume to rebuild, and then healthy because if we don't, the volume might be in the state with 1 failed replica and 1 running replica.

manager/integration/tests/test_scheduling.py
```diff
@@ -899,8 +895,10 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
 
     volume4 = volume4.updateDataLocality(dataLocality="disabled")
     volume4 = volume4.updateReplicaCount(replicaCount=2)
-    wait_for_volume_healthy(client, volume4_name)
+    wait_for_rebuild_start(client, volume4_name)
+    wait_for_rebuild_complete(client, volume4_name)
     volume4 = wait_for_volume_replica_count(client, volume4_name, 2)
+    volume4 = wait_for_volume_healthy(client, volume4_name)
```
PhanLe1010, line 901: Must wait for the volume to rebuild, and then healthy because if we don't, the volume might be in the state with 1 failed replica and 1 running replica.

manager/integration/tests/test_scheduling.py
```diff
@@ -899,8 +895,18 @@ def test_data_locality_basic(client, core_api, volume_name, pod, settings_reset)
...
+    for _ in range(RETRY_COUNTS):
+        volume4 = client.by_id_volume(volume4_name)
+        running_replica_count = 0
+        for r in volume4.replicas:
+            if r.failedAt == "" and r.running is True:
+                running_replica_count += 1
+        if running_replica_count == 2:
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert running_replica_count == 2
```
PhanLe1010, line 909: Must explicitly wait for the volume to have 2 running replicas because if we don't, the volume might be in the state with 1 failed replica and 1 running replica.

shuo-wu (changes requested):

manager/integration/tests/test_scheduling.py
PhanLe1010: Thank you! Fixed it

manager/integration/tests/test_scheduling.py
shuo-wu: It's better not to add the check since the failed replica will be reused automatically. In other words, the replica becoming error is a transient state. Instead, you can directly check if this replica is cleaned up after the volume detachment.

shuo-wu (changes requested):
1. It's better to follow the schema of `test_node_config_annotation_invalid` to split the test to several sub-cases by adding comments. Otherwise, it's hard to understand the whole test since there are more than 40 steps.
2. It seems that the implementation for step11~14 is missed.
3. Why is there a scheduling failed replica after the replica count update in line 678? And it's a little bit similar to the test in case 4. Can you add more comments for that?
```
    volume2.updateReplicaCount(replicaCount=2)

    # 2 Healthy replicas and 1 replica failed to schedule
    volume2 = wait_for_volume_replica_count(client, volume2_name, 3)
    volume2 = client.by_id_volume(volume2_name)
```
4. Actually there is only 800MB rather than 10GB data written to volume(3) in step 29.
```
29. Write 10GB data to volume(3)
```

PhanLe1010:
> 1. It's better to follow the schema of test_node_config_annotation_invalid to split the test to several sub-cases by adding comments. Otherwise, it's hard to understand the whole test since there are more than 40 steps.

Thank you, I fixed it.

> It seems that the implementation for step11~14 is missed.

It is not necessary. I deleted it.

> Why is there a scheduling failed replica after the replica count update in line 678? And it's a little bit similar to the test in case 4. Can you add more comments for that?

I added the comment to explain it

> Actually there is only 800MB rather than 10GB data written to volume(3) in step 29.

Thank you. Fixed the test skeleton 

> It's better not to add the check since the failed replica will be reused automatically. In other words, the replica becoming error is a transient state. Instead, you can directly check if this replica is cleaned up after the volume detachment.

Thank you. Fixed it

shuo-wu (approved):

PR #503, opened 2020-12-17
Add longhorn-share-manager to security scan

yasker:
Signed-off-by: Sheng Yang <sheng.yang@rancher.com>


yasker:
Close this one. Share manager is currently a Fedora based image, which isn't supported by trivy.

We will enable it once we switch the image base for share manager.

PR #513, opened 2021-01-07
Fix failing Nightly tests

meldafrawi:
- `pytest` fixtures are called in the order provided, and teardown is happens in reverse of that specific order, 
Calling set_random_backpustore first, will ensure backupstore_cleanup happens after every volume gets deleted at test teardown.

- some `statefulset` tests was lacking `set_random_backupstore` fixture.


yasker (approved):

khushboo-rancher (approved):
LGTM

PR #516, opened 2021-01-13
use bash instead of ash

meldafrawi:

### DIFF
--- a/test_framework/scripts/longhorn-setup.sh
+++ b/test_framework/scripts/longhorn-setup.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env ash
+#!/usr/bin/env bash
 
 set  -x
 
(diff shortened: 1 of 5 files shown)


yasker:
Is the ash causing the deployment problem? not clear why from the commit message.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -33,7 +33,7 @@ check_longhorn_status() {
 
 install_csi_snapshotter_crds(){                                                 
     git clone --single-branch --branch "master" "https://github.com/kubernetes-csi/external-snapshotter.git" /tmp/k8s-csi-external-snapshotter
-    kubectl apply -f /tmp/k8s-csi-external-snapshotter/client/config/crd \
```
meldafrawi, line 36: @yasker extra spaces here caused the issue, and i decided to switch to bash since we had some other syntax compatibility issues before between `bash` and `ash`

yasker (approved):

PR #517, opened 2021-01-14
Fix timing issue of the recurring job

khushboo-rancher:
Adjusted some sleep time in the test 'test_recurring_job` according to the time of recurring jobs getting triggered.
https://github.com/longhorn/longhorn/issues/2158


meldafrawi (approved):
LGTM

PR #521, opened 2021-01-22
Add ipv6 disable test and add share manager to the image scan list

yasker:
https://github.com/longhorn/longhorn/issues/2111

https://github.com/longhorn/longhorn/issues/2136

### DIFF
--- a/docs/content/manual/release-specific/v1.1.1/disable_ipv6.md
+++ b/docs/content/manual/release-specific/v1.1.1/disable_ipv6.md
@@ -1,0 +1,27 @@
+---
+title: Test Disable IPv6
+---
+
+## Related issue
+
+https://github.com/longhorn/longhorn/issues/2136
+https://github.com/longhorn/longhorn/issues/2197
+
+Longhorn v1.1.1 should work with IPv6 disabled.
+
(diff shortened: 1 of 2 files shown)


docs/content/manual/release-specific/v1.1.1/disable_ipv6.md
```diff
@@ -0,0 +1,19 @@
...
+Longhorn v1.1.1 should work with IPv6 disabled.
+
+## Scenario
+1. Install Kubernetes
+2. Disable IPv6 on all the nodes using the following commands
+```
+sysctl -w net.ipv6.conf.all.disable_ipv6=1
+sysctl -w net.ipv6.conf.default.disable_ipv6=1
+```
+3. Deploy Longhorn and test basic use cases.
```
khushboo-rancher, line 19: Should we add test steps of enabling IPv6 again and verify Longhorn works fine still? And also should we call out `rwx` use cases to test?

yasker: Yeah, good idea.

khushboo-rancher (approved):

PR #522, opened 2021-01-23
add test skeleton for feature `Automatically upgrade volumes' engine to the default engine image after upgrading Longhorn`

PhanLe1010:
longhorn/longhorn#2152

### DIFF
--- a/manager/integration/tests/test_engine_upgrade.py
+++ b/manager/integration/tests/test_engine_upgrade.py
@@ -691,3 +691,76 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
     volume_file_md5sum2 = get_pod_data_md5sum(
         core_api, pod_name, data_path2)
     assert volume_file_md5sum2 == original_md5sum2
+
+
+@pytest.mark.skip(reason="TODO") # NOQA
+def test_auto_upgrade_engine_to_default_version():
+    """
+    Steps:
+
+    Preparation:
+    1. set up a backup store
+    2. Deploy a compatible new engine image
+
+    Case 1: Concurrent engine upgrade
+    1. Create 10 volumes each of 1Gb.
+    2. Attach 5 volumes vol-0 to vol-4. Write data to it
+    3. Upgrade all volumes to the new engine image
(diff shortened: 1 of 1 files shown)


shuo-wu (changes requested):
We can add more tests for the setting:
1. The engine upgrade API is disabled when the setting is enabled
2. The upgrade shouldn't start when the volume is degraded.
3. Offline upgrade (you can create more volumes but do not attach them in case1)

PhanLe1010:
> The engine upgrade API is disabled when the setting is enabled

We test it here https://github.com/longhorn/longhorn-tests/blob/b0d73df12fce9633112aa26d8c1b0c1b158158e4/manager/integration/tests/test_engine_upgrade.py#L724

> The upgrade shouldn't start when the volume is degraded.

I think it is ok to upgrade the degraded volumes

> Offline upgrade (you can create more volumes but do not attach them in case1)

We already has 5 detached volumes in case 1, do we need to create more?

shuo-wu:
Oh sorry for missing some steps. No, 5 detached volumes are enough.

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -691,3 +691,76 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+    5. Set concurrent-automatic-engine-upgrade-per-node-limit setting to 3
+    6. In a retry loop, verify that the number of volumes who
+       is upgrading engine is always smaller or equal to 3
+    7. Wait until the upgrades are completed (volumes' engine image changed,
+       replica mode change to RW for attached volumes, reference count of the
+       new engine image changed, all engine and replicas' engine image changed,
+       etc ...)
+    8. verify the volumes' data
+
+    Case 2: Dr volume
```
shuo-wu, line 721: `DR volume`

shuo-wu (dismissed):
In general LGTM

manager/integration/tests/test_engine_upgrade.py
PhanLe1010: Thank you. Fixed

shuo-wu (approved):

PR #524, opened 2021-01-27
Add tests for Backing Image feature

shuo-wu:
longhorn/longhorn#2006


shuo-wu:
@khushboo-rancher @meldafrawi Can you review this PR?

shuo-wu:
@meldafrawi @khushboo-rancher Can you review this PR so that the backing image won't fail the nightly test

manager/integration/tests/test_backing_image.py
```diff
@@ -0,0 +1,263 @@
...
+    random_disk_id = ""
+    backing_image = client.by_id_backing_image(bi_name)
+    assert backing_image.imageURL == bi_url
+    assert not backing_image.deletionTimestamp
+    assert len(backing_image.diskStateMap) == 3
+    for disk_id, state in iter(backing_image.diskStateMap.items()):
+        assert state == "downloaded"
+        random_disk_id = disk_id
+    assert random_disk_id != ''
+
```
meldafrawi, line 73: I think test is missing step 4 here ?

meldafrawi:
LGTM in general, except for the missing test step

docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
```diff
@@ -0,0 +1,22 @@
...
+
+1. Update the settings:
+   1. Disable Node Soft Anti-affinity.
+   2. Set Replica Replenishment Wait Interval to a relatively long value.
+2. Create a backing image with one of the following URLs:
+```
+   https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.qcow2
+   https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.raw
+```
+3. Create and 2 volumes with the backing image and attach them on different nodes.
```
khushboo-rancher, line 13: `and` typo

khushboo-rancher, line 13: Do we need to check the related disk is added successfully with correct status?

docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
```diff
@@ -0,0 +1,22 @@
...
+4. Verify the backing image content then write random data in the volumes.
+5. Power off a node containing one volume. Verify that
+   - the related disk download state in the backing image will become `failed` once the download pod is removed by Kubernetes.
+   - the volume on the running node still works fine but is state `Degraded`, and the content is correct in the volume.
+   - the volume on the down node become `Unknown`.
+6. Power on the node. Verify
+   - the failed replica of the `Degraded` volume can be reused.
+   - the volume on the down node will be recovered automatically.
+   - the backing image will be recovered automatically.
```
khushboo-rancher, line 22: We should also check the random data written at step 4 by reducing the replica count to one and deleting the replicas from the always active nodes.

docs/content/manual/pre-release/upgrade/backing-image-during-upgrade.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Test Backing Image during Longhorn upgrade
+---
+
+1. Set `Concurrent Automatic Engine Upgrade Per Node Limit` to a positive value to enable volume engine auto upgrade.
```
khushboo-rancher, line 5: Should we add the comment that `Concurrent Automatic Engine Upgrade Per Node Limit` is available from Longhorn v1.1.1 onwards, user upgrading from previous version should set the value while upgrading the Longhorn system?

manager/integration/tests/test_backing_image.py
khushboo-rancher: Similarly the step 7 is missing, I don't see the disk cleanup step.

khushboo-rancher:
In general LGTM, 
One doubt - Should we include the validation of basic functionalities of the volumes created with the image like backup, restore etc in the manual tests.

docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
shuo-wu: Yes. I will update this test

docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
shuo-wu: Good idea. I will add this step to the test

manager/integration/tests/test_backing_image.py
shuo-wu: Step 4 is this part, which is not missed:
```
    with pytest.raises(Exception):
        backing_image.backingImageCleanup(diks=[random_disk_id])
```

But in fact, step 5 will be skipped if when the operation in step4 is failed. I will update this part.

shuo-wu: Step 7 was the last line `test_backing_image_basic_operation(client, volume_name)`, which may be not straightforward. I will add a separate backing image deletion call at the end of **this** function.

shuo-wu:
> In general LGTM, few minor comments.
> One doubt - Should we include the validation of basic functionalities of the volumes created with the image like backup, restore etc in the manual tests.

The test you mentioned should be tested by case `test_backup_with_backing_image`. I don't think we need to do it again manually.

khushboo-rancher (approved):

khushboo-rancher:
LGTM

meldafrawi:
LGTM

meldafrawi (approved):

PR #530, opened 2021-02-08
Add test case for test_recurring_jobs_when_volume_detached_unexpectedly

meldafrawi:
https://github.com/longhorn/longhorn/issues/1922

### DIFF
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -17,10 +17,21 @@ from common import update_statefulset_manifests, get_self_host_id, \
 from common import write_volume_random_data
 from common import write_pod_volume_random_data
 from common import BASE_IMAGE_LABEL, KUBERNETES_STATUS_LABEL
-from common import SIZE, Mi, Gi
+from common import SIZE, Mi, Gi, pvc  # NOQA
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_recurring_job.py
```diff
@@ -636,4 +688,67 @@ def test_recurring_jobs_when_volume_detached_unexpectedly():  # NOQA
...
+    ]
+    vol.recurringUpdate(jobs=jobs)
+
+    wait_for_recurring_backup_to_start(client,
+                                       core_api,
+                                       vol_name,
+                                       expected_snapshot_count=1)
+
+    crash_engine_process_with_sigkill(client, core_api, vol_name)
+    time.sleep(30)
```
PhanLe1010, line 735: Is it correct that we have to wait for 30s so that Longhorn marks volume as faulty and the salvage feature kicks in?
If so, should we put a comment here to explain the reason we need to wait?

PhanLe1010 (commented):
In general, LGTM

PhanLe1010 (approved):
LGTM, except for the flake8 error

meldafrawi:
fixed, Thanks

PhanLe1010:
LGTM

PR #532, opened 2021-02-11
Update tests and add manual tests for the enhanced CPU reservation feature

shuo-wu:
longhorn/longhorn#2207

### DIFF
--- a/docs/content/manual/release-specific/v1.1.1/system-upgrade-with-deprecated-cpu-setting.md
+++ b/docs/content/manual/release-specific/v1.1.1/system-upgrade-with-deprecated-cpu-setting.md
@@ -1,0 +1,20 @@
+---
+title: Test system upgrade with the deprecated CPU setting
+---
+
+## Related issue
+https://github.com/longhorn/longhorn/issues/2207
+
+## Test step
+1. Deploy a cluster that each node has different CPUs.
+2. Launch Longhorn v1.1.0.
+3. Deploy some workloads using Longhorn volumes.
+4. Upgrade to the latest Longhorn version. Validate:
+    1. all workloads work fine and no instance manager pod crash during the upgrade.
+    2. The fields `node.Spec.EngineManagerCPURequest` and `node.Spec.ReplicaManagerCPURequest` of each node are the same as the setting `Guaranteed Engine CPU` value in the old version * 1000.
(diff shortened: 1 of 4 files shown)


shuo-wu:
@khushboo-rancher @meldafrawi @joshimoo Can you take a look?

docs/content/manual/release-specific/v1.1.1/system-upgrade-with-deprecated-cpu-setting.md
```diff
@@ -0,0 +1,21 @@
...
+---
+
+## Related issue
+https://github.com/longhorn/longhorn/issues/2207
+
+## Test step
+1. Deploy a cluster that each node has different CPUs.
+2. Launch Longhorn v1.1.0.
+3. Deploy some workloads using Longhorn volumes.
+4. Upgrade to the latest Longhorn version. Validate the workloads work fine and no instance manager pod crash during the upgrade.
```
khushboo-rancher, line 12: Will the existing value of `Guaranteed Engine Manager CPU` become the new value of `node.Spec.EngineManagerCPURequest` for all the nodes and `Guaranteed Replica Manager CPU` set to a default value? What would be the default value?

khushboo-rancher (dismissed):
LGTM

docs/content/manual/release-specific/v1.1.1/system-upgrade-with-deprecated-cpu-setting.md
shuo-wu: `node.Spec.EngineManagerCPURequest` is a field in node spec, which is different from the setting `Guaranteed Engine Manager CPU`:
1. The setting will be ignored by the specific node only when its spec field `EngineManagerCPURequest` is not 0.
2. The setting is a percentage value. While the node field holds a milli value.

shuo-wu: For more details, you can check the doc PR or the LEP PR.

shuo-wu:
@joshimoo @meldafrawi Do you need to take a look? Since the related manager PR is already merged and there is an old setting deprecated, we need to checkout this PR ASAP in case of the test failures.

meldafrawi (approved):
LGTM

PR #534, opened 2021-02-16
Anupama2501 longhorn

anupama2501:
Added corner test cases for the upgrade engine image verification

### DIFF
--- a/docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
+++ b/docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
@@ -1,6 +1,10 @@
 ---
 title: Automatically Upgrading Longhorn Engine Test
 ---
+<h4>Longhorn version >= 1.1.1 </h4>
+
+[Reference ticket 2152](https://github.com/longhorn/longhorn/issues/2152)
+
 
 ### Test basic upgrade
 1. Install old Longhorn version. E.g., <= `v1.0.2`
@@ -13,12 +17,20 @@ title: Automatically Upgrading Longhorn Engine Test
 1. Create a StatefulSet of scale 10 using 10 Longhorn volume. Set node selector so that all pods land on the same node.
 2. Upgrade Longhorn to use a newer default engine image
 3. In Longhorn UI and Longhorn manager logs, Verify that Longhorn doesn't upgrade all volumes at the same time. Only 3 at a time.
-
+4. In Longhorn UI, while the concurrent value is set to limit 1, change the value to higher number and verify if the upgrade is happening as per the new value.
+5. Create stateful set pods with volume claim template, set the scale to 10 and in the configure options, select RWO and RWX. Verify the volumes created have RWX set as RWX takes precedence.
+6. While the auto upgrade is set, try to manually upgrade the image engine to the latest version. The manual upgrade should work and the image should be upgraded to the default version.
+7. Take backup for a volume when upgrade hasn't started.
+8. Create 10 volumes and set concurrent upgrade limit to 2. When the upgrade starts, delete all the volumes. [Volumes are deleted and Longhorn UI is not corrupted.
+9. Deploy a workload with rwx scale 10 with persistent volume claim template and verify the volume created can be used for all of 10 pods.
+10. While upgrade has started, detach the nodes from the volume and verify the detached volume is upgraded.
+  
 ### Test degraded volume
-1. Verify that Longhorn doesn't upgrade engine image for degraded volume
+1. Verify that Longhorn doesn't upgrade engine image for degraded volume.
+2. While the upgrade is happening, make a volume degraded and verify if the upgrade still happen.
 
 ### Test DR volume
-1. Verify that Longhorn doesn't upgrade engine image for DR volume 
+1. Verify that Longhorn doesn't upgrade engine image for DR volume.
 
 ### Test expanding volume
-1. Verify that Longhorn doesn't upgrade engine image for volume which is expanding
+1. Verify that Longhorn doesn't upgrade engine image for volume which is expanding.
--- a/docs/content/manual/release-specific/v1.1.1/disable_ipv6.md
+++ b/docs/content/manual/release-specific/v1.1.1/disable_ipv6.md
@@ -27,6 +27,5 @@ use the command cat /proc/cmdline to verify "ipv6.disable=1" is reflected in the
 By deleting the "ipv6.disable=1" from the GRUB_CMDLINE_LINUX_DEFAULT accessing grub file from
 /etc/default
 Perform a reboot and once the node comes active 
-use the command cat /proc/cmdline to verify "ipv6.disable=1" is reflected in the values
-```
+use the command cat /proc/cmdline to verify "ipv6.disable=1" is not reflected in the values
 5. Make sure the basic use cases still works.


docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
```diff
@@ -13,9 +13,17 @@ title: Automatically Upgrading Longhorn Engine Test
 1. Create a StatefulSet of scale 10 using 10 Longhorn volume. Set node selector so that all pods land on the same node.
 2. Upgrade Longhorn to use a newer default engine image
 3. In Longhorn UI and Longhorn manager logs, Verify that Longhorn doesn't upgrade all volumes at the same time. Only 3 at a time.
-
+4. In longhorn UI, while the concurrent value is set to limit 1, change the value to higher number and verify if the upgrade is happening as per the new value.
+5. Create stateful set pods with volume claim template, set the scale to 10 and in the configure options, select RWO and RWX. Verify the volumes created have RWO set
```
yasker, line 17: RWO and RWX shouldn't be set at the same time? Not sure what's the logic on the Rancher side.

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
```diff
@@ -13,9 +13,17 @@ title: Automatically Upgrading Longhorn Engine Test
 1. Create a StatefulSet of scale 10 using 10 Longhorn volume. Set node selector so that all pods land on the same node.
 2. Upgrade Longhorn to use a newer default engine image
 3. In Longhorn UI and Longhorn manager logs, Verify that Longhorn doesn't upgrade all volumes at the same time. Only 3 at a time.
-
+4. In longhorn UI, while the concurrent value is set to limit 1, change the value to higher number and verify if the upgrade is happening as per the new value.
```
yasker, line 16: `Longhorn`

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
```diff
@@ -13,9 +13,17 @@ title: Automatically Upgrading Longhorn Engine Test
...
+5. Create stateful set pods with volume claim template, set the scale to 10 and in the configure options, select RWO and RWX. Verify the volumes created have RWO set
+6. While the auto upgrade is set, try to manually upgarde the image engine to the latest version.
```
yasker, line 18: `upgrade`

yasker: What would be the expectation?

yasker (changes requested):
IPv6 part LGTM.

Some of the engine upgrade cases miss the expected results.

@khushboo-rancher @PhanLe1010 Can you also take a look?

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
PhanLe1010: I think that RWX and RWO can be set at the same time. Because [accessModes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) is an array. In this case, Longhorn should create a RWX volume as see in [the code](https://github.com/longhorn/longhorn-manager/blob/c2e1728c2dbfb05f8060bc182c9ad5c23d4171ee/csi/controller_server.go#L129). Can you verify it @anupama2501 

But yeah, they shouldn't be set at the same time.

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
```diff
@@ -13,9 +13,17 @@ title: Automatically Upgrading Longhorn Engine Test
...
+7. Take backup for a volume when upgarde hasn't started
+8. Create 10 volmes and set concurrent upgrade limit to 2. When the upgrade starts, delete all the volumes.
```
PhanLe1010, line 20: What is the expectation?
Is it correct that volumes are deleted and Longhorn system is not corrupted?

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
```diff
@@ -13,9 +13,17 @@ title: Automatically Upgrading Longhorn Engine Test
...
+9. Deploy a workload with rwx scale 10 with persistent volume claim template and verify the volume created can be used for all of 10 pods.
+10.While upgrade has started, detach the nodes from the volume and verfiy the detached volume is not upgraded.
```
PhanLe1010, line 22: I thought that detached volume should be upgraded eventually?

docs/content/manual/release-specific/v1.1.1/disable_ipv6.md
```diff
@@ -10,18 +10,23 @@ https://github.com/longhorn/longhorn/issues/2197
...
     1. Accessing the UI for CRUD RWO/RWX volumes
     2. Use kubectl to create workloads using RWO/RWX volumes.
 4. Re-enable ipv6 on all the nodes using:
 ```
-sysctl -w net.ipv6.conf.all.disable_ipv6=0
-sysctl -w net.ipv6.conf.default.disable_ipv6=0
+By deleting the "ipv6.disable=1" from the GRUB_CMDLINE_LINUX_DEFAULT accessing grub file from
+/etc/default
+Perform a reboot and once the node comes active 
+use the command cat /proc/cmdline to verify "ipv6.disable=1" is reflected in the values
```
PhanLe1010, line 30: Do you mean `ipv6.disable=0` ?

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
anupama2501: When concurrent value is set, and if we try to upgrade the engine image, I saw the engine image upgrade happening only from the automatic upgrade and not while user tries to upgrade

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
anupama2501: Upgrade has started - wanted to say that when the upgrade has started, select a node that wasn't already upgraded and try to detach the volume. Therefore I mentioned the detached volume shouldn't be upgraded. I verified this scenario and saw that the volume is upgraded to the latest version as you mentioned. So, is this expected @PhanLe1010 ?

docs/content/manual/release-specific/v1.1.1/disable_ipv6.md
anupama2501: Original values in the GRUB_CMDLINE_LINUX_DEFAULT:
GRUB_CMDLINE_LINUX=""
GRUB_CMDLINE_LINUX_DEFAULT="quiet splash"
As the original value was empty and after I added this :

GRUB_CMDLINE_LINUX_DEFAULT="ipv6.disable=1"
GRUB_CMDLINE_LINUX="ipv6.disable=1"

I mentioned removed the "ipv6.disable=1" from GRUB_CMDLINE_LINUX and set to the original, GRUB_CMDLINE_LINUX="".

We can also make it  GRUB_CMDLINE_LINUX="ipv6.disable=0" to disable it. Let me know if you want me to change it ipv6.disable=0

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
anupama2501: Sorry, that was a mistake. Yes @PhanLe1010  RWX takes precedence.

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
anupama2501: Yes, the volumes are deleted and the longhorn system is not corrupted when I tried.

khushboo-rancher: `Volumes` typo

khushboo-rancher:
@anupama2501 Looks good in general, there are few points as below.
1. Can you please add reference of the enhancement id https://github.com/longhorn/longhorn/issues/2152 in the test for automatic update?
2. Can you also mention the Longhorn version (v1.1.1 onwards) which supports this because this is being added to the Pre-release testing?
3. Do we want to add the validation of concurrent value can't be set to 0.
4. What happens when the backup is in progress/recurring job or any other operation is in progress and the automatic engine upgrade is in queue for that volume, does the upgrade wait for the operations to finish? Do we want to specifically call out this scenario.

anupama2501:
@khushboo-rancher :
Can you please add reference of the enhancement id longhorn/longhorn#2152 in the test for automatic update?
Can you also mention the Longhorn version (v1.1.1 onwards) which supports this because this is being added to the Pre-release testing?
Sure, will update the above two mentioned values in the test cases document.

**Do we want to add the validation of concurrent value can't be set to 0.** 
As the concurrent value set to 0 doesn't upgrade the image versions automatically, haven't added this in the test cases.

**What happens when the backup is in progress/recurring job or any other operation is in progress and the automatic engine upgrade is in queue for that volume, does the upgrade wait for the operations to finish? Do we want to specifically call out this scenario.**
The concurrent option is not present in the previous version[v1.1.0] but only the latest version[v1.1.1]. So, when the user upgrades the longhorn image from say v1.1.0 to master, the concurrent value is set to 0[no upgrade happens when the value is set to 0]. if users perform backup or any other process before updating the concurrent value, the backup is executed first. If the user updates the concurrent value from 1 to 20[maximum value], and then proceeds with the backup/delete/update, the image is upgraded first. 
If the user tries to perform backup while the upgrade has started and if the volume is not upgraded already, user performed actions backup/delete takes place.  I've added a test case for these in point 10 and 7
Hope this answers.

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
joshimoo: A PVC's access mode is a request for storage (PV) that at least supports the required modes.
Here is some more explanation on this: https://stackoverflow.com/questions/60248790/why-can-you-set-multiple-accessmodes-on-a-persistent-volume

The reason we set longhorns access mode to RWX if the claim contains RWX, is that we need to be able to guarantee that the volume can be accessed in RWX mode if one of the possible access modes is RWX.

khushboo-rancher (dismissed):
LGTM

yasker:
@PhanLe1010 or @joshimoo can you take a final look at this issue? also @anupama2501 we need to rebase the commits to clear up the git history.

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
```diff
@@ -13,9 +17,17 @@ title: Automatically Upgrading Longhorn Engine Test
 1. Create a StatefulSet of scale 10 using 10 Longhorn volume. Set node selector so that all pods land on the same node.
 2. Upgrade Longhorn to use a newer default engine image
 3. In Longhorn UI and Longhorn manager logs, Verify that Longhorn doesn't upgrade all volumes at the same time. Only 3 at a time.
-
+4. In Longhorn UI, while the concurrent value is set to limit 1, change the value to higher number and verify if the upgrade is happening as per the new value.
+5. Create stateful set pods with volume claim template, set the scale to 10 and in the configure options, select RWO and RWX. Verify the volumes created have RWX set as RWX takes precedence
+6. While the auto upgrade is set, try to manually upgrade the image engine to the latest version.
```
PhanLe1010, line 22: We still missing the expectation in this step.
Is it correct that manually upgrade engine image to the default engine image version should be ok?

PhanLe1010 (commented):
LGTM except for a comment

docs/content/manual/pre-release/upgrade/auto-upgrade-engine.md
```diff
@@ -13,9 +17,17 @@ title: Automatically Upgrading Longhorn Engine Test
 1. Create a StatefulSet of scale 10 using 10 Longhorn volume. Set node selector so that all pods land on the same node.
 2. Upgrade Longhorn to use a newer default engine image
 3. In Longhorn UI and Longhorn manager logs, Verify that Longhorn doesn't upgrade all volumes at the same time. Only 3 at a time.
-
+4. In Longhorn UI, while the concurrent value is set to limit 1, change the value to higher number and verify if the upgrade is happening as per the new value.
+5. Create stateful set pods with volume claim template, set the scale to 10 and in the configure options, select RWO and RWX. Verify the volumes created have RWX set as RWX takes precedence
```
c3y1huang, line 21:
```suggestion
5. Create stateful set pods with volume claim template, set the scale to 10 and in the configure options, select RWO and RWX. Verify the volumes created have RWX set as RWX takes precedence.
```

PhanLe1010 (approved):
LGTM

yasker (approved):

PR #541, opened 2021-02-26
Add test skeleton for the case when engine image DaemonSet is not fully deployed

PhanLe1010:
longhorn/longhorn#2081

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -2427,3 +2427,109 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
     del update_disks["extra-disk"]
     node.diskUpdate(disks=update_disks)
     common.wait_for_disk_update(client, node.name, 1)
+
+
+@pytest.mark.skip(reason="TODO") # NOQA
+def test_engine_image_missing_on_some_nodes():
+    """
+    This e2e test follows the manual test steps at:
+    https://github.com/longhorn/longhorn/issues/2081#issuecomment-783747541
+
+    Steps:
+
+    Preparation:
+    1. Set up a backup store
+    2. Let's name the 3 nodes: node-1, node-2, node-3
+
+    Case 1: Test volume operations when engine image DaemonSet is miss
+    scheduled
+    1. Create a volume, vol-1, of 3 replicas
+    2. Create another volume, vol-2, of 3 replicas
+    3. Taint node-1 with the taint: key=value:NoSchedule
+    4. Verify that we can attach, take snapshot, take a backup,
+       detach, then expand vol-1
+
+    Case 2: Test volume operations when engine image DaemonSet is not fully
+    deployed
+    1. Continue from case #1
+    2. Attach vol-1 to node-1. Change the number of replicas of vol-1
+       to 2. Delete the replica on node-1
+    3. Delete the pod on node-1 of the engine image DaemonSet.
+       Or delete the engine image DaemonSet and wait for Longhorn
+       to automatically recreates it.
+    4. Wait for the engine image CR state become deploying
+    5. Verify that functions (snapshot, backup, detach) are working ok
+       for vol-1
+    6. Detach vol-1
+    7. Attach vol-1 to node-1. Verify that Longhorn cannot attach vol-1 to
+       node-1 since there is no engine image on node-1. The attach API call
+       returns error
+    8. Verify that we can attach to another node, take snapshot, take a backup,
+       detach, then expand for vol-1
+    9. Verify that vol-2 cannot be attached to any nodes because one of
+       its replicas is sitting on the node-1 which doesn't have the
+       engine image. The attach API call returns error
+
+    Case 3: Test engine upgrade when engine image DaemonSet is not fully
+    deployed
+    1. Continue from case #2
+    2. Deploy a new engine image, new-ei
+    3. Detach vol-1
+    4. Verify that you can upgrade vol-1 to new-ei
+    5. Attach vol-1 to node-2
+    6. Verify that you can live upgrade vol-1 to back to default engine image
+    7. Try to upgrade vol-2 to new-ei
+    8. Verify that the engineUpgrade API call returns error
+
+    Case 4: Test replicas scheduling when engine image DaemonSet is not fully
+    deployed
+    1. Continue from case #3
+    2. Create a new volume, vol-3, with 2 replicas
+    3. disable the scheduling for node-2
+    4. Verify that there is one replica fail to be scheduled
+    5. enable the scheduling for node-2
+    6. Verify that replicas are scheduled onto node-2 and node-3
+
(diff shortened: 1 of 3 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -2426,3 +2426,94 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
...
+    Preparation:
+    1. Set up a backup store
+    2. Let's name the 3 nodes: node-1, node-2, node-3
+
+    Case 1: Test volume operations when engine image DaemonSet is miss
+    scheduled
+    1. Create a volume, vol-1, of 3 replicas
+    2. Create another volume, vol-2, of 3 replicas
+    3. Taint node-1 with the taint: key=value:NoSchedule
+    4. Verify that all functions (attach, snapshot, backup,
```
shuo-wu, line 2448: snapshot and backup for detached volumes? Do you mean recurring jobs?

manager/integration/tests/test_ha.py
```diff
@@ -2426,3 +2426,94 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
...
+       expand, detach) are working ok for vol-1
+
+    Case 2: Test volume operations when engine image DaemonSet is not fully
```
shuo-wu, line 2451: It seems that all cases are in the same context... I mean, does the taint in case1-steo3 still exist when case2 starts? 
If NO, the engine image CR state `deploying` is a transient state, and we cannot validate anything in that period. 
If YES, it's better to emphasize it in the beginning.

manager/integration/tests/test_ha.py
```diff
@@ -2426,3 +2426,94 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
...
+       to automatically recreates it.
+    3. Wait for the engine image CR state become deploying
+    4. Verify that functions (snapshot, backup, detach) are working ok
+       for vol-1
+    5. Detach vol-1
+    6. Attach vol-1 to node-1. Verify that Longhorn cannot attach vol-1 to
+       node-1 since there is no engine image on node-1
+    7. Verify that all functions (attach to other nodes, snapshot, backup,
+       expand, detach) are working ok for vol-1
+    8. Verify that vol-2 cannot be attached to any nodes because one of
```
shuo-wu, line 2466: Will the volume get stuck in attaching? Or will the `attach` call directly error out?

manager/integration/tests/test_ha.py
```diff
@@ -2426,3 +2426,94 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
...
+    3. Verify that you can upgrade vol-1 to new-ei
+    4. Attach vol-1 to node-2
+    5. Verify that you can live upgrade vol-1 to back to default engine image
+
+    Case 4: Test replicas scheduling when engine image DaemonSet is not fully
+    deployed
+    1. Create a new volume, vol-3, with 2 replicas
+    2. Verify that replicas of vol-3 are on node-2 and node-3
+    3. Check that all functions (attach to node other than node-1, snapshot,
+       backup, expand, detach) are working ok for vol-3
```
shuo-wu, line 2483: This cannot prove that the replica is never scheduled to a node without the engine image. You need to disable the scheduling of node2 or node3, create vol-3, then see if there is one replica scheduling faliure.

shuo-wu: And similar to this logic, you can add one more case for the failed replica reusage senario.

manager/integration/tests/test_ha.py
```diff
@@ -2426,3 +2426,94 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
...
+       is attached to either node-2 or node-3.
+       Let's say it is attached to node-x
+    3. Taint node-x with the taint `key=value:NoSchedule`
+    4. Delete the pod of engine image DeamonSet on node-x. Now, the engine
+       image is missing on node-1 and node-x
+    5. Verify that the owner ID of vol-dr temporarily moves to node-y which
+       is different than node-1 and node-x.
+    6. Verify that vol-dr is auto-attached node-y.
+    7. Restore a volume from backupstore with name vol-rs and replica count
+       is 1
```
shuo-wu, line 2509: Do we expose the volume `ownerID` in API layer? 
Or you mean the DR volume will be detached automatically then reattached to node-y? IIRC, there is no related logic in the implementation PR...

Anyway, the expected behavior of this DR volume is unclear here.

shuo-wu (changes requested):

manager/integration/tests/test_ha.py
PhanLe1010: Sorry for not being clear. It shoud be `Verify that we can attach, take snapshot, take a backup,
       detach, then expand vol-1` 
Fixed

manager/integration/tests/test_ha.py
PhanLe1010: Thank you. Fixed it to be clear that the subsequent case is continued from the previous case

manager/integration/tests/test_ha.py
PhanLe1010: Thank you. Fixed it to specify that we should get the API error

manager/integration/tests/test_ha.py
PhanLe1010: Thank you. Fixed

manager/integration/tests/test_ha.py
PhanLe1010: Yeah, we don't have ownerID in the API layer. Fixed it

Yes, the DR volume will be detached automatically then reattached to node-y since the we automatically attach standby volumes to the ownerID node: https://github.com/longhorn/longhorn-manager/blob/71adb0123b077982a7fe306ec7f986963debf238/controller/volume_controller.go#L1811

khushboo-rancher (approved):
LGTM

manager/integration/tests/test_ha.py
```diff
@@ -2427,3 +2427,109 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
...
+    8. Verify that replica is on node-y and the volume successfully restored.
+    9. Wait for vol-rs to finish restoring
+    10. Expand vol-rs.
+    11. Verify that the expansion is ok
+
+    Case 7: Test replicas scheduling when engine image DaemonSet is not fully
+    deployed
+    1. Continue from case #6
+    2. Set `Replica Replenishment Wait Interval` setting to 600
+    3. Crash the replica of vol-3 on node-x. Wait for the replica to fail
```
shuo-wu, line 2527: One quick question, when the engine image pod is deleted in case6-step5, will the related replicas keep running?

PhanLe1010: Yeah. They will keep running since the binary is loaded into RAM.

shuo-wu (approved):

PR #542, opened 2021-02-27
Add manual test case: `Test Service Account Mounts On Host`

PhanLe1010:
longhorn/longhorn#1877


yasker (changes requested):
Can we mention this should be tested on both yaml installation and chart installation, as well as upgrade scenarios?

Also, we can put them in v1.1.1 pre-release test only.

PhanLe1010:
Thank you. Fixed

innobead (approved):
LGTM

Issue #49, opened 2018-06-25
Add provisioner test with flexvolume


yasker:
Currently we didn't test flexvolume provisioner.

yasker:
The test needs to cover:
1. Volume creation/deletion with provisioner.
2. Test of various parameters can be used with volume.
3. Validate the content of volume.

See https://github.com/rancher/longhorn-tests/blob/master/manager/integration/tests/test_driver.py for some examples how to test with Kubernetes.

PR #552, opened 2021-03-10
Fix backing image tests (w/upgrade disabled)

c3y1huang:
The schema actions could return empty dict even if schema not exist, should look for an attribute instead.

https://ci.longhorn.io/job/public/job/ubuntu/job/amd64/job/longhorn-tests-ubuntu-amd64/285/testReport/tests/test_backing_image/

https://github.com/longhorn/longhorn/issues/2318

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -366,14 +366,11 @@ def create_and_check_volume(client, volume_name,
     :param frontend: The frontend to use for the volume.
     :return: The volume instance created.
     """
-    if backing_image_feature_supported(client):
-        client.create_volume(name=volume_name, size=size,
-                             numberOfReplicas=num_of_replicas,
-                             backingImage=backing_image, frontend=frontend)
-    else:
-        client.create_volume(name=volume_name, size=size,
-                             numberOfReplicas=num_of_replicas,
-                             frontend=frontend)
+    if not backing_image_feature_supported(client):
+        backing_image = None
+    client.create_volume(name=volume_name, size=size,
+                         numberOfReplicas=num_of_replicas,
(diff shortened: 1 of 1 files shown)


c3y1huang:
Test regression [Build #49](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/49/testReport/tests/test_backing_image/)

shuo-wu:
Do we need to skip all backing image tests based on the function? 
https://github.com/longhorn/longhorn-tests/blob/888cb8fc2086a59c72497862cd989d3d2214816e/manager/integration/tests/conftest.py#L27-L28

c3y1huang:
> Do we need to skip all backing image tests based on the function?
> https://github.com/longhorn/longhorn-tests/blob/888cb8fc2086a59c72497862cd989d3d2214816e/manager/integration/tests/conftest.py#L27-L28

If I'm not mistaken I believe @meldafrawi is working on the upgrade failures. This PR is to fix the normal tests.

meldafrawi:
LGTM

meldafrawi (approved):

yasker:
> Test regression [Build #49](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/49/testReport/tests/test_backing_image/).

The regression test has passed. Merge this now.

innobead:
In the future, for any test cases given after the upgrade, we should choose a corresponding test runtime image to run the tests instead of the `master` test runtime image only. This can avoid this kind of common function incompatible between versions. 

before upgrade <-- use `1.1.0` test runtime image
after upgrade  <-- use master image instead

I will create a task for this to follow up.

PR #553, opened 2021-03-10
fix nightly upgrade test

meldafrawi:

Ref: https://github.com/longhorn/longhorn/issues/2321

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -376,7 +376,8 @@ def create_and_check_volume(client, volume_name,
     assert volume.size == size
     assert volume.numberOfReplicas == num_of_replicas
     assert volume.state == "detached"
-    assert volume.backingImage == backing_image
+    if backing_image_feature_supported(client):
+        assert volume.backingImage == backing_image
     assert volume.frontend == frontend
     assert volume.created != ""
     return volume
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2777,7 +2777,7 @@ def reset_engine_image(client):
         ei_list = client.list_engine_image().data
         for ei in ei_list:
             if ei.default:
-                if ei.state != 'deployed':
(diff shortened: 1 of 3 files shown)


manager/integration/tests/common.py
```diff
@@ -2581,184 +2582,23 @@ def reset_disks_for_all_nodes(client):  # NOQA
...
 
-    default_data_locality_setting = \
-        client.by_id_setting(SETTING_DEFAULT_DATA_LOCALITY)
-    try:
-        client.update(default_data_locality_setting, value="disabled")
-    except Exception as e:
-        print("Exception when update Default Data Locality setting",
-              default_data_locality_setting, e)
+    for setting in client.list_setting():
+        setting_name = setting.definition.name
```
c3y1huang, line 2587: I think this should be `setting.name`.

c3y1huang (changes requested):

manager/integration/tests/common.py
shuo-wu: The setting data is like this:
```
{
  "actions": { },
  "definition": {
    "category": "general",
    "default": "false",
    "description": "By default, Longhorn will block `kubectl drain` action on a node if the node contains the last healthy 
  replica of a volume.\n\nIf this setting is enabled, Longhorn will **not** block `kubectl drain` action on a node even if 
  the node contains the last healthy replica of a volume.",
    "displayName": "Allow Node Drain with the Last Healthy Replica",
    "readOnly": false,
    "required": true,
    "type": "bool",
  },
  "id": "allow-node-drain-with-last-healthy-replica",
  "name": "allow-node-drain-with-last-healthy-replica",
  "timestamp": "2021-03-11T02:44:32.632940164Z",
  "type": "setting",
  "value": "false",
},
```

meldafrawi: Fixed, Thanks

c3y1huang (approved):
LGTM

innobead (approved):

PR #554, opened 2021-03-11
Fix test_deleting_backup_volume

c3y1huang:
Missing backupstore targets.

https://github.com/longhorn/longhorn/issues/2323

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -1355,7 +1355,7 @@ def restore_inc_test(client, core_api, volume_name, pod):  # NOQA
     assert len(volumes) == 0
 
 
-def test_deleting_backup_volume(client, volume_name):  # NOQA
+def test_deleting_backup_volume(set_random_backupstore, client, volume_name):  # NOQA
     """
     Test deleting backup volumes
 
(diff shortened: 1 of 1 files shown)


c3y1huang:
Test regression [Build #56](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/56/testReport/tests/test_basic/).

manager/integration/tests/test_basic.py
```diff
@@ -1368,11 +1368,10 @@ def test_deleting_backup_volume(client, volume_name):  # NOQA
     volume.attach(hostId=lht_host_id)
     volume = common.wait_for_volume_healthy(client, volume_name)
 
-    bv, _, snap1, _ = create_backup(client, volume_name)
-    _, _, snap2, _ = create_backup(client, volume_name)
+    create_backup(client, volume_name)
+    create_backup(client, volume_name)
 
     delete_backup_volume(client, volume_name)
-    cleanup_volume(client, volume)
```
innobead, line 1375: Why `create_volume` removed? It will not impact the test result, but we should keep it for now.

innobead (changes requested):

manager/integration/tests/test_basic.py
c3y1huang: I thought this looks duplicated since it's already handled by the cleanup_client. But on another thought, it makes sense to test for volume cleanup.

c3y1huang: Addressed.

c3y1huang:
Test regression [Build #59](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/59/).

innobead (approved):

PR #558, opened 2021-03-15
Fix test_restore_volume_with_invalid_backupstore

c3y1huang:
Volume does not get created with an invalid backupstore URL.
Update test to address to change from https://github.com/longhorn/longhorn/issues/2006.

https://github.com/longhorn/longhorn/issues/2343


c3y1huang:
Test regression [Build #82](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/82/testReport/tests/test_ha/).

innobead (approved):

PR #559, opened 2021-03-15
Fix test_engine_image_daemonset_restart

c3y1huang:
Correct engine image state value.

https://github.com/longhorn/longhorn/issues/2344

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -2195,7 +2195,8 @@ def test_engine_image_daemonset_restart(client, apps_api, volume_name):  # NOQA
     check_volume_data(volume, snap1_data)
 
     # Wait for the restart complete
-    common.wait_for_engine_image_state(client, default_img.name, "ready")
+    ei_state = common.get_engine_image_status_value(client, default_img.name)
+    common.wait_for_engine_image_state(client, default_img.name, ei_state)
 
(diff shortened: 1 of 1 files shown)


c3y1huang:
Test regression [Build #85](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/85/testReport/tests/test_basic/)

manager/integration/tests/test_basic.py
```diff
@@ -2195,7 +2195,8 @@ def test_engine_image_daemonset_restart(client, apps_api, volume_name):  # NOQA
     check_volume_data(volume, snap1_data)
 
     # Wait for the restart complete
-    common.wait_for_engine_image_state(client, default_img.name, "ready")
+    ei_state = common.get_engine_image_status_value(client, default_img.name)
+    common.wait_for_engine_image_state(client, default_img.name, ei_state)
```
innobead, line 2199: This logic has some problem: query the current state of the engine image, then compare the value by querying the engine image state again.

In 1.1.1, I think we should just check the state == `deployed`?

innobead (changes requested):

manager/integration/tests/test_basic.py
innobead: After @c3y1huang explained, I think this is not an issue.

innobead (approved):

PR #562, opened 2021-03-16
Fix test_replica_zone_anti_affinity

c3y1huang:
Remove checking volume condition because allow-volume-creation-with-degraded-availability set to true by default.

https://github.com/longhorn/longhorn/issues/2352


c3y1huang:
Test regression [Build #105](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/105/testReport/tests/test_zone/)

innobead (approved):
LGTM.

But it seems we can add some other test combination w/ same zone label and replicaZoneSoftAntiAffinity is disabled to verify other replicas fale scheduled?

PR #563, opened 2021-03-16
Fix test_disable_scheduling_on_cordoned_node

c3y1huang:
Remove checking volume condition because allow-volume-creation-with-degraded-availability set to true by default.

https://github.com/longhorn/longhorn/issues/2354


c3y1huang:
Test regression [Build#107](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/107/testReport/tests/test_node/)

innobead (approved):
LGTM

ref: https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L958-L976

PR #574, opened 2021-03-19
Split test_recurring_job into 2 cases

c3y1huang:
* Updated the test_recurring_job to test when the write, snapshot, and backup are not happening at the same time.

* Moved original test_recurring_job to test_recurring_job_snapshot_and_backup_concurrent as discovered the unstable backup number created when write and backup is happening at the same time.

https://github.com/longhorn/longhorn/issues/2378


c3y1huang:
Closing. Seems to fixed by https://github.com/longhorn/longhorn/issues/2378.

PR #577, opened 2021-03-22
update test_ha_salvage test to account for revision counter

meldafrawi:
Ref: https://github.com/longhorn/longhorn/issues/508


manager/integration/tests/test_ha.py
```diff
@@ -279,6 +293,100 @@ def ha_salvage_test(client, core_api, # NOQA
...
+
+    data = write_volume_random_data(volume)
+
+    crash_replica_processes(client, core_api, volume_name)
+
+    volume = common.wait_for_volume_faulted(client, volume_name)                
+    assert len(volume.replicas) == 3
+
+    volume = common.wait_for_volume_healthy(client, volume_name)
+    assert len(volume.replicas) == 5
```
meldafrawi, line 330: @shuo-wu according to tests steps, replica count should be 5, but I only get 3
is that expected.

shuo-wu: 3-replica is what we expected now. Because this test skeleton is written before the failed replica reusage feature, the 2 failed replicas mentioned here will be reused automatically now.

You can modify the test skeleton to:
```
7. Verify all 3 replicas are reused successfully.
```

manager/integration/tests/test_ha.py
```diff
@@ -279,6 +293,100 @@ def ha_salvage_test(client, core_api, # NOQA
...
+    # Setting: Enabled auto salvage.
+    # Case 3: Revision counter disabled.
+
+    auto_salvage_setting = client.by_id_setting(SETTING_AUTO_SALVAGE)
+    setting = client.update(auto_salvage_setting, value="true")
+    assert setting.name == SETTING_AUTO_SALVAGE
+    assert setting.value == "true"
+
+    disable_revision_counter_setting = \
+        client.by_id_setting("disable-revision-counter")
```
shuo-wu, line 305: Use a constant here.

shuo-wu (changes requested):

manager/integration/tests/test_ha.py
meldafrawi: Fixed, Thank you.

manager/integration/tests/test_ha.py
meldafrawi: Fixed, Thank you

shuo-wu (approved):
LGTM

innobead (approved):
LGTM

PR #578, opened 2021-03-24
fix empty junit report

meldafrawi:
https://github.com/longhorn/longhorn/issues/2333

### DIFF
--- a/manager/integration/tests/run.sh
+++ b/manager/integration/tests/run.sh
@@ -1,9 +1,7 @@
 #!/bin/bash
 
 if [[ $@ =~ "--junitxml=" ]] ; then 
-  flake8
-
-  pytest -v $@  >> /tmp/longhorn-pytest 2>&1
+  pytest -v $@  > /tmp/longhorn-pytest
   
   cat ${LONGHORN_JUNIT_REPORT_PATH}
 else 
--- a/test_framework/scripts/longhorn-setup.sh
+++ b/test_framework/scripts/longhorn-setup.sh
@@ -150,15 +150,15 @@ run_longhorn_upgrade_test(){
 	kubectl apply -f ${LONGHORN_UPGRADE_TESTS_MANIFEST_FILE_PATH}
 
 	# wait upgrade test pod to start running
-    while [[ -n "`kubectl get pods longhorn-test-upgrade --no-headers=true | awk '{print $3}' | grep -v Running`"  ]]; do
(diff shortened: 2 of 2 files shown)


manager/integration/tests/run.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
 if [[ $@ =~ "--junitxml=" ]] ; then 
-  flake8
+  flake8 >> /dev/null
```
innobead, line 4: Why do we need to run flake8 at runtime? I would suggest having it at build time.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -150,12 +150,12 @@ run_longhorn_upgrade_test(){
...
 
 	# wait upgrade test pod to start running
-    while [[ -n "`kubectl get pods longhorn-test-upgrade --no-headers=true | awk '{print $3}' | grep -v Running`"  ]]; do
+    while [[ -n "`kubectl get pods longhorn-test-upgrade --no-headers=true | awk '{print $3}' | grep -v \"Running\|Completed\"`"  ]]; do
 		echo "waiting upgrade test pod to be in running state ... rechecking in 10s"
 		sleep 10s
     done
 
-	# wait upgrade test to complete
+	 wait upgrade test to complete
```
innobead, line 158: This should be a comment?

manager/integration/tests/run.sh
```diff
@@ -1,9 +1,9 @@
...
 
-  pytest -v $@  >> /tmp/longhorn-pytest 2>&1
+  pytest -v $@  > /tmp/longhorn-pytest
```
innobead, line 6: The output of the testing pod will be used as the JUnit report, so that means it's impossible for now to see any progress or output during running tests via `pytest`, correct?

If yes, we should improve this part later to let people know what's going during the test instead of waiting for the report generated (nightly build will take hours to finish).

innobead (commented):
LGTM, just a few comments to resolve/answer.

manager/integration/tests/run.sh
meldafrawi: Yes we need to wait to the test to finish to get the result.
will try figure something out for it later.

manager/integration/tests/run.sh
meldafrawi: Fixed, Thanks

test_framework/scripts/longhorn-setup.sh
meldafrawi: Fixed, Thank you.

innobead (approved):

PR #580, opened 2021-03-29
E2E Test AWS IAM Role ARN annotation

jenting:


innobead:
Do we still need this PR? maybe we can close and do later if don't. Thanks!

jenting:
Close it first.

PR #581, opened 2021-03-30
wait 30 seconds for docker to completetly start

meldafrawi:
Ref: https://github.com/longhorn/longhorn/issues/2375

RKE somtimes fails to run on worker nodes because the delayed start of docker service.


innobead (approved):
LGTM

Next time, I would suggest adding some comments or logs to know the purpose especially for this kind of general operations. (or you can revisit the current infra code to enhance a bit, this will be helpful for the future troubleshooting)

PR #583, opened 2021-04-01
docs: Update manual tests for enhanced backing image feature

shuo-wu:
longhorn/longhorn#2295

### DIFF
--- a/docs/content/manual/pre-release/upgrade/backing-image-during-upgrade.md
+++ b/docs/content/manual/pre-release/upgrade/backing-image-during-upgrade.md
@@ -2,9 +2,51 @@
 title: Test Backing Image during Longhorn upgrade
 ---
 
-1. Deploy an old version Longhorn. Then set `Concurrent Automatic Engine Upgrade Per Node Limit` to a positive value to enable volume engine auto upgrade.
-2. Create a backing image.
+## System upgrade with compatible backing image manager image
+1. Deploy Longhorn. Then set `Concurrent Automatic Engine Upgrade Per Node Limit` to a positive value to enable volume engine auto upgrade.
+2. Create 2 backing images: a large one and a small one.
+3. Create and attach volumes with the small backing image.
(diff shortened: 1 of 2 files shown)


docs/content/manual/pre-release/upgrade/backing-image-during-upgrade.md
```diff
@@ -2,9 +2,51 @@
...
+    2. The default backing image manager image will be updated.
+    3. The new longhorn manager is compatible with the old backing image manager.
+7. Wait for system upgrade complete. Then verify:
+    1. All old backing image manager and the related pod will be cleaned up automatically after the current downloading is complete. And the existing backing image files won't be removed.
+    2. New default backing image manager will take over all backing image ownerships and show the info in the status map. 
+    3. All attached volumes still work fine without replica crash, and the content is correct in the volumes during/after the upgrade.
+    4. The last volume get attached successfully without replica crash, and the content is correct.
+8. Verify volumes and backing images can be deleted.
+
+## System upgrade with incompatible backing image manager image
```
innobead, line 22: It looks there is no different for system upgrade with compatible/incompatible backing image manager image. Right? because it seems at the end the backing image manager pods will be upgraded and the ownership of images always transfered successfully. 

If there is a difference, do we need to have a negative case for that?

innobead (commented):
LGTM, but have one question to anwser first. Thanks!

docs/content/manual/pre-release/upgrade/backing-image-during-upgrade.md
shuo-wu: From users' perspective, there is no difference. But the actual workflow is different. For compatible old managers, Longhorn will do ownership transfer. For incompatible managers, Longhorn will ask new backing image managers to re-download all necessary images and the new manager will try their best to reuse the files. Besides, we need to make sure all incompatible backing images can be cleaned up automatically and successfully. Hence I put 2 separate cases here.

innobead (approved):

PR #592, opened 2021-04-13
Add and split DR tests regards to backup

c3y1huang:
We want to make sure that when the block deletion for the last backup is aborted by operations such as backups in progress, the DR volume will still pick up the correct last backup.

https://github.com/longhorn/longhorn/issues/2339

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -46,7 +46,7 @@ from common import wait_for_dr_volume_expansion
 from common import check_block_device_size
 from common import wait_for_volume_expansion
 from common import fail_replica_expansion, wait_for_expansion_failure
-from common import wait_for_volume_creation, wait_for_volume_restoration_start
+from common import wait_for_volume_restoration_start
 from common import write_pod_volume_random_data, get_pod_data_md5sum
 from common import prepare_pod_with_data_in_mb
 from common import crash_replica_processes
@@ -57,6 +57,7 @@ from common import VOLUME_FRONTEND_BLOCKDEV, VOLUME_FRONTEND_ISCSI
 from common import VOLUME_CONDITION_SCHEDULED
 from common import MESSAGE_TYPE_ERROR
 from common import DATA_SIZE_IN_MB_1
+from common import update_setting
 from common import SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY
 from common import SETTING_REPLICA_REPLENISHMENT_WAIT_INTERVAL
 from common import CONDITION_REASON_SCHEDULING_FAILURE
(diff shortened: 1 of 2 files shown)


joshimoo (commented):
This might be covered by other tests already, not sure.

I think the DR test you posted needs to be updated though, to create the DR volume then wait for restore of backup(1) before deletion of backup(1). I am assuming the DR volume will stay on backup(1) till the next backup is created, which will update lastBackup.

c3y1huang:
Regression [Build#156](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/156/testReport/tests/test_basic/)

manager/integration/tests/test_basic.py
```diff
@@ -708,6 +708,99 @@ def test_backup_block_deletion(set_random_backupstore, client, core_api, volume_
...
+
+    1.  Create a volume and attach to the current node.
+    2.  Write 4 MB to the beginning of the volume (2 x 2MB backup blocks).
+    3.  Create backup(0) of the volume.
+    4.  Overwrite backup(0) blocks of data on the volume.
+    5.  Create backup(1) of the volume.
+    6.  Verify backup block count == 3.
+    7.  Create DR volume from backup(1).
+    8.  Verify DR volume last backup is backup(1).
+    9.  Create an artificial in progress backup.cfg file.
```
shuo-wu, line 731: Add a simple explanation for this hacky step. e.g.,
```
This cfg file will wrongly convince the longhorn manager that there is a backup being created. Then all subsequent backup block cleanup will be skipped.
```

manager/integration/tests/test_basic.py
```diff
@@ -708,6 +708,99 @@ def test_backup_block_deletion(set_random_backupstore, client, core_api, volume_
...
+    10. Delete backup(1).
+    11. Verify backup block count == 3 (because of the in progress backup).
+    12. Delete the artificial in progress backup.cfg file.
+    13. Verify DR volume last backup is backup(0) or backup(1).
+        s3 -> backup(0), nfs -> backup(1).
```
shuo-wu, line 736: Why is there a difference?

shuo-wu (changes requested):
It seems there are some duplications with test `test_dr_volume_with_last_backup_deletion`. Can you check if we need to merge that with this test, or directly remove it?

manager/integration/tests/test_basic.py
c3y1huang: Actually, I am not entirely sure, will check the test case again.

manager/integration/tests/test_basic.py
c3y1huang: Added.

c3y1huang:
Regression [Build#160](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/160/testReport/tests/test_basic/) - runs `test_dr_volume_with_backup_block_deletion`, `test_dr_volume_with_backup_block_deletion_abort`, `test_dr_volume_with_all_backup_deleted` each 5 times.

manager/integration/tests/test_basic.py
c3y1huang: Cannot reproduce anymore, removed. Test result in https://github.com/longhorn/longhorn-tests/pull/592#issuecomment-819399339.

c3y1huang:
> It seems there are some duplications with test `test_dr_volume_with_last_backup_deletion`. Can you check if we need to merge that with this test, or directly remove it?

Test `test_dr_volume_with_last_backup_deletion` seems to be testing for two cases - one to test when all backups are deleted, the other to test normal backup delete, and in my opinion, also different to this where we are testing the case when block delete is aborted.

I've renamed and separated the `test_dr_volume_with_last_backup_deletion` to:
* test_dr_volume_with_backup_block_deletion
* test_dr_volume_with_backup_block_deletion_abort
* test_dr_volume_with_all_backup_deleted

manager/integration/tests/test_basic.py
```diff
@@ -708,6 +709,243 @@ def test_backup_block_deletion(set_random_backupstore, client, core_api, volume_
...
+
+    check_volume_last_backup(client, dr_vol_name, backup2.name)
+    wait_for_volume_restoration_start(client, dr_vol_name, backup2.name)
+    wait_for_backup_restore_completed(client, dr_vol_name, backup2.name)
+
+    activate_standby_volume(client, dr_vol_name)
+    dr_vol = client.by_id_volume(dr_vol_name)
+    dr_vol.attach(hostId=host_id)
+    dr_vol = common.wait_for_volume_healthy(client, dr_vol_name)
+    check_volume_data(dr_vol, data2, False)
```
shuo-wu, line 792: The 2nd block of data is not verified.
Same below.

manager/integration/tests/test_basic.py
```diff
@@ -708,6 +709,243 @@ def test_backup_block_deletion(set_random_backupstore, client, core_api, volume_
...
+    3.  Create backup(0) of the volume.
+    4.  Overwrite backup(0) blocks of data on the volume.
+    5.  Create backup(1) of the volume.
+    6.  Verify backup block count == 3.
+    7.  Create DR volume from backup(1).
+    8.  Verify DR volume last backup is backup(1).
+    9.  Delete backup(1).
+    10. Verify backup block count == 2.
+    11. Verify DR volume last backup is backup(0).
+    12. Overwrite backup(0) blocks of data on the volume.
```
shuo-wu, line 734: Overwrite the 1st block of data on the volume.
Same below.

shuo-wu (changes requested):
Need to update the commit title as well

manager/integration/tests/test_basic.py
c3y1huang: The case covers `check_volume_last_backup(client, dr_vol_name, backup1.name)`, I think it should be sufficient.
If I understand correctly, to verify data for block(1) is just to ensure the dr is working? I can add another case just for it so it does not overcomplicate what each case is testing for.

c3y1huang: After discussing with @shuo-wu , we have agreed there is no need to verify the 2nd block of data because it involves creating 2 dr and we cannot use the validated dr to continue the rest of the case. There is no need to create another case because there is already a test case for that.

manager/integration/tests/test_basic.py
c3y1huang: Thanks @shuo-wu , this is addressed.

c3y1huang:
Regression [Build#161](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/161/testReport/tests/test_basic/)

c3y1huang:
Thanks @shuo-wu also pointed out the final data validation should also include 2nd block of backup(0). I will update.

innobead (commented):
LGTM, but just heard from Chinya mentioned these new test cases failed, and need to do further investigation first. (after figuring out the problem, it can be approved)

c3y1huang:
> LGTM, but just heard from @c3y1huang mentioned these new test cases failed, and need to do further investigation first. (after figuring out the problem, it can be approved)

Not sure, it seems only failing in my environment. They've passed on Jenkins.

shuo-wu (approved):
LGTM

PR #597, opened 2021-04-15
Fix test_all_replica_restore_failure

c3y1huang:
Update to check the volume manager's response since we're now always [allowed api attach and detach calls](https://github.com/longhorn/longhorn-manager/blob/master/api/model.go#L965-L968).

https://github.com/longhorn/longhorn/issues/2499
https://github.com/longhorn/longhorn-manager/pull/859


c3y1huang:
5 * Regression([Build#162](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/162/testReport/tests/test_ha/))

innobead (approved):
LGTM.

Very good to fix the e2e issues due to the new enhancements introduced.

PR #598, opened 2021-04-15
Fix test_setting_toleration*

c3y1huang:
Check the toleration change for the managed components since we only watch and updated those.

https://github.com/longhorn/longhorn/issues/2500
https://github.com/longhorn/longhorn-manager/pull/842


c3y1huang:
Regression([Build#163](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/163/testReport/tests/test_settings/))

innobead (approved):
LGTM

PR #600, opened 2021-04-16
tests: Update test_restore_inc

shuo-wu:
The backup & restore should involve at least 2 blocks.


shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/167

c3y1huang (approved):
LGTM

PR #601, opened 2021-04-16
Fix test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress

c3y1huang:
The DR last backup should be empty when GC delta block delete is aborted.

https://github.com/longhorn/longhorn-tests/pull/592

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -832,8 +832,8 @@ def test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress(
         skipped.
     10. Delete backup(1).
     11. Verify backup block count == 3 (because of the in progress backup).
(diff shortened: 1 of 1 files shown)


c3y1huang:
Testing with https://github.com/longhorn/longhorn-engine/pull/573.

c3y1huang:
Regression ([Build#175](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/175/testReport/tests/test_basic/))

innobead (approved):
LGTM. The test case change is correct as per the current implementation.

But, just have a little concern in this changed case if the last backup of the volume is empty but there are still some backups in the backup store, does it make sense for users? In fact, there is no any hurt to users, but the last backup does not really reflecting to the user DR in this case.  (probably we need to revisit)

PR #604, opened 2021-04-22
integration: Fix ha_salvage_test

shuo-wu:
longhorn/longhorn#1236

Signed-off-by: Shuo Wu <shuo.wu@suse.com>


shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/198/

innobead (approved):
LGTM

```
If the status is checked when the salvaging hasn’t handled all replicas, the error will be triggered. Vice versa, there will be no errors.
```
@shuo-wu

PR #605, opened 2021-04-27
Implement test_multiple_volumes_creation_with_degraded_availability

c3y1huang:
https://github.com/longhorn/longhorn/issues/2486

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -92,7 +92,7 @@ DEFAULT_BACKUP_TIMEOUT = 100
 DEFAULT_POD_INTERVAL = 1
 DEFAULT_POD_TIMEOUT = 180
 
-DEFAULT_STATEFULSET_INTERVAL = 5
+DEFAULT_STATEFULSET_INTERVAL = 1
 DEFAULT_STATEFULSET_TIMEOUT = 180
 
 DEFAULT_DEPLOYMENT_INTERVAL = 1
@@ -1111,7 +1111,8 @@ def statefulset(request):
         'apiVersion': 'apps/v1',
         'kind': 'StatefulSet',
         'metadata': {
-            'name': 'test-statefulset'
+            'name': 'test-statefulset',
+            'namespace': 'default',
         },
         'spec': {
             'selector': {
@@ -1353,6 +1354,8 @@ def cleanup_client():
 
     if backing_image_feature_supported(client):
         cleanup_all_backing_images(client)
+
+    cleanup_storage_class()
 
     # enable nodes scheduling
     reset_node(client)
@@ -3003,6 +3006,16 @@ def check_csi_expansion(core_api):
     return csi_expansion_enabled
 
 
+def create_statefulset(statefulset_manifest):
+    """
+    Create a new StatefulSet for testing.
+    """
+    api = get_apps_api_client()
+    api.create_namespaced_stateful_set(
+        body=statefulset_manifest,
+        namespace='default')
+
+
 def create_and_wait_statefulset(statefulset_manifest):
     """
     Create a new StatefulSet for testing.
@@ -3010,10 +3023,7 @@ def create_and_wait_statefulset(statefulset_manifest):
     This function will block until all replicas in the StatefulSet are online
     or it times out, whichever occurs first.
     """
-    api = get_apps_api_client()
-    api.create_namespaced_stateful_set(
-        body=statefulset_manifest,
-        namespace='default')
+    create_statefulset(statefulset_manifest)
     wait_statefulset(statefulset_manifest)
 
 
@@ -3043,6 +3053,29 @@ def delete_storage_class(sc_name):
         api.delete_storage_class(sc_name, body=k8sclient.V1DeleteOptions())
     except ApiException as e:
         assert e.status == 404
+
+
+def cleanup_storage_class():
+    skip_sc_deletes = ["longhorn", "local-path"]
+    api = get_storage_api_client()
+    ret = api.list_storage_class()
+    for sc in ret.items:
+        if sc.metadata.name in skip_sc_deletes:
+            continue
+        delete_storage_class(sc.metadata.name)
+
+    ok = False
+    for _ in range(RETRY_COUNTS):
+        ok = True
+        ret = api.list_storage_class()
+        for sc in ret.items:
+            if sc.metadata.name not in skip_sc_deletes:
+                ok = False
+                break
+        if ok:
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert ok
 
 
 def create_pvc(pvc_manifest):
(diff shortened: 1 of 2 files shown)


c3y1huang:
Regression - [Build#208](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/208/testReport/tests/test_basic/)

manager/integration/tests/test_basic.py
```diff
@@ -3324,28 +3326,73 @@ def test_allow_volume_creation_with_degraded_availability_error(
...
+
+    retry_count = int(60 / RETRY_INTERVAL)
+    for _ in range(common.DEFAULT_STATEFULSET_TIMEOUT):
+        try:
+            pod_info = common.get_statefulset_pod_info(core_api, statefulset)
+            break
+        except ApiException as e:
+            assert e.status == 404
+            common.DEFAULT_STATEFULSET_INTERVAL
+    assert pod_info is not None
```
shuo-wu, line 3371: Why is `create_and_wait_statefulset` not used?

And you can add check `assert len(pod_list) == 10`

manager/integration/tests/common.py
```diff
@@ -2859,6 +2867,34 @@ def wait_statefulset(statefulset_manifest):
     assert s_set.status.ready_replicas == replicas
 
 
+def wait_for_statefulset_volume_state(client, manifest, pod_info, field, value,  # NOQA
```
shuo-wu, line 2870: Rename `pod_info` like `pod_list`. Same below

manager/integration/tests/test_basic.py
```diff
@@ -3324,28 +3326,73 @@ def test_allow_volume_creation_with_degraded_availability_error(
...
+
+    common.wait_for_statefulset_volume_state(client, statefulset, pod_info,
+                                             common.VOLUME_FIELD_STATE,
+                                             common.VOLUME_STATE_DETACHED,
+                                             retry_counts=retry_count)
+
+    for p in pod_info:
+        core_api.delete_namespaced_persistent_volume_claim(
+            name=p['pvc_name'], namespace='default',
+            body=k8sclient.V1DeleteOptions())
```
shuo-wu, line 3392: Use `delete_and_wait_pvc`

shuo-wu (changes requested):

manager/integration/tests/test_basic.py
c3y1huang: > Why is `create_and_wait_statefulset` not used?

`create_and_wait_statefulset` loop through 15 mins to wait for replicas to be ready, this looks like it contradicts with the testing step `In a 1-min retry loop, Verify that all 10 volumes are healthy`.

manager/integration/tests/test_basic.py
c3y1huang: `delete_and_wait_pvc` loop through 2.5 mins to wait for pvc to get removed, this looks like it contradicts with the testing step `In a 1-min retry loop, Verify that all 10 volumes are deleted`... do we still want to change this?

manager/integration/tests/test_basic.py
shuo-wu: Why does the function need to wait for 15min? It's too long for a single test... You can directly modify the function if possible.

c3y1huang: We can modify the function and it will still wait for replicas to be ready and won't volume already be healthy?

c3y1huang: @PhanLe1010 since you are the original writer of this test skeleton, can we have your option on this?

manager/integration/tests/common.py
c3y1huang: Addressed.

manager/integration/tests/test_basic.py
shuo-wu: I don't think `1-min` is a hard requirement. @PhanLe1010 Is using a longer wait interval OK in this test?

manager/integration/tests/test_basic.py
c3y1huang: Updated to use `create_and_wait_statefulset` and also update the `DEFAULT_STATEFULSET_INTERVAL` to 1, similar to the timeout for pod and deployment.

manager/integration/tests/test_basic.py
c3y1huang: Since the original goal statement is `We want to verify that multiple volumes with degraded availability can be created, attached, detached, and deleted at nearly the *same time*`, I've updated to use `delete_and_wait_pvc` as well as with loop for 1-min retry loop.

c3y1huang:
Regression [build-451](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/451/testReport/tests/test_basic/)

manager/integration/tests/common.py
```diff
@@ -3045,6 +3084,22 @@ def delete_storage_class(sc_name):
...
 
+def cleanup_storage_class():
+    delete_storage_class(DEFAULT_STORAGECLASS_NAME)
+    ok = False
+    api = get_storage_api_client()
+    for _ in range(RETRY_COUNTS):
+        ok = True
+        ret = api.list_storage_class()
+        for sc in ret.items:
+            if sc.metadata.name == DEFAULT_STORAGECLASS_NAME:
```
shuo-wu, line 3095:
```
if sc.metadata.name != DEFAULT_STORAGECLASS_NAME:
```

manager/integration/tests/common.py
```diff
@@ -3031,6 +3041,35 @@ def wait_statefulset(statefulset_manifest):
     assert s_set.status.ready_replicas == replicas
 
 
+def wait_for_statefulset_pods_volume_state(client, pod_list, field, value,  # NOQA
```
shuo-wu, line 3044: NIT: `wait_for_pods_volume_state` similar below

shuo-wu (changes requested):

manager/integration/tests/common.py
c3y1huang: Don't think that will work because we only want to delete the *extra* storageClass.

manager/integration/tests/common.py
c3y1huang: Addressed.

manager/integration/tests/common.py
shuo-wu: Oh sorry. `DEFAULT_STORAGECLASS_NAME` means the default SC of the integration test rather than `longhorn`. my fault.

c3y1huang: True, it's better to remove everything else. Addressed.

c3y1huang:
Regression [build-453](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/453/testReport/tests/test_basic/)

shuo-wu (approved):

PR #606, opened 2021-04-29
Update test_setting_priority_class and fix toleration

c3y1huang:
https://github.com/longhorn/longhorn/issues/2482

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -1,6 +1,7 @@
 import time
 import pytest
 
+import common
 from common import (  # NOQA
     get_longhorn_api_client, get_self_host_id,
     get_core_api_client, get_apps_api_client,
@@ -30,14 +31,28 @@ from test_infra import wait_for_node_up_longhorn
 KUBERNETES_DEFAULT_TOLERATION = "kubernetes.io"
 
 
-def check_workload_update(core_api, apps_api, count):  # NOQA
+def check_longhorn_managed_workloads(core_api, apps_api, count, backing_images=[]):  # NOQA
+    pod_list = core_api.list_namespaced_pod(LONGHORN_NAMESPACE).items
+    for p in pod_list:
+        if p.status.phase != "Running":
+            return False
+
+    skip_updates = [
+        "longhorn-driver-deployer",
+        "longhorn-manager",
+        "longhorn-ui",
+    ]
     da_list = apps_api.list_namespaced_daemon_set(LONGHORN_NAMESPACE).items
     for da in da_list:
+        if da.metadata.name in skip_updates:
+            continue
         if da.status.updated_number_scheduled != count:
             return False
 
     dp_list = apps_api.list_namespaced_deployment(LONGHORN_NAMESPACE).items
     for dp in dp_list:
+        if dp.metadata.name in skip_updates:
+            continue
         if dp.status.updated_replicas != dp.spec.replicas:
             return False
 
@@ -47,16 +62,18 @@ def check_workload_update(core_api, apps_api, count):  # NOQA
     if len(im_pod_list) != 2 * count:
         return False
 
-    for p in im_pod_list:
-        if p.status.phase != "Running":
-            return False
-
     client = get_longhorn_api_client()  # NOQA
     images = client.list_engine_image()
     assert len(images) == 1
     ei_state = get_engine_image_status_value(client, images[0].name)
     if images[0].state != ei_state:
         return False
+
+    if len(backing_images) != 0:
+        bi_disks = common.list_backing_image_disks(backing_images)
+        bim_pods = common.list_backing_image_manager_pods(core_api)
+        if len(bi_disks) != len(bim_pods):
+            return False
 
     return True
 
@@ -81,7 +98,8 @@ def test_setting_toleration():
 
     1.  Set `taint-toleration` to "key1=value1:NoSchedule; key2:InvalidEffect".
     2.  Verify the request fails.
-    3.  Create a volume and attach it.
+    3.  Create a volume.
+    4.  Attach the volume to self node and wait volume healthy.
     4.  Set `taint-toleration` to "key1=value1:NoSchedule; key2:NoExecute".
     5.  Verify that cannot update toleration setting when any volume is
         attached.
@@ -90,7 +108,7 @@ def test_setting_toleration():
     8.  Set `taint-toleration` to "key1=value1:NoSchedule; key2:NoExecute".
     9.  Wait for all the Longhorn system components to restart with new
         toleration.
-    10. Verify that UI, manager, and drive deployer don't restart and
+    10. Verify that UI, manager, and drive-deployer don't restart and
         don't have new toleration.
     11. Attach the volume again and verify the volume `data1`.
     12. Generate and write `data2` to the volume.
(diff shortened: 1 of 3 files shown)


c3y1huang:
Regression - [Build#209](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/209/testReport/tests/test_settings/)

manager/integration/tests/test_settings.py
```diff
@@ -278,47 +279,41 @@ def test_setting_toleration_extra(core_api, apps_api):  # NOQA
...
         "longhorn-ui",
     ]
-    updated = False
+    ok = False
     for _ in range(RETRY_COUNTS):
         time.sleep(RETRY_INTERVAL_LONG)
 
-        updated = True
+        ok = False
         if not check_workload_update(core_api, apps_api, count):
```
shuo-wu, line 292: 1. It's better to rename this function to `check_longhorn_managed_workloads` to avoid misunderstanding.
2. Need to add check for new components like backing image manager pods and share manager pods.
3. Avoid checking longhorn-manager, longhorn-ui, and longhorn-driver-deployer pods.

manager/integration/tests/test_settings.py
```diff
@@ -278,47 +279,41 @@ def test_setting_toleration_extra(core_api, apps_api):  # NOQA
...
-                assert app_name not in not_managed_apps
-
-            if p.status.phase != "Running" \
-                or not check_tolerations_set(p.spec.tolerations,
-                                             expected_tolerations,
-                                             chk_removed_tolerations):
-                updated = False
+            app_name = str(p.metadata.labels.get('app', ''))
+            if p.status.phase != "Running":
+                ok = False
```
shuo-wu, line 299: 1. I am thinking if this pod-running check should be handled by `check_workload_update`
2. Some longhorn managed workloads like engine image pods do not contain the label `app`
3. Similar to `check_workload_update`, need to make sure all longhorn managed components only are listed and checked here

shuo-wu (changes requested):

manager/integration/tests/test_settings.py
c3y1huang: > 1. It's better to rename this function to `check_longhorn_managed_workloads` to avoid misunderstanding.

Addressed.

> 2. Need to add check for new components like backing image manager pods and share manager pods.

Add to create volume with backing image and this will be checked in `wait_for_toleration_update` and `wait_for_priority_class_update`.

Did not address checking for share manager because volume needs to be detached before the update and doesn't look like it will be there when doing setting updates.

> 3. Avoid checking longhorn-manager, longhorn-ui, and longhorn-driver-deployer pods.

Addressed.

manager/integration/tests/test_settings.py
c3y1huang: Addressed.

c3y1huang:
Regression - [Build#214](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/214/testReport/tests/test_settings/)

manager/integration/tests/test_settings.py
```diff
@@ -113,8 +126,11 @@ def test_setting_toleration():
                       value="key1=value1:NoSchedule; key2:InvalidEffect")
     assert 'invalid effect' in str(e.value)
 
+    common.create_backing_image_with_matching_url(
+        client, common.BACKING_IMAGE_NAME, common.BACKING_IMAGE_QCOW2_URL)
     volume_name = "test-toleration-vol"  # NOQA
-    volume = create_and_check_volume(client, volume_name)
+    volume = create_and_check_volume(client, volume_name,
+                                     backing_image=common.BACKING_IMAGE_NAME)
```
shuo-wu, line 133: It's better to use a separate case to test the backing image. And checking the data is required.
You can refer to the tests in `test_backing_image.py`

Same below

shuo-wu (changes requested):

PhanLe1010 (commented):
On the line https://github.com/longhorn/longhorn-tests/blob/36ed5c7dba4f99f34e936ab827b9bb3d845471b7/manager/integration/tests/test_settings.py#L184

Should we pass in the `chk_removed_tolerations` to verify that tolerations are removed?

manager/integration/tests/test_settings.py
c3y1huang: Separated case to `test_setting_toleration_with_backing_image`.

c3y1huang:
> Should we pass in the `chk_removed_tolerations` to verify that tolerations are removed?

Thanks for the suggestion. Added.

c3y1huang:
test_setting* regression - [build#334](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/334/testReport/)

c3y1huang:
This is an old PR that almost got forgotten. I've addressed all review comments. PTAL @shuo-wu , @PhanLe1010 , @innobead

manager/integration/tests/test_settings.py
```diff
@@ -277,48 +301,43 @@ def test_setting_toleration_extra(core_api, apps_api):  # NOQA
...
+                                                    chk_removed_tolerations)
+            app_name = str(p.metadata.labels.get('app', ''))
+            if app_name in skip_updates and n_update == 0:
+                ok = True
+            elif n_update == len(expected_tolerations):
+                ok = True
+        if ok:
             break
-    assert updated
+    assert ok
```
shuo-wu, line 336: Why not put this backing image manager pod checking into function `check_longhorn_managed_workloads`?

Besides, you can use the following logic to determine if a backing image manager pod **should** exist:
    1. List all BackingImages
    2. For each BackingImage, each entry in `diskStateMap` means there should be a backing image manager pod running for the disk. The key of `diskStateMap` is disk UUID. After iterating BackingImages, we can get the disks that require backing image managers.
    3. List all backing image managers, there should be a backing image manager for each entry of the disk list. And the related pod is running with correct tolerations & priority class.

shuo-wu (changes requested):

manager/integration/tests/test_settings.py
c3y1huang: Fixed!

c3y1huang:
test_setting* regression - [build#334](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/337/)

manager/integration/tests/test_settings.py
```diff
@@ -277,48 +308,46 @@ def test_setting_toleration_extra(core_api, apps_api):  # NOQA
...
+
+            n_update = check_tolerations_set_update(p.spec.tolerations,
+                                                    expected_tolerations,
+                                                    chk_removed_tolerations)
+            app_name = str(p.metadata.labels.get('app', ''))
+            if app_name in skip_updates and n_update == 0:
+                ok = True
+            elif n_update == len(expected_tolerations):
+                ok = True
+        if ok:
```
shuo-wu, line 344: I don't understand this part: 
1. Why doesn't this function check other longhorn-managed pods if there are backing images? 
2. Some of them don't contain label `app`. e.g., engine image pods,

shuo-wu (changes requested):

manager/integration/tests/test_settings.py
c3y1huang: > 1. Why doesn't this function check other longhorn-managed pods if there are backing images?

Since we've decided to have a separate test for backing image I feel like it's a bit duplicated since testing of other pods already covered in test_setting_toleration.

> 2. Some of them don't contain label `app`. e.g., engine image pods,

Ones with no `app` label will be checked. Are there pods we would like to skip with no `app` label?

shuo-wu: Please check all components. We don't mind duplicate checks in integration tests.

The logic looks weird. As long as some of the pods update the tolerations, flag `ok` will be true. It means this function won't wait for all pod update complete. Here, the for loop can be:
```
ok = True
for p in pod_list:
    if p in update_skip_list:
        continue
    if not check_tolerations_set_update():
        ok = False
    if not ok:
        break

```

BTW, the field name `n_update` is confusing as well...

innobead:
@c3y1huang is this still valid? probably @chanow816 can take over.

c3y1huang:
Not working on this, closing.

shuo-wu:
Sorry. Will we ignore/discard this PR and the issue?

PR #610, opened 2021-05-20
Update the integration tests and add a manual test for the backing image upload

shuo-wu:
longhorn/longhorn#2404

Signed-off-by: Shuo Wu <shuo.wu@suse.com>

### DIFF
--- a/docs/content/manual/release-specific/v1.2.0/test-backing-image-upload.md
+++ b/docs/content/manual/release-specific/v1.2.0/test-backing-image-upload.md
@@ -1,0 +1,14 @@
+---
+title: Test backing image upload
+---
+
+1. Prepare a large backing image file (make sure the size is greater than 1Gi and the uploading time is longer than 1 minute) in local.
+2. Click the backing image creation button in UI, choose `Upload From Local`, select the file then start upload.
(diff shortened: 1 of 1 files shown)


khushboo-rancher:
Few below doubts - 
1. On uploading the image, will the file be uploaded to all the nodes?
2. How does setting `Backing Image Cleanup Wait Interval` gonna work now? Will it delete the backing image files from all the nodes except one after the interval time passes by?

shuo-wu:
> Few below doubts -
> 
> 1. On uploading the image, will the file be uploaded to all the nodes?
> 2. How does setting `Backing Image Cleanup Wait Interval` gonna work now? Will it delete the backing image files from all the nodes except one after the interval time passes by?

1. Longhorn will pick up **one** random schedulable node to store the uploaded file
2. For the setting, it works but there is one exception for the uploaded backing image: It won't delete all ready files now. Instead, it will retain one random ready file and remove other files.

khushboo-rancher (approved):
LGTM

innobead:
@shuo-wu is this updated?

innobead (approved):
LGTM

PR #612, opened 2021-05-24
add test for use pvc to expand volume (#2422)

cclhsu:
add test for use pvc to expand volume to resolve ticket 'Application couldn't mount the volume after expanded the volume' (#2422)

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -23,6 +23,7 @@ from common import create_and_wait_pod, delete_and_wait_pod
 from common import delete_and_wait_pvc, delete_and_wait_pv
 from common import CONDITION_STATUS_FALSE, CONDITION_STATUS_TRUE
 from common import RETRY_COUNTS, RETRY_INTERVAL, RETRY_COMMAND_COUNT
+from common import DEFAULT_POD_TIMEOUT, DEFAULT_POD_INTERVAL
 from common import cleanup_volume, create_and_check_volume, create_backup
 from common import DEFAULT_VOLUME_SIZE
 from common import Gi, Mi
@@ -70,6 +71,7 @@ from common import VOLUME_FIELD_ROBUSTNESS, VOLUME_FIELD_READY
 from common import VOLUME_ROBUSTNESS_HEALTHY, VOLUME_ROBUSTNESS_FAULTED
 from common import DATA_SIZE_IN_MB_2, DATA_SIZE_IN_MB_3
 from common import wait_for_backup_to_start
+from common import SETTING_DEFAULT_LONGHORN_STATIC_SC
 
 from backupstore import backupstore_corrupt_backup_cfg_file
 from backupstore import backupstore_delete_volume_cfg_file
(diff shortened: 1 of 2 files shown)


khushboo-rancher:
We can add this in v1.1.2 as it is back ported to v1.1.2.

docs/content/manual/release-specific/v1.2.0/expand.md
```diff
@@ -0,0 +1,140 @@
...
+        Given Kubernetes cluster deployed
+        And Longhorn deployed
+        And a statefulset/deployment deployed with mounted volume
+```
+
+```Gherkin
+    Scenario: statefulset/deployment is replica is zero and expand with Longhorn API
+        Given Background conditions met
+        When scale down replicas for statefulset/deployment to zero
+            """bash
```
khushboo-rancher, line 29: In rich view, `"""bash` is appearing as it is.

manager/integration/tests/test_statefulset.py
```diff
@@ -438,3 +439,133 @@ def test_statefulset_restore(set_random_backupstore, client, core_api, storage_c
...
+
+
+def test_statefulset_volume_expand(client, core_api, storage_class, statefulset): # NOQA
+    """
+    Test that expanded volume can be mounted by a StatefulSet.
+
+    1. Create a StatefulSet with VolumeClaimTemplate and Longhorn.
+    2. Wait for pods to run.
+    3. Scale down replicas to 0
+    4. Expand volume using pvc
```
shuo-wu, line 451: This case is for testing the update HTTP API `Expand`. You need to do expansion via `v.expand()` then check if both PVC and the volume itself get updated.

shuo-wu (changes requested):

innobead:
Can you use the complete name for the test suite file? `expaned.md` is not a good name.. you can refer to existing examples.

innobead:
If your test cases can be covered in automated test cases, I don't think you need to create duplicated test cases for manual regression. 

Your test cases in expand.md seems not really focus on what you changed like why you emphaisze replica is zero or not zero? I would suggest you just need to cover below senarios.

- Successful/Failed volume expansion w/o PVC
- Successful/Failed volume expansion w/ PVC
- Also check the round value of volume size (this is what you changed as well) based on the above cases

Workload types should do not really matter if u know the behavior, because this is more like white box testing already.

innobead (changes requested):
Please update based on my suggestions.

cclhsu:
> If your test cases can be covered in automated test cases, I don't think you need to create duplicated test cases for manual regression.
> 
> Your test cases in expand.md seems not really focus on what you changed like why you emphaisze replica is zero or not zero? I would suggest you just need to cover below senarios.
> 
> * Successful/Failed volume expansion w/o PVC
> * Successful/Failed volume expansion w/ PVC
> * Also check the round value of volume size (this is what you changed as well) based on the above cases
> 
> Workload types should do not really matter if u know the behavior, because this is more like white box testing already.

Change the test cases accordingly.  The new test cases result is in https://ci.longhorn.io/job/private/job/longhorn-tests-regression/241/testReport/tests/test_basic/.

manager/integration/tests/test_basic.py
```diff
@@ -3658,3 +3660,128 @@ def test_volume_toomanysnapshots_condition(client, core_api, volume_name): # NOQ
...
+    volume.attach(hostId=self_hostId, disableFrontend=False)
+    wait_for_volume_healthy(client, volume_name)
+    volume.detach(hostId="")
+    volume = wait_for_volume_detached(client, volume_name)
+
+    client.delete(volume)
+    wait_for_volume_delete(client, volume_name)
+
+
+def test_expand_pvc(client, core_api, volume_name, pod):  # NOQA
```
shuo-wu, line 3711: `test_csi_expansion_with_size_round_up`
And you can put this test in `test_csi.py`

manager/integration/tests/test_basic.py
```diff
@@ -3658,3 +3660,128 @@ def test_volume_toomanysnapshots_condition(client, core_api, volume_name): # NOQ
...
+    """
+    test expand lhv volume
```
shuo-wu, line 3713: expression `lhv volume` is weird. `lhv` is the alias of `longhorn volume`.

manager/integration/tests/test_basic.py
```diff
@@ -3658,3 +3660,128 @@ def test_volume_toomanysnapshots_condition(client, core_api, volume_name): # NOQ
...
+
+    1. Create LHV,PV,PVC volume with size '1Gi'
+    2. attach and detach
```
shuo-wu, line 3716: Attach and mount the volume, write some data into the mount point, then detach the volume.

shuo-wu (changes requested):
1. The target of this feature is asking Longhorn to automatically expand PVC via the HTTP API `expand`. Do we need `test_expand_lhv`? cc @innobead 
2. IIRC, the expansion API (both in HTTP and CSI) won't round up the size. Do we need a manager PR to fix this? Anyway, it's great to test this here.

manager/integration/tests/test_basic.py
cclhsu: changed

manager/integration/tests/test_basic.py
cclhsu: changed

manager/integration/tests/test_basic.py
cclhsu: added

manager/integration/tests/test_basic.py
```diff
@@ -3658,3 +3660,86 @@ def test_volume_toomanysnapshots_condition(client, core_api, volume_name): # NOQ
...
+
+    volume = client.by_id_volume(volume_name)
+    assert volume.size == "2147483648"
+    volume.detach(hostId="")
+    volume = wait_for_volume_detached(client, volume_name)
+
+    self_hostId = get_self_host_id()
+    volume.attach(hostId=self_hostId, disableFrontend=False)
+    volume = wait_for_volume_healthy(client, volume_name)
+    write_volume_random_data(volume)
```
shuo-wu, line 3740: Need to verify existing data first.

Same below

shuo-wu (changes requested):

manager/integration/tests/test_basic.py
cclhsu: changed

shuo-wu (approved):

PR #615, opened 2021-05-26
fix test recurring job with backup of 2 (#2589)

cclhsu:
#### Proposed Changes ####

fix test recurring job with backup of 2 to resolve ticket '[BUG] Tests related to recurring job are failing in nightly integration test runs.' (#2589)

  - tests.test_recurring_job.test_recurring_job[nfs]
  - tests.test_recurring_job.test_recurring_job[s3]

as result of 

  - [[ENHANCEMENT] automatic sucessful backups should only retain the last snapshot #2231](https://github.com/longhorn/longhorn/issues/2231)
  - [Only keep upto 2 snapshots for recurring backup jobs #884](https://github.com/longhorn/longhorn-manager/pull/884)
 
#### Types of Changes ####

Bug Fixing

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/2589

#### Further Comments ####

None

### DIFF
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -44,7 +44,7 @@ def create_jobs1():
     # snapshot every one minute
     job_snap = {"name": "snap", "cron": "* * * * *",
                 "task": "snapshot", "retain": 2}
-    # backup every two minutes
+    # backup every two minutes, retain 1 snapshot from backup and 1 backup
     job_backup = {"name": "backup", "cron": "*/2 * * * *",
                   "task": "backup", "retain": 1}
     return [job_snap, job_backup]
@@ -110,23 +110,23 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
     2. Create a volume.
     3. Create two jobs
         1 job 1: snapshot every one minute, retain 2
-        1 job 2: backup every two minutes, retain 1
+        1 job 2: backup every two minutes, retain 1 snapshot
     4. Attach the volume.
-       Wait until the 10th second since the beginning of an even minute
-    5. Write some data. Sleep 2.5 minutes.
-       Write some data. Sleep 2.5 minutes
+       Wait until the 20th second since the beginning of an even minute
+    5. Write some data. Sleep 2 minutes.
+       Write some data. Sleep 2 minutes
     6. Verify we have 4 snapshots total
-        1. 2 snapshots, 1 backup, 1 volume-head
+        1. 2 snapshots from job_snap, 1 snapshot from job_backup, 1 volume-head
     7. Update jobs to replace the backup job
-        1. New backup job run every one minute, retain 2
-    8. Write some data. Sleep 2.5 minutes.
-       Write some data. Sleep 2.5 minutes
-    9. We should have 6 snapshots
-        1. 2 from job_snap, 1 from job_backup, 2 from job_backup2, 1
-        volume-head
-    10. Make sure there are exactly 4 completed backups.
+        1. New backup job run every one minute, retain 1 snapshot
+    8. Write some data. Sleep 2 minutes.
+       Write some data. Sleep 2 minutes
+    9. We should have 5 snapshots
+        1. 2 from job_snap, 1 snapshot from job_backup,
+           1 snapshot from job_backup2, 1 volume-head
+    10. Make sure there are exactly 5 completed backups.
         1. old backup job completed 2 backups
-        2. new backup job completed 2 backups
+        2. new backup job completed 3 backups
     11. Make sure we have no backup in progress
     """
 
@@ -136,7 +136,7 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
     |W  |   | W |   |   |W  |   | W |   |   |      (write data)
     |   S   |   S   |   |   S   |   S   |   |      (job_snap)
     |   |   B   |   B   |   |   |   |   |   |      (job_backup1)
-    |   |   |   |   |   |   B   |   B   |   |      (job_backup2)
+    |   |   |   |   |   |   B   B   B   |   |      (job_backup2)
     '''
 
     host_id = get_self_host_id()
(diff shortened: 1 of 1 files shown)


khushboo-rancher:
Do we know the reason or issue id because of which the behavior has changed in master from v1.1.0?

cclhsu:
> Do we know the reason or issue id because of which the behavior has changed in master from v1.1.0?

Yes, it is due to PR [Only keep up to 2 snapshots for recurring backup jobs #884](https://github.com/longhorn/longhorn-manager/pull/884).  Before we backup didn't clean up automatically can clog up the local storage space, and now it only keeps 2.  Notice that even when Backup job is retain 1, we will also keep the most recent 2 backup snapshot.

c3y1huang:
@cclhsu , can you check @shuo-wu [comment](https://github.com/longhorn/longhorn/issues/2231#issuecomment-839460780) and see if we still need to keep the check for the snapshot count?

cclhsu:
> @cclhsu , can you check @shuo-wu 's comment [here](https://github.com/longhorn/longhorn/issues/2231#issuecomment-839460780) and see if we still need to keep the check for the snapshot count?

Snapshot count is still relevant, since the condition backup retain of 1 we should check if we keep 2 backup snapshots. However, backup status does not reflect the number of backup we kept instead of number of backup completed.

c3y1huang:
> 
> Snapshot count is still relevant, since the condition backup retain of 1 we should check if we keep 2 backup snapshots. However, backup status does not reflect the number of backup we kept instead of number of backup completed.

Thanks, could you add some notes and reference the cases? it will help us to understand this behavior when reading the cases.

khushboo-rancher:
LGTM, @cclhsu Can you please link the test run results after the change?

cclhsu:
Test on Jenkins: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/228/testReport/tests/test_recurring_job/. Test against 
        TEST_FILE=test_recurring_job.py # test_basic.py | test_cronjob.py | test_recurring_job.py
        TEST_CASE=test_recurring_job
        TEST_CASE=test_recurring_job_in_volume_creation
        TEST_CASE=test_recurring_job_in_storageclass
        TEST_CASE=test_recurring_job_labels

khushboo-rancher (dismissed):
LGTM

manager/integration/tests/test_recurring_job.py
```diff
@@ -110,20 +110,22 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
     2. Create a volume.
     3. Create two jobs
         1 job 1: snapshot every one minute, retain 2
-        1 job 2: backup every two minutes, retain 1
+        1 job 2: backup every two minutes, retain 1 backup and 1 backup
+        snapshots
```
innobead, line 114: 1 job 2: backup every two minutes, retain 1 snapshot

manager/integration/tests/test_recurring_job.py
```diff
@@ -110,20 +110,22 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
...
     4. Attach the volume.
        Wait until the 10th second since the beginning of an even minute
     5. Write some data. Sleep 2.5 minutes.
        Write some data. Sleep 2.5 minutes
     6. Verify we have 4 snapshots total
-        1. 2 snapshots, 1 backup, 1 volume-head
+        1. 2 snapshots, 1 backup snapshot, 1 volume-head
```
innobead, line 120: 1. 2 snapshots from job_snap, 1 snapshot from job_backup, 1 volume-head

manager/integration/tests/test_recurring_job.py
```diff
@@ -110,20 +110,22 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
...
-        1. New backup job run every one minute, retain 2
+        1. New backup job run every one minute, retain 2 backup and 1
+        snapshots
     8. Write some data. Sleep 2.5 minutes.
        Write some data. Sleep 2.5 minutes
-    9. We should have 6 snapshots
-        1. 2 from job_snap, 1 from job_backup, 2 from job_backup2, 1
-        volume-head
+    9. We should have 5 snapshots
+        1. 2 from job_snap, 1 backup snapshot from job_backup,
```
innobead, line 127: 1. 2 from job_snap, 1 snapshot from job_backup,

manager/integration/tests/test_recurring_job.py
```diff
@@ -110,20 +110,22 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
...
+        1 backup snapshot from job_backup2, 1 volume-head
```
innobead, line 128: 1 snapshot from job_backup2, 1 volume-head

manager/integration/tests/test_recurring_job.py
```diff
@@ -110,20 +110,22 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
...
        Wait until the 10th second since the beginning of an even minute
     5. Write some data. Sleep 2.5 minutes.
        Write some data. Sleep 2.5 minutes
     6. Verify we have 4 snapshots total
-        1. 2 snapshots, 1 backup, 1 volume-head
+        1. 2 snapshots, 1 backup snapshot, 1 volume-head
     7. Update jobs to replace the backup job
-        1. New backup job run every one minute, retain 2
+        1. New backup job run every one minute, retain 2 backup and 1
+        snapshots
```
innobead, line 123: 1. New backup job run every one minute, retain 1 snapshot

manager/integration/tests/test_recurring_job.py
```diff
@@ -178,9 +180,10 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
     for snapshot in snapshots:
         if snapshot.removed is False:
             count += 1
-    # 2 from job_snap, 1 from job_backup, 2 from job_backup2, 1 volume-head
-    assert count == 6
+    # 2 from job_snap, 1 from job_backup, 1 from job_backup2, 1 volume-head
+    assert count == 5
 
+    time.sleep(20)
```
innobead, line 186: Why we need to have sleep here? Does the case fail before due to no sleep?

I would suggest having a retry (like 3 times and sleep in less seconds) mechanism instead of just long sleep. (Please check if we already have retry function able to reuse)

innobead (changes requested):
The test description is a little wrong, please correct based on the suggestions. Thanks.

manager/integration/tests/test_recurring_job.py
cclhsu: changed

manager/integration/tests/test_recurring_job.py
cclhsu: changed

manager/integration/tests/test_recurring_job.py
cclhsu: changed

manager/integration/tests/test_recurring_job.py
cclhsu: Changed the logic here.  The problem arises when running NFS, backup seems sometimes slow and cause count to be one less than expected.

cclhsu:
Updated test cases result in Jenkin: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/239/testReport/tests/test_recurring_job/

manager/integration/tests/test_recurring_job.py
```diff
@@ -178,24 +178,31 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
...
+        else:
+            other_backup_state += 1
+
+    assert other_backup_state == 0
+
+    # 2 completed backups from job_backup1
+    # 2 or more completed backups from job_backup2
+    # NOTE: NFS backup can be slow sometimes and error prone
+    assert complete_backup_number_1 == 2
+    assert complete_backup_number_2 >= 2
```
innobead, line 205: Why we can't assert with the exact number?

innobead (commented):
LGTM, but one thing needs to clarify.

manager/integration/tests/test_recurring_job.py
innobead: @cclhsu explained some timing issue possibly causes the latency for complete backups, so need to have a range condition check.

innobead (approved):

PR #618, opened 2021-05-31
Add manual test frontend payload when no resource updated

c3y1huang:
https://github.com/longhorn/longhorn/issues/2646
https://github.com/longhorn/longhorn/issues/2591


khushboo-rancher (changes requested):
I think the test addition in v1.1.2 should be sufficient, same tests need not be added to v1.20

c3y1huang:
> I think the test addition in v1.1.2 should be sufficient, same tests need not be added to v1.20

Addressed, manual test commit dropped for v1.2.0. Currently keeping the commit for the folder, let me know if want to have that dropped too.

khushboo-rancher (approved):
LGTM

innobead (approved):

PR #621, opened 2021-06-10
Update manual test - [v1.1.2] ws data transfer when resource not updated

c3y1huang:
https://github.com/longhorn/longhorn/issues/2672

### DIFF
--- a/docs/content/manual/release-specific/v1.1.2/full-ws-data-tranfer-when-no-updates.md
+++ b/docs/content/manual/release-specific/v1.1.2/full-ws-data-tranfer-when-no-updates.md
@@ -1,16 +1,39 @@
 ---
-title: Test Frontend Payload Data When Resource Not Updated
+title: Test Frontend Web-socket Data Transfer When Resource Not Updated
 ---
 
 ## Related issue
+https://github.com/longhorn/longhorn-manager/pull/918
 https://github.com/longhorn/longhorn/issues/2646
 https://github.com/longhorn/longhorn/issues/2591
 
-## Test Frontend Payload Data
+## Test Data Send Over Web-socket When No Resource Updated
 
 **Given** 1 PVC/Pod created.
+*And* the Pod is not writing to the mounted volume.
 
-**When** monitor traffic in frontend pod with browser inspect tool.
+**When** monitor network traffic with browser inspect tool.
 
 **Then** wait for 3 mins
-*And* should see empty data list in volume payloads. Example: https://github.com/longhorn/longhorn-manager/pull/905#issuecomment-851391374
+*And* should not see data send over web-socket when there are no updates to the resources.
+
+
+## Test Data Send Over Web-socket Resource Updated
+
+**Given** monitor network traffic with browser inspect tool.
+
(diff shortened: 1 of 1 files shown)


innobead (dismissed):

kaxing:
Test case verify with v1.1.2-head(0615) images, WS messages can be still observed for `/v1/ws/1s/volumes` after 6 minutes.

c3y1huang:
> Test case verify with v1.1.2-head(0615) images, WS messages can be still observed for `/v1/ws/1s/volumes` after 6 minutes.

There shouldn't be data transferred if there are no resource updates. Can you diff the last data payload?

c3y1huang:
> There shouldn't be data transferred if there are no resource updates. Can you diff the last data payload?

After checking the data payload in the URL provided by @kaxing , the test pod is constantly updating the volume that resulted in data transfer.

Screenshot-2021-06-15-19:26:40

kaxing:
Hi @c3y1huang , thanks for clarifying my test environment, I was using **dd** inside the target pod container that is why it keep generating changes. It might nice have some note on the test case to use "stationary pod" (not writing the volume while testing).

After retesting with the stationary pod + only 1 volume attached,
no more intense WS messaging under `/volumes`.

HAR files for the reference:
- After creating pod: [after-creating-pod.har.zip](https://github.com/longhorn/longhorn-tests/files/6655912/after-creating-pod.har.zip)
- On deleteing pod: [on-deleting-pod.har.zip](https://github.com/longhorn/longhorn-tests/files/6655913/on-deleting-pod.har.zip)

c3y1huang:
Oops @kaxing I've accidentally deleted your comment while I meant to copy, please help to add it back... sorry....

kaxing:
I think that comment is long gone :)
_
So I retested the test case with stationary pod(the one not writing /data volume), could be nice to put this as hint into the test case.

No more intense WS messaging after creating the testing pod.

And HAR files for reference:
After creating +6 more minutes: [after-creating-pod.har.zip](https://github.com/longhorn/longhorn-tests/files/6656084/after-creating-pod.har.zip)
On deleting: [on-deleting-pod.har.zip](https://github.com/longhorn/longhorn-tests/files/6656081/on-deleting-pod.har.zip)

c3y1huang:
> So I retested the test case with stationary pod(the one not writing /data volume), could be nice to put this as hint into the test case.

Thanks for adding the comment back, and this is addressed.

kaxing:
Still not able to identify the [WS Ping Frame](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers#pings_and_pongs_the_heartbeat_of_websockets), by looking into [the WS library use here](https://github.com/gorilla/websocket/blob/c3dd95aea9779669bb3daafbd84ee0530c8ce1c1/conn.go#L77), the `PingMessage` is a special Unicode character `0x9` which also not found throughout the connection log.

But by observing the connection, there is no additional WS connection when there are no actions until 6mins later.
And the test case is verified now, please approve @khushboo-rancher thank you!

innobead (approved):

PR #623, opened 2021-06-16
Fix expansion test failure due to introuduce of pvc expansion

cclhsu:
Fix expansion test failure due to introduce of pvc expansion to resolve issue 'Application couldn't mount the volume after expanded the volume'

TEST_FILE=test_basic.py
TEST_CASE=test_expansion_canceling
TEST_CASE=test_expansion_with_scheduling_failure

- master-head: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/282/testReport/tests/test_basic/
- v1.1.2-head: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/284/testReport/tests/test_basic/

Longhorn: https://github.com/longhorn/longhorn/issues/2692

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -2543,6 +2543,15 @@ def test_expansion_canceling(client, core_api, volume_name, pod):  # NOQA
     volume = wait_for_volume_detached(client, volume_name)
     volume.expand(size=EXPAND_SIZE)
     wait_for_volume_expansion(client, volume_name)
+    for i in range(RETRY_COUNTS):
+        claim = core_api.read_namespaced_persistent_volume_claim(
+            name=expansion_pvc_name, namespace='default')
+        if claim.spec.resources.requests['storage'] == EXPAND_SIZE and \
+                claim.status.capacity['storage'] == "32Mi":
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert claim.spec.resources.requests['storage'] == EXPAND_SIZE
+    assert claim.status.capacity['storage'] == "32Mi"
     volume = client.by_id_volume(volume_name)
     assert volume.state == "detached"
     assert volume.size == str(EXPAND_SIZE)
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -2743,7 +2763,7 @@ def test_expansion_with_scheduling_failure(
     test_pvc_name = "pvc-" + volume_name
     test_pod_name = "pod-" + volume_name
 
-    volume = create_and_check_volume(client, volume_name, size=str(300 * Mi))
```
shuo-wu, line 2746: Any reason of this size modification?

shuo-wu (changes requested):
You can directly reset the settings in function `reset_settings`.

And I am thinking if we need to rename or remove the setting since right now we won't use a non-existing storageClass (we call this StorageClass as static StorageClass) to link PV with PVC. Instead, we will directly using the default SC.

manager/integration/tests/test_basic.py
cclhsu: changed

cclhsu:
> You can directly reset the settings in function `reset_settings`.
> 
> And I am thinking if we need to rename or remove the setting since right now we won't use a non-existing storageClass (we call this StorageClass as static StorageClass) to link PV with PVC. Instead, we will directly using the default SC.

Since we are keeping the static class for now and for v1.1.2, we can fix the test case failure for now. And revise the test case along with https://github.com/longhorn/longhorn/issues/2723

manager/integration/tests/test_basic.py
```diff
@@ -2496,6 +2496,11 @@ def test_expansion_canceling(client, core_api, volume_name, pod):  # NOQA
     11. Validate the volume content, then check if data writing looks fine
     12. Clean up pod, PVC, and PV
     """
+    static_sc_name = "longhorn"
+    setting = client.by_id_setting(SETTING_DEFAULT_LONGHORN_STATIC_SC)
+    setting = client.update(setting, value=static_sc_name)
+    assert setting.value == static_sc_name
```
shuo-wu, line 2502: We already decided to deprecate this setting. There is no need to care this setting.
Same below

manager/integration/tests/test_basic.py
```diff
@@ -2819,7 +2838,18 @@ def test_expansion_with_scheduling_failure(
...
+            name=test_pvc_name, namespace='default')
+        if claim.spec.resources.requests['storage'] == expanded_size and \
+                claim.status.capacity['storage'] == "400":
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert claim.spec.resources.requests['storage'] == expanded_size
+    assert claim.status.capacity['storage'] == "400Mi"
     volume = client.by_id_volume(volume_name)
+    # volume.detach(hostId="")
+    volume = wait_for_volume_detached(client, volume_name)
```
shuo-wu, line 2852: It's redundant. `wait_for_volume_expansion` already verified that volume detachment.

shuo-wu (changes requested):
You can check PV size as well.

manager/integration/tests/test_basic.py
cclhsu: removed

manager/integration/tests/test_basic.py
cclhsu: changed

cclhsu:
With changes in 
- https://github.com/longhorn/longhorn-manager/pull/948
- https://github.com/longhorn/longhorn-manager/pull/951
The PR longer required.

PR #626, opened 2021-06-17
docs: Add manual test 'Test Instance Manager Streaming Connection Recovery

shuo-wu:
longhorn/longhorn#2561

### DIFF
--- a/docs/content/manual/release-specific/v1.1.2/instance-manager-streaming-connection-recovery.md
+++ b/docs/content/manual/release-specific/v1.1.2/instance-manager-streaming-connection-recovery.md
@@ -1,0 +1,38 @@
+---
+title: Test Instance Manager Streaming Connection Recovery
+---
+
+## Related issue
+https://github.com/longhorn/longhorn/issues/2561
+
+## Test Step
+**Given** A cluster with Longhorn 
+
+*And* create a volume and attach it to a pod.
+
+*And* `exec` into a longhorn manager pod and kill the connection with an engine or replica instance manager pod. The connections are instance manager pods' IP with port `8500`.
+````
+$ kl exec -it longhorn-manager-5z8zn -- bash
+
+root@longhorn-manager-5z8zn:/# ss
+Netid                    State                     Recv-Q                     Send-Q                                           Local Address:Port                                            Peer Address:Port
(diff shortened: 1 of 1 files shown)


docs/content/manual/release-specific/v1.1.2/instance-manager-streaming-connection-recovery.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Test Instance Manager Streaming Connection Recovery
+---
+
+## Related issue
+https://github.com/longhorn/longhorn/issues/2561
+
+## Test Step
+**Given** A cluster with Longhorn 
+
```
khushboo-rancher, line 10: I think, we should add a step here - `Create a volume and attach it to a pod`. Now we can try killing the connection of IM and make sure the volume comes back online healthy.

docs/content/manual/release-specific/v1.1.2/instance-manager-streaming-connection-recovery.md
```diff
@@ -0,0 +1,34 @@
...
+root@longhorn-manager-5z8zn:/# exit
+````
+
+**Then** Check the longhorn manager pod log. There must be following logs:
+```
+[longhorn-manager-5z8zn] time="2021-06-17T11:16:37Z" level=error msg="error receiving next item in engine watch: rpc error: code = Unavailable desc = transport is closing" controller=longhorn-instance-manager instance manager=instance-manager-e-285962e9 node=shuo-cluster-0-worker-3
+......
+[longhorn-manager-5z8zn] time="2021-06-17T11:16:38Z" level=error msg="instance manager monitor streaming continuously errors receiving items for 10 times, will stop the monitor itself" controller=longhorn-instance-manager instance manager=instance-manager-e-285962e9 node=shuo-cluster-0-worker-3
+```
+
```
khushboo-rancher, line 33: 1 qq, If the engine IM was disconnected, will the volume get detached and attached again? And if replica Im was disconnect, will the replica rebuild happen?

shuo-wu: Good question The answer is No. The streaming disconnection is different from node disconnection. All engines and replicas won't be affected here.

I have updated the test step to avoid the misunderstanding about the expected behavior.

jenting (approved):

innobead (approved):

PR #627, opened 2021-06-17
Add kubelet restart test cases

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/2629


innobead (changes requested):
LGTM, but please also make sure we cover RWX volume as well, because we also fix that part via soft umount added in our node plugin.

khushboo-rancher:
> LGTM, but please also make sure we cover RWX volume as well, because we also fix that part via soft umount added in our node plugin.

Thank you for pointing this out. I missed it earlier, added it now.

innobead (approved):

PR #630, opened 2021-06-21
Fix unstable test_replica_auto_balance_*

c3y1huang:
Ensures the n1 dev is created at the test pod host.

https://github.com/longhorn/longhorn/issues/2716

### DIFF
--- a/manager/integration/tests/test_scheduling.py
+++ b/manager/integration/tests/test_scheduling.py
@@ -432,11 +432,10 @@ def test_replica_auto_balance_node_least_effort(client, volume_name):  # NOQA
     And set `replica-auto-balance` to `least_effort`.
     And disable scheduling for node-2.
         disable scheduling for node-3.
-    And create a volume.
-    And attach the volume to node-1.
+    And create a volume with 6 replicas.
+    And attach the volume to self-node.
     And wait for the volume to be healthy.
     And write some data to the volume.
-    And set volume replicas to 6.
     And count replicas running on each nodes.
     And 6 replicas running on node-1.
         0 replicas running on node-2.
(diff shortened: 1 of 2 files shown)


c3y1huang:
Seems fixed by https://github.com/longhorn/longhorn/issues/2378#issuecomment-864702260. Closing.

c3y1huang:
Regression * 10 - [Build#302](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/302/testReport/tests/)

manager/integration/tests/test_scheduling.py
```diff
@@ -468,20 +467,30 @@ def test_replica_auto_balance_node_least_effort(client, volume_name):  # NOQA
...
+    for node in nodes:
+        if node.name == n_names[0]:
+            continue
+        n_names.append(node.name)
+    assert len(n_names) == 3
+    assert n_names[0] == get_self_host_id()
+    n1 = client.by_id_node(n_names[0])
+    n2 = client.by_id_node(n_names[1])
+    n3 = client.by_id_node(n_names[2])
+
```
innobead, line 481: Why we need this? It seems no different than before.

```python
n1, n2, n3 = client.list_node()
```

manager/integration/tests/test_scheduling.py
```diff
@@ -432,11 +432,10 @@ def test_replica_auto_balance_node_least_effort(client, volume_name):  # NOQA
     And set `replica-auto-balance` to `least_effort`.
     And disable scheduling for node-2.
         disable scheduling for node-3.
-    And create a volume.
+    And create a volume with 6 replicas.
```
innobead, line 435: Why we need 6 replicas instead of the default 3?

innobead (changes requested):

manager/integration/tests/test_scheduling.py
c3y1huang: `n1, n2, n3 = client.list_node()` This is the culprit of the unstable test result. This is because the case will write data from the current node, thus need to ensure n1 == get_self_host_id().

manager/integration/tests/test_scheduling.py
c3y1huang: I want to see a clear test result, if using default 3 then is unclear to see if the result is from least effort of best effort.

manager/integration/tests/test_scheduling.py
```diff
@@ -468,20 +467,30 @@ def test_replica_auto_balance_node_least_effort(client, volume_name):  # NOQA
...
+    nodes = client.list_node()
+    for node in nodes:
+        if node.name == n_names[0]:
+            continue
+        n_names.append(node.name)
+    assert len(n_names) == 3
+    assert n_names[0] == get_self_host_id()
+    n1 = client.by_id_node(n_names[0])
+    n2 = client.by_id_node(n_names[1])
+    n3 = client.by_id_node(n_names[2])
```
shuo-wu, line 480: There is no need to either check if the host id is in the node list, or claim a new variable `n1 n2 n3` for each node.
```
host_id = get_self_host_id()
nodes = client.list_node()
assert len(n_names) == 3

......

volume.attach(hostId=host_id)
```

c3y1huang: Thanks, that make sense, fixed.

c3y1huang:
Regression * 5 - [Build#304](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/304/testReport/tests/)

innobead (approved):

innobead:
@c3y1huang thank you for adding and fixing the e2e test cases during the feature implementation, this is really good. 

cc @khushboo-rancher @meldafrawi @kaxing

PR #634, opened 2021-06-23
fix test_rebuild_failure_with_intensive_data

meldafrawi:
https://github.com/longhorn/longhorn/issues/2729

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -667,10 +667,10 @@ def test_rebuild_failure_with_intensive_data(client, core_api, volume_name, csi_
     pod_name, pv_name, pvc_name, original_md5sum_1 = \
         prepare_pod_with_data_in_mb(
             client, core_api, csi_pv, pvc, pod_make, volume_name,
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -667,10 +667,10 @@ def test_rebuild_failure_with_intensive_data(client, core_api, volume_name, csi_
     pod_name, pv_name, pvc_name, original_md5sum_1 = \
         prepare_pod_with_data_in_mb(
             client, core_api, csi_pv, pvc, pod_make, volume_name,
-            data_path=data_path_1, data_size_in_mb=DATA_SIZE_IN_MB_2)
+            data_path=data_path_1, data_size_in_mb=DATA_SIZE_IN_MB_4)
```
innobead, line 670: Why changing the size of data can fix this failed case?

innobead: I believe what you changed is to avoid timing issue by a workaround. Let's see the result.

innobead (approved):

PR #638, opened 2021-06-25
Add manual test for kubectl drain nodes get stock forever

cclhsu:
Add manual test for kubectl drain nodes get stock forever to resolve issue '[BUG] kubectl drain node gets stuck forever'

Longhorn: #2673

### DIFF
--- a/docs/content/manual/pre-release/node/node-drain-deletion.md
+++ b/docs/content/manual/pre-release/node/node-drain-deletion.md
@@ -36,3 +36,24 @@ Make sure the volumes on the drained/removed node can be detached or recovered c
 
 Note: ```--ignore-daemonsets``` should be set to true to ignore some DaemonSets that exist on node such as Longhorn manager, Longhorn CSI plugin, engine image in a Longhorn deployed cluster.
 
+
+## [Test kubectl drain nodes for PVC/PV/LHV is created through Longhorn UI](https://github.com/longhorn/longhorn/issues/2673)
+**Given** 1 PVC/PV/LHV created through Longhorn UI
+_And_ LHV is not yet attached/replicated.
+
+**When** kubectl drain nodes.
+
+```bash
+NODE=centos-worker-0
+kubectl cordon ${NODE}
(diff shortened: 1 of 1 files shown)


khushboo-rancher:
Is this specific to volumes which are not attached?

We have test cases https://longhorn.github.io/longhorn-tests/manual/pre-release/node/node-drain-deletion/ related to drain. Can we add these steps in the existing one? Supporting drain functionality is important and we want test this before every release.

cclhsu:
> Is this specific to volumes which are not attached?
> 
> We have test cases https://longhorn.github.io/longhorn-tests/manual/pre-release/node/node-drain-deletion/ related to drain. Can we add these steps in the existing one? Supporting drain functionality is important and we want test this before every release.

OK, move the test case to the document.

docs/content/manual/pre-release/node/node-drain-deletion.md
```diff
@@ -36,3 +36,22 @@ Make sure the volumes on the drained/removed node can be detached or recovered c
...
+
+**When** kubectl drain nodes.
+
+```bash
+NODE=xxx-worker-0
+kubectl cordon ${NODE}
+kubectl drain --force --ignore-daemonsets --delete-emptydir-data --grace-period=10 ${NODE}
+```
+
+**Then** Success drain node and do not encounter
```
khushboo-rancher, line 52: User can encounter this error if the IM pods are not the last one to get evict, right?

innobead (dismissed):

docs/content/manual/pre-release/node/node-drain-deletion.md
innobead: For the new change, PDB will be not respected if all volumes not attached yet. So, I think the assert should be fine. @cclhsu Can you double confirm this? Thanks.

cclhsu: Yes, user might still see the message a few time, but eventually successfully drain.  I see the patch is not in 1.1.2-rc1 release.

innobead (approved):

PR #639, opened 2021-06-29
Update tests after backing image refactoring & enhancement

shuo-wu:
longhorn/longhorn#2530

### DIFF
--- a/docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
+++ b/docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
@@ -3,18 +3,18 @@ title: Backing Image on a down node
 ---
 
 1. Update the settings:
-   1. Disable Node Soft Anti-affinity.
-   2. Set Replica Replenishment Wait Interval to a relatively long value.
-2. Create a backing image:
+   1. Disable `Node Soft Anti-affinity`.
+   2. Set `Replica Replenishment Wait Interval` to a relatively long value.
+2. Create a backing image. Wait for the backing image being ready in the 1st disk.
 3. Create 2 volumes with the backing image and attach them on different nodes. Verify: 
    - the disk state map of the backing image contains the disks of all replicas, and the state is running for all disks.
-   - the backing image content is correct
+   - the backing image content is correct.
 4. Write random data to the volumes.
 5. Power off 2 nodes. One node should contain one volume engine. Verify that
-   - the related disk download state in the backing image will become `unknown`.
-   - the volume on the running node still works fine but is state `Degraded`, and the content is correct in the volume.
+   - the related disk file state in the backing image will become `Unknown`.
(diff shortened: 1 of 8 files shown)


docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
```diff
@@ -3,18 +3,18 @@ title: Backing Image on a down node
 ---
 
 1. Update the settings:
-   1. Disable Node Soft Anti-affinity.
-   2. Set Replica Replenishment Wait Interval to a relatively long value.
-2. Create a backing image:
+   1. Disable `Node Soft Anti-affinity`.
+   2. Set `Replica Replenishment Wait Interval` to a relatively long value.
+2. Create a backing image. Wait for the backing image being ready in the 1st disk.
```
khushboo-rancher, line 8: Should we expect the backing image pod to be created at this moment?

khushboo-rancher: At this point, if the node containing the backing image is powered down, will the system create another image on another node?
My understanding is if the volume is not attached to backing image, it will maintain at least 1 image on running disk.

khushboo-rancher:
Just to understand, if the backing image download fails on 1st disk, are there retries?

manager/integration/tests/test_debug.py
```diff
@@ -0,0 +1,97 @@
...
+            # check if longhorn manager port is open before calling get_client
+            for ip in mgr_ips:
+                sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+                mgr_port_open = sock.connect_ex((ip, 9500))
+
+                if mgr_port_open == 0:
+                    url = 'http://' + ip + ':9500' + '/v1/schemas'
+                    return longhorn.from_env(url=url)
+        except Exception:
+            time.sleep(1)
```
khushboo-rancher, line 97: Can you please explain in brief the purpose of this file

shuo-wu: Sorry. This file shouldn't be in this PR. My bad!

docs/content/manual/pre-release/node/backing-image-on-a-down-node.md
shuo-wu: 1. A backing image file will be downloaded then stored in a backing-image-manager pod immediately once a backing image is created, no matter if there is a volume using it.
2. For downloaded backing image, YES. Longhorn should re-download it if the node contains the only backing image file is down. But for uploaded backing image in the future, there is no way to do auto recovery

innobead:
@shuo-wu is this updated?

shuo-wu:
@innobead YES. BTW, we need to update the upgrade image after merging this PR.

innobead (approved):
LGTM

PR #645, opened 2021-07-02
Add a manual test for a simulated slow disk.

keithalucas:
I added a manual test for a simulated slow disk because I couldn't get dmsetup to work in the docker environment that the python longhorn-engine tests are run in.

https://github.com/longhorn/longhorn/issues/2206

Signed-off-by: Keith Lucas <keith.lucas@suse.com>

### DIFF
--- a/docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
+++ b/docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
@@ -13,11 +13,21 @@ This case requires the creation of a slow virtual disk with `dmsetup`.
 2. Build longhorn-engine and run it on the slow disk.
     - `make` or use a tag from docker hub in next step.
     - `docker run --privileged -v /dev:/host/dev -v /proc:/host/proc -v /mnt:/volume longhornio/longhorn-engine:<tag> launch-simple-longhorn slow-vol 10g tgt-blockdev`
-3. Perform intense I/O and verify that it doesn't fail.
+3. Perform intense I/O on the block device and verify that it doesn't fail.
     - This uses an iodepth of 16 to maximize the number of threads in tgtd.  An iodepth of 16 or more will maximize the number of threads in tgtd. `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=1280k --direct=1 --size=3000m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=16 --verify=sha256`
     - This exceeds the number of operations tgtd and longhorn can handle simultaneously. `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=16k --direct=1 --size=10m --numjobs=4 --filename=/dev/longhorn/slow-vol --iodepth=128 --verify=sha256`
     - This tests performing one I/O operation at a time: `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=4k --direct=1 --size=10m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=1 --verify=sha256`
+    - This tests with skipping 4k blocks: `fio --name=sparse-writers --ioengine=libaio --rw=write:4k --bs=4k --direct=1 --size=10m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=128 --verify=sha256`
 
+4. Perform intense I/O on a filesystem and verify that it doesn't fail.
+    - `mkfs.ext4 /dev/longhorn/slow-vol`
+    - `mkdir /slow-mnt`
+    - `mount /dev/longhorn/slow-vol /slow-mnt`
+    - Run tests similar to the block device tests:
+        - `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=1280k --direct=1 --size=3000m --numjobs=1 --filename=/slow-mnt/file.dat --iodepth=16 --verify=sha256`
+        - `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=16k --direct=1 --size=10m --numjobs=4 --filename=/slow-mnt/file.dat --iodepth=128 --verify=sha256`
+        - `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=4k --direct=1 --size=10m --numjobs=1 --filename=/slow-mnt/file.dat --iodepth=1 --verify=sha256`
+        - `fio --name=sparse-writers --ioengine=libaio --rw=write:4k --bs=4k --direct=1 --size=10m --numjobs=1 --filename=/slow-mnt/file.dat --iodepth=128 --verify=sha256`
 
 5. Repeat with different filesystems, i.e use `mkfs.xfs` instead of `mkfs.ext4`.
 
--- a/docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
+++ b/docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
@@ -1,0 +1,23 @@
+---
+title: "[#2206](https://github.com/longhorn/longhorn/issues/2206) Fix the spinning disk on Longhorn"
+---
+
+This case requires the creation of a slow virtual disk with `dmsetup`.
+1. Make a slow disk:
+    - Make a disk image file: `truncate -s 10g slow.img`
+    - Create a loopback device: `losetup --show -P -f slow.img`
+    - Get the block size of the loopback device: `blockdev --getsize /dev/loopX`
+    - Create slow device: `echo "0 <blocksize> delay /dev/loopX 0 500" | dmsetup create dm-slow`
+    - Format slow device: `mkfs.ext4 /dev/mapper/dm-slow`
+    - Mount slow device: `mount /dev/mapper/dm-slow /mnt`
+2. Build longhorn-engine and run it on the slow disk.
+    - `make` or use a tag from docker hub in next step.
+    - `docker run --privileged -v /dev:/host/dev -v /proc:/host/proc -v /mnt:/volume longhornio/longhorn-engine:<tag> launch-simple-longhorn slow-vol 10g tgt-blockdev`
+3. Perform intense I/O and verify that it doesn't fail.
+    - This uses an iodepth of 16 to maximize the number of threads in tgtd.  An iodepth of 16 or more will maximize the number of threads in tgtd. `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=1280k --direct=1 --size=3000m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=16 --verify=sha256`
+    - This exceeds the number of operations tgtd and longhorn can handle simultaneously. `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=16k --direct=1 --size=10m --numjobs=4 --filename=/dev/longhorn/slow-vol --iodepth=128 --verify=sha256`
+    - This tests performing one I/O operation at a time: `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=4k --direct=1 --size=10m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=1 --verify=sha256`
+
+
+5. Repeat with different filesystems, i.e use `mkfs.xfs` instead of `mkfs.ext4`.
+
--- a/docs/content/manual/pre-release/resiliency/timeout.md
+++ b/docs/content/manual/pre-release/resiliency/timeout.md
@@ -1,0 +1,87 @@
+---
+title: "Test timeout on loss of network connectivity"
+---
+
+## Ping Timeout
+
+1. Create a docker network:
+
+```shell
+docker network create -d bridge --subnet 192.168.22.0/24 longhorn-network
+```
+
+2. Start a replica:
+
+```shell
+docker run --net longhorn-network --ip 192.168.22.2 \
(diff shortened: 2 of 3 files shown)


docs/content/manual/pre-release/spinning-disk/simulated-slow-disk.md
```diff
@@ -0,0 +1,18 @@
...
+  - Create a loopback device: `losetup --show -P -f slow.img`
+  - Get the block size of the loopback device: `blockdev --getsize /dev/loopX`
+  - Create slow device: `echo "0 <blocksize> delay /dev/loopX 0 500" | dmsetup create dm-slow`
+  - Format slow device: `mkfs.ext4 /dev/mapper/dm-slow`
+  - Mount slow device: `mount /dev/mapper/dm-slow /mnt`
+2. Build longhorn-engine and run it on the slow disk.
+  - `make` or use a tag from docker hub in next step.
+  - `docker run --privileged -v /dev:/host/dev -v /proc:/host/proc -v /mnt:/volume longhornio/longhorn-engine:<tag> launch-simple-longhorn slow-vol 10g tgt-blockdev`
+3. Peform intense I/O and verify that it doesn't fail.
+  - `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=1280k --direct=1 --size=3000m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=16`
```
yasker, line 17: Need to try a higher iodepth and smaller block size.

docs/content/manual/pre-release/spinning-disk/_index.md
```diff
@@ -0,0 +1,3 @@
+---
+title: Spinning Disk
```
yasker, line 2: We can call it `resiliency` instead of spinning disk, which is too narrowly defined. This should also applies to the scenarios with low network or low CPU situations.

yasker (changes requested):

docs/content/manual/pre-release/spinning-disk/_index.md
keithalucas: I renamed it ``resiliency``.

docs/content/manual/pre-release/spinning-disk/simulated-slow-disk.md
keithalucas: I added a test to run with higher iodepth and smaller block size.  I also added a test with an iodepth of 1.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
```diff
@@ -0,0 +1,21 @@
...
+  - Get the block size of the loopback device: `blockdev --getsize /dev/loopX`
+  - Create slow device: `echo "0 <blocksize> delay /dev/loopX 0 500" | dmsetup create dm-slow`
+  - Format slow device: `mkfs.ext4 /dev/mapper/dm-slow`
+  - Mount slow device: `mount /dev/mapper/dm-slow /mnt`
+2. Build longhorn-engine and run it on the slow disk.
+  - `make` or use a tag from docker hub in next step.
+  - `docker run --privileged -v /dev:/host/dev -v /proc:/host/proc -v /mnt:/volume longhornio/longhorn-engine:<tag> launch-simple-longhorn slow-vol 10g tgt-blockdev`
+3. Peform intense I/O and verify that it doesn't fail.
+  - This uses an iodepth of 16 to maximize the number of threads in tgtd: `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=1280k --direct=1 --size=3000m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=16`
+  - This exceeds the number of operations tgtd and longhorn can handle simultaneously. `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=16k --direct=1 --size=10m --numjobs=4 --filename=/dev/longhorn/slow-vol --iodepth=128`
```
yasker, line 18: use `4k` for the small IO.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
```diff
@@ -0,0 +1,21 @@
...
+  - Create a loopback device: `losetup --show -P -f slow.img`
+  - Get the block size of the loopback device: `blockdev --getsize /dev/loopX`
+  - Create slow device: `echo "0 <blocksize> delay /dev/loopX 0 500" | dmsetup create dm-slow`
+  - Format slow device: `mkfs.ext4 /dev/mapper/dm-slow`
+  - Mount slow device: `mount /dev/mapper/dm-slow /mnt`
+2. Build longhorn-engine and run it on the slow disk.
+  - `make` or use a tag from docker hub in next step.
+  - `docker run --privileged -v /dev:/host/dev -v /proc:/host/proc -v /mnt:/volume longhornio/longhorn-engine:<tag> launch-simple-longhorn slow-vol 10g tgt-blockdev`
+3. Peform intense I/O and verify that it doesn't fail.
+  - This uses an iodepth of 16 to maximize the number of threads in tgtd: `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=1280k --direct=1 --size=3000m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=16`
```
yasker, line 17: In fact any IODepth more than 16 should maximize the number of threads in tgtd.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
```diff
@@ -0,0 +1,21 @@
...
+  - This exceeds the number of operations tgtd and longhorn can handle simultaneously. `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=16k --direct=1 --size=10m --numjobs=4 --filename=/dev/longhorn/slow-vol --iodepth=128`
+  - This tests performing one I/O operation at a time: `fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=16k --direct=1 --size=10m --numjobs=1 --filename=/dev/longhorn/slow-vol --iodepth=1`
+
```
yasker, line 20: Since we're removing the timeout test from the engine integration test, we also need to check the replicas are properly timed out if anything went wrong.

Also, all the tests we ran here should have the checksum/verification option to ensure the data correctness.

yasker (changes requested):

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
keithalucas: I  added the timeout tests, and I added verify to the fio commands.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
keithalucas: I change it to 4k.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
keithalucas: I noted that 16 or more with maximum the threads.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
keithalucas: I also change the timeout test to have two replicas.

docs/content/manual/pre-release/resiliency/timeout.md
```diff
@@ -0,0 +1,65 @@
...
+```shell
+fio --name=random-writers --ioengine=libaio --rw=randwrite --bs=16k \
+           --direct=1 --size=1000m --numjobs=1 \
+           --filename=/dev/longhorn/timeout-test --iodepth=1
+```
+3. Stop the network on the replica with steps 5-6 in the Ping Timeout Section.
+
+4. Note that the controller output displays ""Setting replica tcp://192.168.22.2:9502 to ERR due to: r/w timeout" after eight seconds.
+
+5. Note that the fio operation continues to work and completes without error.
```
yasker, line 65: Need to check `dmesg` to ensure there is no error from the kernel or filesystem level.

yasker:
@khushboo-rancher can you review this test plan as well?

docs/content/manual/pre-release/resiliency/timeout.md
keithalucas: I added a step to check dmesg.  This test doesn't use a filesystem on the block device.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
```diff
@@ -0,0 +1,21 @@
...
+    - Make a disk image file: `truncate -s 10g slow.img`
+    - Create a loopback device: `losetup --show -P -f slow.img`
+    - Get the block size of the loopback device: `blockdev --getsize /dev/loopX`
+    - Create slow device: `echo "0 <blocksize> delay /dev/loopX 0 500" | dmsetup create dm-slow`
+    - Format slow device: `mkfs.ext4 /dev/mapper/dm-slow`
+    - Mount slow device: `mount /dev/mapper/dm-slow /mnt`
+2. Build longhorn-engine and run it on the slow disk.
+    - `make` or use a tag from docker hub in next step.
+    - `docker run --privileged -v /dev:/host/dev -v /proc:/host/proc -v /mnt:/volume longhornio/longhorn-engine:<tag> launch-simple-longhorn slow-vol 10g tgt-blockdev`
+3. Perform intense I/O and verify that it doesn't fail.
```
khushboo-rancher, line 16: I'm not sure but do we want to add a test case here with holes in the sparse file? In the past we've seen it impacts the performance.

khushboo-rancher:
LGTM, left a small doubt comment.

docs/content/manual/pre-release/resiliency/timeout.md
yasker: Hmm? Should we add filesystem then run top of it? That's how normal user use it. And it should be able to expose more problem e.g. filesystem might be more sensitivity.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
keithalucas: I'm not sure.  The sparse file used to store the replica should have holes in it while the fio command is running.   Do you have a description of a test with the sparse file having holes impacting performance?  I'm not sure how to recreate it with this direct I/O on the block device file example.

khushboo-rancher: I used below fio job to create sparse file having holes, I assume that `--rw=write:4k` will help.
```
[global]
bs=4k
iodepth=128
direct=1
sync=0
end_fsync=1
ioengine=libaio
randseed=1
randrepeat=0
# buffer_pattern="X"
group_reporting

# 4k bs
[job1]
bs=4k
rw=write:4k
filename=/dev/longhorn/rebuild-4k-old
name=data4k-holes4k-test
size=5G
```

docs/content/manual/pre-release/resiliency/timeout.md
keithalucas: When I used fio on a file on a file system and the replica was getting timed out with old code, the journal of the last 1000 operations showed mostly the operations with the block size of the fio with periodic write operations of 768k and 1280k.  Also, running mkfs.ext4 seemed to use the largest possible block sizes; it was similar to running fio with a blocksize of 1280k.  

It will valuable to use a filesystem because will perform different sized I/O operations and it is what the end user will actually be using.  I'll add a test to create an ext4 filesystem, mount it, and use fio on it.

docs/content/manual/pre-release/resiliency/simulated-slow-disk.md
keithalucas: I added a test with --rw=write:4k

docs/content/manual/pre-release/resiliency/timeout.md
yasker: We need filesystem level check on timeout as well, in addition to block level check.

docs/content/manual/pre-release/resiliency/timeout.md
```diff
@@ -0,0 +1,67 @@
...
+
+4. In another terminal, start the controller:
+
+```shell
+docker run --net longhorn-network --ip 192.168.22.4 --privileged \
+         -v /dev:/dev -v /proc:/host/proc \
+         longhornio/longhorn-engine:<tag> \
+         longhorn controller --replica tcp://192.168.22.2:9502 \
+         --replica tcp://192.168.22.3:9502 \
+         --frontend tgt-blockdev timeout-test
```
yasker, line 37: We shouldn't use `tgt-blockdev` frontend since the logic has been moved into the instance manager.

docs/content/manual/pre-release/resiliency/timeout.md
keithalucas: Added filesystem check.

docs/content/manual/pre-release/resiliency/timeout.md
keithalucas: I'm not using the instance manager for this test.  I will wait until the instance manager being removed from longhornio/longhorn-engine is resolved to resolve this.

yasker: Sounds good

yasker (approved):
LGTM. @khushboo-rancher ?

khushboo-rancher (approved):
LGTM

PR #647, opened 2021-07-07
Remove VA related test cases

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/2745
https://github.com/longhorn/longhorn/issues/2756


innobead (approved):
LGTM

PR #648, opened 2021-07-08
Use set_random_backupstore to set S3/NFS

jenting:
Use `set_random_backupstore` test fixture to simplifies configure set S3/NFS in each test case.

- test_basic.py::backup_status_for_unavailable_replicas_test
- test_basic.py::test_backup
- test_basic.py::test_backup_labels
- test_csi.py::test_csi_backup
- test_ha.py::test_ha_backup_deletion_recovery


khushboo-rancher (approved):
LGTM

PR #658, opened 2021-07-26
Automate and manual tests for label-driven recurring job

c3y1huang:
https://github.com/longhorn/longhorn/issues/467

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -54,6 +54,7 @@ RETRY_INTERVAL = 0.5
 RETRY_INTERVAL_LONG = 2
 RETRY_BACKUP_COUNTS = 300
 RETRY_BACKUP_INTERVAL = 1
+RETRY_SNAPSHOT_INTERVAL = 1
 RETRY_EXEC_COUNTS = 30
 RETRY_EXEC_INTERVAL = 5
 
@@ -212,6 +213,8 @@ K8S_ZONE_LABEL = "topology.kubernetes.io/zone"
 
 BACKING_IMAGE_SOURCE_TYPE_DOWNLOAD = "download"
 
+JOB_LABEL = "recurring-job.longhorn.io"
+
 
 def load_k8s_config():
     c = Configuration()
@@ -297,7 +300,6 @@ def cleanup_all_volumes(client):
         except Exception as e:
             print("\nException when cleanup volume ", v)
             print(e)
-            pass
     for i in range(RETRY_COUNTS):
         volumes = client.list_volume()
         if len(volumes) == 0:
@@ -352,6 +354,16 @@ def create_backup(client, volname, data={}, labels={}):
     assert volume.lastBackupAt != ""
 
     return bv, b, snap, data
+
+
+def wait_for_backup_count(backup_volume, number, retry_counts=120):
+    ok = False
+    for _ in range(retry_counts):
+        if len(backup_volume.backupList()) == number:
+            ok = True
+            break
+        time.sleep(RETRY_BACKUP_INTERVAL)
+    assert ok
 
 
 def delete_backup(client, volume_name, backup_name):
@@ -953,6 +965,22 @@ def apps_api(request):
     return apps_api
 
 
+@pytest.fixture
+def batch_v1_beta_api(request):
+    """
+    Create a new BatchV1beta1Api instance.
+    Returns:
+        A new BatchV1beta1Api Instance.
+    """
+    c = Configuration()
+    c.assert_hostname = False
+    Configuration.set_default(c)
+    k8sconfig.load_incluster_config()
+    api = k8sclient.BatchV1beta1Api()
+
+    return api
+
+
 def get_pv_manifest(request):
     volume_name = generate_volume_name()
     pv_manifest = {
@@ -1318,6 +1346,8 @@ def cleanup_client():
     # cleanup test disks
     cleanup_test_disks(client)
 
+    cleanup_all_recurring_jobs(client)
+
     cleanup_all_volumes(client)
 
     if backing_image_feature_supported(client):
@@ -1536,7 +1566,7 @@ def wait_for_volume_delete(client, name):
 
 
 def wait_for_backup_volume_delete(client, name):
-    for _ in range(RETRY_COUNTS):
+    for _ in range(RETRY_BACKUP_COUNTS):
         bvs = client.list_backupVolume()
         found = False
         for bv in bvs:
@@ -1545,7 +1575,7 @@ def wait_for_backup_volume_delete(client, name):
                 break
         if not found:
             break
-        time.sleep(RETRY_INTERVAL)
+        time.sleep(RETRY_BACKUP_INTERVAL)
     assert not found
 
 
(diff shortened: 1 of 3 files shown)


manager/integration/tests/common.py
```diff
@@ -4039,6 +4080,27 @@ def backing_image_feature_supported(client):
...
 
+def cleanup_all_recurring_jobs(client):
+    recurring_jobs = client.list_recurring_job()
+    for recurring_job in recurring_jobs:
+        try:
+            client.delete(recurring_job)
+        except Exception as e:
+            print("\nException when cleanup recurring job ", recurring_job)
+            print(e)
+            pass
```
innobead, line 4091: Pass should be needless.

manager/integration/tests/common.py
```diff
@@ -4039,6 +4080,27 @@ def backing_image_feature_supported(client):
...
 
 
+def cleanup_all_recurring_jobs(client):
+    recurring_jobs = client.list_recurring_job()
+    for recurring_job in recurring_jobs:
+        try:
+            client.delete(recurring_job)
+        except Exception as e:
+            print("\nException when cleanup recurring job ", recurring_job)
+            print(e)
```
innobead, line 4090: Minor, but u can merge two lines as one print statement.

manager/integration/tests/common.py
```diff
@@ -4052,3 +4114,111 @@ def get_engine_image_status_value(client, ei_name):
...
+                                    Cron=spec["cron"],
+                                    Retain=spec["retain"],
+                                    Concurrency=spec["concurrency"])
+        policy = client.by_id_recurring_job(name)
+        assert policy.name == name
+        assert policy.task == spec["task"]
+        assert policy.groups == spec["groups"]
+        assert policy.cron == spec["cron"]
+        assert policy.retain == spec["retain"]
+        assert policy.concurrency == spec["concurrency"]
```
innobead, line 4133: Assert here could be optional in common libs but usually used in test cases condition check instead.

manager/integration/tests/test_recurring_job.py
```diff
@@ -759,3 +794,641 @@ def backup_complete_predicate(b):
...
+    # wait_until_begin_of_a_minute()
+    common.write_volume_random_data(volume1)
+    common.write_volume_random_data(volume2)
+    time.sleep(60)
+    wait_for_snapshot_count(volume1, 3)
+    wait_for_snapshot_count(volume2, 3)
+
+
+@pytest.mark.recurring_job  # NOQA
+def test_recurring_job_policy_backup(set_random_backupstore, client, batch_v1_beta_api):  # NOQA
```
innobead, line 1332: Probably we can remove policy keyword

manager/integration/tests/test_recurring_job.py
```diff
@@ -759,3 +794,641 @@ def backup_complete_predicate(b):
...
+            TASK: BACKUP,
+            GROUPS: [group1],
+            CRON: "* * * * *",
+            RETAIN: 1,
+            CONCURRENCY: 2
+        },
+    }
+    create_recurring_jobs(client, recurring_jobs)
+
+    volume1.recurringJobAdd(name=group1, isGroup=True)
```
innobead, line 843: Python method name should be snake case

innobead (commented):
In general, LGTM. Some existing recurring job test cases needed to modify as u mentioned by TODO.

manager/integration/tests/test_recurring_job.py
c3y1huang: This POST queried for [actions](https://github.com/longhorn/longhorn-manager/blob/master/api/router.go#L68-L95).

manager/integration/tests/common.py
c3y1huang: Without going over the line character's limit, I think the error would be easier to read on a separate line. And this is currently consistent with the others.

innobead: I mean `print("....", var, e)` or even use .format. But that's fine if most of the code has been done with this style already.

manager/integration/tests/test_recurring_job.py
innobead: Thanks :), although this does not follow up with the language convention, that's fine for now based on our own convention.

manager/integration/tests/common.py
c3y1huang: Removed.

manager/integration/tests/common.py
c3y1huang: Make sense. Splited the checking to another function.

manager/integration/tests/test_recurring_job.py
c3y1huang: Oops. Thanks, I've missed a few... fixed.

c3y1huang:
Regression Build - [398](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/398/)

c3y1huang:
Regression - [407](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/407/testReport/)

manager/integration/tests/test_recurring_job.py
```diff
@@ -4,60 +4,94 @@
...
+RETAIN = "retain"
+SNAPSHOT = "snapshot"
+BACKUP = "backup"
+CONCURRENCY = "concurrency"
+LABELS = "labels"
+DEFAULT = "default"
+SCHEDULE_1MIN = "* * * * *"
+
+
+def wait_until_begin_of_a_minute():
```
innobead, line 89: Why do we need this?

manager/integration/tests/test_recurring_job.py
```diff
@@ -658,103 +951,619 @@ def test_recurring_jobs_on_nodes_with_taints():  # NOQA
...
+
+    Given volume `test-job-1` created, attached and healthy.
+    And create `backup1`   recurring job with `default` in groups.
+        create `backup2`   recurring job with ``        in groups.
+    And `default` group exist in `test-job-1` volume recurring job label.
+    And `backup1` cron job exist.
+        `backup2` cron job exist.
+    And write data to  `test-job-1` volume.
+    And 2 snapshot exist in `test-job-1` volume.
+    And 1 backup   exist in `test-job-2` volume.
```
innobead, line 1354: There is no `given` to describe `test-job-2`? why we have this `And`?

manager/integration/tests/test_recurring_job.py
```diff
@@ -658,103 +951,619 @@ def test_recurring_jobs_on_nodes_with_taints():  # NOQA
...
+
+    When create and attach volume `test-job-2`.
+         wait for volume `test-job-2` to be healthy.
+    And `default` group exist in `test-job-2` volume recurring job label.
+    And write data to  `test-job-1` volume.
+    Then 2 snapshot exist in `test-job-1` volume.
+         1 backup   exist in `test-job-2` volume.
```
innobead, line 1361: These should be for `test-job-2`

innobead (approved):
In general, LGTM. Merging to resolve for the change of code change. 

P.S. some test description maybe needs to adjust, but @c3y1huang can update it later.

cc @khushboo-rancher @meldafrawi @kaxing

PR #661, opened 2021-08-02
Add dev scripts (scale, 2818 backup test) to test repo 

joshimoo:
In the future we can turn longhorn.py and common functionality in to a module
that can easily be imported. The scale test was moved from the longhorn/longhorn repo.

longhorn/longhorn#2818

### DIFF
--- a/manager/dev/backup-test/backup-test.py
+++ b/manager/dev/backup-test/backup-test.py
@@ -1,0 +1,116 @@
+import sys
+import logging
+
+import longhorn
+import common
+
+# env setup
+NODE_NAME = "jmoody-2818-post-work1"
+VOLUME_PREFIX = "test-1859-post"
+SNAPSHOT_NAME = "test-1859-post"
+
+# local connection forwarded to cluster
+# kubectl port-forward services/longhorn-frontend 8080:http -n longhorn-system
+LONGHORN_URL = 'http://localhost:8080/v1'  # forwarded to cluster
+
+NAMESPACE = "default"
+VOLUME_SIZE = str(16 * 1024 * 1024)
+VOLUME_COUNT = 1
+
+# we only keep 5 completed backups per replica
+# https://github.com/longhorn/longhorn-engine/blob/42283ca52d02838dfea437fa0552a2e2ee95447d/pkg/sync/rpc/server.go#L40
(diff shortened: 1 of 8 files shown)


khushboo-rancher:
This is great!

Can we add readme file for backup test as well so that anyone using it will know the longhorn url, volume count etc variables should be set before running it.

manager/dev/backup-test/backup-test.py
```diff
@@ -0,0 +1,116 @@
+import sys
+import logging
+
+import longhorn
+import common
+
+# env setup
+NODE_NAME = "jmoody-2818-post-work1"
```
khushboo-rancher, line 8: Can we use a generalized name here?

manager/dev/backup-test/backup-test.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def create_backup(volume, snapshot_name):
+    log = logging.getLogger()
+    backup = volume.snapshotBackup(name=snapshot_name)
+    log.info("volume %s created backup %s" % (volume.name, backup.id))
+    return backup
+
+
+def test_1859_backup_status():
```
khushboo-rancher, line 76: Any specific reason for this naming?

innobead:
@joshimoo could you help answer @khushboo-rancher questions ☝️ ?

joshimoo:
Not sure if this is still needed, I opened that back then so it doesn't get lost and I can link to the scale scipts as a basis for @PhanLe1010 :)

manager/dev/backup-test/backup-test.py
joshimoo: It was supposed to refer to a repro/eval of the solution for a specific issue.
But I think I used the wrong issue number, it should have been the grpc socket issue.
https://github.com/longhorn/longhorn/issues/2818

joshimoo:
For now we can close this, issue withhout merging. 
Since the scale test will be / has been taken over by @PhanLe1010 
And the socket backup issue https://github.com/longhorn/longhorn/issues/2818 is no longer relevant.

PhanLe1010:
Scale test script PR is at https://github.com/longhorn/longhorn-tests/pull/834.

PR #672, opened 2021-08-24
Fix test_backup_labels_with_backing_image

jenting:
#### Proposal Change
Because the backup volume CR changes to async pull from the remote backup target.
Add retry to get the backing image information from backup volume CR to
verify the backing image name and checksum.

#### Issue
https://github.com/longhorn/longhorn/issues/2876


jenting:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/411/

innobead (approved):
LGTM

PR #673, opened 2021-08-24
Fix test_recurring_job_labels_with_backing_image

jenting:
#### Proposal Change
The test `test_recurring_job_labels_with_backing_image` tests 2 backing image files.

When calling `recurring_job_labels_test`, clean up the recurring job setting to prevent `create_recurring_jobs` failed due to the previous recurring job still there when running the 2nd backing image file test.

#### Issues
https://github.com/longhorn/longhorn/issues/2876


c3y1huang:
Because `test_recurring_job_labels` also shares this function and it doesn't require extra cleanups. 

I prefer this fix go into `test_recurring_job_labels_with_backing_image` with the other cleanups [here](https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_backing_image.py#L357-L359).

innobead (approved):

PR #690, opened 2021-08-30
docs: Add a manual test 'cluster-using-customized-kubelet-root-directory'

shuo-wu:
Longhorn/longhorn#2746


innobead (approved):
LGTM. @kaxing Please help check and verify it does.

kaxing (approved):
Reviewed and tested on local environment.

Would be nice to point the reader to the section of ` longhorn-driver-deployer.yaml` but this should be good.
```yaml
kind: Deployment
metadata:
...
  name: longhorn-driver-deployer
          - name: KUBELET_ROOT_DIR
            value: /var/lib/longhorn-test
 ...
```

PR #692, opened 2021-08-31
Fix test_statefulset_recurring_backup

jenting:
Run the test with recurring snapshots since the backup store did not configure.
Otherwise, the orphan Backup CR would be staying in the cluster.

Note that, this PR did not fix the flakiness of this test case. But focus on fixing the orphan Backup CR.

### DIFF
--- a/manager/integration/tests/test_statefulset.py
+++ b/manager/integration/tests/test_statefulset.py
@@ -272,8 +272,7 @@ def test_statefulset_backup(set_random_backupstore, client, core_api, storage_cl
 
 
 @pytest.mark.recurring_job  # NOQA
-def test_statefulset_recurring_backup(client, core_api, storage_class,  # NOQA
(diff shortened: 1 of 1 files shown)


jenting:
Run 1 time: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/434/testReport/
Run 5 times: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/435/testReport/

innobead (dismissed):
LGTM, but would like to let @c3y1huang take a look.

c3y1huang (commented):
This is altering the original test case. I'm not sure if it matters if the backup store is set or not because the original case is only counting snapshots for backup. You try to set the backup store with set_random_backupstore.

innobead (approved):

innobead:
@jenting please backport to 1.2.x.

PR #700, opened 2021-09-09
fix secscan ubuntu update lock

meldafrawi:


innobead (approved):
This workaround is fine for now, but we probably need to have a better way later to avoid encounting waiting apt daily services down forever.

PR #705, opened 2021-09-10
Set poll interval to 0 to test backup/restore works correctly

jenting:
The backup/restore function _must_ works correctly when configuring poll interval to 0 excepts for the DR volume feature.

However, the current E2E tests did not launch 2 clusters and running the DR volume testing.
So, change the poll interval to 0 won't affect the DR volume related tests.

If we could test DR volume with 2 clusters in E2E tests in the future, we should different poll interval setting
- poll interval = 0 for normal backup/restore function
- poll interval > 0 (e.g. 300 secs) for DR volume function


innobead:
@jenting still valid and working on it?

jenting:
Close this PR because not working on this one now.

PR #709, opened 2021-09-14
docs: Add test skeleton 'test_setting_concurrent_rebuild_limit'

shuo-wu:
Longhorn/longhorn#2460

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -658,3 +658,54 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
        and all entries will be removed from the map later.
     9. Delete the backing image.
     """
+
+
+@pytest.mark.skip(reason="TODO")
+def test_setting_concurrent_rebuild_limit():  # NOQA
+    """
+    Test if setting Concurrent Replica Rebuild Per Node Limit works correctly.
+
+    The default setting value is 0, which means no limit.
+
+    Case 1 - the setting will limit the rebuilding correctly:
+    1. Set `ConcurrentReplicaRebuildPerNodeLimit` to 1.
+    2. Create 2 volumes then attach both volumes.
+    3. Write a large amount of data into both volumes,
+       so that the rebuilding will take a while.
+    4. Delete one replica for volume 1 then the replica on the same node for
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_settings.py
```diff
@@ -658,3 +658,49 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
        and all entries will be removed from the map later.
     9. Delete the backing image.
     """
+
+
+@pytest.mark.skip(reason="TODO")
+@pytest.mark.backing_image  # NOQA
```
khushboo-rancher, line 664: Why is this marked as backing_image?

manager/integration/tests/test_settings.py
```diff
@@ -658,3 +658,49 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+    5. Verify the new replica of volume 2 won't be started until volume 1
+       rebuilding complete.
+       And the new replica of volume 2 will be started immediately once
+       the 1st rebuilding is done.
+    6. Wait for rebuilding complete then repeat step 4.
+    7. Set `ConcurrentReplicaRebuildPerNodeLimit` to 0 or 2 while the volume 1
+       rebuilding is still in progress.
+       Then the new replica of volume 2 will be started immediately before
+       the 1st rebuilding is done.
+
```
khushboo-rancher, line 687: Should we add a case where the replica rebuilding is failed for the volume 1 while the setting `ConcurrentReplicaRebuildPerNodeLimit` to 1?

In this case, I'm assuming the replica rebuilding for volume 2 should start as failure case should not stop other rebuilding to progress.

manager/integration/tests/test_settings.py
shuo-wu: Sorry, typo here...

manager/integration/tests/test_settings.py
shuo-wu: Yes, good suggestion! Let me add this case to the skeleton.

innobead (approved):
LGTM

khushboo-rancher (approved):
Thanks @shuo-wu for adding the failed replica scenario.

PR #710, opened 2021-09-14
Add fsGroup test

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/2967

I've not included the check for `ls -l` as per test skeleton because writing operation and reading operation are already there in the steps. So just avoided extra line of code

Fixed some flake8 checks warnings.

Signed-off-by: Khushboo fnu.khushboo@suse.com

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -72,6 +72,7 @@ from common import DATA_SIZE_IN_MB_2, DATA_SIZE_IN_MB_3
 from common import wait_for_backup_to_start
 from common import SETTING_DEFAULT_LONGHORN_STATIC_SC
 from common import wait_for_backup_volume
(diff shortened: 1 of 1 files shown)


chanow816 (approved):
It's very clear and helpful for me to get in. thank you

Below are steps I understood from this function:
1. Deploy a pod with securityContext configured and fsGroup value specified
2. After the deployment is finished, go into the pod to execute some commands to see if the permissions of the mounted folder is working properly

innobead (approved):
LGTM.

Updating the dict is quite tedious a bit. In the future, probably we can leverage some lib to improve it. (ref example: https://pypi.org/project/dotty-dict/)

innobead:
@khushboo-rancher Please help backport to 1.2.x.

PR #716, opened 2021-09-16
integration: fix wait_for_engine_image_deletion()

shuo-wu:
There is no need to continue waiting if the deletion is done.

Before this fix, `test_engine_live_upgrade` takes 223s. After the fix, the test spends 62s only.


innobead (approved):

innobead:
@shuo-wu Please backport to 1.2.x.

PR #717, opened 2021-09-16
[Backport v1.2.x] Fix all recurring job related test failed

jenting:
https://ci.longhorn.io/job/public/job/v1.2.x/job/v1.2.x-longhorn-tests-ubuntu-amd64/2/

The recurring job `spec.groups` is an optional field https://github.com/longhorn/longhorn-manager/blob/3df7b91/types/resource.go#L190

Check the argument contains the `groups` field and then verify it's equal compares to the recurring job `spec.groups`.


innobead (approved):

innobead:
cc @longhorn/qa

PR #721, opened 2021-09-16
fix secscan: disable apt-upgrade timers for secscan provisioning and move upgrade to provision script

meldafrawi:

### DIFF
--- a/secscan/scripts/provision.sh
+++ b/secscan/scripts/provision.sh
@@ -3,7 +3,13 @@
 DOCKER_VERSION=20.10
 
 systemctl stop apt-daily.service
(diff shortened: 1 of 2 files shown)


secscan/scripts/provision.sh
```diff
@@ -17,6 +23,8 @@ do
   sleep 1;
 done
 
+apt-get update
```
yasker, line 26: We already have `apt-get update` below?

meldafrawi: fixed, Thanks

yasker (approved):

PR #724, opened 2021-09-17
Avoid waiting for instance manager during setting reset

shuo-wu:
The nodes and disks have not been reset. Then the instance manager wait may fail the test (race condition).


khushboo-rancher (approved):
LGTM

yasker (approved):

PR #736, opened 2021-10-08
Flakiness test case : test_node_config_annotation_missing (test_node.py)

chriscchien:
This test case is flakiness because during testing run time , instance-manager-r restarted and in teardown step check instance-manager-r running well.

Sometimes, test completed and entering teardown steps, the instance-manager-r still in terminating then got a new name, but teardown still use old name(terminated one), script got api response code 500 and direct raise error.

So I added try - except section in wait_for_all_instance_manager_running(), let the retry mechanism can still work.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2795,25 +2795,31 @@ def wait_for_all_instance_manager_running(client):
     for i in range(RETRY_COUNTS):
         instance_managers = client.list_instance_manager()
         node_to_engine_manager_map, node_to_replica_manager_map = {}, {}
-        for im in instance_managers:
-            if im.managerType == "engine" and im.currentState == "running":
-                node_to_engine_manager_map[im.nodeID] = im
-            elif im.managerType == "replica" and im.currentState == "running":
-                node_to_replica_manager_map[im.nodeID] = im
-            else:
-                print("\nFound unknown instance manager:", im)
-        if len(node_to_engine_manager_map) != len(nodes) or \
-                len(node_to_replica_manager_map) != len(nodes):
-            time.sleep(RETRY_INTERVAL)
-            continue
+        try:
+            for im in instance_managers:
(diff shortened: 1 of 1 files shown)


jenting (changes requested):
Could you please change your signoff-by email address to use the suse.com email? Thank u.

manager/integration/tests/common.py
```diff
@@ -2795,25 +2795,31 @@ def wait_for_all_instance_manager_running(client):
...
+                wait_for_instance_manager_desire_state(client, core_api,
+                                                       im.name, "Running",
+                                                       True)
+            for _, im in node_to_replica_manager_map.items():
+                wait_for_instance_manager_desire_state(client, core_api,
+                                                       im.name, "Running",
+                                                       True)
+            break
+        except ApiException:
+            continue
```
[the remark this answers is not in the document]
chanow816: Yes If I add time.sleep(15) before `for im in instance_managers:`, all instance-manager-r will ready for next steps

manager/integration/tests/common.py
```diff
@@ -2795,25 +2795,31 @@ def wait_for_all_instance_manager_running(client):
...
+            for _, im in node_to_engine_manager_map.items():
+                wait_for_instance_manager_desire_state(client, core_api,
+                                                       im.name, "Running",
+                                                       True)
+            for _, im in node_to_replica_manager_map.items():
+                wait_for_instance_manager_desire_state(client, core_api,
+                                                       im.name, "Running",
+                                                       True)
+            break
+        except ApiException:
```
jenting, line 2821: If we change to catch the `ApiException` error under function `wait_for_instance_manager_desire_state`, would it be better?

jenting: We should update the `node_to_engine_manager_map` since it'd change for the newly created instance/replica instance manager Pods name.

PR #744, opened 2021-10-14
Extend OS support matrix 

meldafrawi:

- [x] rocky linux
- [x] oracle linux
- [x] centos
ref: https://github.com/longhorn/longhorn/issues/3085


innobead:
LGTM, but rew questions. 

1. Did you verify them already and successfully ran the regression
2. Right now, we are use RKE1 for amd64 but K3s for arm64. I think we should change this by using K3s by default in both. Also, having capabilities to run on different distro.

meldafrawi:
1. I've verified provisioning is working, but we need to merge and start few runs of the integration tests to fix any issues that might arise.
2.created https://github.com/longhorn/longhorn/issues/3251 to track k3s and rke2 support efforts.

innobead (approved):

PR #750, opened 2021-10-21
Change except type in common:wait_for_all_instance_manager_running 

chanow816:
From #736 
Should use Exception not ApiException make the the loop wroking


innobead:
@chanow816 backport to 1.2.x if needed.

PR #765, opened 2021-10-29
Update manual test case "Test S3 backupstore in a cluster sitting behind a HTTP proxy"

kaxing:
Make the test steps simpler and add related information.
- Point the Squid proxy configuration to a comment
- For the network configuration, since it is depending on cloud provider, I am leaving this out of test case

Also, I am thinking about renaming the document to `test-s3-behind-proxy.md`. There is no interlink within the docs/ but I am not sure if changing the docs name will impact others.


chanow816 (approved):

innobead (approved):

innobead:
@kaxing please work with @khushboo-rancher later if you think some test cases should be renamed correctly.

PR #779, opened 2021-11-02
Reorganize the jobs for nightly test.

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/3200
Remove the two files for separate configuration for nfs and s3 instead decide using Jenkins parameter.

Note: Should be merged together with https://github.com/longhorn/infra/pull/63


innobead (approved):
LGTM

PR #784, opened 2021-11-03
integration: Add a test skeleton for restore volume migration

shuo-wu:
Longhorn/longhorn#3210

### DIFF
--- a/manager/integration/tests/test_migration.py
+++ b/manager/integration/tests/test_migration.py
@@ -176,6 +176,26 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
     return
 
 
+@pytest.mark.coretest  # NOQA
+@pytest.mark.migration # NOQA
+@pytest.mark.skip(reason="TODO") # NOQA
+def test_migration_with_restore_volume(clients, volume_name):  # NOQA
+    """
+    Test that a restored volume can be migrated.
+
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_migration.py
```diff
@@ -176,6 +176,26 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
...
+    Test that a restored volume can be migrated.
+
+    1. Prepare one backup.
+    2. Create a StorageClass with `migratable` being enabled and
+       `fromBackup` pointing to the above backup.
+    3. Create a new RWX migratable volume using the StorageClass.
+    2. Attach to node 1, then write some data.
+    4. Attach the volume to node 2 (migration target).
+    5. Wait for the migration ready. Verify that field
+       `volume.controllers[0].requestedBackupRestore` is empty.
```
innobead, line 193: Right now, we are using the payload of API call to check the result, but maybe we can also validate the longhorn engine resource as well.

innobead (approved):
LGTM and leave to @longhorn/qa review before merging.

manager/integration/tests/test_migration.py
shuo-wu: This is the field in the API response rather than a field in the CR.

chanow816 (approved):

PR #785, opened 2021-11-05
Fix flaky case : test_node_eviction_soft_anti_affinity

chanow816:
Origin from [tests.test_node.test_node_eviction_soft_anti_affinity](https://ci.longhorn.io/job/public/job/master/job/ubuntu/job/amd64/job/longhorn-tests-ubuntu-amd64-nfs/12/testReport/junit/tests/test_node/test_node_eviction_soft_anti_affinity/) and can reproduce on my local environment.

All fail happened at step 13, when replicas stop at node 2 and start to schedule on node 1 and node 3, at very few rate volume.robustness become **False** make script assert error and flaky.

In my observation, this volume.robustness become **False** behavior stay at few seconds and did not really effect replica schedule on node 1 and node 3, so I would like to change wait_for_replica_scheduled in common.py let the for loop work on volume and check in the neat end step.
```
    1. Disable scheduling on node 3.
    2. Create pv, pvc, pod with volume of 2 replicas.
    3. Write some data and get the checksum.
    7. Set 'Eviction Requested' to 'true' and disable scheduling on node 1.
    8. Set 'Replica Node Level Soft Anti-Affinity' to 'true'.
    9. Check volume 'healthy' and wait for replicas running on node 2
    Case #2: node 2 to node 1, 3 eviction
    10. Enable scheduling on node 1 and 3.
    11. Set 'Replica Node Level Soft Anti-Affinity' to 'false'.
    12. Set 'Eviction Requested' to 'true' and disable scheduling on node 2.
    13. Check volume 'healthy' and wait for replicas running on node 1 and 3.
    14. Check volume data checksum.
```

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -2485,7 +2485,8 @@ def test_node_eviction_soft_anti_affinity(client, core_api, csi_pv, pvc, pod_mak
     common.wait_for_volume_healthy(client, volume_name)
 
     common.wait_for_replica_scheduled(client, volume_name,
-                                      to_nodes=[node1.name, node3.name])
+                                      to_nodes=[node1.name, node3.name],
+                                      chk_vol_healthy=False)
 
     expect_md5sum = get_pod_data_md5sum(core_api, pod_name, data_path)
     assert expect_md5sum == created_md5sum


innobead (changes requested):
I would suggest to use `wait_for_volume_healthy` before the flaky step 13 instead of modifying `wait_for_replica_scheduled` because you will change the original pre-condition of `wait_for_replica_scheduled` which is the volume is healy.

chanow816:
Hi @innobead, 
In our current design, there already exist `wait_for_volume_healthy` before `wait_for_replica_scheduled` just like you mentioned.
To avoid modify `wait_for_replica_scheduled` , I add argument `chk_vol_healthy=False` when calling it.
Because we checked volume healthy before, and in the end we still check replicas running on node 1 and node 3.

innobead (approved):
LGTM, it's ok for now, but it is still a little confusing why the volume is not healthy in a short time sometimes.

innobead:
@chanow816 Please backport to 1.2.x and 1.1.x.

PR #786, opened 2021-11-06
Fix backup store type handling

khushboo-rancher:
The variable `backup_store_type` is not available directly in the test from the Jenkins, so updating the test.yaml file while the Longhorn setup takes place based on Jenkins variable.


innobead (approved):
LGTM.

PR #791, opened 2021-11-09
Install longhorn-nfs before all rwx cases

chanow816:
Our RWX caes were flaky and often failed at early test step related to statefulset.
We think this may due to nfs-common not installed well on all nodes
This PR create an fixture to install longhorn-nfs before each RWX cases to see if the flaky will improved.

Task for tracking rwx flaky cases: https://github.com/longhorn/longhorn/issues/2748

cc @khushboo-rancher

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4573,8 +4573,8 @@ def nfs(request):
     subprocess.check_output(cmd)
 
     node_list = api.list_node()
-    cmd = ["kubectl get pod -l app=longhorn-nfs-installation | "
-           "awk 'NR>1 {print $1}'"]
+    cmd = ["kubectl get pod -l app={} | "
+           "awk 'NR>1 {{print $1}}'".format(LONGHORN_NFS_DAEMONSET_NAME)]
     for i in range(RETRY_COUNTS):
         nfs_pods = {}
         pod_list = \
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -215,12 +215,6 @@ BACKING_IMAGE_SOURCE_TYPE_DOWNLOAD = "download"
 
 JOB_LABEL = "recurring-job.longhorn.io"
 
-LONGHORN_NFS_INSTALLATION_URL = \
-    "https://raw.githubusercontent.com/longhorn/" \
-    "longhorn/master/deploy/prerequisite/longhorn" \
-    "-nfs-installation.yaml"
-LONGHORN_NFS_DAEMONSET_NAME = "longhorn-nfs-installation"
-
 
 def load_k8s_config():
     c = Configuration()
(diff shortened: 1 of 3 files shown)


manager/integration/tests/common.py
```diff
@@ -4555,3 +4561,42 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
...
+    for i in range(RETRY_COUNTS):
+        nfs_pods = {}
+        pod_list = api.list_namespaced_pod("default")
+
+        for pod in pod_list.items:
+            pod_name = pod.metadata.name
+            if "longhorn-nfs" in pod_name:
+                nfs_pods[pod_name] = \
+                    api.read_namespaced_pod(name=pod_name,
+                                            namespace='default').status.phase
```
jenting, line 4585: Probably you could consider using the label selector to filter longhorn-nfs Pods. For example
```shell
kubectl get pod -l app=longhorn-nfs-installation
```

khushboo-rancher (changes requested):
Can you please use fixture with module scope so that at the beginning of the tests, only once nfs-common can be installed and removed after all the test cases have ran.
`@pytest.fixture(scope="module", autouse="True")`

innobead:
Could we have this kind of setup once instead of running in each nfs test case via fixture? It seems repeated applying install manifest even the package is already installed after the first applying. WDYT?

Do we have global setup before running test suites/cases? Also what's the cause the nfs package sometimes can't be installed via terraform?

manager/integration/tests/test_rwx.py
```diff
@@ -30,7 +30,7 @@ def write_data_into_pod(pod_name_and_data_path):
                                  DATA_SIZE_IN_MB_3)
 
 
-def test_rwx_with_statefulset_multi_pods(core_api, statefulset):  # NOQA
+def test_rwx_with_statefulset_multi_pods(nfs, core_api, statefulset):  # NOQA
```
khushboo-rancher, line 33: If you're using the fixture, no need to call it for single test. Module here is all the tests in the file test_rwx, so the fixture with scope module will run automatically before starting of any test in this file.

manager/integration/tests/test_rwx.py
```diff
@@ -96,7 +96,7 @@ def test_rwx_with_statefulset_multi_pods(core_api, statefulset):  # NOQA
         assert pod_data == md5sum_pod[i]
 
 
-def test_rwx_multi_statefulset_with_same_pvc(core_api, pvc, statefulset, pod):  # NOQA
+def test_rwx_multi_statefulset_with_same_pvc(nfs, core_api, pvc, statefulset, pod):  # NOQA
```
khushboo-rancher, line 99: Same as above comment.

manager/integration/tests/test_rwx.py
```diff
@@ -159,7 +159,7 @@ def test_rwx_multi_statefulset_with_same_pvc(core_api, pvc, statefulset, pod):
         core_api, command2, share_manager_name, LONGHORN_NAMESPACE)
 
 
-def test_rwx_parallel_writing(core_api, statefulset, pod):  # NOQA
+def test_rwx_parallel_writing(nfs, core_api, statefulset, pod):  # NOQA
```
khushboo-rancher, line 162: Same as above comment.

manager/integration/tests/test_rwx.py
```diff
@@ -224,7 +224,7 @@ def test_rwx_parallel_writing(core_api, statefulset, pod):  # NOQA
     assert md5sum2 == share_manager_data2
 
 
-def test_rwx_statefulset_scale_down_up(core_api, statefulset):  # NOQA
+def test_rwx_statefulset_scale_down_up(nfs, core_api, statefulset):  # NOQA
```
khushboo-rancher, line 227: Same as above and applicable for all the below tests as well.

khushboo-rancher (changes requested):

manager/integration/tests/test_rwx.py
```diff
@@ -15,7 +15,7 @@
 from common import wait_delete_pod, wait_for_pod_remount
 from common import get_core_api_client, write_pod_volume_random_data
 from common import create_pvc_spec, make_deployment_with_pvc  # NOQA
-from common import core_api, statefulset, pvc, pod, client  # NOQA
+from common import core_api, statefulset, pvc, pod, client, nfs  # NOQA
```
jenting, line 18: This line change is no more needed.

manager/integration/tests/common.py
```diff
@@ -4555,3 +4561,43 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
...
+
+        if len(nfs_pods) == len(node_list.items):
+            if all(value == "Running" for value in nfs_pods.values()):
+                break
+
+        time.sleep(RETRY_INTERVAL)
+
+    if not \
+            all(value == "Running" for value in nfs_pods.values()) \
+            or len(nfs_pods) == 0:
```
jenting, line 4596: Should it be?
```python
or len(nfs_pods) == len(node_list.items):
```

manager/integration/tests/test_rwx.py
chanow816: Hello @khushboo-rancher, I kept import fixture nfs and removed all nfs function call.

manager/integration/tests/common.py
chanow816: I modified to `or not len(nfs_pods) == len(node_list.items)`

chanow816:
Hi @innobead ,

I modified code, now longhorn-nfs will just install one time before module test_rwx started.

When creating nodes, terraform have command to install nfs-common but I didn't found code to check installation status. `def nfs()` will check longhorn-nfs install successfuly on all nodes. thank you.

innobead (approved):
LGTM

jenting (approved):
LGTM

innobead:
@chanow816 please backport to 1.2.x or even 1.1.3.

PR #794, opened 2021-11-12
Recurring job start just after minute start

chanow816:
From https://github.com/longhorn/longhorn/issues/3233, we have some flaky recurring job cases in daily run.

Most cases fail reason was timing issue. Longhorn recurring job started at  beginning of a minute, but our test cases will set recurring job directly (maybe at 55th second), then do some test steps then do sleep **whole** minute and check result in the end

Modified `create_recurring_jobs` and reduce sleep time in test cases.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -20,6 +20,7 @@ from kubernetes.client import Configuration
 from kubernetes.stream import stream
 
 from kubernetes.client.rest import ApiException
+from datetime import datetime
 
 Ki = 1024
 Mi = (1024 * 1024)
@@ -4326,6 +4327,12 @@ def update_setting(client, name, value):
 
 
 def create_recurring_jobs(client, recurring_jobs):
+    for i in range(60):
+        current_time = datetime.utcnow()
+        if current_time.second == 0:
+            break
+        time.sleep(1)
+
     for name, spec in recurring_jobs.items():
         client.create_recurring_job(Name=name,
                                     Task=spec["task"],
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4327,11 +4327,7 @@ def update_setting(client, name, value):
 
 
 def create_recurring_jobs(client, recurring_jobs):
-    for i in range(60):
-        current_time = datetime.utcnow()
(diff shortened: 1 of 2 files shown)


jenting:
>  Longhorn recurring job started at beginning of a minute, but our test cases will set recurring job directly 

Nice catch!

manager/integration/tests/common.py
```diff
@@ -4326,6 +4327,12 @@ def update_setting(client, name, value):
 
 
 def create_recurring_jobs(client, recurring_jobs):
+    for i in range(60):
```
innobead, line 4330: Why not just wait for the rest of seconds in a minute?

Something like

```python
time.sleep(60 - datetime.utcnow().second)
```

manager/integration/tests/test_recurring_job.py
```diff
@@ -421,7 +421,7 @@ def test_recurring_job_in_storageclass(set_random_backupstore, client, core_api,
     for pod_name in pod_names:
         write_pod_volume_random_data(core_api, pod_name, volume_data_path, 2)
 
-    time.sleep(150)  # 2.5 minutes
+    time.sleep(140)  # 2.5 minutes - 10 secends
```
innobead, line 424: If you already made the change in create_recurring_jobs, why need to change the sleep time? (including the all changes in this module)

innobead (changes requested):

innobead:
> > Longhorn recurring job started at beginning of a minute, but our test cases will set recurring job directly
> 
> Nice catch!

Indeed 👍

chanow816:
Hi @innobead ,

The most flaky scenario was we set recurring job retain count to 2, scheduled each minute, after 1 whole minute passed we check backup should equal 1, however somehow the back count equal 2 (Because we checked too late, second retain already completed), this caused by:
- Actions between set recurring job and check backup count need time.
- We set recurring at end of minutes , like 55th second 

So I try to set recurrent job just after minute start and reduce sleep time in each flaky case, thank you.

innobead (approved):

innobead:
@chanow816 plz backport to 1.2.x and 1.1.3. Thanks.

PR #806, opened 2021-11-17
docs: Add HA volume migration tests

shuo-wu:
Longhorn/longhorn#3206

### DIFF
--- a/docs/content/manual/pre-release/ha/ha-volume-migration.md
+++ b/docs/content/manual/pre-release/ha/ha-volume-migration.md
@@ -1,0 +1,31 @@
+---
+title: HA Volume Migration
+---
+
+1. Create a migratable volume:
+    1. Deploy a migratable StorageClass. e.g., https://github.com/longhorn/longhorn/blob/master/examples/rwx/storageclass-migratable.yaml
+    2. Create a PVC with access mode `ReadWriteMany` via this StorageClass.
+2. Attach a volume to a node and wait for volume running. Then write some data into the volume. Here I would recommend directly restoring a volume (set `fromBackup` in the StorageClass) and attach it instead.
+3. Start the migration by request attaching to another node for the volume.
+4. Trigger the following scenarios then confirm or rollback the migration:
+
+   --------------------------
+   | **#** | **Scenario**          | **Migration Confirmation** | **Migration Rollback** |
+   | ----- | --------------------- | -------------------------- | ---------------------- |
+   | 1     | New engine crash      | If the request is successfully sent before the migration engine & replicas being removed, the volume should be auto reattached after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 2     | All new replica crash | If the request is successfully sent before the migration engine & replicas being removed, the volume should keep running (then may start rebuilding) after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 3     | Old engine crash      | If the request is successfully sent before the volume becoming Faulted, the volume should keep running after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 4     | All old replica crash | If the request is successfully sent before the volume becoming Faulted, the volume should keep running (then may start rebuilding) after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 5     | Degraded volume       | Should succeed | Should succeed |
+   | 6     | One replica node down | Should succeed | Should succeed |
+   | 7     | New engine node down  | The request should be rejected | Should succeed |
+   | 8     | Old engine node down  | The request should be rejected | The request should be accepted. But the volume should keep state Unknown after the rollback since it's on a down node. |
+   
+   Note:
+    1. Scenario 1-5 can be handled by integration test. But we need to manually test them before the the implementation.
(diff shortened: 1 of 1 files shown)


innobead (dismissed):
LGTM

docs/content/manual/pre-release/ha/ha-volume-migration.md
```diff
@@ -0,0 +1,31 @@
+---
+title: HA Volume Migration
+---
+
+1. Create a migratable volume:
+    1. Deploy a migratable StorageClass. e.g., https://github.com/longhorn/longhorn/blob/master/examples/rwx/storageclass-migratable.yaml
+    2. Create a PVC with access mode `ReadWriteMany` via this StorageClass.
+2. Attach a volume to a node and wait for volume running. Then write some data into the volume. Here I would recommend directly restoring a volume (set `fromBackup` in the StorageClass) and attach it instead.
+3. Start the migration by request attaching to another node for the volume.
+4. Trigger the following scenarios then onfirm or rollback the migration:
```
innobead, line 10:
```suggestion
4. Trigger the following scenarios then confirm or rollback the migration:
```

docs/content/manual/pre-release/ha/ha-volume-migration.md
```diff
@@ -0,0 +1,31 @@
...
+   | 3     | Old engine crash      | If the request is successfully sent before the volume becoming Faulted, the volume should keep running after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 4     | All old replica crash | If the request is successfully sent before the volume becoming Faulted, the volume should keep running (then may start rebuilding) after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 5     | Degraded volume       | Should succeed | Should succeed |
+   | 6     | One replica node down | Should succeed | Should succeed |
+   | 7     | New engine node down  | The request should be rejected | Should succeed |
+   | 8     | Old engine node down  | The request should be rejected | The request should be accepted. But the volume should keep state Unknown after the rollback since it's on a down node. |
```
khushboo-rancher, line 22: Will this be same as old engine crash, if we are trying to power down the node after the migration is triggered?
If yes, the expectation for migration confirmation should also be `the request will be rejected until the volume is reattached and restarts the migration components`?

docs/content/manual/pre-release/ha/ha-volume-migration.md
```diff
@@ -0,0 +1,31 @@
...
+   | 2     | All new replica crash | If the request is successfully sent before the migration engine & replicas being removed, the volume should keep running (then may start rebuilding) after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 3     | Old engine crash      | If the request is successfully sent before the volume becoming Faulted, the volume should keep running after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
+   | 4     | All old replica crash | If the request is successfully sent before the volume becoming Faulted, the volume should keep running (then may start rebuilding) after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
```
khushboo-rancher, line 18: If Auto-salvage is disabled, the volume will remain in faulted, right?

kaxing (approved):

docs/content/manual/pre-release/ha/ha-volume-migration.md
```diff
@@ -0,0 +1,31 @@
...
+   | 5     | Degraded volume       | Should succeed | Should succeed |
+   | 6     | One replica node down | Should succeed | Should succeed |
+   | 7     | New engine node down  | The request should be rejected | Should succeed |
+   | 8     | Old engine node down  | The request should be rejected | The request should be accepted. But the volume should keep state Unknown after the rollback since it's on a down node. |
+
```
khushboo-rancher, line 23: Can there be a case where disk space is not available for the new replicas?

khushboo-rancher (commented):
LGTM in general, left some doubts.

docs/content/manual/pre-release/ha/ha-volume-migration.md
shuo-wu: No. In the old engine crash case, the volume will be reattached automatically and immediately. But for the node down case, the volume keeps state Unknown then there is no chance to confirm the migration. The node reboot case should be similar to the engine crash hence I didn't add it in this test plan.

docs/content/manual/pre-release/ha/ha-volume-migration.md
shuo-wu: Yes. But I didn't test this case, can you help check it?

docs/content/manual/pre-release/ha/ha-volume-migration.md
shuo-wu: NO. All migration replica CRs are directly cloned from the old replicas without regular scheduling progress. And each migration replica will directly use the old replica data directory rather than copying all data.

docs/content/manual/pre-release/ha/ha-volume-migration.md
khushboo-rancher: Yes, Checked it. If auto-salvage is disabled, the volume will remain in faulted state.

khushboo-rancher (approved):

chanow816 (approved):

PR #807, opened 2021-11-17
Check longhorn-nfs installed on worker node

chanow816:
From https://github.com/longhorn/longhorn/issues/2748

longhorn-nfs install failed in [nightly build](https://ci.longhorn.io/job/public/job/v1.2.x/job/v1.2.x-lh-tests-ubuntu-amd64/48/testReport/junit/tests/test_rwx/test_rwx_with_statefulset_multi_pods/), because `node_list = api.list_node()` will get non-worker node(we have 1 non-worker node in nightly build), but longhoron-nfs only install on worker node make `if len(nfs_pods) == len(node_list)` raise exception

### DIFF
--- a/manager/integration/tests/test_rwx.py
+++ b/manager/integration/tests/test_rwx.py
@@ -41,40 +41,14 @@ def write_data_into_pod(pod_name_and_data_path):
 @pytest.fixture(scope="module", autouse="True")
 def nfs(request):
 
-    api = get_core_api_client()
     apps_api = get_apps_api_client()
 
     cmd = ["kubectl", "apply", "-f", LONGHORN_NFS_INSTALLATION_URL]
     subprocess.check_output(cmd)
 
-    worker_cnt = 0
-    node_list = api.list_node()
-    for node in node_list.items:
-        if "csi.volume.kubernetes.io/nodeid" in node.metadata.annotations:
-            worker_cnt = worker_cnt + 1
-
-    cmd = ["kubectl get pod -l app={} | "
-           "awk 'NR>1 {{print $1}}'".format(LONGHORN_NFS_DAEMONSET_NAME)]
-    for i in range(RETRY_COUNTS):
-        nfs_pods = {}
-        pod_list = \
-            subprocess.check_output(cmd, shell=True).decode("utf-8").split()
-
-        for pod_name in pod_list:
-            nfs_pods[pod_name] = \
-                api.read_namespaced_pod(name=pod_name,
-                                        namespace='default').status.phase
-
-        if len(nfs_pods) == worker_cnt:
-            if all(value == "Running" for value in nfs_pods.values()):
-                break
-
-        time.sleep(RETRY_INTERVAL)
-
-    if not \
-            all(value == "Running" for value in nfs_pods.values()) \
-            or not len(nfs_pods) == worker_cnt:
-        raise Exception("Longhorn-nfs not installed on all nodes")
+    cmd = ["kubectl", "rollout", "status",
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_rwx.py
```diff
@@ -47,7 +47,12 @@ def nfs(request):
     cmd = ["kubectl", "apply", "-f", LONGHORN_NFS_INSTALLATION_URL]
     subprocess.check_output(cmd)
 
+    worker_cnt = 0
     node_list = api.list_node()
+    for node in node_list.items:
+        if "csi.volume.kubernetes.io/nodeid" in node.metadata.annotations:
```
jenting, line 53: consider using the label `node-role.kubernetes.io/master=true` to check whether it's a control plane node or not.

innobead:
> Signed-off-by: Chris Chien [chris.chien@suse.com](mailto:chris.chien@suse.com) From [longhorn/longhorn#2748](https://github.com/longhorn/longhorn/issues/2748)
> 
> longhorn-nfs install failed in [nightly build](https://ci.longhorn.io/job/public/job/v1.2.x/job/v1.2.x-lh-tests-ubuntu-amd64/48/testReport/junit/tests/test_rwx/test_rwx_with_statefulset_multi_pods/), because `node_list = api.list_node()` will get non-worker node(we have 1 non-worker node in nightly build), but longhoron-nfs only install on worker node make `if len(nfs_pods) == len(node_list)` raise exception

> node_list = api.list_node()

IIRC, `node_list = api.list_node()` should return the nodes recognized by longhorn, so it should be the list of `lhn` resources. Those nodes don't mean `non-worker node`.

I think the root cause is probably you run out the retry, then the number of running is not matched. Please check again and also simplify the code by using `kubectl rollout status ds/longhorn-nfs-installation --timeout=5m` (blindly waiting is always not efficient) or find out the real root cause of why the package is sometimes not installed correctly when provisioning nodes.

BTW, we are running different distros on different arches, so maybe this makes you confusing. 

- amd64: we deploy rke1, the master nodes of which are trained by default, so that means the NFS install ds pod will not run there, but worker nodes only.
- arm64: we deploy k3s, the master nodes are not tainted, so that means NFS install ds pod run on all types of nodes

The change you made

innobead (changes requested):
as I said, please help check again and come out with the fix.

manager/integration/tests/test_rwx.py
```diff
@@ -41,40 +41,14 @@ def write_data_into_pod(pod_name_and_data_path):
...
-    if not \
-            all(value == "Running" for value in nfs_pods.values()) \
-            or not len(nfs_pods) == worker_cnt:
-        raise Exception("Longhorn-nfs not installed on all nodes")
+    cmd = ["kubectl", "rollout", "status",
+           "ds/{}".format(LONGHORN_NFS_DAEMONSET_NAME), "--timeout=5m"]
+    subprocess.check_output(cmd)
 
     def finalizer():
         apps_api.delete_namespaced_daemon_set(name=LONGHORN_NFS_DAEMONSET_NAME,
```
jenting, line 54: nit: we could consider deleting the DaemonSet by `kubectl delete -f LONGHORN_NFS_INSTALLATION_URL`.

jenting (approved):

chanow816:
Hi @innobead ,

Thanks for your suggestion, use daemonset to check longhorn-nfs status is really more efficient than previous design.
And I verified new submit on both rke and k3s, thank you.

jenting (approved):

innobead (approved):

innobead:
@chanow816 Please help backport this to 1.2.x and 1.1.3. Thanks.

PR #811, opened 2021-11-19
Update the assert check message

jenting:
Since the longhorn manager changes the error message for the invalid
backup target setting, update the assert check message to address it.

https://github.com/longhorn/longhorn/issues/3280


jenting:
Not sure if there is any better way to tackle it.

innobead (approved):

innobead (changes requested):
Can we just keep `name+" with invalid` only in the assert? 

How about this `name+" with invalid value "`?

jenting:
> Can we just keep `name+" with invalid` only in the assert?

How about this `name+" with invalid value "`?

innobead (approved):

innobead:
@jenting need to backport?

jenting:
Yes

khushboo-rancher (commented):
LGTM

PR #823, opened 2021-11-25
Fix test_backup_status_for_unavailable_replicas

jenting:
#### Proposal Change

The behavior changes after longhorn 3035.

Previous, the volume's backupStatus reflects the engine's backup status directly.
So, even the backup is completed, if we delete the corresponding replica, the
volume's backupStatus shows the engine's backup status with an error message.

Now, after longhorn 3035 was introduced, the volume's backupStatus reflects
the Backup CR status, not the engine's backup status anymore.
So, after the backup is completed, even we delete the corresponding replica,
the volume's backupStatus shows the Backup state is Completed.

We change the test case to during the backup creation in progress,
deletes the responsible replica to make the backup falls into Error
state and with an error message.

#### Issue

https://github.com/longhorn/longhorn/issues/3035

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -507,15 +507,8 @@ def test_backup_status_for_unavailable_replicas(set_random_backupstore, client, 
 
     Context:
 
-    We want to make sure that we do not try to retrieve the backup status
-    of no longer valid replicas (offline, deleted, etc). The reason for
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead:
@jenting after merging, need to backport to 1.2.x.

chanow816 (approved):

PR #832, opened 2021-12-02
docs: Add backing image related scalability test

shuo-wu:
Longhorn/longhorn#3242

ref: https://github.com/longhorn/longhorn/issues/3242#issuecomment-984409310

### DIFF
--- a/docs/content/manual/release-specific/v1.2.3/test-scalability-with-backing-image.md
+++ b/docs/content/manual/release-specific/v1.2.3/test-scalability-with-backing-image.md
@@ -1,0 +1,337 @@
+---
+title: Test scalability with backing image 
+---
+
+### Test step
+1. Deploy a cluster with 3 worker nodes. The recommended nodes is 4v cores CPU + 8G memory at least.
+2. Deploy Longhorn.
+3. Launch 10 backing images with the following YAML:
+    ```yaml
+    apiVersion: longhorn.io/v1beta1
+    kind: BackingImage
+    metadata:
+      name: bi-test1
+      namespace: longhorn-system
+    spec:
+      sourceType: download
+      sourceParameters:
+        url: https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.qcow2
+    ---
+    apiVersion: longhorn.io/v1beta1
+    kind: BackingImage
+    metadata:
+      name: bi-test2
+      namespace: longhorn-system
+    spec:
+      sourceType: download
+      sourceParameters:
+        url: https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.qcow2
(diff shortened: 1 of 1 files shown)


PhanLe1010:
One quick question, why do you create 10 backing images but only use 5 in the SCs? What is the intention?

shuo-wu:
> One quick question, why do you create 10 backing images but only use 5 in the SCs? What is the intention?

Before the fix, the extra backing images will lead to replica controller enqueuing regardless of it is not used. Hence I want to test if the unused backing images still affect the enqueuing.

docs/content/manual/release-specific/v1.2.3/test-scalability-with-backing-image.md
```diff
@@ -0,0 +1,338 @@
...
+            requests:
+              storage: 32Mi
+    ```
+    
+   Notice:  
+     - Each replica/pod of this StatefulSet contains 10 volumes. You just adjust the volume count by scaling this StatefulSet.
+     - Launching this StatefulSet will take a long time.
+5. Wait for all pods up and running.
+6. Create and attach a volume via UI. This should not take a long time.
+7. Upgrade the longhorn manager version to `v1.2.3-rc1`. Notice that each longhorn manager pod will take some time to walk through the ugprade path. Please do not operate Longhorn before each manager pod printing out the log like `"Debug Server listening on 127.0.0.1:6060" node=shuo-k8s-worker1`.
```
innobead, line 336: For step7 to 8, I think we don't need to include in this test case, but just have a note this test case will fail in the previous version (< 1.2.3 or 1.3.0 if the issue does not be fixed in 1.2.3)

innobead (changes requested):
LGTM, but few comments needed to resolve and also what Phan asked is suggested to explain in this test case well to give more context for QA.

PhanLe1010 (approved):

PhanLe1010 (dismissed):
LGTM

docs/content/manual/release-specific/v1.2.3/test-scalability-with-backing-image.md
shuo-wu: Got it. I will remove these 2 steps then add more comments for explanation.

innobead (approved):

kaxing:
@shuo-wu, thanks for the clear steps and context!
One idea, can we add those manifests text block into actual yaml file? It will also make deploying the workload easier.

shuo-wu:
> @shuo-wu, thanks for the clear steps and context! One idea, can we add those manifests text block into actual yaml file? It will also make deploying the workload easier.

I am not sure if it would affect the directory structure. If NO, YES we can.

PR #836, opened 2021-12-21
Add a test for the new cluster restore instruction 

shuo-wu:
https://github.com/longhorn/longhorn/issues/3367


chanow816 (dismissed):

docs/content/manual/pre-release/cluster-restore/restore-to-a-new-cluster.md
```diff
@@ -0,0 +1,34 @@
...
+1. Deploy the 2nd cluster then install Velero only. You can try with different cluster config (more nodes or disks) here.
+2. Restore the cluster backup. e.g.,
+    ```bash
+    velero restore create --from-backup lh-cluster
+    ```
+3. Removing all old instance manager pods from namespace `longhorn-system`.
+4. Re-config nodes and disks for the restored Longhorn system if necessary.
+5. Re-create backing images.
+6. Restore all Longhorn volumes from the remote backup target.
+7. Update the access mode to `ReadWriteMany` since all restored volumes are mode `ReadWriteOnce` by default.
```
innobead, line 28: How about if there is a volume actually RWO instead of RWX?

docs/content/manual/pre-release/cluster-restore/restore-to-a-new-cluster.md
```diff
@@ -0,0 +1,34 @@
...
+    velero backup create lh-cluster --exclude-resources persistentvolumes,persistentvolumeclaims,backuptargets.longhorn.io,backupvolumes.longhorn.io,backups.longhorn.io,nodes.longhorn.io,volumes.longhorn.io,engines.longhorn.io,replicas.longhorn.io,backingimagedatasources.longhorn.io,backingimagemanagers.longhorn.io,backingimages.longhorn.io,sharemanagers.longhorn.io,instancemanagers.longhorn.io,engineimages.longhorn.io
+    ```
+   
+#### Restore to a new cluster
+1. Deploy the 2nd cluster then install Velero only. You can try with different cluster config (more nodes or disks) here.
+2. Restore the cluster backup. e.g.,
+    ```bash
+    velero restore create --from-backup lh-cluster
+    ```
+3. Removing all old instance manager pods from namespace `longhorn-system`.
```
innobead, line 24: Could you explain why we need removing all old instance manager pods here? I guess you would like to clean up old IM pods recreated by Velero but instead just delete them first then let longhorn managers recreate them.

For managed pods, is it possible any other cases similar to IM pods? like backing image?

innobead (commented):
LGTM, but need to answer my questions first.

docs/content/manual/pre-release/cluster-restore/restore-to-a-new-cluster.md
shuo-wu: We don't need to do anything extra for RWO volumes. And after fixing https://github.com/longhorn/longhorn/issues/3444, there is no need to apply this step, either.

docs/content/manual/pre-release/cluster-restore/restore-to-a-new-cluster.md
shuo-wu: Yes. We need to clean up the instance manager pods created by Velero. There is no corresponding InstanceManager CR for these old pods.
And thank you for reminding me. It's better to remove backing image manager pods as well.

innobead (approved):

innobead:
@chanow816 already reviewed, so just merge. 

cc @longhorn/qa

PR #838, opened 2021-12-24
Change backup count to reflect retain count in test_recurring_job

chanow816:
After https://github.com/longhorn/longhorn/issues/627 and https://github.com/longhorn/longhorn/issues/3035 closed, backup count in recurring job will reflect by retain count correctly.

Modified test_recurring_job(), make it match longhorn behavior.

### DIFF
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -259,12 +259,11 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
         elif "backup2-" in b.snapshot:
             complete_backup_2_count += 1
 
-    # 2 completed backups from backup1
-    # 2 or more completed backups from backup2
+    # 1 completed backups from backup1
+    # 2 completed backups from backup2
     # NOTE: NFS backup can be slow sometimes and error prone
-    assert complete_backup_1_count == 2
-    assert complete_backup_2_count >= 2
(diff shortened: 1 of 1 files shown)


jenting (approved):

manager/integration/tests/test_recurring_job.py
```diff
@@ -259,12 +259,11 @@ def test_recurring_job(set_random_backupstore, client, volume_name):  # NOQA
...
             complete_backup_2_count += 1
 
-    # 2 completed backups from backup1
-    # 2 or more completed backups from backup2
+    # 1 completed backups from backup1
+    # 2 completed backups from backup2
     # NOTE: NFS backup can be slow sometimes and error prone
-    assert complete_backup_1_count == 2
-    assert complete_backup_2_count >= 2
-    assert complete_backup_2_count < 4
```
innobead, line 267: Could you explain why we had these assertions before?

jenting: Maybe @c3y1huang know the reason.

innobead: ya, forgot the details, but it seems like a timing issue so we made this tolerance for the counts. 

@chanow816 But why https://github.com/longhorn/longhorn/issues/3035 can fix this issue if the timing problem still exists?

innobead (approved):

PR #839, opened 2021-12-29
change assertion string insensitive

chanow816:

Change assertion string to case insensitive. 
Because currently our assertion is `Job Can\\'t retain more than 50 snapshots` and master-head API returned `job Can\\'t retain more than 50 snapshots`


jenting (approved):

khushboo-rancher (approved):
LGTM

PR #843, opened 2022-01-03
Add case skeleton for checking kubelet status while workload with subpath

kaxing:
Context: https://github.com/longhorn/longhorn/issues/3080

Competition checklist:
- [X] Manual steps
- [X] Case skeleton
- [ ] Working test case
  - [ ] Deploy the testing workload programmatically 
  - [X] Functions to restart and check k3s kubelet agent logs

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4564,9 +4564,10 @@ def restart_k3s_agent():
     subprocess.check_call(cmd)
 
 def check_k3s_agent_log(keyword, minutes=60*3):
-    # Check the keyword in k3s-agent log for x minutes every second of the output
+    # Check the keyword in k3s-agent log for 3 minutes then check RETRY_COMMAND_COUNT times of the log
     cmd = ['journalctl -u k3s-agent | grep -i ' + keyword]
-    for _ in range(minutes*60):
+    time.sleep(minutes*60)
+    for _ in range(RETRY_COMMAND_COUNT):
         output = subprocess.check_output(cmd)
         if keyword in output:
             return True
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4563,9 +4563,12 @@ def restart_k3s_agent():
     cmd = ['service k3s-agent restart']
     subprocess.check_call(cmd)
 
-def check_k3s_agent_log(keyword):
-    cmd = ['journalctl -u k3s-agent']
-    output = subprocess.check_output(cmd)
-    if keyword in output:
-        return True
+def check_k3s_agent_log(keyword, minutes=60*3):
+    # Check the keyword in k3s-agent log for x minutes every second of the output
+    cmd = ['journalctl -u k3s-agent | grep -i ' + keyword]
+    for _ in range(minutes*60):
+        output = subprocess.check_output(cmd)
+        if keyword in output:
+            return True
+        time.sleep(1)
     return False
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4558,3 +4558,14 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
                                      'default')
 
     return data_checksum, output, restore_pod_name
+
+def restart_k3s_agent():
+    cmd = ['service k3s-agent restart']
+    subprocess.check_call(cmd)
+
+def check_k3s_agent_log(keyword):
+    cmd = ['journalctl -u k3s-agent']
+    output = subprocess.check_output(cmd)
+    if keyword in output:
+        return True
+    return False
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4558,18 +4558,3 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
                                      'default')
 
     return data_checksum, output, restore_pod_name
-
-def restart_k3s_agent():
-    cmd = ['service k3s-agent restart']
-    subprocess.check_call(cmd)
-
-def check_k3s_agent_log(keyword, minutes=60*3):
-    # Check the keyword in k3s-agent log for 3 minutes then check RETRY_COMMAND_COUNT times of the log
-    cmd = ['journalctl -u k3s-agent | grep -i ' + keyword]
-    time.sleep(minutes*60)
-    for _ in range(RETRY_COMMAND_COUNT):
-        output = subprocess.check_output(cmd)
-        if keyword in output:
-            return True
-        time.sleep(1)
-    return False
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -2141,6 +2141,23 @@ def test_auto_remount_with_subpath(client, core_api, storage_class, sts_name, st
     assert expect_md5sum == md5sum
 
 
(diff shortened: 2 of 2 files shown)


innobead:
@kaxing DCO check failed. Please fix this w/ git sign.

kaxing:
@innobead Thanks for the reminder. I am aware of it.

There is already something I need to fix. I will include signing in my next commit.

manager/integration/tests/common.py
```diff
@@ -4558,3 +4558,17 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
...
     return data_checksum, output, restore_pod_name
+
+def restart_k3s_agent():
+    cmd = ['service k3s-agent restart']
+    subprocess.check_call(cmd)
+
+def check_k3s_agent_log(keyword, minutes=60*3):
+    # Check the keyword in k3s-agent log for x minutes every second of the output
+    cmd = ['journalctl -u k3s-agent | grep -i ' + keyword]
+    for _ in range(minutes*60):
```
khushboo-rancher, line 4569: We have common retry range variables like `RETRY_COMMAND_COUNT` etc

manager/integration/tests/test_ha.py
```diff
@@ -2141,6 +2141,46 @@ def test_auto_remount_with_subpath(client, core_api, storage_class, sts_name, st
     assert expect_md5sum == md5sum
 
 
+@pytest.mark.skip(reason="WAIT_FOR_UPSTREAM_FIX") # NOQA
+@pytest.mark.skipif( "k3s" not in k8s_api_client.status.node_info.kubelet_version, reason='This test is only compaitable with k3s')  # NOQA
```
khushboo-rancher, line 2145: Is `k8s_api_client.status.node_info.kubelet_version` available before our test env is set up? If not, the pdoc check will fail. 
I think, it's better if we include the check in the test.

manager/integration/tests/test_ha.py
```diff
@@ -2141,6 +2141,46 @@ def test_auto_remount_with_subpath(client, core_api, storage_class, sts_name, st
...
+def test_restart_kubelete_when_pod_with_subpath():
+    """
+    Related issue 3080:
+    https://github.com/longhorn/longhorn/issues/3080
+
+    Steps:
+    0. Check if the cluster runs with k3s
+    1. Create deployment with subpath with pods on the same targeting-node
+    2. Wait for pods become Running
```
khushboo-rancher, line 2154: 3. Write some data in the volume, so that we can check the data is intact after the kubelet restart.

manager/integration/tests/test_ha.py
```diff
@@ -2141,6 +2141,46 @@ def test_auto_remount_with_subpath(client, core_api, storage_class, sts_name, st
...
+    3. Restart k3s kubelet-agent service
+    4. Wait for new pods become Running
+    5. Check kubelet log from the targeting-node
```
khushboo-rancher, line 2157: 6. Assert the data once the pod is back running.

manager/integration/tests/test_ha.py
```diff
@@ -2141,6 +2141,46 @@ def test_auto_remount_with_subpath(client, core_api, storage_class, sts_name, st
...
+    deployment.spec.template.spec.containers[0].volume_mounts[0].sub_path = "sub"
+
+    # Wait for deployment become ready
+    wait_for_deployment_ready(k8s_api_client, deployment)
+    pod_name = deployment.spec.template.metadata.name
+
+    # TODO: Specify node to run the command
+
+    common.restart_k3s_agent
+    common.wait_for_pod_phase(core_api, pod_name, pod_phase="Running")
```
khushboo-rancher, line 2177: Immediately, after the kubelet restart, the pod status might be in running state for few seconds. So, first we need to check the pod status changed to `pending` or `terminating. Then we should check the running status.

manager/integration/tests/test_ha.py
```diff
@@ -2141,6 +2141,46 @@ def test_auto_remount_with_subpath(client, core_api, storage_class, sts_name, st
...
+
+    # Check kubelet log from the targeting-node
+    expect_restult = not common.check_k3s_agent_log("Path does not exist", 60*5)
+    assert expect_restult == True
+
```
khushboo-rancher, line 2182: Assert the data of the volume.

khushboo-rancher (changes requested):

manager/integration/tests/common.py
kaxing: Hey Khushboo, I think what I trying to do here is to tailing the log after waiting for certain amount of time. I will see what I can weak this function with **RETRY_COMMAND_COUNT**.

manager/integration/tests/test_ha.py
kaxing: Okay, I will check again, and see if we want a variable that stores the cluster version before the test case.

manager/integration/tests/test_ha.py
kaxing: Added get_container_data_md5sum function to validate the volume data in the latest commit.

manager/integration/tests/test_ha.py
kaxing: I am now leveraging `wait_deployment_replica_ready` to check on all pods ready status after destroying the pods(as simulating kubelet restart)

manager/integration/tests/test_ha.py
kaxing: Thanks Chris for the hint. In latest commit, I am using subprocess to call kubectl and obtain the node version.

kaxing:
In the latest change, 

- Rework the implementation, importing the necessary modules.
- Reuse code from [get_pod_data_md5sum](https://github.com/longhorn/longhorn-tests/blob/dfdaff305fc331891737511b1bfeb3db9bb1a373/manager/integration/tests/common.py#L714) and [write_pod_volume_data](https://github.com/longhorn/longhorn-tests/blob/dfdaff305fc331891737511b1bfeb3db9bb1a373/manager/integration/tests/common.py#L629) and add the container parameter to access the pod with multiple containers.

Things still need to be done in the future:
- Find a way to restart kubelet, k3s-agenet from inside the longhorn-test pod
  - Also avoid deploy workload and volume on the same node as longhorn-test
- To minify/shrink _deployment_ manifest dictionary to a smaller format

manager/integration/tests/test_ha.py
```diff
@@ -2141,6 +2145,192 @@ def test_auto_remount_with_subpath(client, core_api, storage_class, sts_name, st
...
+    Related issue 3080:
+    https://github.com/longhorn/longhorn/issues/3080
+
+    Steps:
+    0. Check if the all nodes runs with k3s
+    1. Label one worker node with "targeting-node: true"
+    2. Create deployment with subpath and deploy to the targeting-node
+    3. Wait for pods become ready
+    4. Write some random data to the volume
+    5. Restart k3s kubelet-agent service
```
khushboo-rancher, line 2163: Can you please specify restart k3s kubelet-agent on which node?

khushboo-rancher (changes requested):

innobead:
I don't think we need this anymore, because this is more about a kubelet issue. https://github.com/longhorn/longhorn/issues/3080#issuecomment-1069121270

cc @derekbit

PR #844, opened 2022-01-03
doc: Add a test for the longhorn manager NPE issue

shuo-wu:
Longhorn/longhorn#3484


khushboo-rancher:
I think, we can add this test in the integration test rather than in the manual.

khushboo-rancher (commented):
I think, we can add this test in the integration test rather than manual.

shuo-wu:
> I think, we can add this test in the integration test rather than manual.

The operations in this test are not invalid use cases in the real world... I don't think we need to put it in Longhorn nightly integration test plan.

innobead (approved):
LGTM but this is more like chaos testing instead of e2e testing. 

I think it's fine to have it in the manual regression for now.

PR #849, opened 2022-01-10
modify security group to allow udp connection for longhorn webhooks

meldafrawi:

### DIFF
--- a/test_framework/terraform/aws/centos/main.tf
+++ b/test_framework/terraform/aws/centos/main.tf
@@ -69,6 +69,13 @@ resource "aws_security_group" "lh_aws_secgrp_controlplane" {
     cidr_blocks = ["0.0.0.0/0"]
   }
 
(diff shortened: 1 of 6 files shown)


innobead (commented):
LGTM, but I would like to where to find out UDP traffic is needed for webhook.

jenting:
> LGTM, but I would like to where to find out UDP traffic is needed for webhook.

+1, I'll double-check the code as well.

innobead (approved):

innobead:
Merged first, but will rever if needed after @jenting figures out the root cause later.

innobead:
@meldafrawi said

while debugging the issue,
i suspected that it is a networking issue, opened all communications to/from controller node, tests passed
tried to open only TCP connections, the tests fail again.
that narrowed it down to UDP communication, but i couldn't find which port is required for that communication.

PR #853, opened 2022-01-13
allow k3s on amd64 instances

meldafrawi:


meldafrawi:
This PR should be merged after: https://github.com/longhorn/infra/pull/76

test_framework/terraform/aws/centos/main.tf
```diff
@@ -261,7 +261,7 @@ resource "aws_instance" "lh_aws_instance_controlplane" {
   }
 
   key_name = aws_key_pair.lh_aws_pair_key.key_name
-  user_data = var.arch == "arm64" ? data.template_file.provision_arm64_server.rendered : data.template_file.provision_amd64.rendered
+  user_data = var.k8s_distro_name == "k3s" ? data.template_file.provision_k3s_server.rendered : data.template_file.provision_rke.rendered
```
innobead, line 264: This ternary expression will be adjusted when supporting rke2 in near future.

test_framework/terraform/aws/centos/variables.tf
```diff
@@ -94,10 +94,20 @@ variable "rke_k8s_version" {
...
-variable "k3s_version" {
+variable "k8s_distro_name" {
+  type        = string
+  default     = "k3s"
+  description = "kubernetes distro version to install [rke, k3s]  (default: k3s)"
+}
+
+variable "k8s_distro_version" {
   type        = string
   default     = "v1.20.8+k3s1"
```
innobead, line 105: We should update to the rather recent version later.

test_framework/terraform/aws/rockylinux/user-data-scripts/provision_rke.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
innobead, line 1: this is different from the other template file like .tpl. Does it matter? If not, probably we can change all of them to .sh instead.

innobead (changes requested):
LGTM, just few comments to resolve.

test_framework/terraform/aws/rockylinux/user-data-scripts/provision_rke.sh
meldafrawi: fixed, Thank you.

test_framework/terraform/aws/centos/variables.tf
meldafrawi: fixed

test_framework/terraform/aws/centos/main.tf
meldafrawi: Yes, this will be modified when adding RKE2 support.

innobead (approved):

PR #858, opened 2022-01-19
Build test engine image for amd64 and arm 64 

chanow816:
For https://github.com/longhorn/longhorn/issues/1961
After below environment set complete

- docker_id
- docker_pwd
- docker_repo
- docker_tag (need to automatically get) 
- TF_VAR_build_engine_aws_access_key
- TF_VAR_build_engine_aws_secret_key

Simple run `python3 ./test_engine/build_test_engine.py` script will create arm64 / amd64 longhorn-engine docker images and push to docker hub for testing, and use dock manifest to combine two images into one as `<docker_repo>:<docker_tag>`

And we can use `<docker_repo>:<docker_tag>` to do engine upgrade e2e test.

### DIFF
--- a/test_engine/.gitignore
+++ b/test_engine/.gitignore
@@ -1,0 +1,18 @@
+.tox
+.cache
+__pycache__
+*.pyc
+MANIFEST
+validationTestsJunit.xml
+.bash_history
+*.DS_Store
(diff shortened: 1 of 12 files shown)


chanow816:
reopen btn grayed out

PR #860, opened 2022-01-21
Add volume cloning test

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/2941

This is implementation of tests in the volume cloning module..

I did below modifications in the test skeleton:
1. Renamed test case name.
2. Previously cloning from attached and detached volume was part one test, now divided them in two.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -79,6 +79,9 @@ VOLUME_ROBUSTNESS_UNKNOWN = "unknown"
 VOLUME_FIELD_RESTOREREQUIRED = "restoreRequired"
 VOLUME_FIELD_READY = "ready"
 
+VOLUME_FIELD_CLONE_STATUS = "cloneStatus"
+VOLUME_FIELD_CLONE_COMPLETED = "completed"
+
 VOLUME_REPLICA_WO_LIMIT = 1
 
 DEFAULT_STORAGECLASS_NAME = 'longhorn-test'
@@ -1983,6 +1986,11 @@ def generate_sts_name():
     return STATEFULSET_NAME + "-" + \
         ''.join(random.choice(string.ascii_lowercase + string.digits)
                 for _ in range(6))
+
+
+def generate_random_suffix():
+    return "-" + ''.join(random.choice(string.ascii_lowercase + string.digits)
+                         for _ in range(6))
 
 
 def get_default_engine_image(client):
@@ -4465,6 +4473,30 @@ def wait_for_pod_annotation(core_api,
                 break
         time.sleep(RETRY_INTERVAL)
     assert matches is True
+
+
+def wait_for_volume_clone_status(client, name, key, value):
+    for _ in range(RETRY_COUNTS):
+        volume = client.by_id_volume(name)
+        if volume[VOLUME_FIELD_CLONE_STATUS][key] == value:
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert volume[VOLUME_FIELD_CLONE_STATUS][key] == value, \
+        f" Expected value={value}\n. " \
+        f" Got volume[{VOLUME_FIELD_CLONE_STATUS}][{key}]= " \
+        f"{volume[VOLUME_FIELD_CLONE_STATUS][key]}\n. volume={volume}"
+    return volume
+
+
+def get_clone_volume_name(client, source_volume_name):
+    for _ in range(RETRY_EXEC_COUNTS):
+        volumes = client.list_volume()
+        for volume in volumes:
+            if volume['cloneStatus']['sourceVolume'] == \
+                    source_volume_name:
+                return volume.name
+        time.sleep(RETRY_INTERVAL_LONG)
+    return None
 
 
 def create_backup_from_volume_attached_to_pod(client, core_api,
--- a/manager/integration/tests/test_cloning.py
+++ b/manager/integration/tests/test_cloning.py
@@ -1,8 +1,66 @@
 import pytest
 
-@pytest.mark.skip(reason="TODO") # NOQA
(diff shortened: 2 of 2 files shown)


chanow816 (dismissed):

manager/integration/tests/common.py
```diff
@@ -4467,6 +4475,31 @@ def wait_for_pod_annotation(core_api,
...
 
 
+def wait_for_volume_clone_status(client, name, key, value):
+    for _ in range(RETRY_COUNTS):
+        volume = client.by_id_volume(name)
+        if volume[VOLUME_FIELD_CLONE_STATUS][key] == value:
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert volume[VOLUME_FIELD_CLONE_STATUS][key] == value, \
+        f" value={value}\n. volume[cloneStatus][key]=" \
```
PhanLe1010, line 4485:
```suggestion
        f" value={value}\n. volume[{VOLUME_FIELD_CLONE_STATUS}][{key}]=" \
```

PhanLe1010, line 4485:
```suggestion
        f" Expecting value={value}\n. Got volume[{VOLUME_FIELD_CLONE_STATUS}][{key}]=" \
```

manager/integration/tests/test_cloning.py
```diff
@@ -36,49 +94,178 @@ def test_cloning_without_backing_image():  # NOQA
...
-    6. Wait for the `CloneStatus.State` in `cloned-pvc` to be `completed`
-    7. In 3-min retry loop, wait for the cloned pod to be running
-    8. Verify the data in `cloned-pvc` is the same as in `source-pvc`
-    9. In 2-min retry loop, verify the volume of the `clone-pvc` eventually
+    5. Wait for the `CloneStatus.State` in `cloned-pvc` to be `completed`
+    6. Clone volume should get detached after cloning completion, wait for it.
+    7. Specify the `cloned-pvc` in a cloned pod yaml and deploy the cloned pod
+    8. In 3-min retry loop, wait for the cloned pod to be running
+    9. Verify the data in `cloned-pvc` is the same as in `source-pvc`
+    10. In 2-min retry loop, verify the volume of the `clone-pvc` eventually
```
PhanLe1010, line 104: Just want to double check, are we cleaning up the pods and the PVC at the end of the test?

manager/integration/tests/test_cloning.py
```diff
@@ -36,49 +94,178 @@ def test_cloning_without_backing_image():  # NOQA
...
+                name: source-pvc
+                kind: PersistentVolumeClaim
+              accessModes:
+                - ReadWriteOnce
+              resources:
+                requests:
+                  storage: 10Gi
+            ```
+        6. Wait for `source-pvc` to be attached
+        7. Wait for a new snapshot created in `source-pvc` volume created
```
PhanLe1010, line 198: If the cloning happens fast enough, we might not see these intermediate steps. Should we skip them?

manager/integration/tests/test_cloning.py
```diff
@@ -140,14 +327,90 @@ def test_cloning_interrupted():  # NOQA
...
+
+    # Step-2
+    pod_name = 'source-pod' + generate_random_suffix()
+    pod['metadata']['name'] = pod_name
+    pod['spec']['volumes'] = [create_pvc_spec(source_pvc_name)]
+    create_and_wait_pod(core_api, pod)
+
+    # Step-3
+    write_pod_volume_random_data(core_api, pod_name,
+                                 '/data/test', DATA_SIZE_IN_MB_3)
```
PhanLe1010, line 356: Can we modify the test description to specify that we only write 500MB of data?

PhanLe1010 (commented):
In general LGTM.

Thank you for implementing the test 🚀

manager/integration/tests/common.py
khushboo-rancher: Fixed, thanks

manager/integration/tests/test_cloning.py
khushboo-rancher: Fixed, thanks

manager/integration/tests/test_cloning.py
khushboo-rancher: Yes, the pods and pvc are getting cleaned up in the fixtures.

manager/integration/tests/test_cloning.py
khushboo-rancher: I ran this multiple times, we get enough time to get these states and I think these are important to check. 
If encounter problem with these checks in the Jenkins job runs, we can skip these checks.

PhanLe1010: Okay. Thank you for the explanation

PhanLe1010 (approved):
LGTM

manager/integration/tests/common.py
```diff
@@ -4467,6 +4475,32 @@ def wait_for_pod_annotation(core_api,
...
+
+def get_clone_volume_name(client, source_volume_name):
+    for _ in range(RETRY_EXEC_COUNTS):
+        volumes = client.list_volume()
+        for volume in volumes:
+            if volume['cloneStatus']['sourceVolume'] == \
+                    source_volume_name:
+                return volume.name
+            else:
+                continue
```
innobead, line 4499: nit: we don't need this else condition, because at the end of loop, it will just reiterate the next.

innobead (changes requested):
LGTM, but just one minor comment needed to fix.

manager/integration/tests/common.py
khushboo-rancher: Right, we don't need it, removed it. :)

innobead (approved):

PR #864, opened 2022-01-25
[v1.1.3-backport]  fix create_aws_secret in longhorn-setup script

meldafrawi:


innobead (approved):

innobead:
Just found we should backport to 1.1.x instead of 1.1.3, but that’s fine for now.

PR #874, opened 2022-02-04
Skip the nfs installation in the test

khushboo-rancher:
Skipping the installation of the nfs in the test

### DIFF
--- a/manager/integration/tests/test_rwx.py
+++ b/manager/integration/tests/test_rwx.py
@@ -21,8 +21,8 @@ from backupstore import set_random_backupstore # NOQA
 from multiprocessing import Pool
 
 import time
-import pytest
-import subprocess
+# import pytest
+# import subprocess
 
 LONGHORN_NFS_INSTALLATION_URL = \
     "https://raw.githubusercontent.com/longhorn/" \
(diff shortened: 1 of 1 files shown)


innobead:
@khushboo-rancher I guess this was added to ensure nfs client lib was actually installed for some reason before (@chanow816 ), so right now we would like to remove it back but just rely on the host packages installed by the terraform script. right?

khushboo-rancher:
@innobead Yes, you're right. After the infra changes, we expect the nfs client should be installed properly. If not, I think this is good time to fix it in infra setup as @meldafrawi is putting efforts to fix all the problems we have in infra.
So, for now just commented the lines and if everything is working fine on Jenkins, we can remove the lines from the code.

innobead (approved):

khushboo-rancher:
The test `test_rwx` has passed on latest run of the integration test, that's mean no need to install the nfs-client from the test. Test setup worked fine on SLES. https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/33/testReport/tests/test_rwx/

Triggered test_rwx on Ubuntu to check this before we can permanently remove the nfs-client installation from the test. https://ci.longhorn.io/job/public/job/master/job/ubuntu/job/amd64/job/longhorn-tests-ubuntu-amd64/563/

PR #875, opened 2022-02-05
Add print statement with assert

khushboo-rancher:
Added print statement in case assertion failures identify the problems in our integration tests.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1458,7 +1458,9 @@ def wait_scheduling_failure(client, volume_name):
         if scheduling_failure:
             break
         time.sleep(RETRY_INTERVAL)
-    assert scheduling_failure
+    assert scheduling_failure, f" Scheduled Status = " \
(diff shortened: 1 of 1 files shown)


manager/integration/tests/common.py
```diff
@@ -1935,9 +1937,10 @@ def wait_for_replica_scheduled(client, volume_name, to_nodes,
 
         time.sleep(RETRY_INTERVAL)
 
-    assert scheduled == expect_success
-    assert unexpect_fail == 0
-    assert len(volume.replicas) == expect_success + expect_fail
+    assert scheduled == expect_success, f" Volume = {volume}"
+    assert unexpect_fail == 0, f" Volume = {volume}"
+    assert len(volume.replicas) == expect_success + expect_fail, \
+        f" Volume = {volume}"
```
PhanLe1010, line 1943: We already has the volume name from the error trace but it is ok to print it out here too

PhanLe1010 (approved):
LGTM

PR #879, opened 2022-02-09
Generate support bundle when e2e case fail

cchien816:
For https://github.com/longhorn/longhorn/issues/3557
Add feature to generate support bundle in longhorn-test pod when e2e test case fail

When test case fail , script will create folder `support_bundle` which the abs path is `/integration/tests/support_bundle` in pod
And will generate support bundle renamed `"test case name.zip"` into folder support_bundle

This is current design, for become Jenkins artifact for download, @meldafrawi  could you advise me should I modify code to match your Jenkins design, thank you

Screenshot from 2022-02-09 16-44-54

cc @innobead , @khushboo-rancher

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -14,6 +14,8 @@ import socket
 import pytest
 
 import longhorn
+import requests
+import warnings
 
 from kubernetes import client as k8sclient, config as k8sconfig
 from kubernetes.client import Configuration
@@ -218,6 +220,8 @@ K8S_ZONE_LABEL = "topology.kubernetes.io/zone"
 BACKING_IMAGE_SOURCE_TYPE_DOWNLOAD = "download"
 
 JOB_LABEL = "recurring-job.longhorn.io"
+
+MAX_SUPPORT_BINDLE_NUMBER = 20
 
 
 def load_k8s_config():
@@ -4598,3 +4602,55 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
                                      'default')
 
     return data_checksum, output, restore_pod_name
+
+
+def generate_support_bundle(case_name):
+    """
+        Generate support bundle into folder ./support_bundle/case_name.zip
+
+        Won't generate support bundle if current support bundle count
+        greate than MAX_SUPPORT_BINDLE_NUMBER.
+        Args:
+            case_name: support bundle will named case_name.zip
+    """
+
+    os.makedirs("support_bundle", exist_ok=True)
+    file_cnt = len(os.listdir("support_bundle"))
+
+    if file_cnt >= MAX_SUPPORT_BINDLE_NUMBER:
+        warnings.warn("Ignoring the bundle download because of \
+                            avoiding overwhelming the disk usage.")
+        return
+
+    # Use API gen support bundle
+    client = get_longhorn_api_client()
+    url = client._url.replace('schemas', 'supportbundles')
+    data = {'description': case_name, 'issueURL': case_name}
+    res = requests.post(url, json=data).json()
+
+    id = res['id']
+    name = res['name']
+
+    support_bundle_url = '{}/{}/{}'.format(url, id, name)
+    for i in range(RETRY_EXEC_COUNTS):
+        res = requests.get(support_bundle_url).json()
+
+        if res['progressPercentage'] == 100:
+            break
+        else:
+            time.sleep(RETRY_INTERVAL_LONG)
+
+    if res['progressPercentage'] != 100:
+        warnings.warn("Timeout to wait support bundle ready, skip download")
+        return
+
+    # Download support bundle
+    download_url = '{}/download'.format(support_bundle_url)
+    try:
+        r = requests.get(download_url, allow_redirects=True, timeout=300)
+        r.raise_for_status()
+        with open('./support_bundle/{0}.zip'.format(case_name), 'wb') as f:
+            f.write(r.content)
+    except Exception as e:
+        warnings.warn("Error occured while downloading support bundle {}.zip\n\
+            The error was {}".format(case_name, e))
--- a/manager/integration/tests/conftest.py
+++ b/manager/integration/tests/conftest.py
@@ -7,7 +7,7 @@ from common import get_longhorn_api_client, \
     NODE_CONDITION_MOUNTPROPAGATION, CONDITION_STATUS_TRUE
 from common import wait_for_node_mountpropagation_condition
 from common import check_longhorn, check_csi_expansion
-
+from common import generate_support_bundle
 
 SKIP_BACKING_IMAGE_OPT = "--skip-backing-image-test"
 SKIP_RECURRING_JOB_OPT = "--skip-recurring-job-test"
@@ -173,3 +173,16 @@ def pytest_collection_modifyitems(config, items):
         for item in items:
             if "upgrade" in item.keywords:
                 item.add_marker(skip_upgrade)
+
+
+def pytest_exception_interact(call, report):
+
+    # Only work on TestReport, not on CollectReport
+    if type(report).__name__ != "TestReport":
+        return
+
+    # Get case name
+    case_name = str(report).split()[1]
+    case_name = case_name.split('/')[1].replace('\'', '')
+
+    generate_support_bundle(case_name)


innobead:
@chanow816 Do we consider if tons of cases failed, having a threshold/check avoids lots of bundles possibly overwhelming the disk space?

chanow816:
Hi @innobead ,

For tons of cases failed scenario, I did not consider that before, will start work on it today

cchien816:
Update script to let support bundle max count = 20 in case tons of cases failed scenario happen.

cc @innobead

manager/integration/tests/conftest.py
```diff
@@ -173,3 +177,38 @@ def pytest_collection_modifyitems(config, items):
...
                 item.add_marker(skip_upgrade)
+
+
+def pytest_exception_interact(call, report):
+
+    # Get case name
+    case_name = str(report).split()[1]
+    case_name = case_name.split('/')[1].replace('\'', '')
+
+    # Use API gen support bundle
```
innobead, line 190: Create a function for support bundle download in common.py, then reuse it here instead.

manager/integration/tests/conftest.py
```diff
@@ -173,3 +177,38 @@ def pytest_collection_modifyitems(config, items):
...
+
+        if res['progressPercentage'] == 100:
+            break
+        else:
+            time.sleep(RETRY_INTERVAL_LONG)
+
+    # Download support bindle
+    download_url = '{}/download'.format(support_bundle_url)
+
+    r = requests.get(download_url, allow_redirects=True)
```
innobead, line 212: Please handle if the support bundle is failed to download like timeout from retry period or return code is not 2xx.

If can't download, just leave a warning message.

manager/integration/tests/conftest.py
```diff
@@ -173,3 +177,38 @@ def pytest_collection_modifyitems(config, items):
...
+
+    os.makedirs("support_bundle", exist_ok=True)
+
+    open('./support_bundle/{0}.zip'.format(case_name), 'wb').write(r.content)
```
innobead: Please make sure handle any exception from all operations except failing in this hook.

manager/integration/tests/conftest.py
```diff
@@ -173,3 +177,38 @@ def pytest_collection_modifyitems(config, items):
         for item in items:
             if "upgrade" in item.keywords:
                 item.add_marker(skip_upgrade)
+
+
+def pytest_exception_interact(call, report):
```
innobead, line 184: From the [doc](https://docs.pytest.org/en/6.2.x/reference.html#pytest.hookspec.pytest_exception_interact), please only do this hook for test report case instead of collection repot.

manager/integration/tests/conftest.py
```diff
@@ -173,3 +179,42 @@ def pytest_collection_modifyitems(config, items):
...
+    # Download support bindle
+    download_url = '{}/download'.format(support_bundle_url)
+
+    r = requests.get(download_url, allow_redirects=True)
+
+    os.makedirs("support_bundle", exist_ok=True)
+
+    file_cnt = len(os.listdir("support_bundle"))
+
+    if file_cnt < MAX_SUPPORT_BINDLE_NUMBER:
```
innobead, line 218: If >= max, please leave a message about ignoring the bundle download because of avoiding overwhelming the disk usage.

innobead (changes requested):

cchien816:
Hi @innobead 

Follow comments I have changed below code for you review, thank you

- Create function `generate_support_bundle` in common.py for generate support bundle
- Handle support bundle download fail scenario and pop warning message
- The generate support bundle only report type is `TestReport`
- If support bundle number greater then we expected, will pop warning message and return

manager/integration/tests/conftest.py
cchien816: Modified, please see https://github.com/longhorn/longhorn-tests/pull/879#issuecomment-1040301027

manager/integration/tests/conftest.py
cchien816: Modified, please see https://github.com/longhorn/longhorn-tests/pull/879#issuecomment-1040301027

manager/integration/tests/conftest.py
cchien816: Modified, please see https://github.com/longhorn/longhorn-tests/pull/879#issuecomment-1040301027

yangchiu (approved):

manager/integration/tests/common.py
```diff
@@ -4598,3 +4602,51 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
...
+            break
+        else:
+            time.sleep(RETRY_INTERVAL_LONG)
+
+    # Download support bundle
+    download_url = '{}/download'.format(support_bundle_url)
+    try:
+        r = requests.get(download_url, allow_redirects=True, timeout=300)
+        r.raise_for_status()
+        open('./support_bundle/{0}.zip'.format(case_name), 'wb')\
```
innobead, line 4648: Could you use `with` to ensure the file get closed after writing?

manager/integration/tests/common.py
```diff
@@ -4598,3 +4602,51 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
...
+    for i in range(RETRY_EXEC_COUNTS):
+        res = requests.get(support_bundle_url).json()
+
+        if res['progressPercentage'] == 100:
+            break
+        else:
+            time.sleep(RETRY_INTERVAL_LONG)
+
+    # Download support bundle
+    download_url = '{}/download'.format(support_bundle_url)
```
innobead, line 4644: If timeout to wait for the support bundle ready, should have a log then ignore downloading.

innobead (changes requested):
Few minor fixed needed, also too many fragile commits in this PR. This PR is quite simple, so I would suggest having one commit to make the contributon understandable.

innobead:
@meldafrawi Please ensure we can download the bundles from the CI task build page later. Thanks.

manager/integration/tests/common.py
cchien816: HI @innobead ,

I did test on local environment to simulate timeout scenario, current code `r.raise_for_status()` will catch timeout error as below log and ignore download. 
`HTTPConnectionPool(host='10.10.10.100', port=80): Max retries exceeded with url: /download (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x7f2941b91850>, 'Connection to 10.10.10.100 timed out. (connect timeout=10)'))`

If there have anything I missed, I will update asap, thank you.

innobead: I mean, if running out to wait for `res['progressPercentage'] == 100`, you don't need to do the bundle download because it should not be there.

innobead:
@cchien816 this is the last one https://github.com/longhorn/longhorn-tests/pull/879/files#r817650274, then we can merge after this is resolved.

manager/integration/tests/common.py
cchien816: Hi @innobead , 

I modified code to skip download when `res['progressPercentage'] != 100` for review, thank you

manager/integration/tests/common.py
```diff
@@ -4598,3 +4602,55 @@ def restore_backup_and_get_data_checksum(client, core_api, backup, pod,
...
+            r = requests.get(download_url, allow_redirects=True, timeout=300)
+            r.raise_for_status()
+            with open('./support_bundle/{0}.zip'.format(case_name), 'wb') as f:
+                f.write(r.content)
+
+        except Exception as e:
+            warnings.warn("Error occured while downloading support bundle {}.zip\n\
+                The error was {}".format(case_name, e))
+    else:
+        warnings.warn("Timeout to wait support bundle ready, skip download")
```
innobead, line 4656:
```suggestion
    if res['progressPercentage'] != 100:
        warnings.warn("Timeout to wait support bundle ready, skip download")
        return
    
    # Download support bundle
    download_url = '{}/download'.format(support_bundle_url)
    try:
        r = requests.get(download_url, allow_redirects=True, timeout=300)
        r.raise_for_status()
        with open('./support_bundle/{0}.zip'.format(case_name), 'wb') as f:
            f.write(r.content)

    except Exception as e:
        warnings.warn("Error occured while downloading support bundle {}.zip\n\
            The error was {}".format(case_name, e))
```

innobead (changes requested):

manager/integration/tests/common.py
cchien816: Hi @innobead ,

I updated PR content for review, thank you

innobead (approved):

innobead:
@meldafrawi please work with @cchien816 to make sure the support bundle is able to download from the CI build.

PR #883, opened 2022-02-14
Fix the tagging tests

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/3625

**Cause of failure**:
The CRD is changed to have tags as a list instead of dictionary. 

**Solution**:
Changed the assert to check the list is empty.

cc: @longhorn/qa 

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1286,10 +1286,11 @@ def node_default_tags():
         update_disks[list(update_disks)[0]].tags = []
         new_node = node.diskUpdate(disks=update_disks)
         disks = get_update_disks(new_node.disks)
-        assert disks[list(new_node.disks)[0]].tags is None
+        assert len(disks[list(new_node.disks)[0]].tags) == 0, \
+            f" disk = {disks}"
 
         new_node = set_node_tags(client, node)
-        assert new_node.tags is None
+        assert len(new_node.tags) == 0, f" Node = {new_node}"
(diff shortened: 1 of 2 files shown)


meldafrawi (approved):
LGTM

cchien816 (approved):
LGTM

innobead (approved):
Good fix, and the type of tags actually does not change, but the dfault value gets changed w/ an empy value instead of None.

ref: http://github.com/longhorn/longhorn-manager/blob/d88579453a06b9824c6c28d9feab6cfda7996b74/datastore/longhorn.go#L1793-L1793

innobead:
btw, this should only happen to the master branch.

derekbit:
> Good fix, and the type of tags actually does not change, but the dfault value gets changed w/ an empy value instead of None.
> 
> ref: [http://github.com/longhorn/longhorn-manager/blob/d88579453a06b9824c6c28d9feab6cfda7996b74/datastore/longhorn.go#L1793-L1793](https://github.com/longhorn/longhorn-manager/blob/d88579453a06b9824c6c28d9feab6cfda7996b74/datastore/longhorn.go#L1793-L1793)
> 
> @derekbit it seems some values are initialized in the master, so probably some e2e auto regression will fail. (@longhorn/qa needs to be aware of this)

Yes. Need to change it to the correct value (empty list) during initialization because of CRD validation rules.

PR #885, opened 2022-02-17
Remove nfs-installation from the test

khushboo-rancher:
We don't need the installation of nfs-common inside the test as the nfs-common installation is taken care in infra set up.


khushboo-rancher:
We'll need to back port this to v1.2.x

innobead (approved):

PR #886, opened 2022-02-18
Fix the integration/README.md ordered list indentation 

kaxing:
Turns out ordered list begin with `1.1` doesn't supported in the GitHub Flavored Markdown<sup>*</sup> , and there is a couple trouble<sup>**</sup> with the ordered nested list.

<sup>*</sup> https://github.github.com/gfm/#ordered-list-marker
<sup>**</sup> https://github.com/github/markup/issues/210#issuecomment-470705342

### DIFF
--- a/manager/integration/README.md
+++ b/manager/integration/README.md
@@ -4,7 +4,10 @@
 ## Integration test
 
 Requirement:
-1. A Kubernetes cluster with at least 3 nodes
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead (changes requested):
Some conflicts need to fix

innobead (approved):

PR #894, opened 2022-03-01
add RKE2 support and enable support bundle in nightly tests

meldafrawi:
Ref: https://github.com/longhorn/longhorn/issues/3251


innobead (approved):
In general, LGTM.

test_framework/Jenkinsfile
```diff
@@ -61,6 +61,13 @@ node {
             stage ('longhorn setup & tests') {
                 sh  " docker exec ${JOB_BASE_NAME}-${BUILD_NUMBER} ${TF_VAR_tf_workspace}/scripts/longhorn-setup.sh"
             }
+
+            stage ('download support bundle') {
```
innobead, line 65: We also have support bundle download for each failed test case by Chris. Please also ensure that works as expected and be able to download from the corresponding CI job build.

PR #896, opened 2022-03-03
Update Scalability Test configuration and doc

kaxing:
- Update terraform configuration to allow launching cluster on empty or non-default VPC AWS
- Add Pipfile to address require module versions and avoid mismatch Kubernetes API
- Add a small change allow scale_test.py to recognize **$KUBECONFIG**

### DIFF
--- a/scalability_test/script/Pipfile
+++ b/scalability_test/script/Pipfile
@@ -1,0 +1,13 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+matplotlib = "*"
+kubernetes = "==21.7.0"
+
+[dev-packages]
+
+[requires]
+python_version = "3"
--- a/scalability_test/script/scale-test.py
+++ b/scalability_test/script/scale-test.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python3
+
+import os
 from os import times
 from typing import Dict
 from matplotlib.pyplot import get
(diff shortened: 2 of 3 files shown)


PhanLe1010:
@kaxing ,there was a question from you about why the public IP of worker node:

> I am a not entirely sure why worker nodes require external public IP to work with control nodes.

Did you try to create a private subnet, a round table for the private subnet, and put the worker nodes into the private subnet? I see that @meldafrawi have some example about it at https://github.com/longhorn/longhorn-tests/blob/master/test_framework/terraform/aws/ubuntu/main.tf

kaxing:
@PhanLe1010 I am able to make the cluster work with 2 _aws_subnets_(private, public) setup, by mixing up configuration from `test_framework/terraform/aws/*/main.tf`.

With only one _aws_subnet_ and 1 public IP(EIP), we will need to route the worker nodes through/with control node(s). Which for this setup, control node and worker node will have IP in different private subnets(10.0.1.x and 10.0.2.x). Maybe there is a way to put them with in the same private IP subnet. Let me check with @meldafrawi.

So, to bypass this, I am using `map_public_ip_on_launch` to give each worker node to obtain public IPv4 address, which allows them talk to control node(s).

PhanLe1010:
> I am able to make the cluster work with 2 aws_subnets(private, public) setup, by mixing up configuration from test_framework/terraform/aws/*/main.tf.

Is it the same as what we are doing in `test_framework/terraform/aws/*/main.tf` ? Why we don't use this approach?

kaxing:
@PhanLe1010 I think we can. I am just wondering if there is a way to reduce the resources managed/created by terraform. To have a simpler setup.

PhanLe1010:
@kaxing Can we make 2 aws_subnets(private, public) setup similar what we are doing in `test_framework/terraform/aws/*/main.tf` so we can merge this PR?

innobead:
@yangchiu Please help take over this and work with @PhanLe1010 . The goal for us is to make this automated as a pipeline.

innobead:
Closing, and leave @yangchiu to follow up. 

Thanks, @kaxing contribution!

PR #897, opened 2022-03-09
Fix test_allow_volume_creation_with_degraded_availability_restore

c3y1huang:
Set replica-replenishment-wait-interval to 0 so the test will not get blocked.

https://github.com/longhorn/longhorn/issues/3183

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -3324,7 +3324,8 @@ def test_allow_volume_creation_with_degraded_availability_restore(set_random_bac
     Requirement:
     1. Set `allow-volume-creation-with-degraded-availability` to true.
     2. `node-level-soft-anti-affinity` to false.
-    3. Create a backup of 800MB.
+    3. `replica-replenishment-wait-interval` to 0.
+    4. Create a backup of 800MB.
 
     Steps:
     (restore)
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -3351,6 +3352,10 @@ def test_allow_volume_creation_with_degraded_availability_restore(set_random_bac
         client.by_id_setting(SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY)
     client.update(replica_soft_anti_affinity_setting, value="false")
 
+    replenish_wait_setting = \
+        client.by_id_setting(SETTING_REPLICA_REPLENISHMENT_WAIT_INTERVAL)
+    client.update(replenish_wait_setting, value="0")
```
innobead, line 3357: This reminds me if we will reset all settings after the test case is finished to ensure there will be not any cascading impact for the following test cases? This is more like a general question.

innobead (approved):
LGTM, but have one comment which we probably need to enhance later.

manager/integration/tests/test_basic.py
c3y1huang: Yes, it will get reset, ref [here](https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/common.py#L1378).

innobead: Do we use that fixture in this case?

c3y1huang: test_allow_volume_creation_with_degraded_availability_restore(set_random_backupstore, [client](https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/common.py#L1337), core_api, volume_name, csi_pv, pvc, pod, pod_make)

innobead: ah thanks :)

innobead:
@c3y1huang let's backport to 1.2 and 1.1 if needed.

PR #905, opened 2022-03-15
Update test_recurring_jobs_maximum_retain

derekbit:
Capture and compare the error message from the schema validation.

Longhorn 3562


innobead (approved):

innobead:
@longhorn/qa FYI. Some changes are probably needed like this after we introduce CRD OpenAPI schema validation.

PR #912, opened 2022-03-22
Print pytest logs in jenkins console output

yangchiu:
For task [#3740](https://github.com/longhorn/longhorn/issues/3740)

To fetch the test report, we have to "kubectl cp" the report, but there is no way to "kubectl cp" a completed pod, so another keepalive pod is created and we can get the test report from it.
    
After resolved the "kubectl cp" issue above, we don't have to use "kubectl logs" to print out the test report anymore. We can use "kubectl logs" to print pytest logs normally.

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -60,7 +60,7 @@ metadata:
     longhorn-test: test-job
 spec:
   containers:
-  - name: longhorn-test-pod
+  - name: longhorn-test
     image: longhornio/longhorn-manager-test:master-head
 #    args: [
 #           "-x", "-s",
@@ -74,7 +74,7 @@ spec:
       privileged: true
     env:
     - name: LONGHORN_JUNIT_REPORT_PATH
-      value: /integration/tests/longhorn-test-junit-report.xml
+      value: /tmp/test-report/longhorn-test-junit-report.xml
     - name: LONGHORN_BACKUPSTORES
       value: "s3://backupbucket@us-east-1/backupstore$minio-secret, nfs://longhorn-test-nfs-svc.default:/opt/backupstore"
     - name: LONGHORN_BACKUPSTORE_POLL_INTERVAL
@@ -94,6 +94,14 @@ spec:
     - name: longhorn
       mountPath: /var/lib/longhorn/
       mountPropagation: Bidirectional
+    - name: test-report
+      mountPath: /tmp/test-report
+  - name: longhorn-test-report
+    image: busybox
+    command: [ "tail", "-f", "/dev/null" ]
+    volumeMounts:
+    - name: test-report
+      mountPath: /tmp/test-report
   dnsConfig:
     nameservers:
       - 8.8.8.8
@@ -111,5 +119,8 @@ spec:
   - name: longhorn
     hostPath:
       path: /var/lib/longhorn/
+  - name: test-report
+    hostPath:
+      path: /tmp/test-report/
   restartPolicy: Never
   serviceAccountName: longhorn-test-service-account
--- a/manager/integration/tests/run.sh
+++ b/manager/integration/tests/run.sh
@@ -1,13 +1,3 @@
 #!/bin/bash
 
-if [[ $@ =~ "--junitxml=" ]] ; then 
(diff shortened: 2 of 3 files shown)


cchien816 (changes requested):
LGTM but maybe reduce the commit numbers will be a good idea, thank you

manager/integration/tests/run.sh
```diff
@@ -1,11 +1,9 @@
 #!/bin/bash
 
-if [[ $@ =~ "--junitxml=" ]] ; then 
-  pytest -v "$@" > /tmp/longhorn-pytest 2>&1
-  
-  cat ${LONGHORN_JUNIT_REPORT_PATH}
+if [[ $@ =~ "--junitxml=" ]] ; then
```
innobead, line 3: The implementation of this script is a little confusing because both conditions are almost the same except set option & flake8, and actually flake8 here seems not quite useful.

I would suggest have `pytest -v "$@" here to simplify it.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -188,16 +188,16 @@ run_longhorn_upgrade_test(){
                               '''
 
 	## generate upgrade_test pod manifest
-    yq e 'select(.spec.containers[0] != null).spec.containers[0].args=['"${PYTEST_COMMAND_ARGS}"']' "${LONGHORN_TESTS_MANIFEST_FILE_PATH}" > ${LONGHORN_UPGRADE_TESTS_MANIFEST_FILE_PATH}
-    yq e -i 'select(.spec.containers[0] != null).spec.containers[0].image="'${LONGHORN_TESTS_CUSTOM_IMAGE}'"' ${LONGHORN_UPGRADE_TESTS_MANIFEST_FILE_PATH}
-    yq e -i 'select(.spec.containers[0] != null).metadata.name="'${LONGHORN_UPGRADE_TEST_POD_NAME}'"' ${LONGHORN_UPGRADE_TESTS_MANIFEST_FILE_PATH}
+    yq e 'select(.metadata.labels.test-report == null and .spec.containers[0] != null).spec.containers[0].args=['"${PYTEST_COMMAND_ARGS}"']' "${LONGHORN_TESTS_MANIFEST_FILE_PATH}" > ${LONGHORN_UPGRADE_TESTS_MANIFEST_FILE_PATH}
```
innobead, line 191: No need to add this selection if you move the test report container to the original pod.

innobead (changes requested):

innobead:
> LGTM but maybe reduce the commit numbers will be a good idea, thank you

@yangchiu this is important to make comments in the PR meaningful, then it will be easy for review.

manager/integration/deploy/test.yaml
```diff
@@ -60,7 +60,7 @@ metadata:
     longhorn-test: test-job
 spec:
   containers:
-  - name: longhorn-test-pod
+  - name: longhorn-test-container
```
innobead, line 63:
```suggestion
  - name: longhorn-test
```

manager/integration/deploy/test.yaml
```diff
@@ -94,6 +94,14 @@ spec:
     - name: longhorn
       mountPath: /var/lib/longhorn/
       mountPropagation: Bidirectional
+    - name: test-report
```
innobead, line 97:
```suggestion
    - name: longhorn-test-report
```

innobead (changes requested):
LGTM. Minor requests to change.

manager/integration/deploy/test.yaml
innobead: Remember to update other places accordingly as well.

yangchiu: sorry but just want to confirm something:
rename container name "keepalive" to "longhorn-test-report"
or
rename volume name "test-report" to "longhorn-test-report" ?

innobead: yeah, container name, I think I marked the wrong place :)

innobead (approved):

yangchiu:
Should we backport this to v1.1.x and v1.2.x ?
Since when we run test for v1.1.x or v1.2.x, it would fetch old longhornio/longhorn-manager-test image contains old run.sh which has been modified in this commit.

PR #913, opened 2022-03-22
Implement test_autosalvage_with_data_locality_enabled()

cchien816:

Implement test_autosalvage_with_data_locality_enabled()

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -61,6 +61,7 @@ from common import make_deployment_with_pvc # NOQA
 from common import get_apps_api_client, create_and_wait_deployment
 from common import wait_delete_pod
 from common import wait_pod, exec_command_in_pod
+from common import RETRY_EXEC_COUNTS, RETRY_EXEC_INTERVAL
 
 from backupstore import set_random_backupstore # NOQA
 from backupstore import backupstore_cleanup
@@ -2586,8 +2587,7 @@ def test_engine_image_missing_on_some_nodes():
     pass
 
 
-@pytest.mark.skip(reason="TODO")  # NOQA
-def test_autosalvage_with_data_locality_enabled():
+def test_autosalvage_with_data_locality_enabled(client, core_api, make_deployment_with_pvc, volume_name, pvc): # NOQA
     """
     This e2e test follows the manual test steps at:
     https://github.com/longhorn/longhorn/issues/2778#issue-939331805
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,88 @@ def test_autosalvage_with_data_locality_enabled():
...
+    test_data = generate_random_data(VOLUME_RWTEST_SIZE)
+    write_pod_volume_data(core_api, pod_names[0], test_data, filename='test')
+    create_snapshot(client, volume_name)
+
+    # Step6
+    node_1_name = node_1["name"]
+    ret = core_api.list_namespaced_pod(namespace="longhorn-system")
+    for item in ret.items:
+        if "instance-manager-r" in item.metadata.name:
+            if item.metadata.labels["longhorn.io/node"] == node_1_name:
```
innobead, line 2661: You can try to use `label_selector` in `list_namespaced_pod`, so you don't need to have this forloop to find out the matched IM pod.

manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,88 @@ def test_autosalvage_with_data_locality_enabled():
...
+                imr_name = item.metadata.name
+                break
+
+    target_pod = \
+        core_api.read_namespaced_pod(
+            name=imr_name, namespace='longhorn-system')
+    exec_cmd = ["kubectl", "delete", "pod",  imr_name, "-n", "longhorn-system"]
```
innobead, line 2668: Because we already use python lib, so if it's a general case, suggest just use `delete_namespaced_pod` instead. You can even use sync way to wait by this function as well, so probably you don't need to use `wait_delete_pod`.

manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,88 @@ def test_autosalvage_with_data_locality_enabled():
...
+    subprocess.check_output(exec_cmd)
+    wait_delete_pod(core_api, target_pod.metadata.uid)
+
+    # Step7
+    target_pod = \
+        core_api.read_namespaced_pod(name=pod_names[0], namespace='default')
+    wait_delete_pod(core_api, target_pod.metadata.uid)
```
innobead, line 2675: Why this repeated? You have waited for the deletion above.

manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,88 @@ def test_autosalvage_with_data_locality_enabled():
...
+
+    command = 'cat /data/test'
+    to_be_verified_data = exec_command_in_pod(
+        core_api, command, pod_names[0], 'default')
+
+    assert test_data == to_be_verified_data
+
+    # Step8
+    node_2_name = node_2["name"]
+    ret = core_api.list_namespaced_pod(namespace="longhorn-system")
```
innobead, line 2687: The same. You probably can leverage `field_selector` to find out the pod.

innobead (changes requested):
In general, LGTM. A few comments needed to resolve, also because we already use k8s client lib, so it's better to leverage it well if possible.

manager/integration/tests/test_ha.py
cchien816: HI @innobead ,
I modified code to use label_selector to find specific pod without iterate by for-loop

manager/integration/tests/test_ha.py
cchien816: HI @innobead ,
I modified code to use label_selector and field_selector to find specific pod without iterate by for-loop , thank you

manager/integration/tests/test_ha.py
cchien816: HI @innobead ,
This pod is deployment pod which not the same with previous instance manager pod.
I think script should wait deployment pod restart when auto salvage happen

manager/integration/tests/test_ha.py
cchien816: Follow our suggestion, I modified code to use `delete_and_wait_pod` to simplify code

manager/integration/tests/test_ha.py
innobead: Yeah, you right 👍

manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,86 @@ def test_autosalvage_with_data_locality_enabled():
...
+
+    # Step7
+    target_pod = \
+        core_api.read_namespaced_pod(name=pod_names[0], namespace='default')
+    wait_delete_pod(core_api, target_pod.metadata.uid)
+    pod_names = common.get_deployment_pod_names(core_api, deployment)
+    wait_pod(pod_names[0])
+
+    command = 'cat /data/test'
+    to_be_verified_data = exec_command_in_pod(
```
innobead, line 2677:
```suggestion
    verified_data = exec_command_in_pod(
```

manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,86 @@ def test_autosalvage_with_data_locality_enabled():
...
+        core_api, command, pod_names[0], 'default')
+
+    assert test_data == to_be_verified_data
```
innobead, line 2680:
```suggestion
    assert test_data == verified_data
```

manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,86 @@ def test_autosalvage_with_data_locality_enabled():
...
+    apps_api = get_apps_api_client()
+    create_and_wait_deployment(apps_api, deployment)
+
+    pod_names = common.get_deployment_pod_names(core_api, deployment)
+    test_data = generate_random_data(VOLUME_RWTEST_SIZE)
+    write_pod_volume_data(core_api, pod_names[0], test_data, filename='test')
+    create_snapshot(client, volume_name)
+
+    # Step6
+    node_1_name = node_1["name"]
```
innobead, line 2657:
```suggestion
```

manager/integration/tests/test_ha.py
```diff
@@ -2616,7 +2616,86 @@ def test_autosalvage_with_data_locality_enabled():
...
+    labels = "longhorn.io/node={}\
+            , longhorn.io/instance-manager-type=replica"\
+            .format(node_1_name)
```
innobead, line 2660:
```suggestion
            .format(node_1["name"])
```

innobead (changes requested):
Just a few convention request to change.

manager/integration/tests/test_ha.py
cchien816: HI @innobead , 
I modified some conventions you mentioned for review, thank you.

innobead (approved):
LGTM. Another test case added, well done 👍

innobead:
@cchien816 Please backport to 1.2.x and 1.1.x.

PR #921, opened 2022-03-29
Fix integration tests fail to publish junit report

yangchiu:
Deal with streaming-connection-idle-timeout to fix integration tests fail to publish junit report.

By default, streaming-connection-idle-timeout is set to 4 hours, so the "kubectl logs -f" would lost the connection after 4 hours even though the pod is still running. But it also isn't recommended to disable the streaming-connection-idle-timeout.

So keep tracking the status of the longhorn-tests pod and re-trigger "kubectl logs -f".


meldafrawi (approved):
LGTM

innobead (approved):

PR #922, opened 2022-03-30
Implement test_cloning_with_backing_image

cchien816:

Implement `test_cloning_with_backing_image`
https://github.com/longhorn/longhorn/issues/3797

### DIFF
--- a/manager/integration/tests/test_cloning.py
+++ b/manager/integration/tests/test_cloning.py
@@ -14,6 +14,7 @@ from common import delete_and_wait_pvc, wait_for_volume_attached
 from common import generate_random_suffix, wait_for_volume_endpoint
 from common import wait_for_snapshot_count, DATA_SIZE_IN_MB_3
 from common import get_clone_volume_name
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_cloning.py
```diff
@@ -289,6 +289,64 @@ def test_cloning_with_backing_image():  # NOQA
     3. Clean up the test
     """
 
+    # Create storage class with backing image
```
innobead, line 292: I believe this test is the same as `test_cloning_basic` except the storage class is different. 

I would suggest reusing `test_cloning_basic` w/ some changes by providing a storage class parameter of backing image instead of repeating the similar steps again?

innobead (changes requested):

manager/integration/tests/test_cloning.py
cchien816: Hi @innobead ,

Yes this case is from `test_cloning_basic`, I will modify code to use `test_cloning_basic` to achieve test goal

innobead (approved):

PR #926, opened 2022-04-01
Fix timing issue in test_backup_status_for_unavailable_replicas

yangchiu:
For [#3808](https://github.com/longhorn/longhorn/issues/3808)

when trying to find replica for a backup, the replica name is empty at first. 
Have to wait a few seconds (~5s) to get the correct value.


innobead (approved):

innobead:
See if need to backport to 1.2

PR #930, opened 2022-04-06
integration: Add skeleton test_default_storage_class_syncup

shuo-wu:
Longhorn/longhorn#3821


innobead (approved):
cc @cchien816 @yangchiu

PR #932, opened 2022-04-07
Implement test csi snapshot with bak param

cchien816:
Implement `test_csi_snapshot_with_bak_param`

- Remove `deploy_csi_snapshot_and_classes` from common.py because  `volumesnapshotclass` in test_csi_snapshotter.py can deploy specific volumesnapshotclass and CSI snapshot CRDs already installed in `longhorn-setup.sh`
- Reuse `test_csi_volumesnapshot_basic`

### DIFF
--- a/manager/integration/tests/test_csi_snapshotter.py
+++ b/manager/integration/tests/test_csi_snapshotter.py
@@ -32,7 +32,7 @@ def volumesnapshotclass(request):
         manifests = []
 
         @staticmethod
-        def create_volumesnapshotclass(name, deletepolicy):
+        def create_volumesnapshotclass(name, deletepolicy, snapshot_type=None):
             manifest = {
                 'kind': 'VolumeSnapshotClass',
                 'apiVersion': 'snapshot.storage.k8s.io/v1beta1',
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_csi_snapshotter.py
```diff
@@ -32,7 +32,7 @@ class VolumeSnapshotClassFactory():
         manifests = []
 
         @staticmethod
-        def create_volumesnapshotclass(name, deletepolicy):
+        def create_volumesnapshotclass(name, deletepolicy, param=None):
```
innobead, line 35: Suggest using an explanatory parameter instead of a general one. ex: `snapshot_type` instead of `param`.

You should change other places as well.

innobead (changes requested):

manager/integration/tests/test_csi_snapshotter.py
cchien816: Hi @innobead ,
I changed parameter name for you review, thanks

innobead (approved):

PR #936, opened 2022-04-12
Add tests for cluster autoscaler

c3y1huang:
https://github.com/longhorn/longhorn/issues/2203

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -87,6 +87,8 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: spec.nodeName
+    - name: MANAGED_K8S_CLUSTER
+      value: "false"
     volumeMounts:
     - name: dev
       mountPath: /dev
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -29,7 +29,7 @@ rules:
   resources: ["customresourcedefinitions"]
   verbs: ["*"]
 - apiGroups: ["policy"]
-  resources: ["podsecuritypolicies"]
+  resources: ["podsecuritypolicies", "poddisruptionbudgets"]
   verbs: ["*"]
 - apiGroups: ["snapshot.storage.k8s.io"]
   resources: ["*"]
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -69,6 +69,8 @@ RETRY_BACKUP_INTERVAL = 1
 RETRY_SNAPSHOT_INTERVAL = 1
 RETRY_EXEC_COUNTS = 30
 RETRY_EXEC_INTERVAL = 5
+RETRY_AUTOSCALER_INTERVAL = 30
+RETRY_AUTOSCALER_COUNTS = 10*60//RETRY_AUTOSCALER_INTERVAL  # 10 minutes
 
 LONGHORN_NAMESPACE = "longhorn-system"
 
@@ -195,6 +197,8 @@ SETTING_SNAPSHOT_FAST_REPLICA_REBUILD_ENABLED = "fast-replica-rebuild-enabled"
 SETTING_CONCURRENT_VOLUME_BACKUP_RESTORE = \
     "concurrent-volume-backup-restore-per-node-limit"
 SETTING_NODE_SELECTOR = "system-managed-components-node-selector"
+SETTING_K8S_CLUSTER_AUTOSCALER_ENABLED = \
+    "kubernetes-cluster-autoscaler-enabled"
 
 SNAPSHOT_DATA_INTEGRITY_IGNORED = "ignored"
 SNAPSHOT_DATA_INTEGRITY_DISABLED = "disabled"
@@ -245,6 +249,11 @@ BACKUP_BLOCK_SIZE = 2 * Mi
 DEPRECATED_K8S_ZONE_LABEL = "failure-domain.beta.kubernetes.io/zone"
 
 K8S_ZONE_LABEL = "topology.kubernetes.io/zone"
+
+K8S_CLUSTER_AUTOSCALER_EVICT_KEY = \
+    "cluster-autoscaler.kubernetes.io/safe-to-evict"
+K8S_CLUSTER_AUTOSCALER_SCALE_DOWN_DISABLED_KEY = \
+    "cluster-autoscaler.kubernetes.io/scale-down-disabled"
 
 BACKING_IMAGE_SOURCE_TYPE_DOWNLOAD = "download"
 BACKING_IMAGE_SOURCE_TYPE_FROM_VOLUME = "export-from-volume"
@@ -3001,7 +3010,9 @@ def reset_node(client, core_api):
             print("\nException when reset node scheduling and tags", node)
             print(e)
 
-    reset_longhorn_node_zone(client)
+    managed_k8s_cluster = os.getenv("MANAGED_K8S_CLUSTER").lower() == 'true'
+    if not managed_k8s_cluster:
+        reset_longhorn_node_zone(client)
 
 
 def reset_longhorn_node_zone(client):
@@ -3014,7 +3025,6 @@ def reset_longhorn_node_zone(client):
 
 
 def wait_longhorn_node_zone_reset(client):
-
     lh_nodes = client.list_node()
     node_names = map(lambda node: node.name, lh_nodes)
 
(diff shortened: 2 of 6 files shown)


manager/integration/tests/common.py
```diff
@@ -60,6 +60,8 @@
 RETRY_SNAPSHOT_INTERVAL = 1
 RETRY_EXEC_COUNTS = 30
 RETRY_EXEC_INTERVAL = 5
+RETRY_AUTOSCALER_INTERVAL = 30
+RETRY_AUTOSCALER_COUNTS = 5*60//RETRY_AUTOSCALER_INTERVAL  # 5 minutes
```
innobead, line 64: Should it be `RETRY_AUTOSCALER_COUNTS = 5*60` instead?

kaxing (approved):
@c3y1huang  Thanks for the improvement and the part to cleanup the cluster.

The case had been throughly verified on GKE with following spec:
```
gcloud container clusters create longhorn-ca-test \
--zone asia-east1-a \
--machine-type n1-standard-4 \
--image-type ubuntu \
--cluster-version 1.18 \
--enable-autoscaling \
--num-nodes 3 \
--min-nodes 2 \
--max-nodes 9
```
A note for @longhorn/qa, do make sure the minimum node is less than 3 and the beginning with 3 nodes, so the test can see the nodepool number changes.

kaxing:
Tested latest revision on GKE with following spec:
```
gcloud container clusters create $GKE_CLUSTER_NAME \
--zone asia-east1-a \
--machine-type n1-standard-4 \
--image-type ubuntu \
--cluster-version 1.20 \
--enable-autoscaling \
--autoscaling-profile optimize-utilization \
--num-nodes 3 \
--min-nodes 2 \
--max-nodes 9
```
Time elapsed: 2616.83s (0:43:36)

kaxing:
Tested latest revision on AKS with following spec:
```
az aks create \
  --resource-group $AZ_RESOURCE_GROUP \
  --name $AZ_CLUSTER_NAME \
  --vm-set-type VirtualMachineScaleSets \
  --node-vm-size Standard_DS2_v2 \
  --node-count 3 \
  --min-count 2 \
  --max-count 7 \
  --enable-cluster-autoscaler \
  --cluster-autoscaler-profile skip-nodes-with-system-pods=false scan-interval=10s scale-down-delay-after-add=1m scale-down-unneeded-time=1m skip-nodes-with-local-storage=false
```
Time elapsed: 3548.73s (0:59:08)

c3y1huang:
Verified `master-6280eb0` on GKE. I'm also testing on AKS, however, the scaling seems to take a lot longer and makes the test result flaky, especially test_cluster_autoscaler.py::test_cluster_autoscaler_backing_image. Probably need to check on the AKS cluster setup command.
```
platform linux -- Python 3.9.10, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/bin/python3.9
cachedir: .pytest_cache
plugins: order-1.0.1, repeat-0.9.1
collecting ... collected 3 items

test_cluster_autoscaler.py::test_cluster_autoscaler PASSED
test_cluster_autoscaler.py::test_cluster_autoscaler_all_nodes_with_volume_replicas PASSED
test_cluster_autoscaler.py::test_cluster_autoscaler_backing_image PASSED

======================== 3 passed in 2837.56s (0:47:17) ========================
```

c3y1huang:
> Verified `master-6280eb0` on GKE. I'm also testing on AKS, however, the scaling seems to take a lot longer and makes the test result flaky, especially `test_cluster_autoscaler.py::test_cluster_autoscaler_backing_image`. Probably need to check on the AKS cluster setup command.

Found the culprit `I0610 18:02:02.382332 1 cluster.go:169] Fast evaluation: node aks-c3y1amuwtn-29569835-vmss000002 cannot be removed: default/longhorn-test-minio is not replicated`. MinIO was on the node waiting to get downscaled, after removing MinIO the scaling seems to be working.

Testing in progress...

c3y1huang:
> AKS testing is in progress, I will update the result later.

```
platform linux -- Python 3.9.10, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/bin/python3.9
cachedir: .pytest_cache
plugins: repeat-0.9.1, order-1.0.1
collecting ... collected 3 items

test_cluster_autoscaler.py::test_cluster_autoscaler PASSED
test_cluster_autoscaler.py::test_cluster_autoscaler_all_nodes_with_volume_replicas PASSED
test_cluster_autoscaler.py::test_cluster_autoscaler_backing_image PASSED

======================== 3 passed in 3575.82s (0:59:35) ========================
```

manager/integration/tests/test_cluster_autoscaler.py
```diff
@@ -0,0 +1,389 @@
...
+    # Cleanup
+    def finalizer():
+        configure_node_scale_down(core_api, client.list_node(), disable="")
+        set_node_cordon(core_api, host_id, False)
+    request.addfinalizer(finalizer)
+
+    host_id = get_self_host_id()
+    host_node = client.by_id_node(host_id)
+    configure_node_scale_down(core_api, [host_node], disable="true")
+    set_node_cordon(core_api, host_id, True)
```
shuo-wu, line 76: Why not do the scaling down disabling and cordon for all existing/original nodes?

manager/integration/tests/test_cluster_autoscaler.py
```diff
@@ -0,0 +1,389 @@
...
+
+    update_setting(client, SETTING_K8S_CLUSTER_AUTOSCALER_ENABLED, "true")
+
+    nodes = client.list_node()
```
shuo-wu, line 80: NIT: Rename to original_nodes or existing_nodes.

manager/integration/tests/test_cluster_autoscaler.py
```diff
@@ -0,0 +1,389 @@
...
+         (double the node number, not including host node)
+    Then Cluster should have double the node number.
+
+    When Trigger CA to scale-down by decrease deployment replicas.
+    Then Cluster should scale-down to original node number.
+    """
+    # Cleanup
+    def finalizer():
+        configure_node_scale_down(core_api, client.list_node(), disable="")
+        set_node_cordon(core_api, host_id, False)
```
shuo-wu, line 70: Should wait for the node count back to normal as well.

manager/integration/tests/test_cluster_autoscaler.py
```diff
@@ -0,0 +1,389 @@
...
+
+    When Trigger CA to scale-up by increase deployment replicas.
+         (double the node number, not including host node)
+    Then Cluster should have double the node number.
+
+    When Annotate new nodes with
+         `cluster-autoscaler.kubernetes.io/scale-down-disabled`.
+         (this ensures scale-down only the old nodes)
+    And Trigger CA to scale-down by decrease deployment replicas.
+    Then Cluster should have original node number + 1 blocked node.
```
shuo-wu, line 130: I prefer not to touch/scale down the old nodes. The backup target pods are running on those old nodes.

shuo-wu (commented):
How can we guarantee that the cluster autoscaler won't shut down the original (3) nodes when the annotation does not exist? Sorry for forgetting some implementation details...

innobead:
@c3y1huang resolve the conflict, then let's move this forward. cc @longhorn/qa

c3y1huang:
> @c3y1huang resolve the conflict, then let's move this forward. cc @longhorn/qa

Resolved.

innobead (changes requested):
In general, LGTM. Just a few things needed to fix before merging. 

This PR also cover a test case to let Cluster Autoscaler to respect data-related component, backing image manager pod on nodes when an image exists.

manager/integration/tests/common.py
```diff
@@ -3029,7 +3039,9 @@ def get_k8s_zone_label():
 
     k8s_ver_minor = k8s_ver_data.minor
 
-    if int(k8s_ver_minor) >= 17:
+    # minor version is not always integer convertable:
+    #   ValueError: invalid literal for int() with base 10: '20+'
```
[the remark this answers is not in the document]
c3y1huang:
```
    # k8s_ver_minor no needs to be an int
    # it could be "24+" in eks
    if int(re.sub('\\D', '', k8s_ver_minor)) >= 17:
```

Ref: https://github.com/longhorn/longhorn-tests/pull/1255/files#diff-a2154691ec50c6925b7f385cf30c42f33b2a3c935192c894bfe4b32fe3ee49dcR3034-R3036

manager/integration/deploy/test.yaml
```diff
@@ -87,6 +87,8 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: spec.nodeName
+    - name: MANAGED_K8S_CLUSTER
+      value: "true"
```
yangchiu, line 91: LGTM. But need to set `MANAGED_K8S_CLUSTER` to `false` to avoid it running by default.

yangchiu (approved):
LGTM. Suggest merge this first. If it needs different options in different managed k8s services to let these test cases pass, we can tackle them later.

PR #939, opened 2022-04-15
Add Benchmark test

yangchiu:
to run kbench and record results periodically

for issue [#3540](https://github.com/longhorn/longhorn/issues/3540)

### DIFF
--- a/benchmark_test/scripts/benchmark-test-setup.sh
+++ b/benchmark_test/scripts/benchmark-test-setup.sh
@@ -1,0 +1,154 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+
+set_kubeconfig_envvar(){
+  ARCH=${1}
+  BASEDIR=${2}
+
+  if [[ ${ARCH} == "amd64" ]] ; then
+    if [[ ${TF_VAR_k8s_distro_name} == [rR][kK][eE] ]]; then
+      export KUBECONFIG="${BASEDIR}/kube_config_rke.yml"
+    elif [[ ${TF_VAR_k8s_distro_name} == [rR][kK][eE]2 ]]; then
+      export KUBECONFIG="${BASEDIR}/terraform/aws/${DISTRO}/rke2.yaml"
+    else
+      export KUBECONFIG="${BASEDIR}/terraform/aws/${DISTRO}/k3s.yaml"
+    fi
+  elif [[ ${ARCH} == "arm64"  ]]; then
+    export KUBECONFIG="${BASEDIR}/terraform/aws/${DISTRO}/k3s.yaml"
+  fi
+}
+
+
+wait_local_path_provisioner_status_running(){
+  local RETRY_COUNTS=10  # in seconds
+  local RETRY_INTERVAL="10s"
+
+  RETRIES=0
+  while [[ -n `kubectl get pods -n local-path-storage --no-headers | awk '{print $3}' | grep -v Running` ]]; do
+    echo "local-path-provisioner is still installing ... re-checking in 10s"
+    sleep ${RETRY_INTERVAL}
+    RETRIES=$((RETRIES+1))
+
+    if [[ ${RETRIES} -eq ${RETRY_COUNTS} ]]; then echo "Error: local-path-provisioner installation timeout"; exit 1 ; fi
+  done
+}
+
+
+install_local_path_provisioner(){
+  kubectl apply -f https://raw.githubusercontent.com/rancher/local-path-provisioner/v0.0.22/deploy/local-path-storage.yaml
+  wait_local_path_provisioner_status_running
+}
+
+
+wait_longhorn_status_running(){
+  local RETRY_COUNTS=10  # in minutes
+  local RETRY_INTERVAL="1m"
+
+  RETRIES=0
+  while [[ -n `kubectl get pods -n longhorn-system --no-headers | awk '{print $3}' | grep -v Running` ]]; do
+    echo "Longhorn is still installing ... re-checking in 1m"
+    sleep ${RETRY_INTERVAL}
+    RETRIES=$((RETRIES+1))
+
+    if [[ ${RETRIES} -eq ${RETRY_COUNTS} ]]; then echo "Error: longhorn installation timeout"; exit 1 ; fi
+  done
+}
+
+
+install_longhorn(){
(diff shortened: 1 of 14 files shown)


derekbit:
I'm thinking if we can measure the average CPU usage of each pod.

performance_test/scripts/fio-cmp.yaml
```diff
@@ -0,0 +1,81 @@
...
+spec:
+  volumeMode: Filesystem
+  #volumeMode: Block
+  #storageClassName: longhorn  # replace with your storage class
+  storageClassName: local-path
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 55Gi
```
derekbit, line 14: What's the memory size in each node?
Typically, we use 5x of memory size as the storage size, because we hope the io sometimes from on-disk data and sometime from memory in buffer io test.

yangchiu: the default ec2 instance used is c5d.2xlarge, it has 16GB of memory

derekbit: Then we can update the storage to `80Gi`.

yangchiu:
> I'm thinking if we can measure the average CPU usage of each pod.
> 
> There are some issues related to the high CPU usage of instance-managers. Maybe we can provide more statistics in the benchmark. [longhorn/longhorn#3889](https://github.com/longhorn/longhorn/issues/3889)

There is a CPU idleness profiling option in [kbench](https://github.com/yasker/kbench), but it's disabled by default because:

`CPU_IDLE_PROF environmental variable: the CPU idleness profiling measures the CPU idleness, but it introduces extra overhead and reduces the storage performance. By default, the flag is disabled.`

Should we enable it?

performance_test/scripts/fio-cmp.yaml
yangchiu: [test size must be 10% smaller than the PVC size due to filesystem also took space](https://github.com/yasker/kbench/blob/main/deploy/fio.yaml), so maybe test size = 80Gi, storage size ~= 89Gi ?

derekbit: Yeah. But this is a rough value. 4x-5x is okay.

derekbit:
> > I'm thinking if we can measure the average CPU usage of each pod.
> > There are some issues related to the high CPU usage of instance-managers. Maybe we can provide more statistics in the benchmark. [longhorn/longhorn#3889](https://github.com/longhorn/longhorn/issues/3889)
> 
> There is a CPU idleness profiling option in [kbench](https://github.com/yasker/kbench), but it's disabled by default because:
> 
> `CPU_IDLE_PROF environmental variable: the CPU idleness profiling measures the CPU idleness, but it introduces extra overhead and reduces the storage performance. By default, the flag is disabled.`
> 
> Should we enable it?

No, we can't.
Except for this way, maybe need a pod measuring CPU usage using `kubectl top` periodically.

innobead (commented):
Is it possible to use symoblic link terraform/aws to https://github.com/longhorn/longhorn-tests/tree/master/test_framework/terraform/aws? so we don't need to copy the same TF files again?

performance_test/scripts/fio-cmp.yaml
innobead: Can we make this configurable via Jenkins job?

performance_test/scripts/fio-cmp.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: kbench-pvc-1
+spec:
+  volumeMode: Filesystem
+  #volumeMode: Block
+  #storageClassName: longhorn  # replace with your storage class
+  storageClassName: local-path
```
innobead, line 9: For comparison, we probably can make the value of storage class configurable later, so it will not just be limited to local path and longhorn, or even only 2 volumes. 

We can do this when we add other solutions to compare in the future.

innobead (changes requested):
In general, LGTM, but some comments needed to resolve first.

yangchiu:
Fixed:
(1) make TEST_SIZE configurable
(2) reuse terraform scripts from regression tests

For CPU usage measurement, tracked by https://github.com/longhorn/longhorn/issues/3903

innobead (dismissed):
LGTM

benchmark_test/scripts/benchmark-test-setup.sh
```diff
@@ -0,0 +1,154 @@
...
+    if [[ ${RETRIES} -eq ${RETRY_COUNTS} ]]; then echo "Error: local-path-provisioner installation timeout"; exit 1 ; fi
+  done
+}
+
+
+install_local_path_provisioner(){
+	kubectl apply -f https://raw.githubusercontent.com/rancher/local-path-provisioner/v0.0.22/deploy/local-path-storage.yaml
+	wait_local_path_provisioner_status_running
+}
+
```
innobead, line 44:
```suggestion
install_local_path_provisioner(){
  kubectl apply -f https://raw.githubusercontent.com/rancher/local-path-provisioner/v0.0.22/deploy/local-path-storage.yaml
  wait_local_path_provisioner_status_running
}

```

benchmark_test/scripts/terraform-setup.sh
```diff
@@ -0,0 +1,20 @@
...
+	if [[ ${TF_VAR_k8s_distro_name} == [rR][kK][eE] ]]; then
+    	terraform -chdir=${TF_VAR_tf_workspace}/terraform/aws/${DISTRO} apply -auto-approve -no-color -refresh-only
+		terraform -chdir=${TF_VAR_tf_workspace}/terraform/aws/${DISTRO} output -raw rke_config > ${TF_VAR_tf_workspace}/rke.yml
+		sleep 30
+		rke up --config ${TF_VAR_tf_workspace}/rke.yml
+	fi
+else
+	terraform -chdir=${TF_VAR_tf_workspace}/terraform/aws/${DISTRO} init
+	terraform -chdir=${TF_VAR_tf_workspace}/terraform/aws/${DISTRO} apply -auto-approve -no-color
+fi
```
innobead, line 18: The formatting is tab instead of space. Suggest having the same way as other scripts.

innobead:
@yangchiu Two comments needed to resolve, then we can merge this. Thanks!

innobead (approved):

yangchiu:
Tabs are replaced with spaces.

For cleaning up orphaned resources, [#3913](https://github.com/longhorn/longhorn/issues/3913) has been created to track this.

PR #940, opened 2022-04-18
integration: Add orphan tests

derekbit:
[Longhorn 685](https://github.com/longhorn/longhorn/issues/685)

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -173,6 +173,7 @@ SETTING_TAINT_TOLERATION = "taint-toleration"
 SETTING_BACKING_IMAGE_CLEANUP_WAIT_INTERVAL = \
     "backing-image-cleanup-wait-interval"
 SETTING_DISABLE_REVISION_COUNTER = "disable-revision-counter"
+SETTING_ORPHAN_AUTO_DELETION = "orphan-auto-deletion"
 
 CSI_UNKNOWN = 0
 CSI_TRUE = 1
--- a/manager/integration/tests/test_orphan.py
+++ b/manager/integration/tests/test_orphan.py
@@ -1,0 +1,762 @@
+import pytest
+import os
+import time
+import random
+import string
+
+from common import core_api, client # NOQA
+from common import Gi, SIZE
+from common import volume_name # NOQA
+from common import SETTING_ORPHAN_AUTO_DELETION
+from common import RETRY_COUNTS, RETRY_INTERVAL_LONG
+from common import exec_nsenter
+from common import get_self_host_id
+from common import get_update_disks, wait_for_disk_update, cleanup_node_disks
+from common import create_and_check_volume, wait_for_volume_healthy
+from common import cleanup_volume_by_name
+from common import create_host_disk, cleanup_host_disks
+from common import wait_for_node_update
+from common import wait_for_disk_status
+
+
+def generate_random_id(num_bytes):
+    return ''.join(random.choice(string.ascii_lowercase + string.digits)
+                   for _ in range(num_bytes))
+
+
+def crate_disks_on_host(client, disk_names, request):  # NOQA
+    disk_paths = []
+
+    lht_hostId = get_self_host_id()
+    node = client.by_id_node(lht_hostId)
+    update_disks = get_update_disks(node.disks)
+
+    for name in disk_names:
+        disk_path = create_host_disk(client,
+                                     name,
+                                     str(Gi),
+                                     lht_hostId)
+        disk = {"path": disk_path, "allowScheduling": True}
+        update_disks[name] = disk
+        disk_paths.append(disk_path)
+
+    node = node.diskUpdate(disks=update_disks)
+    node = wait_for_disk_update(client, node.name, len(update_disks))
+
+    def finalizer():
+        delete_extra_disks_on_host(client, disk_names)
+        for disk_name in disk_names:
+            cleanup_host_disks(client, disk_name)
+
+    request.addfinalizer(finalizer)
+
+    return disk_paths
+
+
+def create_volume_with_replica_on_each_node(client, volume_name):  # NOQA
+    lht_hostId = get_self_host_id()
+
+    nodes = client.list_node()
+
+    volume = create_and_check_volume(client, volume_name, len(nodes), SIZE)
+    volume.attach(hostId=lht_hostId, disableFrontend=False)
+    wait_for_volume_healthy(client, volume_name)
+
+    return volume
+
+
+def create_orphaned_directories_on_host(volume, disk_paths, num_orphans):  # NOQA
+    lht_hostId = get_self_host_id()
+    paths = []
+    for replica in volume.replicas:
+        if replica.hostId != lht_hostId:
+            continue
+        for _ in range(num_orphans):
+            for i, disk_path in enumerate(disk_paths):
+                replica_dir_name = volume.name + "-" + generate_random_id(8)
+                path = os.path.join(disk_path, "replicas", replica_dir_name)
+                paths.append(path)
+                exec_nsenter("cp -a {} {}".format(replica.dataPath, path))
+
+    return paths
+
+
+def wait_for_orphan_delete(client, name):  # NOQA
+    for _ in range(RETRY_COUNTS):
+        orphans = client.list_orphan()
+        found = False
+        for orphan in orphans:
+            if orphan.name == name:
+                found = True
+                break
+        if not found:
+            break
+        time.sleep(RETRY_INTERVAL_LONG)
+    assert not found
+
+
+def delete_orphan(client, orphan_name):  # NOQA
+    orphan = client.by_id_orphan(orphan_name)
+    client.delete(orphan)
+    wait_for_orphan_delete(client, orphan_name)
+
+
+def delete_orphans(client):  # NOQA
+    orphans = client.list_orphan()
+    for orphan in orphans:
+        delete_orphan(client, orphan.name)
+
+
+def wait_for_orphan_count(client, number, retry_counts=120):  # NOQA
+    for _ in range(retry_counts):
+        orphans = client.list_orphan()
+        if len(orphans) == number:
+            break
+        time.sleep(RETRY_INTERVAL_LONG)
+    return len(orphans)
+
+
+def wait_for_file_count(path, number, retry_counts=120):
+    for _ in range(retry_counts):
+        count = exec_nsenter("ls {} | wc -l".format(path))
+        if int(count) == number:
+            break
+        time.sleep(RETRY_INTERVAL_LONG)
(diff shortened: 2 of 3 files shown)


innobead:
@derekbit PR check failed, need to fix that.

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,940 @@
...
+def cleanup_host_disk(client, *args):  # NOQA
+    # clean disk
+    for vol_name in args:
+        # umount disk
+        common.cleanup_host_disk(vol_name)
+        # clean volume
+        cleanup_volume(client, vol_name)
+
+
+def create_volume(client, vol_name, size, node_id, r_num):  # NOQA
```
innobead, line 45: This seems reusable. Could we move this to common.py instead?

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,940 @@
...
+    assert volumeByName.state == volume.state
+    assert volumeByName.created == volume.created
+
+    volume.attach(hostId=node_id)
+    volume = wait_for_volume_healthy(client, vol_name)
+
+    return volume
+
+
+def cleanup_volume(client, vol_name):  # NOQA
```
innobead, line 70: Same here, move to common.py.

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,940 @@
...
+    2. Create a volume and then attach to the current node
+    3. Create one valid orphaned replica directories by copying the active
+       replica directory
+    4. Create multiple invalid orphan replica directories with wrong naming
+       format
+    5. Verify orphan list only contains the orphan CR for valid orphaned
+       replica directory
+    """
+
+    lht_hostId = get_self_host_id()
```
innobead, line 93: Because we don't have BBD test case atm, please add a comment for each step at the beginning of it like "# Step 1" and it will be helpful for reading/development/debugging.

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,940 @@
...
+
+    update_disks = get_update_disks(node.disks)
+    update_disks[new_disk_volume_name] = new_disk
+    node = node.diskUpdate(disks=update_disks)
+    node = wait_for_disk_update(client, node.name, len(update_disks))
+
+    nodes = client.list_node()
+    volume = create_and_check_volume(client, volume_name, len(nodes), SIZE)
+    volume.attach(hostId=lht_hostId, disableFrontend=False)
+    wait_for_volume_healthy(client, volume_name)
```
innobead, line 114: This seems repeated in other test function as well, so I would suggest to add a helper function in this test module.

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,940 @@
...
+    for orphan in orphans:
+        delete_orphan(client, orphan.name)
+
+    # wait for 30 sec before finishing the deletion of on-disk data
+    time.sleep(30)
+
+    # The orphaned directories in the ready disk should be deleted
+    for path in orphaned_directories:
+        with pytest.raises(subprocess.CalledProcessError):
+            exec_nsenter("ls {}".format(path))
```
innobead, line 631: Could we use assert instead to verify the number of dirs?

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,940 @@
...
+
+    # The orphaned directories in the duplicated disk should not be deleted
+    for path in orphaned_directories_in_duplicated_disk:
+        exec_nsenter("ls {}".format(path))
```
innobead, line 635: Use assert instead.

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,940 @@
...
+            for disk_path in disk_paths:
+                path = os.path.join(disk_path, "replicas",
+                                    volume_name + "-" + generate_random_id(8))
+                orphaned_directories.append(path)
+                exec_nsenter("cp -a {} {}".format(replica.dataPath, path))
+
+    cleanup_volume(client, volume_name)
+
+    # Check orphan CRs
+    count = wait_for_orphan_count(client,
```
innobead, line 777: `wait_for_orphan_count` actually does the number check w/o assertion, so probably we can enhance this to provide more reusable assertions instead of having another following general assert to do the duplicated check again. 

However, It's fine for now and leave to the later test cases refactoring.

innobead (commented):
In general, LGTM. Left some comments especially for repeated code in some test functions. See if we can do some improvement before merging.

innobead:
cc @longhorn/qa for orphan replica cleanup, probably we can rely on this test suite directly.

manager/integration/tests/test_orphan.py
innobead: cc @longhorn/qa try to consider introducing assertion util functions in the future.

manager/integration/tests/test_orphan.py
derekbit: `exec_nsenter("ls {}".format(path))` is trying to check if the file exists one by one.
Using assert is more intuitive and undertandable, but `exec_nsenter` can raise the error, so I don't use `assert` here.

manager/integration/tests/test_orphan.py
derekbit: Addressed!

manager/integration/tests/test_orphan.py
derekbit: Addressed.

yangchiu (approved):

manager/integration/tests/common.py
```diff
@@ -4680,3 +4681,103 @@ def get_volume_running_replica_cnt(client, volume_name):  # NOQA
             client, volume_name, node.name, chk_running=True)
 
     return cnt
+
+
+def wait_for_orphan_count(client, number, retry_counts=120):
```
innobead, line 4686: I think we should organize common.py better based on different resources and util functions in the future for better maintenance. 

cc @yangchiu @cchien816

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,757 @@
...
+    nodes = client.list_node()
+
+    volume = create_and_check_volume(client, volume_name, len(nodes), SIZE)
+    volume.attach(hostId=lht_hostId, disableFrontend=False)
+    wait_for_volume_healthy(client, volume_name)
+
+    return volume
+
+
+def create_orphaned_directories(volume, disk_paths, num_orphans):  # NOQA
```
innobead, line 55: Change to `create_orphaned_directories_on_test_node`.

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,757 @@
...
+from common import generate_random_id
+from common import wait_for_orphan_count, delete_orphan
+from common import wait_for_file_count
+from common import cleanup_volume_by_name
+from common import create_host_disk, cleanup_host_disks
+from common import wait_for_node_update
+from common import wait_for_disk_status
+
+
+def create_disks(client, disk_names):  # NOQA
```
innobead, line 21: Change to `crate_disks_on_test_node`

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,757 @@
...
+    return paths
+
+
+def delete_orphans(client):  # NOQA
+    orphans = client.list_orphan()
+    for orphan in orphans:
+        delete_orphan(client, orphan.name)
+
+
+def delete_orphaned_directory(directories):  # NOQA
```
innobead, line 77: Change to `delete_orphaned_directory_on_test_node`

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,757 @@
...
+    for path in directories:
+        exec_nsenter("rm -rf {}".format(path))
+
+
+def delete_extra_disks(client, disk_names):  # NOQA
```
innobead, line 82: Change to `delete_extra_disks_on_test_node`

manager/integration/tests/common.py
```diff
@@ -4680,3 +4681,103 @@ def get_volume_running_replica_cnt(client, volume_name):  # NOQA
...
+    assert not found
+
+
+def delete_orphan(client, orphan_name):
+    orphan = client.by_id_orphan(orphan_name)
+    client.delete(orphan)
+    wait_for_orphan_delete(client, orphan_name)
+
+
+def create_volume(client, vol_name, size, node_id, r_num):
```
innobead, line 4731: This is good to stay in common.py.

manager/integration/tests/common.py
```diff
@@ -4680,3 +4681,103 @@ def get_volume_running_replica_cnt(client, volume_name):  # NOQA
...
+            if orphan.name == name:
+                found = True
+                break
+        if not found:
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert not found
+
+
+def delete_orphan(client, orphan_name):
```
innobead, line 4725: This is probably better to stay in test_orphan.py first until it's actually reusable in other modules, then we can move to common.py. The same idea applies to other new added functions. 

In general, some refactoring needs for common.py :)

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,757 @@
...
+        exec_nsenter("cp -a {} {}; echo xxx > {}".format(
+            replica.dataPath, path, os.path.join(path, "volume.meta")))
+
+    # Step 5
+    cleanup_volume_by_name(client, volume_name)
+
+    # Step 6
+    assert wait_for_orphan_count(client, 1, 180) == 1
+
+    # Step 7
```
innobead, line 165: For teardown, can we use `def finalizer():` as other test modules?

innobead (changes requested):
In general, LGTM. Some comments needed to resolve before merge.

yangchiu (dismissed):

innobead (approved):

manager/integration/tests/test_orphan.py
```diff
@@ -0,0 +1,757 @@
...
+        update_disks[name] = disk
+        disk_paths.append(disk_path)
+
+    node = node.diskUpdate(disks=update_disks)
+    node = wait_for_disk_update(client, node.name, len(update_disks))
+
+    return disk_paths
+
+
+def create_volume_with_replica_on_each_node(client, volume_name):  # NOQA
```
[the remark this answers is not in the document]
innobead: This seems should be `create_volume_with_replica_on_host`.

@derekbit could you create another PR to update this later? Thanks. cc @longhorn/qa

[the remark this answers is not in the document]
derekbit: okay!

PR #942, opened 2022-04-20
Print cloud-init log if jq command not found

yangchiu:
to debug regression test

for [#3856](https://github.com/longhorn/longhorn/issues/3856)

### DIFF
--- a/test_framework/terraform/aws/sles/k3s_instaces.tf
+++ b/test_framework/terraform/aws/sles/k3s_instaces.tf
@@ -98,8 +98,12 @@ resource "null_resource" "rsync_kubeconfig_file" {
   ]
 
   provisioner "remote-exec" {
-    inline = ["until([ -f /etc/rancher/k3s/k3s.yaml ] && [ `sudo /usr/local/bin/kubectl get node -o jsonpath='{.items[*].status.conditions}'  | jq '.[] | select(.type  == \"Ready\").status' | grep -ci true` -eq $((${var.lh_aws_instance_count_controlplane} + ${var.lh_aws_instance_count_worker})) ]); do echo \"waiting for k3s cluster nodes to be running\"; sleep 2; done"]
 
+    inline = [
+      "cloud-init status --wait",
+      "if [ \"`cloud-init status | grep error`\" ]; then cat /var/log/cloud-init-output.log; fi",
+      "until([ -f /etc/rancher/k3s/k3s.yaml ] && [ `sudo /usr/local/bin/kubectl get node -o jsonpath='{.items[*].status.conditions}'  | jq '.[] | select(.type  == \"Ready\").status' | grep -ci true` -eq $((${var.lh_aws_instance_count_controlplane} + ${var.lh_aws_instance_count_worker})) ]); do echo \"waiting for k3s cluster nodes to be running\"; sleep 2; done"
+    ]
 
     connection {
       type     = "ssh"
--- a/test_framework/terraform/aws/sles/user-data-scripts/provision_k3s_server.sh.tpl
+++ b/test_framework/terraform/aws/sles/user-data-scripts/provision_k3s_server.sh.tpl
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 sudo zypper ref
 sudo zypper install -y -t pattern devel_basis
 sudo zypper install -y open-iscsi nfs-client jq
--- a/test_framework/terraform/aws/centos/k3s_instances.tf
+++ b/test_framework/terraform/aws/centos/k3s_instances.tf
@@ -98,7 +98,11 @@ resource "null_resource" "rsync_kubeconfig_file" {
   ]
 
   provisioner "remote-exec" {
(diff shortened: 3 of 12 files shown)


test_framework/terraform/aws/sles/user-data-scripts/provision_k3s_server.sh.tpl
```diff
@@ -6,6 +6,8 @@ sudo zypper install -y open-iscsi nfs-client jq
 sudo systemctl -q enable iscsid
 sudo systemctl start iscsid
 
+touch /tmp/signal
```
innobead, line 9: I don't think we need this, but rather use cloud-init to check provisioning ready or not later.

test_framework/terraform/aws/sles/k3s_instaces.tf
```diff
@@ -98,8 +98,13 @@ resource "null_resource" "rsync_kubeconfig_file" {
   ]
 
   provisioner "remote-exec" {
-    inline = ["until([ -f /etc/rancher/k3s/k3s.yaml ] && [ `sudo /usr/local/bin/kubectl get node -o jsonpath='{.items[*].status.conditions}'  | jq '.[] | select(.type  == \"Ready\").status' | grep -ci true` -eq $((${var.lh_aws_instance_count_controlplane} + ${var.lh_aws_instance_count_worker})) ]); do echo \"waiting for k3s cluster nodes to be running\"; sleep 2; done"]
 
+    inline = [
+      "while [ ! -f /tmp/signal ]; do echo \"waiting for packages installed\"; sleep 2; done",
```
innobead, line 103: Suggest seeing if it's possible to use `cloud-init status` to check the provisioning status. 

ref: 
- https://cloudinit.readthedocs.io/en/latest/topics/cli.html#status
- (error code will be returned if error) https://github.com/delphix/cloud-init/blob/master/cloudinit/cmd/status.py#L178

test_framework/terraform/aws/sles/k3s_instaces.tf
```diff
@@ -98,8 +98,13 @@ resource "null_resource" "rsync_kubeconfig_file" {
...
+      "jq --version",
```
innobead, line 104: I don't think we need to have this, but rather a general check for cloud init

test_framework/terraform/aws/sles/k3s_instaces.tf
```diff
@@ -98,8 +98,13 @@ resource "null_resource" "rsync_kubeconfig_file" {
...
+      "if [ $? -eq 127 ]; then cat /var/log/cloud-init-output.log; fi",
```
innobead, line 105: You can base on `cloud-init status` to see if need to dump the log.

innobead (changes requested):
We need to make this change a general check for cloud-init instead of a specific pkg only.

Also, this general check needed to apply to all distro provisioning.

test_framework/terraform/aws/sles/k3s_instaces.tf
kaxing: To this, I am suspecting the `$PATH` inside the terraform_*.sh, as I am not certain how remote-exec runs.

I would have a simple check like:
```
"command -v jq || echo $PATH",
```
And maybe we should specify bash as the interpreter:
```
    inline = [
...
    ]
    interpreter = "/bin/bash"
```
ref.: https://github.com/hashicorp/terraform/issues/5397#issuecomment-190793340

innobead: @kaxing not sure if I get your point, but I mean we don't need to check jq, but rather make this check as a general check to see if packages all installed ready by cloud-init CLI. However, I feel we should add `set -e` in the scripts used in the user_data, so we will know the TF provisioning failed before any test execution. 

About how remote-exec runs, it's not really a point for us here.

kaxing: @innobead just thinking if this is for an one-time investigation.

We can certainly enable `set -e` even `set -x` if there are no security related output in-between TF provisioning logs.
About cloud-init I'v just notice the subcommand: [collect-logs ](https://cloudinit.readthedocs.io/en/latest/topics/cli.html?highlight=collect-logs#collect-logs) grabs quite a bit information if there is suitable place to keep the logs for each job:
```
$ cloud-init collect-logs --verbose
collected cloud-init --version
collected dpkg version
collected cloud-init version: /usr/bin/cloud-init 22.1-14-g2e17a0d6-0ubuntu1~20.04.3

collected dmesg output
collected systemd journal of current boot
collected file: /var/log/cloud-init.log
collected file: /var/log/cloud-init-output.log
collected dir /run/cloud-init
Wrote /home/ubuntu/cloud-init.tar.gz
```

innobead: Yeah, this seems helpful, but not sure if we need this. If want, you can have another TF file provisioner to copy this tar back.

innobead (approved):

innobead:
@yangchiu please backport to 1.2.x and 1.1.x, then it will be easier to find out the root cause because other branches probably will happen as well. Thank you.

PR #943, opened 2022-04-20
Implement test case test_setting_backing_image_auto_cleanup

cchien816:

Implement test case `test_setting_backing_image_auto_cleanup`

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4261,6 +4261,25 @@ def create_backing_image_with_matching_url(client, name, url):
     return bi
 
 
+def check_backing_image_disk_map_status(client, bi_name, expect_cnt, expect_disk_state): # NOQA
+    # Number of expect_cnt should equal to number of disk map
+    # that have expect_disk_state
+
+    for i in range(RETRY_COUNTS):
+        backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+
+        count = 0
+        for disk_id, status in iter(backing_image.diskFileStatusMap.items()):
+            if status.state == expect_disk_state:
+                count = count + 1
+
+        if expect_cnt == count:
+            break
+        time.sleep(RETRY_INTERVAL)
+
+    assert expect_cnt == count
+
+
 def wait_for_backing_image_disk_cleanup(client, bi_name, disk_id):
     found = False
     for i in range(RETRY_COUNTS):
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -1,5 +1,6 @@
 import time
 import pytest
+import os
 
 from common import (  # NOQA
     get_longhorn_api_client, get_self_host_id,
@@ -23,11 +24,16 @@ from common import (  # NOQA
     SETTING_GUARANTEED_REPLICA_MANAGER_CPU,
     SETTING_PRIORITY_CLASS,
     SIZE, RETRY_COUNTS, RETRY_INTERVAL, RETRY_INTERVAL_LONG,
+    update_setting, BACKING_IMAGE_QCOW2_URL, BACKING_IMAGE_NAME,
+    create_backing_image_with_matching_url, BACKING_IMAGE_EXT4_SIZE,
+    check_backing_image_disk_map_status, wait_for_volume_delete,
+    SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY, wait_for_node_update,
 )
 
 from test_infra import wait_for_node_up_longhorn
 
 KUBERNETES_DEFAULT_TOLERATION = "kubernetes.io"
+BACKING_IMAGE_CLEANUP_WAIT_INTERVAL = "50"
 
 
 def check_workload_update(core_api, apps_api, count):  # NOQA
@@ -624,9 +630,8 @@ def wait_for_priority_class_update(core_api, apps_api, count, priority_class=Non
     assert updated
 
 
-@pytest.mark.skip(reason="TODO")
 @pytest.mark.backing_image  # NOQA
-def test_setting_backing_image_auto_cleanup():  # NOQA
+def test_setting_backing_image_auto_cleanup(client, core_api, volume_name):  # NOQA
     """
     Test that the Backing Image Cleanup Wait Interval setting works correctly.
 
@@ -642,9 +647,10 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
         4. Verify the only backing image file in each disk is reused by
            multiple replicas. The backing image file path is
            `<Data path>/<The backing image name>/backing`
-    5. Decrease the replica count by 1 for all volumes.
+    5. Unschedule test node to guarantee when replica removed from test node,
+       no new replica can be rebuilt on the test node.
     6. Remove all replicas in one disk.
-       Wait for 1 minute.
+       Wait for 50 seconds.
        Then verify nothing changes in the backing image disk state map
        (before the cleanup wait interval is passed).
     7. Modify `BackingImageCleanupWaitInterval` to a small value. Then verify:
@@ -653,11 +659,82 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
         2. The related backing image file is removed.
         3. The download state of other disks keep unchanged.
            All volumes still work fine.
-    8. Delete all volumes. Verify that all states in the backing image disk map
-       will become terminating first,
-       and all entries will be removed from the map later.
+    8. Delete all volumes. Verify that there will only remain 1 entry in the
+       backing image disk map
     9. Delete the backing image.
     """
+    # Step 1
+    update_setting(client,
+                   "backing-image-cleanup-wait-interval",
+                   BACKING_IMAGE_CLEANUP_WAIT_INTERVAL)
+
+    # Step 2
+    create_backing_image_with_matching_url(
+            client, BACKING_IMAGE_NAME, BACKING_IMAGE_QCOW2_URL)
+
+    # Step 3
+    volume_names = [
+        volume_name + "-1",
+        volume_name + "-2",
+        volume_name + "-3"
+    ]
+
+    for volume_name in volume_names:
+        volume = create_and_check_volume(
+                            client, volume_name, 3,
+                            str(BACKING_IMAGE_EXT4_SIZE),
+                            BACKING_IMAGE_NAME)
+
+    # Step 4
+    lht_host_id = get_self_host_id()
+    for volume_name in volume_names:
+        volume = client.by_id_volume(volume_name)
+        volume.attach(hostId=lht_host_id)
+        wait_for_volume_healthy(client, volume_name)
+        assert volume.backingImage == BACKING_IMAGE_NAME
+
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    assert len(backing_image.diskFileStatusMap) == 3
+    for disk_id, status in iter(backing_image.diskFileStatusMap.items()):
+        assert status.state == "ready"
+
+    backing_images_in_disk = os.listdir("/var/lib/longhorn/backing-images")
+    assert len(backing_images_in_disk) == 1
+    assert os.path.exists("/var/lib/longhorn/backing-images/{}/backing"
+                          .format(backing_images_in_disk[0]))
+    assert os.path.exists("/var/lib/longhorn/backing-images/{}/backing.cfg"
+                          .format(backing_images_in_disk[0]))
+
+    # Step 5
+    current_host = client.by_id_node(id=lht_host_id)
+    client.update(current_host, allowScheduling=False)
+    wait_for_node_update(client, lht_host_id, "allowScheduling", False)
+
+    # Step 6
+    for volume_name in volume_names:
+        volume = client.by_id_volume(volume_name)
+        for replica in volume.replicas:
+            if replica.hostId == lht_host_id:
+                replica_name = replica.name
+        volume.replicaRemove(name=replica_name)
+
+    time.sleep(int(BACKING_IMAGE_CLEANUP_WAIT_INTERVAL))
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 3, "ready")
+
+    # Step 7
+    update_setting(client, "backing-image-cleanup-wait-interval", "1")
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 2, "ready")
+
+    backing_images_in_disk = os.listdir("/var/lib/longhorn/backing-images")
+    assert len(backing_images_in_disk) == 0
+
+    # Step 8
+    for volume_name in volume_names:
+        volume = client.by_id_volume(volume_name)
+        client.delete(volume)
+        wait_for_volume_delete(client, volume_name)
+
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 1, "ready")
 
 
 @pytest.mark.skip(reason="TODO")


kaxing (approved):
Reviewed and also check with local VM setup with longhorn-manager-test:master-head and longhorn-manager:master-head.

```
root@longhorn-test:/integration/tests# pytest -v test_settings.py::test_setting_backing_image_auto_cleanup
=========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.9.12, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
plugins: repeat-0.9.1
collected 1 item

test_settings.py::test_setting_backing_image_auto_cleanup PASSED                                                                                                                                    [100%]

====================================================================================== 1 passed in 253.32s (0:04:13) ======================================================================================
```

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +656,78 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
         2. The related backing image file is removed.
         3. The download state of other disks keep unchanged.
            All volumes still work fine.
-    8. Delete all volumes. Verify that all states in the backing image disk map
-       will become terminating first,
-       and all entries will be removed from the map later.
+    8. Delete all volumes. Verify that there will only remain 1 entry in the
+       backing image disk map
```
innobead, line 660: Why still remaining 1 entry if all volumes deleted and BIM auto cleanup executed?

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +656,78 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+        assert status.state == "ready"
+        assert disk_id != ''
+
+    backing_images_in_disk = os.listdir("/var/lib/longhorn/backing-images")
+    assert len(backing_images_in_disk) == 1
+
+    # Step 5
+    for volume_name in volume_names:
+        volume = client.by_id_volume(volume_name)
+        volume.updateReplicaCount(replicaCount=1)
```
innobead, line 705: I feel this step is unnecessary? because what we want to verify is `if a backing image in one disk is not referenced by any replicas, then it should be cleaned up after the wait interval. 

Basically, it's not related to how many replicas we need to change to test this.

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +656,78 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 2, "ready")
+
+    # Step 8
+    for volume_name in volume_names:
+        volume = client.by_id_volume(volume_name)
+        client.delete(volume)
+        wait_for_volume_delete(client, volume_name)
+
+    time.sleep(int(BACKING_IMAGE_CLEANUP_WAIT_INTERVAL))
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 1, "ready")
```
innobead, line 730: This should be `check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 0, "ready")`? because all volumes gone, and all backing images are not referenced by any replicas anymore, the info should be all cleaned up?

innobead (changes requested):
A few comments needed to clarify.

manager/integration/tests/test_settings.py
innobead: cc @shuo-wu

manager/integration/tests/test_settings.py
cchien816: Hi, @innobead ,

I checked with @shuo-wu, by default there will be 1 file in backing image data map to avoid download again if user want to re use the backing image, so if we delete all volumes which using that backing image, there should be still 1

manager/integration/tests/test_settings.py
cchien816: If I understanding correctly, this step is for the next step `6. Remove all replicas in one disk.` to make sure on a node there will be no replica exist to observe backing image on that node will auto clean up

But actually the replicaCount can set to 2, thank you

manager/integration/tests/test_settings.py
innobead: Yeah, I remembered that now. Thanks, @shuo-wu and @cchien816 .

manager/integration/tests/test_settings.py
innobead: Yeah, thanks. Keeping 1 is for the downloading case.

manager/integration/tests/test_settings.py
innobead: Your understanding is partially correct, but the implementation here is missing some steps about replica scheduling. 

Firstly, removing replicas from the node does not mean new replicas rebuilt will not happen in the same node unless that node is unscheduled. 

Also, in your step 5, when scaling down the number of replicas, actually some replicas will be removed from any nodes, then you do the step 6, so it will probably cause some new replicas rebuilding and they are possibly recreated in the same node.

So, what you need to do can be: (however, need your verification)
- set your testing node (`lht_host_id`) unschedulable (I assume there is only one disk on that node, but technically you should make a disk unschedulable)
- then remove replicas on that node/disk, all new replicas will be created in other nodes
- then step 7

or 
- the same as the above first step
- scale down the number of replicas
- then step 7

cchien816: Hi @innobead ,

I have changed step 5 to disable test node and enable `Replica Node Level Soft Anti-Affinity ` to make sure no new replica will generate at next few steps.

innobead: Actually U don't need to enable soft node anti affinity, but just keep it as default value.

shuo-wu: Here, what I want to clarify is that decreasing the replica count of a volume won't lead to Longhorn actively removing the redundant replica. Then the replica removal below is just deleting the extra/redundant replica on the current node without rebuilding.

Disabling the scheduling of the current node also works but there will be rebuilding triggered. I am fine with either solution. 

Yes, you don't need to enable soft node anti-affinity.

manager/integration/tests/common.py
```diff
@@ -4261,6 +4261,20 @@ def create_backing_image_with_matching_url(client, name, url):
     return bi
 
 
+def check_backing_image_disk_map_status(client, bi_name, expect_cnt, diskmap_status): # NOQA
```
shuo-wu, line 4264: 1. Rename input `diskmap_status` -> `expect_disk_state`
2. This function means that the state of all disk files of a backing image should be the same. And the count should match. Or you just want to check if there are **enough** disk files being in the expected state, regardless of some other disk files being in an arbitrary state.

manager/integration/tests/test_settings.py
```diff
@@ -642,7 +646,9 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
         4. Verify the only backing image file in each disk is reused by
            multiple replicas. The backing image file path is
            `<Data path>/<The backing image name>/backing`
-    5. Decrease the replica count by 1 for all volumes.
+    5. Enable 'Replica Node Level Soft Anti-Affinity' and unschedule test node
```
shuo-wu, line 649: Don't need this

manager/integration/tests/test_settings.py
```diff
@@ -642,7 +646,9 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+       to guarantee if replica removed from test node, no new replica will
+       generte on it.
```
shuo-wu, line 651: ` no new replica will generte on it` -> `no new replica can be rebuilt on the test node`

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +659,82 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+        assert status.state == "ready"
+        assert disk_id != ''
+
+    backing_images_in_disk = os.listdir("/var/lib/longhorn/backing-images")
+    assert len(backing_images_in_disk) == 1
+
+    # Step 5
+    replica_node_soft_anti_affinity_setting = \
+        client.by_id_setting(SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY)
+    client.update(replica_node_soft_anti_affinity_setting, value="true")
```
shuo-wu, line 708: Don't need this.

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +659,82 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+        wait_for_volume_healthy(client, volume_name)
+        assert volume.backingImage == BACKING_IMAGE_NAME
+
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    assert len(backing_image.diskFileStatusMap) == 3
+    for disk_id, status in iter(backing_image.diskFileStatusMap.items()):
+        assert status.state == "ready"
+        assert disk_id != ''
+
+    backing_images_in_disk = os.listdir("/var/lib/longhorn/backing-images")
```
shuo-wu, line 702: NIT: It's better to check the actual file as well.

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +659,82 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+
+    # Step 6
+    for volume_name in volume_names:
+        volume = client.by_id_volume(volume_name)
+        for replica in volume.replicas:
+            if replica.hostId == lht_host_id:
+                replica_name = replica.name
+        volume.replicaRemove(name=replica_name)
+
+    time.sleep(int(BACKING_IMAGE_CLEANUP_WAIT_INTERVAL))
```
shuo-wu, line 722: Wait... This is exactly the current backing image cleanup wait interval. After the step, the disk file on the test node may be removed. In brief, you shouldn't wait for so long here.

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +659,82 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 3, "ready")
+
+    # Step 7
+    update_setting(client, "backing-image-cleanup-wait-interval", "1")
+    time.sleep(int(BACKING_IMAGE_CLEANUP_WAIT_INTERVAL))
```
shuo-wu, line 727: You don't need to wait for so long there. 5~30s may be enough.

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +659,82 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 2, "ready")
```
shuo-wu, line 728: Need to verify the disk of the test node is not in the map now.

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +659,82 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+
+    # Step 8
+    for volume_name in volume_names:
+        volume = client.by_id_volume(volume_name)
+        client.delete(volume)
+        wait_for_volume_delete(client, volume_name)
+
+    time.sleep(int(BACKING_IMAGE_CLEANUP_WAIT_INTERVAL))
```
shuo-wu, line 736: Same here. You don't need to wait for so long there. 5~30s may be enough.

shuo-wu (changes requested):

manager/integration/tests/test_settings.py
innobead: > Here, what I want to clarify is that decreasing the replica count of a volume won't lead to Longhorn actively removing the redundant replica. Then the replica removal below is just deleting the extra/redundant replica on the current node without rebuilding.
> 

Thanks for the clarification on this! Outside the test here, so retaining the redundant ones seems intended even the number decreased, but why? Any context?

innobead: Interesting, the context is here https://github.com/longhorn/longhorn-manager/pull/672#issuecomment-686045343. 

@shuo-wu Do we have any doc for this behavior? If not, I feel users will not know the redundant healthy replicas keeping alive even if the number of replicas has been decreased.

shuo-wu: No... The link may be not the context. IIRC, this behavior exists since the alpha version.

innobead: @cchien816 ☝️ see this clarification from Shuo.

manager/integration/tests/test_settings.py
cchien816: I midified the code

manager/integration/tests/test_settings.py
cchien816: Code removed

manager/integration/tests/test_settings.py
cchien816: Added code to check `backing` and `backing.cfg` exist in backing image folder

manager/integration/tests/test_settings.py
cchien816: I reduced `BACKING_IMAGE_CLEANUP_WAIT_INTERVAL` to 5, thank you

manager/integration/tests/test_settings.py
cchien816: Added below code to check test node have no backing image files
```python
    backing_images_in_disk = os.listdir("/var/lib/longhorn/backing-images")
    assert len(backing_images_in_disk) == 0
```

manager/integration/tests/common.py
cchien816: HI @shuo-wu ,
I changed the variable name to `expect_disk_state` and modified function `check_backing_image_disk_map_status`, let the function check counts of specific disk map status were equal to we expected or not. thank you.

manager/integration/tests/test_settings.py
shuo-wu: No. I mean, in this line, the wait interval should be smaller than the value of setting `backing-image-cleanup-wait-interval`. So that the auto cleanup won't be triggered and the backing image disk map still holds 3 files.
On the other hand, **all** the **below** wait intervals should be greater than the updated value of the setting, which is "1".

manager/integration/tests/test_settings.py
cchien816: Hi @shuo-wu ,
I removed the sleep tine because next step `check_backing_image_disk_map_status` will have a 150s loop to check if the clean up really work

manager/integration/tests/test_settings.py
cchien816: Hi @shuo-wu ,
I removed the sleep tine because next step `check_backing_image_disk_map_status` will have a 150s loop to check if the clean up really work

manager/integration/tests/test_settings.py
cchien816: Hi @shuo-wu ,
I changed the sleep time to 1 minute to match the test step to avoid uncertain situation happen for review, thanks

shuo-wu: It's better to set a smaller value here (50s). The value should be slightly smaller than the setting value. Sorry for the confusion. And remember to update the skeleton description as well.

shuo-wu (commented):
Others LGTM

manager/integration/tests/common.py
```diff
@@ -4261,6 +4261,25 @@ def create_backing_image_with_matching_url(client, name, url):
...
+    # that have expect_disk_state
+
+    for i in range(RETRY_COUNTS):
+        backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+
+        count = 0
+        for disk_id, status in iter(backing_image.diskFileStatusMap.items()):
+            if status.state == expect_disk_state:
+                count = count + 1
+                assert disk_id != ''
```
innobead, line 4275: Probably we don't need to worry about this specially, because each status should always map to a disk.

innobead (dismissed):
LGTM except 2 extra comments to resolve.

manager/integration/tests/common.py
cchien816: HI @innobead , I removed the code, thank you

manager/integration/tests/test_settings.py
```diff
@@ -653,11 +659,83 @@ def test_setting_backing_image_auto_cleanup():  # NOQA
...
+        volume = client.by_id_volume(volume_name)
+        volume.attach(hostId=lht_host_id)
+        wait_for_volume_healthy(client, volume_name)
+        assert volume.backingImage == BACKING_IMAGE_NAME
+
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    assert len(backing_image.diskFileStatusMap) == 3
+    for disk_id, status in iter(backing_image.diskFileStatusMap.items()):
+        assert status.state == "ready"
+        assert disk_id != ''
```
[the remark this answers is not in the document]
cchien816: HI @innobead , I removed the code, thank you

innobead (approved):

innobead:
@cchien816 see if need to backport to 1.2.x.

PR #948, opened 2022-04-25
fix secret expected length to be at least (1)

cchien816:

From https://github.com/longhorn/longhorn/issues/3892
Newest terraform not accept length = 0 cluster secret, modify code to cope it

cc @meldafrawi

### DIFF
--- a/test_framework/terraform/aws/sles/main.tf
+++ b/test_framework/terraform/aws/sles/main.tf
@@ -21,6 +21,12 @@ resource "random_string" "random_suffix" {
   upper            = false
 }
 
+# Create a random string for the cluster secret
+resource "random_password" "cluster_secret" {
+  length = 64
+  special = false
+}
+
 # Create a VPC
 resource "aws_vpc" "lh_aws_vpc" {
   cidr_block = "10.0.0.0/16"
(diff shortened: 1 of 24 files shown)


test_framework/terraform/aws/sles/main.tf
```diff
@@ -250,3 +250,8 @@ resource "aws_eip" "lh_aws_eip_controlplane" {
   count    = var.lh_aws_instance_count_controlplane
   vpc      = true
 }
+
+resource "random_password" "cluster_secret" {
+  length = 64
+  special = false
+}
```
innobead, line 257: Could you move this under the below resource call, and add a comment (ex: `Create a random string for the cluster secret`) for it. 

```hcl
resource "random_string" "random_suffix" {
...
}

<-- add here
```

innobead (changes requested):
Good job!

Please also apply this change to other distros as well.

test_framework/terraform/aws/sles/main.tf
cchien816: Hi @innobead ,

I moved the code and added comment. And apply those changes on all distro for review, thank you.

innobead (approved):

innobead:
@cchien816 See if need to backport to 1.2.x and 1.1.x. I believe we have the same code bases for the testing infra provision.

cchien816:
Hi @innobead ,

I looked into v1.2.x and find that v1.2.x currently only support k3s, hence the password were fixed in length 64. 
We won't hit length = 0 error unless we support rke2 cluster on v1.2.x, thank you.

PR #949, opened 2022-04-25
[v1.1.x-backport] Fix test_allow_volume_creation_with_degraded_availability_restore

kaxing:
- from longhorn/longhorn-tests#756


innobead:
We need to backport to 1.2.x as well.

innobead (approved):

PR #952, opened 2022-04-29
Fix findmnt notfound after migrating to bci

innobead:
Signed-off-by: David Ko <dko@suse.com>


kaxing:
Didn't notice this when migrating to bci, thanks for the quick fix. These change looks good to me.

kaxing (approved):

PR #959, opened 2022-05-06
Run benchmark test for customized longhorn

yangchiu:
Run benchmark test for customized longhorn

For [#3919](https://github.com/longhorn/longhorn/issues/3919)

### DIFF
--- a/benchmark_test/scripts/benchmark-test-setup.sh
+++ b/benchmark_test/scripts/benchmark-test-setup.sh
@@ -3,7 +3,6 @@
 set -x
 set -e
 
-
 set_kubeconfig_envvar(){
   ARCH=${1}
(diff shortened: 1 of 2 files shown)


benchmark_test/scripts/benchmark-test-setup.sh
```diff
@@ -43,6 +42,65 @@ install_local_path_provisioner(){
 }
 
 
+generate_longhorn_yaml_manifest(){
```
innobead, line 45: This is actually similar to test_framework/scripts. For now, it's OK, but probably you can think about making them shareable in the future, so that will be easy for you for later refactoring.

innobead (approved):
LGTM

PR #960, opened 2022-05-10
Retry node/disk update if "too many retries error" happened.

derekbit:
There are some errors in the e2e test after introducing the node monitor, like
```
...cannot finish API request due to too many error retries
```

Currently, the tests are not waiting long enough for the disks to be reconciled by the node controller.
The subsequent actions might encounter unexpected errors due to the non-ready disks.

The fix is retrying node/disk update if "too many retries error" happened.

[Longhorn 3939](https://github.com/longhorn/longhorn/issues/3939)

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -226,6 +226,10 @@ JOB_LABEL = "recurring-job.longhorn.io"
 
 MAX_SUPPORT_BINDLE_NUMBER = 20
 
+NODE_UPDATE_RETRY_INTERVAL = 6
+NODE_UPDATE_RETRY_COUNT = 30
+retry_error = "cannot finish API request due to too many error retries"
+
 
 def load_k8s_config():
     c = Configuration()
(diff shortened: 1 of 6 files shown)


kaxing (approved):
Derek, thanks for the retry mechanism!

manager/integration/tests/test_basic.py
```diff
@@ -563,7 +564,7 @@ def backup_status_for_unavailable_replicas_test(client, volume_name,  # NOQA
     for r in volume.replicas:
         if r.name == replica_name:
             node = client.by_id_node(r.hostId)
-            node = client.update(node, allowScheduling=False)
+            node = set_node_scheduling(client, node, allowScheduling=False)
```
innobead, line 567: Because the tests here are special to change disks frequently, I would suggest not blindly making related util functions just retrying but rather using a specific parameter to instruct we do this on purpose like below. 

```python
set_node_scheduling(client, node, allowScheduling=False, retry=True)
```
This should apply to other node/disk util functions.

innobead (changes requested):

manager/integration/tests/test_basic.py
derekbit: okay

innobead:
To give more context for @longhorn/qa 

There is a node monitoring on each node and they will update the disk status of the node resource consistent with the disks on the node. (by default 30 seconds to sync)

In some node-related test cases here, we will change node disk status to simulate different cases, so it will probably cause some delayed sync of disk status of nodes. At that time, when doing some node spec updates via API, our validator webhook will reject the request when the disk spec and disk status of the node resource do not match. This is the reason why @derekbit needs to add retry on purpose. However, in general, if no disks change, updating node spec will not hit any issues. 

https://github.com/longhorn/longhorn-manager/blob/4e25888fd7acdd6fa73bede0f85de1d28e9b9421/webhook/resources/node/validator.go#L75-L77

manager/integration/tests/test_basic.py
derekbit: Addressed. By default, the `retry` is `False`. `retry` is `True` in test_node.py test cases and cleanup disks functions in common.py.

derekbit:
> To give more context for @longhorn/qa
> 
> There is a node monitoring on each node and they will update the disk status of the node resource consistent with the disks on the node. (by default 30 seconds to sync)
> 
> In some node-related test cases here, we will change node disk status to simulate different cases, so it will probably cause some delayed sync of disk status of nodes. At that time, when doing some node spec updates via API, our validator webhook will reject the request when the disk spec and disk status of the node resource do not match. This is the reason why @derekbit needs to add retry on purpose. However, in general, if no disks change, updating node spec will not hit any issues.
> 
> https://github.com/longhorn/longhorn-manager/blob/4e25888fd7acdd6fa73bede0f85de1d28e9b9421/webhook/resources/node/validator.go#L75-L77

The UI error message
image

The message is not friendly to users.
```
func RetryOnErrorCondition(fn func() (interface{}, error), predicate func(error) bool) (interface{}, error) {
	for i := 0; i < APIRetryCounts; i++ {
		obj, err := fn()
		if err == nil {
			return obj, nil
		}
		if !predicate(err) {
			return nil, err
		}
		time.Sleep(APIRetryInterval + APIRetryJitterInterval*time.Duration(rand.Intn(5)))
	}
	return nil, fmt.Errorf("cannot finish API request due to too many error retries")
} 
```

Maybe we should add more information in the error message, e.g
```
return nil, fmt.Errorf("cannot finish API request due to too many error retries since %v", err)
```

manager/integration/tests/common.py
```diff
@@ -839,23 +843,85 @@ def check_volume_replicas(volume, spec, tag_mapping):
...
+                    continue
+                print(e)
+                raise
+            else:
+                break
 
+        return node
+    else:
+        return client.update(node, allowScheduling=node.allowScheduling,
+                             tags=tags)
```
innobead, line 873: Suggest changing to this, so don't need to have a long if-else statement. The same for other functions. 

```suggestion
    if not retry:
        return client.update(node, allowScheduling=node.allowScheduling, tags=tags)
                   
    for _ in range(NODE_UPDATE_RETRY_COUNT):
        try:
            node = client.update(node,
                                 allowScheduling=node.allowScheduling,
                                 tags=tags)
        except Exception as e:
            if retry_error in str(e.error.message):
                time.sleep(NODE_UPDATE_RETRY_INTERVAL)
                continue
            print(e)
            raise
        else:
            break

    return node
```

innobead (changes requested):

manager/integration/tests/common.py
derekbit: Addressed.

innobead (approved):

PR #961, opened 2022-05-10
Use more precise error message for comparison

derekbit:
Use more precise error message for comparison.

[Longhorn 3939](https://github.com/longhorn/longhorn/issues/3939)


innobead (approved):

derekbit:
@innobead Can we get this PR merged?

PR #964, opened 2022-05-12
test_orphan.py: retry update_node_disks

derekbit:
test cases create and teardown disks frequently, so the retry is necessary.

Longhorn 685

### DIFF
--- a/manager/integration/tests/test_orphan.py
+++ b/manager/integration/tests/test_orphan.py
@@ -17,6 +17,7 @@ from common import cleanup_volume_by_name
 from common import create_host_disk, cleanup_host_disks
 from common import wait_for_node_update
 from common import wait_for_disk_status
+from common import update_node_disks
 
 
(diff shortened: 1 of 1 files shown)


innobead:
@derekbit PR check failed.

manager/integration/tests/test_orphan.py
```diff
@@ -40,7 +41,8 @@ def crate_disks_on_host(client, disk_names, request):  # NOQA
         update_disks[name] = disk
         disk_paths.append(disk_path)
 
-    node = node.diskUpdate(disks=update_disks)
+    node = update_node_disks(client, node.name, disks=update_disks,
```
innobead, line 44: I found the enclosed function name is typo :slightly_smiling_face: 

`crate_disks_on_host(` should be `create_disks_on_host(`

innobead (changes requested):

innobead (approved):

PR #966, opened 2022-05-18
Update node-deletion.md: Fix the ordered list style and add example to crash replicas via cli

kaxing:
Fix the ordered list for sublists with css, adding to the html > header > style tag block.

It should like this, as tested from my local environment:
image


docs/content/manual/pre-release/node/node-deletion.md
```diff
@@ -27,3 +27,8 @@ title: Test node deletion
     4. Crash all replicas of the 3rd volume and trigger the auto salvage. --> The auto salvage should work. The volume works fine and the data is correct after the salvage.
```
mantissahz, line 27: Could we give an example for "Crash all replicas" such as delete the IM-r pods?

kaxing: Sure, I will add more details to this step.

innobead (approved):

PR #973, opened 2022-05-25
Add test case for test external csi snapshot with snap param

cchien816:
https://github.com/longhorn/longhorn/issues/3845

Separate `test_csi_snapshot_with_snap_param` into

- `test_csi_snapshot_snap_create_csi_snapshot`
- `test_csi_snapshot_snap_create_volume_from_snapshot` 
- `test_csi_snapshot_snap_delete_csi_snapshot` (only skeleton)

to reduce long script and avoid hard maintain and debug

### DIFF
--- a/manager/integration/tests/test_csi_snapshotter.py
+++ b/manager/integration/tests/test_csi_snapshotter.py
@@ -2,6 +2,7 @@ import time
 from urllib.parse import urlparse
 
 import pytest
+import common
 from backupstore import backupstore_s3  # NOQA
 from backupstore import set_random_backupstore  # NOQA
 from common import RETRY_COUNTS
@@ -24,6 +25,10 @@ from common import wait_for_backup_delete
 from common import wait_for_volume_detached
 from common import wait_for_volume_restoration_completed
 from kubernetes.client.rest import ApiException
+from common import generate_volume_name, Mi, wait_and_get_pv_for_pvc, create_pvc # NOQA
+from common import make_deployment_with_pvc, apps_api # NOQA
+from common import check_pvc_in_specific_status # NOQA
+from common import wait_for_pvc_phase
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_csi_snapshotter.py
```diff
@@ -674,23 +732,213 @@ def test_csi_snapshot_with_snap_param(client, volume_name): # NOQA
...
+    csivolsnap = volumesnapshot(vol.name + "-volumesnapshot",
+                                "default",
+                                csisnapclass["metadata"]["name"],
+                                "persistentVolumeClaimName",
+                                pvc_name)
+
+    wait_for_volumesnapshot_ready(
+                            volumesnapshot_name=csivolsnap["metadata"]["name"],
+                            namespace='default',
+                            ready_to_use=False)
```
PhanLe1010, line 765: This method name is a bit confused but I think it is ok

manager/integration/tests/test_csi_snapshotter.py
```diff
@@ -674,23 +732,213 @@ def test_csi_snapshot_with_snap_param(client, volume_name): # NOQA
...
+                                         name=deployment_name)
+    common.wait_for_volume_status(client, vol.name,
+                                  common.VOLUME_FIELD_STATE,
+                                  common.VOLUME_STATE_DETACHED)
+
+    new_pvc1 = pvc
+    new_pvc1['metadata']['name'] = pvc['metadata']['name'] + "new-pvc1"
+    create_pvc(new_pvc1)
+    check_pvc_in_specific_status(core_api,
+                                 new_pvc1['metadata']['name'], "Pending")
```
PhanLe1010, line 893: This could be a false positive because initially every PVC is in the `pending` state. We need to wait for a while and confirm that the PVC is still in pending state. One way to achieve this by modifying the function `check_pvc_in_specific_status` to have similar logic to function https://github.com/longhorn/longhorn-tests/blob/5f89bafbb6acbcd1ace11c1d468f2f7cd95573ff/manager/integration/tests/test_csi_snapshotter.py#L287-L306

manager/integration/tests/test_csi_snapshotter.py
```diff
@@ -674,23 +732,213 @@ def test_csi_snapshot_with_snap_param(client, volume_name): # NOQA
...
+        - Create VolumeSnapshot with class longhorn-snap
         - Create volume from longhorn-snapshot
             - Source volume is attached && Longhorn snapshot exist
                 - Create PVC from snapshot generated from step 1
                 - Attach the PVC and verify data
             - Source volume is detached
                 - Scale down the workload
-                - Create PVC from VolumeSnapshot generated from step 1
+                - Create PVC from VolumeSnapshot generated from step beggining
                 - Verify PVC provision failed
```
PhanLe1010, line 819: Can we add the steps:
* Scale up the workload
* Wait for PVC to finish provisioning and be bounded
* Attach the PVC test-restore-pvc and verify the data

manager/integration/tests/test_csi_snapshotter.py
```diff
@@ -674,23 +732,213 @@ def test_csi_snapshot_with_snap_param(client, volume_name): # NOQA
...
+    deployment["spec"]["replicas"] = 1
+    apps_api.patch_namespaced_deployment(body=deployment,
+                                         namespace='default',
+                                         name=deployment_name)
+    common.wait_for_volume_status(client, vol.name,
+                                  common.VOLUME_FIELD_STATE,
+                                  common.VOLUME_STATE_ATTACHED)
+
+    vol = client.by_id_volume(vol.name)
+    # create new snapshot aovid only exist 1 snpashot and can not deleted
```
PhanLe1010, line 905:
```suggestion
    # create new snapshot to avoid the case the volume only has 1 snapshot so the snapshot can not deleted
```

manager/integration/tests/test_csi_snapshotter.py
```diff
@@ -674,23 +732,213 @@ def test_csi_snapshot_with_snap_param(client, volume_name): # NOQA
...
+    vol.snapshotCreate()
+    snapshots = vol.snapshotList()
+    for item in snapshots:
+        if "snap" in item.name:
+            vol.snapshotDelete(name=item.name)
+    vol.snapshotPurge()
+
+    new_pvc2 = pvc
+    new_pvc2['metadata']['name'] = pvc['metadata']['name'] + "new-pvc2"
```
PhanLe1010, line 914: Can we make the logic here a bit clearer for the reader by
```suggestion
    # Find the VolumeSnapshotContent of the volumesnapshot_name.
    # Find the Longhorn snapshot name inside the field VolumeSnapshotContent.snapshotHandle, let's call it snap_name
    snapshots = vol.snapshotList()
    for item in snapshots:
        if snap_name == item.name:
            vol.snapshotDelete(name=snap_name)
    vol.snapshotPurge()
    # wait for snap_name to be removed from the volume

    new_pvc2 = pvc
    new_pvc2['metadata']['name'] = pvc['metadata']['name'] + "new-pvc2"
```

cchien816: HI @PhanLe1010 
I modified coed for review, thanks

manager/integration/tests/test_csi_snapshotter.py
cchien816: HI @PhanLe1010 
I added steps into code, thank you

manager/integration/tests/test_csi_snapshotter.py
cchien816: I modified coed for review, thank  you

innobead:
@PhanLe1010 Please help review this to see if anything is missing. Thanks.

manager/integration/tests/test_csi_snapshotter.py
PhanLe1010: I think this function still have a bit problem:
```go
def check_pvc_in_specific_status(api, pvc_name, status):
    for i in range(RETRY_EXEC_COUNTS):
        pvcs = api.list_namespaced_persistent_volume_claim("default")
        for item in pvcs.items:
            if item.metadata.name == pvc_name:
                break
        time.sleep(RETRY_INTERVAL)

    assert item.status.phase == status
```
Let say we create a new PVC, initially the state of the PVC is "pending". The above function check and return immediately that the PVC is in pending. That is not what we want. We want to make sure that the PVC remaining in pending for a period of time. Therefore, how do you think about this:
```go
def check_pvc_in_specific_status(api, pvc_name, status):
    for i in range(RETRY_EXEC_COUNTS):
        claim = api.read_namespaced_persistent_volume_claim(name=pvc_name, namespace='default')
        if claim.status.phase == "bound":
            break
        time.sleep(RETRY_INTERVAL)

    assert item.status.phase == status
```

PhanLe1010:
In general, LGTM except for a small comment

manager/integration/tests/test_csi_snapshotter.py
cchien816: HI @PhanLe1010.

PhanLe1010: Thank you for the fix!

PhanLe1010 (approved):
LGTM

PR #978, opened 2022-05-27
flaky-results: fix backupstore cleanup

c3y1huang:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1362/testReport/junit/tests/test_ha/test_dr_volume_with_restore_command_error_s3__2/

https://github.com/longhorn/longhorn-manager/pull/1338#issuecomment-1138380264

https://github.com/longhorn/longhorn/issues/3546

### DIFF
--- a/manager/integration/tests/backupstore.py
+++ b/manager/integration/tests/backupstore.py
@@ -14,6 +14,7 @@ from common import SETTING_BACKUP_TARGET
 from common import SETTING_BACKUP_TARGET_CREDENTIAL_SECRET
 from common import SETTING_BACKUPSTORE_POLL_INTERVAL
 from common import LONGHORN_NAMESPACE
+from common import cleanup_all_volumes
 from common import is_backupTarget_s3
 from common import is_backupTarget_nfs
 from common import get_longhorn_api_client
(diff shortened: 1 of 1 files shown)


kaxing (approved):

manager/integration/tests/backupstore.py
```diff
@@ -53,6 +54,7 @@ def set_random_backupstore(request, client):
         mount_nfs_backupstore(client)
 
     yield
+    cleanup_all_volumes(client)
```
innobead, line 57: I know this is a workaround, but we should have a better way instead of adding non-backup clean up logic here?

Do we have any teardown place for each test case? probably we can add this there?

c3y1huang: Do you think we should have a consolidated place for all teardown logic?
AFAIK ATM we are depending on the fixture to clean up itself after each test case. This `set_random_backupstore` is a fixture so it seems to be ok to add it here.

innobead: Probably no side effect but it's just more about the segregated responsibility for each fixture.

If mixing things (here cleanup/teardown) in different fixtures, u will probably don't know what teardown each fixture does.

However, I think let's merge this first to see how it works first then revisit this when reviewing all fixtures.

innobead (approved):

PR #979, opened 2022-05-30
Add test case : test case test_disk_eviction_with_node_level_soft_anti_affinity_disabled

cchien816:

Implement test case test_disk_eviction_with_node_level_soft_anti_affinity_disabled

https://github.com/longhorn/longhorn/issues/4040

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -2597,8 +2597,11 @@ def test_node_eviction_multiple_volume(client, core_api, csi_pv, pvc, pod_make, 
     assert expect_md5sum == created_md5sum2
 
 
-@pytest.mark.skip(reason="TODO") # NOQA
-def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(): # NOQA
+def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(client, # NOQA
+                                                                   volume_name, # NOQA
+                                                                   request, # NOQA
+                                                                   settings_reset, # NOQA
+                                                                   reset_disk_settings): # NOQA
     """
     Steps:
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_node.py
```diff
@@ -2611,3 +2614,80 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(): # NOQA
...
+
+    # Step 6
+    replica_path = test_disk_path + '/replicas'
+    assert os.path.isdir(replica_path)
+
+    for i in range(common.RETRY_COMMAND_COUNT):
+        if len(os.listdir(replica_path)) > 0:
+            break
+        time.sleep(common.RETRY_EXEC_INTERVAL)
+    assert len(os.listdir(replica_path)) > 0
```
innobead, line 2665: Do you think we can do this way?

- wait_for_volume_degraded
- wait_for_volume_healthy
- then just `assert len(os.listdir(replica_path)) > 0`

manager/integration/tests/test_node.py
```diff
@@ -2611,3 +2614,80 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(): # NOQA
...
+        node = client.by_id_node(lht_hostId)
+        update_disks = get_update_disks(node.disks)
+        for disk in update_disks.values():
+            if disk["path"] == test_disk_path:
+                disk.allowScheduling = False
+            else:
+                disk.allowScheduling = True
+                disk.evictionRequested = False
+
+        update_node_disks(client, node.name, disks=update_disks, retry=True)
```
innobead, line 2690: Can we just reuse this? Probably you can move `delete_extra_disks_on_host` to common.py.

https://github.com/longhorn/longhorn-tests/blob/a7f2d31378ec3ce5aabb353f5c8618d7495b518d/manager/integration/tests/test_orphan.py#L48-L51

manager/integration/tests/test_node.py
```diff
@@ -2611,3 +2614,80 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(): # NOQA
...
+        common.cleanup_all_volumes(client)
```
innobead, line 2691: Can we move this to the general clean up? https://github.com/longhorn/longhorn/issues/4066

innobead (changes requested):

manager/integration/tests/test_node.py
cchien816: Hi @innobead ,
From observing product behavior, when disk eviction in progress, script can not catch volume degraded status(maybe it's too quick), so currently I kept old code for review.

manager/integration/tests/test_node.py
cchien816: HI @innobead ,
I found we already have `cleanup_test_disks` in common can reuse.
Therefore In finalizer I only kept `cleanup_all_volumes` let no volume mounted to extra disk and let `cleanup_test_disks` can work properly at test case teardown.
Also add enable_default_disk in common teardown for review.

manager/integration/tests/test_node.py
cchien816: HI @innobead ,
I found we already have `cleanup_test_disks` in common can reuse.
Therefore In finalizer I only kept `cleanup_all_volumes` let no volume mounted to extra disk and let `cleanup_test_disks` can work properly at test case teardown.
Also add enable_default_disk in common teardown for review.

innobead (approved):

innobead:
@cchien816 Please backport to 1.3.x and 1.2.x.

PR #981, opened 2022-05-31
Setup pipeline for testing longhorn air gap installation.

yangchiu:
Setup pipeline for testing longhorn air gap installation.

for [#3926](https://github.com/longhorn/longhorn/issues/3926)

### DIFF
--- a/airgap/terraform/main.tf
+++ b/airgap/terraform/main.tf
@@ -1,0 +1,307 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 3.0"
+    }
+    digitalocean = {
+      source = "digitalocean/digitalocean"
+      version = "~> 2.0"
+    }
+  }
+}
+
+provider "aws" {
+  region = var.aws_region
+  access_key = var.aws_access_key
+  secret_key = var.aws_secret_key
+}
+
+data "aws_availability_zones" "available" {
+  state = "available"
+}
+
+provider "digitalocean" {
+  token = var.do_token
+}
+
+locals {
+  registry_username = "registry_user"
+  registry_password = random_string.registry_password.id
+}
+
+resource "random_string" "registry_password" {
+  length           = 16
+  special          = false
+  lower            = true
+  upper            = false
+}
+
+# Create a random string suffix for instance names
+resource "random_string" "random_suffix" {
+  length           = 8
+  special          = false
+  lower            = true
+  upper            = false
+}
+
+# Create a VPC
+resource "aws_vpc" "lh_registry_aws_vpc" {
+
+  cidr_block = "10.0.0.0/16"
+  enable_dns_hostnames = true
+
+  tags = {
+    Name = "lh-registry-vpc-${random_string.random_suffix.id}"
+  }
+}
+
+# Create internet gateway
+resource "aws_internet_gateway" "lh_registry_aws_igw" {
+  vpc_id = aws_vpc.lh_registry_aws_vpc.id
+
+  tags = {
+    Name = "lh-registry-igw-${random_string.random_suffix.id}"
+  }
+}
+
+# Create registry security group
+resource "aws_security_group" "lh_registry_aws_secgrp" {
+  name        = "lh_registry_aws_secgrp_${random_string.random_suffix.id}"
+  description = "Allow all inbound traffic"
+  vpc_id      = aws_vpc.lh_registry_aws_vpc.id
+
+  ingress {
+    description = "Allow SSH"
+    from_port   = 22
+    to_port     = 22
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  ingress {
+    description = "Allow HTTP"
+    from_port   = 80
+    to_port     = 80
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  ingress {
+    description = "Allow HTTPS"
+    from_port   = 443
+    to_port     = 443
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  ingress {
+    description = "Allow TCP connection for longhorn-webhooks"
+    from_port   = 0
+    to_port     = 65535
+    protocol    = "tcp"
+    cidr_blocks = ["10.0.0.0/8"]
+  }
+
+  ingress {
+    description = "Allow UDP connection for longhorn-webhooks"
+    from_port   = 0
+    to_port     = 65535
+    protocol    = "udp"
+    cidr_blocks = ["10.0.0.0/8"]
+  }
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  tags = {
+    Name = "lh-registry-secgrp-${random_string.random_suffix.id}"
+  }
+}
+
+# Create Public subnet
+resource "aws_subnet" "lh_registry_aws_public_subnet" {
+  vpc_id     = aws_vpc.lh_registry_aws_vpc.id
+  availability_zone = data.aws_availability_zones.available.names[0]
+  cidr_block = "10.0.1.0/24"
+
+  map_public_ip_on_launch = true
+
+  tags = {
+    Name = "lh-registry-public-subnet-${random_string.random_suffix.id}"
+  }
+}
+
+# Create private subnet
+resource "aws_subnet" "lh_registry_aws_private_subnet" {
+  vpc_id     = aws_vpc.lh_registry_aws_vpc.id
+  availability_zone = data.aws_availability_zones.available.names[1]
+  cidr_block = "10.0.2.0/24"
+
+  map_public_ip_on_launch = true
+
+  tags = {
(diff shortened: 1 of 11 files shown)


innobead (changes requested):
Can we use airgap as the pipeline/folder name instead of `private-registry`>

yangchiu:
> Can we use airgap as the pipeline/folder name instead of `private-registry`>

modified!

airgap/terraform/main.tf
```diff
@@ -0,0 +1,305 @@
...
+}
+
+# Create instance for registry
+resource "aws_instance" "lh_registry_aws_instance" {
+
+  depends_on = [
+    aws_subnet.lh_registry_aws_public_subnet,
+  ]
+
+  availability_zone = "us-east-1a"
```
innobead, line 219: why we hard code zone here?

airgap/terraform/user-data-scripts/provision_registry_server.sh.tpl
```diff
@@ -0,0 +1,58 @@
...
+cat << EOF > /etc/nginx/letsencrypt.conf
+location ^~ /.well-known {
+    root /var/www/letsencrypt;
+    default_type "text/plain";
+}
+EOF
+
+# (5) create a htpasswd file to be used for authentication
+
+USER=${registry_username}; PASSWORD=${registry_password}; echo "${registry_username}:$(openssl passwd -stdin -apr1 <<< ${registry_password})" > /etc/nginx/registry.password
```
innobead, line 53: Why we need `USER=${registry_username}; PASSWORD=${registry_password};`? you already echo the generated user:password to the file.

airgap/terraform/user-data-scripts/script.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+# (7) create nginx site configuration to be as proxy for the registry container and reload Nginx configuration
```
innobead, line 3: Why separate to two scripts here instead of having one?

airgap/terraform/user-data-scripts/script.sh
```diff
@@ -0,0 +1,87 @@
...
+nginx -s reload
+
+# (10) test registry with docker login command.
+
+docker login -u ${registry_username} -p ${registry_password} ${registry_url}
+
+# (11) pull longhorn images into registry
+
+wget https://raw.githubusercontent.com/longhorn/longhorn/${longhorn_version}/deploy/longhorn-images.txt
+wget https://raw.githubusercontent.com/longhorn/longhorn/v1.2.4/scripts/save-images.sh
```
innobead, line 78: `v1.2.4` this should be configured instead.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -290,16 +337,29 @@ main(){
...
-		run_longhorn_tests ${WORKSPACE}
-	else
-		install_longhorn_master "${TF_VAR_tf_workspace}/longhorn.yaml"
-		run_longhorn_tests ${WORKSPACE}
-	fi
+  if [[ "${AIR_GAP_INSTALLATION}" == true ]]; then
+    create_registry_secret
+    if [[ "${LONGHORN_INSTALL_METHOD}" == "manifest-file" ]]; then
+      get_longhorn_manifest
+      customize_longhorn_manifest_for_airgap
```
innobead, line 345: We need to ensure to set the setting (https://longhorn.io/docs/1.3.0/references/settings/#registry-secret) as well, otherwise the following pods created by longhorn-manager will have errors. (like share manager pods, backing image manager pods, etc)

innobead:
@yangchiu In general, LGTM. Some comments left and please also resolve the conflict.

airgap/terraform/main.tf
yangchiu: > why we hard code zone here?

Just for simplicity.
Actually the AWS account longhorn-intergration-service-account used to run regression tests can only work well in us-east-1, because the resources limits for longhorn-intergration-service-account are only increased in us-east-1. If we change the regression test to run in other zone, we would encounter errors like "resource limit exceeded".

airgap/terraform/user-data-scripts/provision_registry_server.sh.tpl
```diff
@@ -0,0 +1,58 @@
...
+cp ~ubuntu/.ssh/authorized_keys ~root/.ssh/authorized_keys
+service sshd reload
+
+# (1) install docker and run registry container
+
+apt-get update
+apt-get install -y docker.io
+docker run -d -p 127.0.0.1:5000:5000 --name registry registry:2
+
+# (2) install nginx and disable the default site configuration
```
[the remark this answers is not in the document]
yangchiu: > I am curious why not just use container way to run nginx?

I just follow the documentation of setup private registry.
But, Yes, I'll try to do this.

airgap/terraform/main.tf
yangchiu: Modified to use a variable with default value "us-east-1".

airgap/terraform/user-data-scripts/provision_registry_server.sh.tpl
yangchiu: Refined

airgap/terraform/user-data-scripts/script.sh
yangchiu: let it be ${longhorn_version}

test_framework/scripts/longhorn-setup.sh
yangchiu: Refined

airgap/terraform/user-data-scripts/provision_registry_server.sh.tpl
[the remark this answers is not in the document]
yangchiu: Removed nginx and tls setup. Use docker instead.

airgap/terraform/user-data-scripts/script.sh
yangchiu: Moved all the setup code from provision script to script.sh

test_framework/scripts/longhorn-setup.sh
yangchiu: **For installation from manifest file**, setting `registry-secret` in `default-setting.yaml` works as expected. If the value is set, longhorn regression test can pass as normal, otherwise there would be lots of ImagePullBackOff errors during the test.

**But for installation from helm chart**, even though we set the `defaultSettings.registrySecret` value in `chart/values.yaml `according to the [documentation](https://longhorn.io/docs/1.3.0/advanced-resources/deploy/airgap/#use-default-image-name), the value would not be set in `default-setting.yaml` since it's been removed in [this commit](https://github.com/longhorn/longhorn/commit/b79b914967a84d66c5bb141e8056cd638d52ace2#diff-b4292a87e40a1ece0e015a805cf4ccc28516e54f3bc3e5a36e03526f8179004b), so longhorn-manager would still have errors during the test.

innobead: Thanks, @yangchiu found this. For the chart install/upgrade, the setting will be implicitly updated as per the secret from the manage daemonset as below, so we don't need to update that setting separately. 

https://github.com/longhorn/longhorn-manager/blob/ddf1b8dfe8f3ccceeeafff2909d9c5763bf33122/app/daemon.go#L155

yangchiu:
> @yangchiu In general, LGTM. Some comments left and please also resolve the conflict.

code refined and conflicts resolved.

airgap/terraform/main.tf
```diff
@@ -0,0 +1,309 @@
...
+  associate_public_ip_address = true
+
+  subnet_id = aws_subnet.lh_registry_aws_public_subnet.id
+  vpc_security_group_ids = [
+    aws_security_group.lh_registry_aws_secgrp.id
+  ]
+
+  key_name = aws_key_pair.lh_registry_aws_pair_key.key_name
+
+  user_data = data.template_file.provision_registry_server.rendered
```
innobead, line 236: The content in `provision_registry_server` is not really about the register server provisioning, so probably the name can be changed to a meaningful name?

airgap/terraform/user-data-scripts/provision_registry_server.sh.tpl
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cp ~ubuntu/.ssh/authorized_keys ~root/.ssh/authorized_keys
```
innobead, line 3: Why we need to do this for the root account? Don't we just use the default account to do the ssh operation only in terraform script?

innobead (commented):
LGTM, just have 2 comments. After answering, then we can merge.

airgap/terraform/user-data-scripts/provision_registry_server.sh.tpl
yangchiu: The original version of commit had lots of sudo operations in the post-setup scripts, but it's not needed anymore. 
Remove this part and test again.

airgap/terraform/main.tf
yangchiu: Remove provision_registry_server template file, and rename script.sh to provision_registry_server.sh, then test again.

yangchiu:
> LGTM, just have 2 comments. After answering, then we can merge.

code refined and conflicts resolved.

PR #982, opened 2022-05-31
Fix test case test_cloning_interrupted

cchien816:

https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/139/testReport/junit/tests/test_cloning/test_cloning_interrupted/
https://github.com/longhorn/longhorn/issues/4047

Fix test case test_cloning_interrupted

Add `wait_for_volume_degraded` for clone volume before crash source volume replica process.
Or the cloning won't interrupt in recent build.

### DIFF
--- a/manager/integration/tests/test_cloning.py
+++ b/manager/integration/tests/test_cloning.py
@@ -15,6 +15,7 @@ from common import generate_random_suffix, wait_for_volume_endpoint
 from common import wait_for_snapshot_count, DATA_SIZE_IN_MB_3
 from common import get_clone_volume_name
 from common import create_storage_class, storage_class  # NOQA
+from common import wait_for_volume_degraded
 
 
 # Kept some fixtures specifically for volume cloning module to avoid cleaning
@@ -387,6 +388,7 @@ def test_cloning_interrupted(client, core_api, pvc, pod, clone_pvc, clone_pod): 
                                  'initiated')
 
     # Step-6
+    wait_for_volume_degraded(client, clone_volume_name)
     crash_replica_processes(client, core_api, source_volume_name)
 
     # Step-7
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4767,9 +4767,14 @@ def wait_for_pod_annotation(core_api,
 def wait_for_volume_clone_status(client, name, key, value):
     for _ in range(RETRY_COUNTS):
         volume = client.by_id_volume(name)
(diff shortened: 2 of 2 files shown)


manager/integration/tests/test_cloning.py
```diff
@@ -387,6 +388,7 @@ def test_cloning_interrupted(client, core_api, pvc, pod, clone_pvc, clone_pod):
                                  'initiated')
 
     # Step-6
+    wait_for_volume_degraded(client, clone_volume_name)
```
innobead, line 391: Why not `wait_for_volume_healthy`?

innobead (changes requested):

manager/integration/tests/test_cloning.py
innobead, line 391: This seems not right, because we don't need to check if the cloning volume is degraded or not. The process is just:
- crash replicas of the source volume
- then check if the cloning volume is faulted and clone status is failed (here, sometimes this step is failed)

About failing to check the clone status is failed, it's a side effect of https://github.com/longhorn/longhorn-tests/commit/831ba4fb10e5b14ad1a4c723e6fff43d2932c481 which causes some crashed replicas have been recreated before finishing `crash_replica_processes`...

cc @c3y1huang

innobead:
@cchien816 You don't need to fix this PR as my explanation https://github.com/longhorn/longhorn-tests/pull/982#discussion_r887620154.

manager/integration/tests/test_cloning.py
c3y1huang: Adding reference, https://github.com/longhorn/longhorn/issues/4045.

innobead:
https://github.com/longhorn/longhorn-tests/pull/987 cc @mantissahz

cchien816:
Hi @innobead ,

After https://github.com/longhorn/longhorn/issues/4045 fixed, this case still fail, from [log](https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/188/testReport/junit/tests/test_cloning/test_cloning_interrupted/) I can see the clone status were `'state': 'completed'` not `faulted` due to timing issue, did local test of this PR an result pass, thank you.

cchien816:
Hi @innobead ,

I updated PR for `VOLUME_FIELD_CLONE_STATUS` is not ready issue, thank you

innobead (approved):

innobead:
@cchien816 Please backport to 1.3. Thanks.

PR #983, opened 2022-05-31
integration: Use a longer timeout for test_exporting_backing_image_from_volume

shuo-wu:
Creating a backing image from a volume then copying it to other nodes during volume attachment would take a longer time. Hence we need to use a longer timeout.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -56,8 +56,8 @@ BACKING_IMAGE_EXT4_SIZE = 32 * Mi
 PORT = ":9500"
 
 RETRY_COMMAND_COUNT = 3
-RETRY_COUNTS = 300
-RETRY_INTERVAL = 0.5
+RETRY_COUNTS = 150
+RETRY_INTERVAL = 1
 RETRY_INTERVAL_LONG = 2
 RETRY_BACKUP_COUNTS = 300
 RETRY_BACKUP_INTERVAL = 1
@@ -1656,13 +1656,13 @@ def wait_for_volume_detached_unknown(client, name):
     return wait_for_volume_detached(client, name)
 
 
-def wait_for_volume_healthy(client, name):
+def wait_for_volume_healthy(client, name, retry_count=RETRY_COUNTS):
     wait_for_volume_status(client, name,
                            VOLUME_FIELD_STATE,
-                           VOLUME_STATE_ATTACHED)
+                           VOLUME_STATE_ATTACHED, retry_count)
     wait_for_volume_status(client, name,
                            VOLUME_FIELD_ROBUSTNESS,
-                           VOLUME_ROBUSTNESS_HEALTHY)
+                           VOLUME_ROBUSTNESS_HEALTHY, retry_count)
     return wait_for_volume_endpoint(client, name)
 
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -1677,9 +1677,10 @@ def wait_for_volume_faulted(client, name):
                                   VOLUME_ROBUSTNESS_FAULTED)
 
 
-def wait_for_volume_status(client, name, key, value):
+def wait_for_volume_status(client, name, key, value,
+                           timeout_in_sec=RETRY_COUNTS//2):
```
innobead, line 1681: Why not just 

`def wait_for_volume_status(client, name, key, value, timeout_in_sec=RETRY_COUNTS):`

manager/integration/tests/test_backing_image.py
```diff
@@ -471,7 +471,7 @@ def test_exporting_backing_image_from_volume(client, volume_name):  # NOQA
         client, volume_name=volume2_name, size=str(1 * Gi),
         backing_image=backing_img1["name"])
     volume2 = volume2.attach(hostId=hostId)
-    volume2 = wait_for_volume_healthy(client, volume2_name)
+    volume2 = wait_for_volume_healthy(client, volume2_name, 300)
```
innobead, line 474: Then, `wait_for_volume_healthy(client, volume2_name, RETRY_COUNTS * 2)`

innobead (changes requested):

manager/integration/tests/common.py
```diff
@@ -1640,13 +1640,13 @@ def wait_for_volume_detached_unknown(client, name):
     return wait_for_volume_detached(client, name)
 
 
-def wait_for_volume_healthy(client, name):
+def wait_for_volume_healthy(client, name, timeout_in_sec=RETRY_COUNTS//2):
```
innobead, line 1643: `def wait_for_volume_healthy(client, name, timeout_in_sec=RETRY_COUNTS):`

manager/integration/tests/common.py
shuo-wu: Since the wait interval is 0.5s rather than 1s.
I don't want to affect other test cases or slow down the status check/wait since I choose not to modify the interval.

shuo-wu:
I just changed the default retry interval from 0.5s to 1s. Not sure if there is any other impact.

manager/integration/tests/common.py
```diff
@@ -1656,13 +1656,13 @@ def wait_for_volume_detached_unknown(client, name):
     return wait_for_volume_detached(client, name)
 
 
-def wait_for_volume_healthy(client, name):
+def wait_for_volume_healthy(client, name, timeout_in_sec=RETRY_COUNTS):
```
innobead, line 1659: `timeout_in_sec` seems confusing, suggest `retry_count` instead.

manager/integration/tests/common.py
```diff
@@ -1693,9 +1693,10 @@ def wait_for_volume_faulted(client, name):
                                   VOLUME_ROBUSTNESS_FAULTED)
 
 
-def wait_for_volume_status(client, name, key, value):
+def wait_for_volume_status(client, name, key, value,
+                           timeout_in_sec=RETRY_COUNTS):
```
innobead, line 1697: The same, suggest `retry_count`.

cchien816:
HI @shuo-wu ,

I use this PR code to run test case `test_exporting_backing_image_from_volume` and got fail result the same with https://github.com/longhorn/longhorn/issues/4256 logged, so maybe extend time won't be the best solution, thank you

cchien816 (commented):
HI @shuo-wu ,

I use this PR code to run test case test_exporting_backing_image_from_volume and got fail result the same with https://github.com/longhorn/longhorn/issues/4256 logged, so maybe extend time won't be the best solution, thank you

innobead (approved):

innobead:
@shuo-wu Please backport to 1.3.x and 1.2.x. Thanks.

shuo-wu:
> 

It seems that something else is wrong with the longhorn-manager. I will check it then.

PR #986, opened 2022-06-02
backup: implement backup ownership transfer test

c3y1huang:
https://github.com/longhorn/longhorn/issues/4064

https://github.com/longhorn/longhorn-manager/pull/1361

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -93,6 +93,8 @@ from backupstore import set_backupstore_url, set_backupstore_credential_secret, 
 from backupstore import reset_backupstore_setting  # NOQA
 from backupstore import set_backupstore_s3, backupstore_get_secret  # NOQA
 
+from test_node import set_node_cordon
+
 from kubernetes import client as k8sclient
 
 BACKUPSTORE = get_backupstores()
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -3006,6 +3008,163 @@ def test_backup_lock_creation_during_deletion(set_random_backupstore, client, co
...
+
+    new_data = {
+        'pos': 0,
+        'content': generate_random_data(20 * Mi),
+    }
+
+    _, backup, snapshot, written_data = create_backup(
+        client, volume_name, data=new_data)
+
+    original_ownerID = backup.ownerID
```
innobead, line 3050: oh, `ownerID` is for testing purpose.

manager/integration/tests/test_basic.py
```diff
@@ -3006,6 +3008,163 @@ def test_backup_lock_creation_during_deletion(set_random_backupstore, client, co
...
+    instance_managers = client.list_instance_manager()
+    engine_instance_manager = None
+    for im in instance_managers:
+        if im.managerType != "engine":
+            continue
+
+        if im.nodeID != backup.ownerID:
+            continue
+
+        engine_instance_manager = im
```
innobead, line 3061: you can break here when finding out the matched IM-E or you just use the below way.

```python
if im.managerType == "engine" and im.nodeID == backup.ownerID:
    engine_instance_manager = im
     break
```

manager/integration/tests/test_basic.py
```diff
@@ -3006,6 +3008,163 @@ def test_backup_lock_creation_during_deletion(set_random_backupstore, client, co
...
+        time.sleep(RETRY_INTERVAL)
+
+    backup_volume, backup = find_backup(client, volume.name, snapshot.name)
+    assert backup.ownerID != original_ownerID, \
+        f'expecting {backup.name} ownerID {original_ownerID} to change'
+
+    wait_for_backup_completion(client, volume.name, snapshot.name)
+    volume.detach(hostId="")
+
+    if recover:
```
innobead, line 3100: the node will be uncordoned anyway because it will be done int he reqeust finalizer, why we need this here?

manager/integration/tests/test_basic.py
```diff
@@ -3006,6 +3008,163 @@ def test_backup_lock_creation_during_deletion(set_random_backupstore, client, co
...
+            client, core_api, volume_name, request,
+        )
+
+    common.delete_backup(client, volume.name, backup.name)
+
+    try:
+        _, backup = common.find_backup(client, volume.name, snapshot.name)
+    except AssertionError:
+        backup = None
+    assert backup is None
```
innobead, line 3127: This did not test after deleting the backup, and it's IM-E down, then the backup will be still deleted, so I feel this test function still testing the same thing `backup_ownership_transfer_when_instance_manager_is_down`

manager/integration/tests/test_basic.py
```diff
@@ -3006,6 +3008,163 @@ def test_backup_lock_creation_during_deletion(set_random_backupstore, client, co
...
+
+
+def test_backup_restore_when_instance_manager_is_down(set_random_backupstore, client, core_api, volume_name, request):  # NOQA
```
innobead, line 3130: The same reason as the previous comment for `test_backup_delete_when_instance_manager_is_down`.

Restoring a backup to a volume should be related to volume controller instead of backup controller? Should we need to test this here?

innobead (changes requested):

manager/integration/tests/test_basic.py
c3y1huang: Ok, we can run by `kubectl`.
```
def get_backup_ownerID(backup_name):
    return subprocess.Popen(
        "kubectl get backup " + backup_name + " -n longhorn-system -o jsonpath='{.status.ownerID}'",  # NOQA,
        shell=True,
        stdout=subprocess.PIPE
    ).communicate()[0].decode('utf-8').strip()
```

manager/integration/tests/test_basic.py
c3y1huang: Updated.

c3y1huang:
Close in favor of https://github.com/longhorn/longhorn/issues/4076.

PR #987, opened 2022-06-02
Check replicas failed in parallel.

mantissahz:
Implement parallel checks for replica failures in crash_replica_processes.

longhorn/longhorn#4045

Signed-off-by: James Lu <jamesluhz@gmail.com>

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -9,6 +9,8 @@ import subprocess
 import json
 import hashlib
 import signal
+import types
+import threading
 
 import socket
 import pytest
@@ -1937,8 +1939,72 @@ def delete_replica_processes(client, api, volname):
         exec_instance_manager(api, rm_name, delete_command)
 
 
+class AssertErrorCheckThread(threading.Thread):
+    """
+        This class is used for catching exception caused in threads,
+        especially for AssertionError now.
+
+        Parameters:
+            target  :       The threading function.
+            args    :       Arguments of the target fucntion.
+    """
+    def __init__(self, target, args):
+        threading.Thread.__init__(self)
+        self.target = target
+        self.args = args
+        self.asserted = None
+
+    def run(self):
+        try:
+            self.target(*self.args)
+        except AssertionError as e:
+            self.asserted = e
+
+    def join(self):
+        threading.Thread.join(self)
+        if self.asserted:
+            raise self.asserted
+
+
+def create_assert_error_check_thread(func, *args):
+    """
+        Do func by threading with arguments
+
+        Parameters:
+            func:   function that want to do things in parallel.
+            args:   arguments for function.
+    """
+    assert isinstance(func, types.FunctionType), "First arg is not a function."
+
+    t = AssertErrorCheckThread(target=func, args=args)
+    t.start()
+
+    return t
+
+
+def assert_from_assert_error_check_threads(thrd_list):
+    """
+        Check all threads in thrd_list are done and their status
+
+        Parameters:
+            thrd_list: thread list created by create_assert_error_check_thread.
+    """
+    assert isinstance(thrd_list, list), "thrd_list is not a list"
+
+    err_list = []
+    for t in thrd_list:
+        try:
+            t.join()
+        except AssertionError as e:
+            print(e)
+            err_list.append(e)
+    if err_list:
+        assert False, err_list
+
+
 def crash_replica_processes(client, api, volname, replicas=None,
                             wait_to_fail=True):
+    threads = []
 
     if replicas is None:
         volume = client.by_id_volume(volname)
@@ -1950,10 +2016,15 @@ def crash_replica_processes(client, api, volname, replicas=None,
                        "' | grep -v grep | awk '{print $2}'`"
         exec_instance_manager(api, r.instanceManagerName, kill_command)
 
-        # FIXME: this is a work around, should check this in-parallel.
-        #        https://github.com/longhorn/longhorn/issues/4045
         if wait_to_fail is True:
-            wait_for_replica_failed(client, volname, r['name'])
+            thread = create_assert_error_check_thread(
+                wait_for_replica_failed,
+                client, volname, r['name'], RETRY_COUNTS*2, RETRY_INTERVAL/2
+            )
+            threads.append(thread)
+
+    if wait_to_fail:
+        assert_from_assert_error_check_threads(threads)
 
 
 def exec_instance_manager(api, im_name, cmd):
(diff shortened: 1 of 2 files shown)


c3y1huang:
nit: can we have a meaningful commit message instead of the issue link? For example.
```
Implement parallel checks for replica failures in crash_replica_processes.

Longhorn-4045
```

manager/integration/tests/common.py
```diff
@@ -1950,10 +1952,19 @@ def crash_replica_processes(client, api, volname, replicas=None,
...
         exec_instance_manager(api, r.instanceManagerName, kill_command)
 
-        # FIXME: this is a work around, should check this in-parallel.
-        #        https://github.com/longhorn/longhorn/issues/4045
+        # XXX: check all replica status in-parallel but it might not make sure
+        #      that replica is failed after checking and quickly recovering.
+        #      https://github.com/longhorn/longhorn/issues/4045
         if wait_to_fail is True:
-            wait_for_replica_failed(client, volname, r['name'])
+            check_t = threading.Thread(target=wait_for_replica_failed,
```
innobead, line 1959: The assertion except can be caught in the main thread? how to know if `wait_for_replica_failed` failed?

manager/integration/tests/common.py
```diff
@@ -1950,10 +1952,19 @@ def crash_replica_processes(client, api, volname, replicas=None,
                        "' | grep -v grep | awk '{print $2}'`"
         exec_instance_manager(api, r.instanceManagerName, kill_command)
 
-        # FIXME: this is a work around, should check this in-parallel.
-        #        https://github.com/longhorn/longhorn/issues/4045
+        # XXX: check all replica status in-parallel but it might not make sure
```
c3y1huang, line 1955: Out of curiosity, what is `XXX`?

Can we have a couple of tests run for this to see if tests are still flaky?

manager/integration/tests/common.py
innobead: If can't, probably you cound consider using https://pypi.org/project/atomic/ or similar, create an atomic count and an ad-hoc closure to the `wait_for_replica_failed`, then increase the count if any failure.  (of course, you should modify `wait_for_replica_failed`.

When all threads are finished, check the count to understand if all replicas crashed already. 

Also, I don't think you should customize rety_* variables in `wait_for_replica_failed`. You can adjust it in the future if needed.

manager/integration/tests/common.py
mantissahz: XXX, I used for something needed to be noticed.
Here is one test run this morning [1444](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1444/), yesterday I run two or three runs for this.
I could run another one to test.

innobead: You definitely use eclipse :slightly_smiling_face:

c3y1huang: Ah :slightly_smiling_face:

manager/integration/tests/common.py
mantissahz: It should be caught in the main thread, let me do some tests.

manager/integration/tests/common.py
mantissahz: Agree, I would remove `XXX`.

manager/integration/tests/common.py
mantissahz: I added a Thread class and the assertion except would be caught by this thread class then asserted the exception again.

manager/integration/tests/common.py
mantissahz: updated.

manager/integration/tests/common.py
```diff
@@ -1937,8 +1938,37 @@ def delete_replica_processes(client, api, volname):
...
+        try:
+            self.target(*self.args)
+        except AssertionError as e:
+            self.asserted = True
+            print("[DEBUG] Err: {} in the thread".format(e))
+
+    def join(self):
+        threading.Thread.join(self)
+        if self.asserted:
+            assert False
```
innobead, line 1966: Can we just save the throed assert error here? then rethrow out instead of `assert False` which is too general, and would prefer assertion w/ the correct info instead of just checking the printed logs. WDYT?

manager/integration/tests/common.py
```diff
@@ -1950,10 +1980,19 @@ def crash_replica_processes(client, api, volname, replicas=None,
...
-            wait_for_replica_failed(client, volname, r['name'])
+            check_t = replicaThread(target=wait_for_replica_failed,
+                                    args=(client, volname, r['name'],
+                                          RETRY_COUNTS*2, RETRY_INTERVAL/2))
+            threads.append(check_t)
+            check_t.start()
+
+    if wait_to_fail:
+        for t in threads:
+            t.join()
```
innobead, line 1995: Suggest just raise an accumulated or one except when all threads finished instead of leaving some unhandled threads behind. This way can help us find some issues earlier like generally orphaned threads.

innobead (changes requested):

manager/integration/tests/common.py
mantissahz: It seems we only catch the exception at run(), but we can keep the exception and error message at run() time and reraise the exception at join() time.

manager/integration/tests/common.py
mantissahz: Agree, I will modify codes for this.

mantissahz:
More general version for threading to check function status in parallel.
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1497/

manager/integration/tests/common.py
```diff
@@ -1950,10 +2016,18 @@ def crash_replica_processes(client, api, volname, replicas=None,
                        "' | grep -v grep | awk '{print $2}'`"
         exec_instance_manager(api, r.instanceManagerName, kill_command)
 
-        # FIXME: this is a work around, should check this in-parallel.
-        #        https://github.com/longhorn/longhorn/issues/4045
+        # Check all replica status in-parallel but it might not make sure that
+        # replica is failed after checking and quickly recovering.
+        # https://github.com/longhorn/longhorn/issues/4045
```
innobead, line 2021: I think we can remove these comments.

manager/integration/tests/common.py
```diff
@@ -1937,8 +1939,72 @@ def delete_replica_processes(client, api, volname):
...
+    """
+    assert isinstance(func, types.FunctionType), "First arg is not a function."
+
+    t = CatchExceptThread(target=func, args=args)
+    t.start()
+
+    return t
+
+
+def check_all_jobs_done_and_status(thrd_list):
```
innobead, line 1985: Suggest renaming to `assert_from_assert_error_check_threads`

manager/integration/tests/common.py
mantissahz: Removed.

manager/integration/tests/common.py
```diff
@@ -1937,8 +1939,72 @@ def delete_replica_processes(client, api, volname):
         exec_instance_manager(api, rm_name, delete_command)
 
 
+class CatchExceptThread(threading.Thread):
```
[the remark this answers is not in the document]
mantissahz: Renamed

manager/integration/tests/common.py
```diff
@@ -1937,8 +1939,72 @@ def delete_replica_processes(client, api, volname):
...
+        except AssertionError as e:
+            self.asserted = e
+
+    def join(self):
+        threading.Thread.join(self)
+        if self.asserted:
+            raise self.asserted
+
+
+def do_job_in_parallel(func, *args):
```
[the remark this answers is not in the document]
mantissahz: Renamed.

manager/integration/tests/common.py
mantissahz: Renamed.

innobead:
Thank you. Waiting for the testing result before merging. @mantissahz

innobead (approved):

mantissahz:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1526/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1538/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1539/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1540/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1545/
still a bit flaky when doing wait_for_volume_faulted

innobead:
@mantissahz Please backport to 1.3.x and 1.2.x. Thanks.

PR #993, opened 2022-06-15
Fix os distro version is not passed to terraform

yangchiu:
(1) Fix os distro version is not passed to terraform
(2) Fix rockylinux cloud-init status permission issue
(3) Update default rockylinux version


innobead (approved):

innobead:
See if need to backport to 1.2.x

PR #995, opened 2022-06-16
Manual test replica scale-down warning

c3y1huang:
https://github.com/longhorn/longhorn/issues/4120


docs/content/manual/release-specific/v1.4.0/test-storage-network.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Test replica scale-down warning
+---
+
+## Related issue
+https://github.com/longhorn/longhorn/issues/4120
+
+## Steps
+
+**Given** `Replica Auto Balance` set to `least-effort` or `best-effort`.
```
innobead, line 10: Why we need to care `replica auto balance` setting here?

c3y1huang: Observed from the support bundle from the original issue, the environment had the replica auto-balance set with best-effort.

innobead: After discussing with @c3y1huang , this setting should be the pre-condition, because only when auto-rebalancing is enabled, the extra replicas will be automatically removed during the attached status. Otherwise, users need to manually delete the extra replicas.

innobead (approved):

PR #999, opened 2022-06-17
fix flaky test_allow_volume_creation_with_degraded_availability_csi

cchien816:

1. When volume status was False, after deployment scale up from 0 to 1, get pod data  immediately may cause `Reason: Handshake status 500 Internal Server Error`, add retry mechanism.
2. When volume status was True, after deployment scale up from 0 to 1, wait volume healthy then get pod data to avoid flaky status (Previously pass case all at volume healthy status).

https://github.com/longhorn/longhorn/issues/4121

### DIFF
--- a/manager/integration/tests/test_csi.py
+++ b/manager/integration/tests/test_csi.py
@@ -4,6 +4,7 @@ import subprocess
 import random
 
 import common
+import time
 from common import client, core_api, apps_api # NOQA
 from common import csi_pv, pod_make, pvc, storage_class  # NOQA
 from common import make_deployment_with_pvc  # NOQA
@@ -620,15 +621,25 @@ def test_allow_volume_creation_with_degraded_availability_csi(
                                   common.VOLUME_STATE_ATTACHED)
     common.wait_for_volume_condition_scheduled(client, vol.name, "status",
                                                common.CONDITION_STATUS_FALSE)
-    pod = common.wait_and_get_any_deployment_pod(core_api, deployment_name)
-    assert created_md5sum == get_pod_data_md5sum(core_api,
-                                                 pod.metadata.name,
-                                                 data_path)
+
+    # This is a workaround, this flaky step is because sometimes, pod
+    # just changed to running state can not get data
+    # https://github.com/longhorn/longhorn-tests/pull/999#discussion_r902125882 # NOQA
+    for i in range(common.RETRY_COMMAND_COUNT):
+        try:
+            pod = common.wait_and_get_any_deployment_pod(core_api,
+                                                         deployment_name)
+            pod_md5 = get_pod_data_md5sum(core_api,
+                                          pod.metadata.name,
+                                          data_path)
+        except Exception as e:
+            print(e)
+            time.sleep(common.RETRY_INTERVAL)
+            continue
+    assert created_md5sum == pod_md5
 
     client.update(node3, allowScheduling=True)
-    common.wait_for_rebuild_start(client, vol.name)
     vol = client.by_id_volume(vol.name)
-    assert vol.conditions[VOLUME_CONDITION_SCHEDULED]['status'] == "True"
     common.wait_for_rebuild_complete(client, vol.name)
 
     deployment['spec']['replicas'] = 0
@@ -642,8 +653,8 @@ def test_allow_volume_creation_with_degraded_availability_csi(
                                          namespace='default',
                                          name=deployment_name)
     common.wait_for_volume_status(client, vol.name,
-                                  common.VOLUME_FIELD_STATE,
-                                  common.VOLUME_STATE_ATTACHED)
+                                  common.VOLUME_FIELD_ROBUSTNESS,
+                                  common.VOLUME_ROBUSTNESS_HEALTHY)
 
     pod = common.wait_and_get_any_deployment_pod(core_api, deployment_name)
     assert created_md5sum == get_pod_data_md5sum(core_api,


manager/integration/tests/test_csi.py
```diff
@@ -620,15 +621,24 @@ def test_allow_volume_creation_with_degraded_availability_csi(
...
+        try:
+            pod = common.wait_and_get_any_deployment_pod(core_api,
+                                                         deployment_name)
+            pod_md5 = get_pod_data_md5sum(core_api,
+                                          pod.metadata.name,
+                                          data_path)
+        except Exception as e:
+            print(e)
+            time.sleep(common.RETRY_INTERVAL)
+            continue
```
innobead, line 637: Know what you are doing, but I am curious why this will happen?

cc @shuo-wu @derekbit have you seen this before, after the workload running, but encountered an error as below described by @cchien816 at https://github.com/longhorn/longhorn-tests/pull/999#issue-1274738394?

```
When volume status was False, after deployment scale up from 0 to 1, get pod data immediately may cause Reason: Handshake status 500 Internal Server Error, add retry mechanism.
```

innobead: @cchien816 do you have the complete error info when the flaky happened?

cchien816: Below are error logs when flaky `When volume status was False, after deployment scale up from 0 to 1, get pod data immediately may cause Reason: Handshake status 500 Internal Server Error, add retry mechanism.` happened. Which can see the error happened at kubernetes socket call

```
test_csi.py:624: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
common.py:737: in get_pod_data_md5sum
    return stream(
/usr/local/lib/python3.9/site-packages/kubernetes/stream/stream.py:35: in stream
    return func(*args, **kwargs)
/usr/local/lib/python3.9/site-packages/kubernetes/client/api/core_v1_api.py:841: in connect_get_namespaced_pod_exec
    (data) = self.connect_get_namespaced_pod_exec_with_http_info(name, namespace, **kwargs)  # noqa: E501
/usr/local/lib/python3.9/site-packages/kubernetes/client/api/core_v1_api.py:927: in connect_get_namespaced_pod_exec_with_http_info
    return self.api_client.call_api(
/usr/local/lib/python3.9/site-packages/kubernetes/client/api_client.py:340: in call_api
    return self.__call_api(resource_path, method,
/usr/local/lib/python3.9/site-packages/kubernetes/client/api_client.py:172: in __call_api
    response_data = self.request(
/usr/local/lib/python3.9/site-packages/kubernetes/stream/stream.py:30: in _intercept_request_call
    return ws_client.websocket_call(config, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

configuration = <kubernetes.client.configuration.Configuration object at 0x7f719bdf2b80>
args = ('GET', 'https://10.43.0.1:443/api/v1/namespaces/default/pods/longhorn-testvol-03wkb7-dep-5fcc659ff7-nwvb2/exec')
kwargs = {'_preload_content': True, '_request_timeout': None, 'body': None, 'headers': {'Accept': '*/*', 'User-Agent': 'OpenAPI...7dLaeo4SCVRC8GfXJh5krd3mTfLTjAWEnNWUl3BJvpXvzDQ9QwZsv2ULb4trnWjbAwoSX5togdY_Kk-MG1s4XxfX3mGqTsYeGhNkYQAaPJGuKNA'}, ...}
url = 'https://10.43.0.1:443/api/v1/namespaces/default/pods/longhorn-testvol-03wkb7-dep-5fcc659ff7-nwvb2/exec?command=%2Fbin...&command=-c&command=md5sum+%2Fdata%2Ftest+%7C+awk+%27%7Bprint+%241%7D%27&stderr=True&stdin=False&stdout=True&tty=False'
_request_timeout = None, _preload_content = True, capture_all = True
headers = {'Accept': '*/*', 'User-Agent': 'OpenAPI-Generator/11.0.0/python', 'authorization': 'bearer eyJhbGciOiJSUzI1NiIsImtpZC...TAkjjI7dLaeo4SCVRC8GfXJh5krd3mTfLTjAWEnNWUl3BJvpXvzDQ9QwZsv2ULb4trnWjbAwoSX5togdY_Kk-MG1s4XxfX3mGqTsYeGhNkYQAaPJGuKNA'}
query_params = [('command', '/bin/sh'), ('command', '-c'), ('command', "md5sum /data/test | awk '{print $1}'"), ('stderr', True), ('stdin', False), ('stdout', True), ...]
key = 'tty', value = False, command = "md5sum /data/test | awk '{print $1}'"
```
And below are exception logs for reference
```
kubernetes.client.rest.ApiException: (0)
Reason: Handshake status 500 Internal Server Error
```
cc @innobead , @shuo-wu

shuo-wu: I am thinking if we can bypass this by adding a liveness or readiness probe (that can detect the mount point availability) for the deployment.

innobead: Or you can just simply check the status of deployment is ready via API before `get_pod_data_md5sum`? this should be caused by the pod is not ready yet.

shuo-wu: Function `wait_and_get_any_deployment_pod `already guarantees the deployment pod is running.

shuo-wu: Based on @cchien816's testing result, the liveness probe does not help. It seems that we have to this workaround.

cchien816: Reference to https://github.com/longhorn/longhorn-manager/blob/ddf1b8dfe8f3ccceeeafff2909d9c5763bf33122/examples/pod_with_pvc.yaml#L25-L31 and function `get_liveness_probe_spec`  from common,py, I add livenessprobe into deployment pod and did some test but the k8s API call error still happen

PR #1003, opened 2022-06-20
Add case test_migration_with_failed_replica

cchien816:

Add test case `test_migration_with_failed_replica`

### DIFF
--- a/manager/integration/tests/test_migration.py
+++ b/manager/integration/tests/test_migration.py
@@ -1,5 +1,5 @@
 import pytest
-
+import subprocess
 import common
 from common import clients, volume_name  # NOQA
 from common import get_random_client
@@ -47,7 +47,7 @@ def migration_confirm_test(clients, volume_name, backing_image=""):  # NOQA
     volume = common.wait_for_volume_detached(client, volume_name)
 
     # verify test data
-    check_volume_data(client, volume_name, data)
+    check_detached_volume_data(client, volume_name, data)
 
     client.delete(volume)
     wait_for_volume_delete(client, volume_name)
@@ -92,7 +92,7 @@ def migration_rollback_test(clients, volume_name, backing_image=""):  # NOQA
     volume = common.wait_for_volume_detached(client, volume_name)
 
     # verify test data
-    check_volume_data(client, volume_name, data)
+    check_detached_volume_data(client, volume_name, data)
 
     client.delete(volume)
     wait_for_volume_delete(client, volume_name)
@@ -128,8 +128,7 @@ def test_migration_with_unscheduled_replica(clients, volume_name):  # NOQA
 
 @pytest.mark.coretest  # NOQA
 @pytest.mark.migration # NOQA
-@pytest.mark.skip(reason="TODO") # NOQA
-def test_migration_with_failed_replica(clients, volume_name):  # NOQA
+def test_migration_with_failed_replica(clients, request, volume_name):  # NOQA
     """
     Test that a degraded migratable RWX volume that contain an failed replica
     can be migrated.
@@ -147,7 +146,55 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
        And the old failed replica will be cleaned up.
     8. Validate initially written test data.
     """
-    return
+    def finalizer():
+        exec_cmd = ["mkdir", "-p", "/var/lib/longhorn/replicas"]
+        subprocess.check_output(exec_cmd)
+
+    request.addfinalizer(finalizer)
+
+    client, volume, data = setup_migration_test(clients,
+                                                volume_name,
+                                                replica_cnt=3)
+
+    current_node = common.get_self_host_id()
+    hosts = get_hosts_for_migration_test(clients)
+    migrate_target = hosts[0]
+    volume.attach(hostId=current_node)
+    volume = common.wait_for_volume_healthy(client, volume_name)
+
+    old_replicas = []
+    for replica in volume.replicas:
+        old_replicas.append(replica.name)
+
+    exec_cmd = ["rm", "-rf",  "/var/lib/longhorn/replicas"]
+    subprocess.check_output(exec_cmd)
+    volume = common.wait_for_volume_degraded(client, volume_name)
+
+    volume.attach(hostId=migrate_target)
+    volume = common.wait_for_volume_migration_ready(client, volume_name)
+
+    new_replicas = 0
+    for replica in volume.replicas:
+        if replica.hostId == current_node:
+            assert replica.running is False
+        else:
+            assert replica.running is True
+            if replica.name not in old_replicas:
+                new_replicas = new_replicas + 1
+    assert new_replicas == 2
+
+    volume.detach(hostId=current_node)
+
+    volume = common.wait_for_volume_migration_node(client,
+                                                   volume_name,
+                                                   migrate_target)
+
+    volume.updateReplicaCount(replicaCount=2)
+    volume = common.wait_for_volume_healthy(client, volume_name)
+    volume.detach(hostId="")
+    volume = common.wait_for_volume_detached(client, volume_name)
+    volume = client.by_id_volume(volume_name)
+    check_detached_volume_data(client, volume_name, data)
 
 
 @pytest.mark.coretest  # NOQA
@@ -206,12 +253,12 @@ def get_hosts_for_migration_test(clients): # NOQA
     hosts = []
     current_host = common.get_self_host_id()
     for host in list(clients):
-        if host is not current_host:
+        if host != current_host:
             hosts.append(host)
     return hosts[0], hosts[1]
 
 
-def check_volume_data(client, volume_name, data): # NOQA
+def check_detached_volume_data(client, volume_name, data): # NOQA
     """
     Attaches the volume to the current node
     then compares the volumes data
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_migration.py
```diff
@@ -5,6 +5,7 @@
 from common import get_random_client
 from common import SIZE
 from common import wait_for_volume_delete
+import subprocess
```
shuo-wu, line 8: NIT: Move this up

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +147,40 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+
+    volume.attach(hostId=host1)
+
+    volume = common.wait_for_volume_migration_ready(client, volume_name)
+
+    volume.detach(hostId=current_node)
+
+    volume = common.wait_for_volume_migration_node(client,
+                                                   volume_name,
+                                                   host1)
```
shuo-wu, line 171: Need to verify the old failed replica is removed.

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +147,40 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+
+    volume.updateReplicaCount(replicaCount=2)
+    volume.detach(hostId="")
+    volume = common.wait_for_volume_detached(client, volume_name)
+    volume = client.by_id_volume(volume_name)
```
shuo-wu, line 176: Why does the test need this? And how can it verify the data if the volume is detached?

shuo-wu (changes requested):

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +147,40 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+    check_volume_data(client, volume_name, data)
+
+    def finalizer():
+        exec_cmd = ["mkdir",  "/var/lib/longhorn/replicas"]
+        subprocess.check_output(exec_cmd)
+
+    request.addfinalizer(finalizer)
```
c3y1huang, line 183: I think the finalizer should move up to `exec_cmd = ["rm", "-rf",  "/var/lib/longhorn/replicas"]` in case the test fails midway.

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +147,40 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
     8. Validate initially written test data.
     """
-    return
+    client, volume, data = setup_migration_test(clients,
+                                                volume_name,
+                                                replica_cnt=3)
+
+    current_node = common.get_self_host_id()
+    host1, host2 = get_hosts_for_migration_test(clients)
+    volume.attach(hostId=current_node)
```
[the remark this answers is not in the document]
cchien816: I modified code, use `migrate_target` to replace `host1`

manager/integration/tests/test_migration.py
cchien816: Hi @shuo-wu, `check_volume_data` will attach the volume and doing data check

manager/integration/tests/test_migration.py
cchien816: Hi @c3y1huang , I modified code, thank you

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +147,40 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+    volume = common.wait_for_volume_healthy(client, volume_name)
+
+    exec_cmd = ["rm", "-rf",  "/var/lib/longhorn/replicas"]
+    subprocess.check_output(exec_cmd)
+    volume = common.wait_for_volume_degraded(client, volume_name)
+
+    volume.attach(hostId=host1)
+
+    volume = common.wait_for_volume_migration_ready(client, volume_name)
```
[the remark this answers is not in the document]
cchien816: Hi @shuo-wu ,
I add below code to make sure failed replica is on test node
```
for replica in volume.replicas:
        if replica.hostId == current_node:
            assert replica.running is False
        else:
            assert replica.running is True
```

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +146,47 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+    volume = common.wait_for_volume_degraded(client, volume_name)
+
+    volume.attach(hostId=migrate_target)
+
+    volume = common.wait_for_volume_migration_ready(client, volume_name)
+    for replica in volume.replicas:
+        if replica.hostId == current_node:
+            assert replica.running is False
+        else:
+            assert replica.running is True
```
shuo-wu, line 176: Make sure the current running replicas are newly created replicas.
You can verify this by recording the old replica list then checking if the new replica name is in the list

shuo-wu (commented):
Others LGTM

manager/integration/tests/test_migration.py
shuo-wu: This comment is missed

manager/integration/tests/test_migration.py
cchien816: Modified function name from `check_volume_data` to `check_detached_volume_data`

manager/integration/tests/test_migration.py
cchien816: All replicas will in running state except replica on current node. so use a count to check new running replica = 2

manager/integration/tests/test_migration.py
cchien816: Failed replica will keep exist since the replica count was 3 in my test environment, so I did not change code for this comment

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +146,55 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+                new_replicas = new_replicas + 1
+    assert new_replicas == 2
+
+    volume.detach(hostId=current_node)
+
+    volume = common.wait_for_volume_migration_node(client,
+                                                   volume_name,
+                                                   migrate_target)
+
+    volume.updateReplicaCount(replicaCount=2)
```
innobead, line 192: A little confusing, so after migration, the number of replica will not be back to 3? Why we need to specially update the replica count to 2?

shuo-wu: This verifies that the old/un-migrated replica is still removable.

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +146,55 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+    hosts = get_hosts_for_migration_test(clients)
+    migrate_target = hosts[0]
+    volume.attach(hostId=current_node)
+    volume = common.wait_for_volume_healthy(client, volume_name)
+
+    old_replicas = []
+    for replica in volume.replicas:
+        old_replicas.append(replica.name)
+
+    exec_cmd = ["rm", "-rf",  "/var/lib/longhorn/replicas"]
```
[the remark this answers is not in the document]
shuo-wu: Probably Yes. The replica directory name is already exposed to HTTP API

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +146,55 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
        And the old failed replica will be cleaned up.
     8. Validate initially written test data.
     """
-    return
+    def finalizer():
+        exec_cmd = ["mkdir", "/var/lib/longhorn/replicas"]
```
shuo-wu, line 150: Use `["mkdir", "-p", "/var/lib/longhorn/replicas"]` to avoid the directory-already-exist error

manager/integration/tests/test_migration.py
shuo-wu: IIRC, after the detachment, all failed replicas will be removed and new replicas will be replenished.

manager/integration/tests/test_migration.py
cchien816: Hi @shuo-wu .

I modified the code, thank you

manager/integration/tests/test_migration.py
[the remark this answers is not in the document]
cchien816: Hi @innobead ,
In my environment this command will keep` /var/lib/longhorn `exist, could you guide more detail for this comment? thank you.

manager/integration/tests/test_migration.py
```diff
@@ -147,7 +146,55 @@ def test_migration_with_failed_replica(clients, volume_name):  # NOQA
...
+
+    new_replicas = 0
+    for replica in volume.replicas:
+        if replica.hostId == current_node:
+            assert replica.running is False
+        else:
+            assert replica.running is True
+            if replica.name not in old_replicas:
+                new_replicas = new_replicas + 1
+    assert new_replicas == 2
```
innobead, line 184: The migration process only creates a migrated replica from each healthy replica only. 

https://github.com/longhorn/longhorn-manager/blob/73e3705d651681d38bda026146e302cad45d8baa/controller/volume_controller.go#L3189-L3220

innobead (approved):
LGTM

innobead:
@cchien816 Please backport to 1.3.x and 1.2.x. Thanks.

manager/integration/tests/test_migration.py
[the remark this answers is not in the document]
shuo-wu: We can crash the replica/make the volume degraded by removing the specific replica directory only. Removing the whole data directory `/var/lib/longhorn` may affect all replicas in this disk.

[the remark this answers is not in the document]
innobead: Exactly. @cchien816 Because right now our e2e testing does not run in parallel, so it should be fine, but suggest updating the case via deleting the specific replica folder in the future.

PR #1006, opened 2022-07-04
Refine benchmark test to let volume and replicas on the same node

yangchiu:
(1) Fix rke2 control plane schedulable
(2) Cordon redundant nodes when test replica count < cluster worker node count

For https://github.com/longhorn/longhorn/issues/4200

### DIFF
--- a/benchmark_test/scripts/benchmark-test-setup.sh
+++ b/benchmark_test/scripts/benchmark-test-setup.sh
@@ -3,6 +3,10 @@
 set -x
 set -e
 
+# create and clean tmpdir
+TMPDIR="/tmp/longhorn"
+mkdir -p ${TMPDIR}
+
 set_kubeconfig_envvar(){
   ARCH=${1}
   BASEDIR=${2}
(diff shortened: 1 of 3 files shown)


benchmark_test/scripts/benchmark-test-setup.sh
```diff
@@ -3,6 +3,11 @@
 set -x
 set -e
 
+# create and clean tmpdir
```
c3y1huang, line 6: Can you explain what this is for?

yangchiu: It's just a temp folder for git clone longhorn-manager

yangchiu: forgot to add this in the initial commit, so currently longhorn-manager would be cloned into root directory, which is not ideal.

c3y1huang: What is the purpose of removing the directory immediately after `mkdir`?

yangchiu: No need to do this. I've removed this line. Thank you!

derekbit (approved):
LGTM. It makes the result more reliable.

benchmark_test/scripts/benchmark-test-setup.sh
```diff
@@ -201,14 +205,16 @@ main(){
 
   run_fio_local_path_test
 
-  adjust_replica_count 1
-  run_fio_longhorn_test 1
+  adjust_replica_count 3
+  run_fio_longhorn_test 3
 
   adjust_replica_count 2
+  kubectl cordon "$(kubectl get nodes | awk 'NR!=1 && $3!~/control-plane/ {print $1}' | awk 'NR==3')"
```
innobead, line 212: This is based on the assumption we have 3 worker nodes.

innobead (approved):
LGTM

innobead:
Good finding 👍 !

benchmark_test/scripts/benchmark-test-setup.sh
yangchiu: Yes. It's temporarily OK since there's no parameter to change the worker count in the current Jenkins longhorn-benchmark-test pipeline, so the worker count is temporarily always 3.

PR #1032, opened 2022-07-26
change test_node_eviction_multiple_volume order

cchien816:

https://github.com/longhorn/longhorn/issues/4271
Change test_node_eviction_multiple_volume order to next to last, because this case will cause cascading failure

When issue resolved we can remove the pytest order mark


innobead:
@cchien816 Do you mean if this test runs at the second last, the result of node test suites will be better?

cchien816:
It may improve cascading error situation because undeleted volume created by this case will only generate at near end of regression test.

innobead:
Let's figure out the root cause or revisit here after other issues are fixed.

innobead:
Closing first. Revisit this change later if needed.

PR #1036, opened 2022-07-28
Add replica removal retry for test_disk_eviction_with_node_level_soft_anti_affinity_disabled

shuo-wu:
When deleting the replica at step 7 of the test, the new replica
rebuilding may be not done yet. Then the ReplicaRemove API would
be rejected until the new replica is healthy.

Here I prefer to use a retry rather than directly checking and
waiting for the new replica rebuilding. Since this retry operation
would help verify that the last replica won't disappear
accidentally.

Longhorn/longhorn#4238

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -36,6 +36,7 @@ from common import create_and_wait_pod
 from common import get_pod_data_md5sum
 from common import wait_for_volume_healthy
 from common import wait_for_volume_replica_count
+from common import is_replica_available
 from common import delete_and_wait_pod
 from common import wait_for_volume_detached
 from common import RETRY_COUNTS, RETRY_INTERVAL
@@ -2621,14 +2622,13 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(client, # NOQ
 
     # Step 2
     nodes = client.list_node()
-    vol_name = common.generate_volume_name()
-    volume = client.create_volume(name=vol_name,
+    volume = client.create_volume(name=volume_name,
                                   size=SIZE, numberOfReplicas=len(nodes))
-    common.wait_for_volume_detached(client, vol_name)
+    common.wait_for_volume_detached(client, volume_name)
 
     lht_hostId = get_self_host_id()
     volume.attach(hostId=lht_hostId)
-    volume = common.wait_for_volume_healthy(client, vol_name)
+    volume = common.wait_for_volume_healthy(client, volume_name)
 
     # Step 3
     data = common.write_volume_random_data(volume)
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_node.py
```diff
@@ -2663,16 +2663,21 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(client, # NOQ
             break
         time.sleep(common.RETRY_EXEC_INTERVAL)
     assert len(os.listdir(replica_path)) > 0
+    volume = common.wait_for_volume_healthy(client, vol_name)
```
c3y1huang, line 2666:
```suggestion
    volume = wait_for_volume_healthy(client, vol_name)
```

manager/integration/tests/test_node.py
```diff
@@ -2663,16 +2663,21 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(client, # NOQ
...
-        if r.hostId != lht_hostId:
-            volume.replicaRemove(name=r.name)
+    for i in range(RETRY_COUNTS):
+        for r in volume.replicas:
+            if r.hostId != lht_hostId:
+                try:
+                    volume.replicaRemove(name=r.name)
+                except Exception:
+                    pass
+                time.sleep(RETRY_INTERVAL)
```
c3y1huang, line 2680: Should check if the volume replicas are deleted to break out of the retry loop.

c3y1huang (changes requested):

manager/integration/tests/test_node.py
shuo-wu: Yes. Thank you for pointing this out! Addressed

manager/integration/tests/test_node.py
```diff
@@ -2663,16 +2662,25 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(client, # NOQ
...
+        deleted = True
+        volume = client.by_id_volume(volume_name)
+        for r in volume.replicas:
+            if r.hostId != lht_hostId:
+                try:
+                    volume.replicaRemove(name=r.name)
+                except Exception:
+                    deleted = False
+        if deleted:
+            break
```
c3y1huang, line 2682: Should also assert the volume has only 1 replica after the loop to ensure replicas got removed.

c3y1huang (changes requested):

manager/integration/tests/test_node.py
shuo-wu: Good idea! Yes the test can verify the only replica is the newly rebuilt replica then the following data r/w would validate the availability of the new replica.

c3y1huang (dismissed):
LGTM

c3y1huang (approved):

innobead (approved):

innobead:
@shuo-wu Please backport to 1.3.x and 1.2.x.

PR #1037, opened 2022-07-29
Fix test_engine_crash_for_restore_volum

cchien816:

From https://github.com/longhorn/longhorn/issues/4309
Root cause : https://github.com/longhorn/longhorn/issues/4309#issuecomment-1197897496

Remove code for catch volume detach after `crash_engine_process_with_sigkill`
Without that line, test script still checked `RestoreInProgress` after engine crash and check restored volume's md5, which matched the test goal

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4427,7 +4427,7 @@ def wait_for_instance_manager_desire_state(client, core_api, im_name,
                 continue
             # Report any other error
             else:
-                assert(not e)
+                assert (not e)
         if desire:
             if im.currentState == state.lower() and pod.status.phase == state:
                 break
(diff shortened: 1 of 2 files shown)


innobead (commented):
LGTM, but can we comment out instead of removing, and add some comment around to know why we skiped this. 

This can be a shorterm solution, but we need to think about how to test the state transition of a volume in e2e.

innobead:
@cchien816 this is actually the same as @shuo-wu PR. Let's close this and follow up there.

shuo-wu (approved):

innobead:
@cchien816 Please fix the PR check error about flake8 check.

innobead (approved):

innobead:
@cchien816 Need to backport to 1.3 and 1.2.

PR #1042, opened 2022-08-02
Fix RockyLinux 8.5 arm64 has no AMI results

yangchiu:
Fix RockyLinux 8.5 arm64 has no AMI results

For https://github.com/longhorn/longhorn/issues/4341


innobead (approved):

innobead:
@yangchiu This needs to backport to 1.3 and 1.2.

PR #1046, opened 2022-08-02
Fix no enough disk space on air gap registry server

yangchiu:
Fix no enough disk space on air gap registry server

For https://github.com/longhorn/longhorn/issues/4350

### DIFF
--- a/airgap/terraform/main.tf
+++ b/airgap/terraform/main.tf
@@ -231,6 +231,11 @@ resource "aws_instance" "lh_registry_aws_instance" {
     aws_security_group.lh_registry_aws_secgrp.id
   ]
 
+  root_block_device {
+    delete_on_termination = true
+    volume_size = 40
+  }
+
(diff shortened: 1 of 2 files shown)


airgap/terraform/main.tf
```diff
@@ -231,6 +231,11 @@ resource "aws_instance" "lh_registry_aws_instance" {
     aws_security_group.lh_registry_aws_secgrp.id
   ]
 
+  root_block_device {
+    delete_on_termination = true
+    volume_size = 40
```
innobead, line 236: curious: what's the default value if not specified?

innobead (approved):

innobead:
@yangchiu Need to backport to 1.3.x and 1.2.x.

airgap/terraform/main.tf
yangchiu: default value is 8 GiB

PR #1013, opened 2022-07-11
e2e regression test using HDD

yangchiu:
e2e regression test using HDD

For https://github.com/longhorn/longhorn/issues/4182

Signed-off-by: Yang Chiu <yang.chiu@suse.com>

PR #1050, opened 2022-08-03
Fix bug: make sure to cleanup all deployment created by the tests and Allow customization for timeouts when running tests on hdd disk

PhanLe1010:
longhorn/longhorn#4227

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -79,6 +79,8 @@ spec:
       value: "s3://backupbucket@us-east-1/backupstore$minio-secret, nfs://longhorn-test-nfs-svc.default:/opt/backupstore"
     - name: LONGHORN_BACKUPSTORE_POLL_INTERVAL
       value: "30"
+    - name: LONGHORN_DISK_TYPE
+      value: "ssd"
     - name: NODE_NAME
       valueFrom:
         fieldRef:
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -233,6 +233,13 @@ NODE_UPDATE_RETRY_INTERVAL = 6
 NODE_UPDATE_RETRY_COUNT = 30
 disk_being_syncing = "being syncing and please retry later"
 
+# customize the timeout for HDD
+disktype = os.environ['LONGHORN_DISK_TYPE']
+if disktype == "hdd":
+    RETRY_COUNTS *= 32
+    DEFAULT_DEPLOYMENT_TIMEOUT *= 32
+    STREAM_EXEC_TIMEOUT *= 32
+
 
 def load_k8s_config():
     c = Configuration()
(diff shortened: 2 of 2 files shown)


manager/integration/deploy/test.yaml
```diff
@@ -79,6 +79,8 @@ spec:
       value: "s3://backupbucket@us-east-1/backupstore$minio-secret, nfs://longhorn-test-nfs-svc.default:/opt/backupstore"
     - name: LONGHORN_BACKUPSTORE_POLL_INTERVAL
       value: "30"
+    - name: LONGHORN_DISK_TYPE
+      value: "ssd"
```
PhanLe1010, line 83: How do think about this approach @yangchiu . If yes, can we modify the Jenkin pipeline to specify this env variable?

innobead (approved):
LGTM, but @yangchiu has correct knowledge about HDD testing. Leave it to him for the final review.

manager/integration/deploy/test.yaml
yangchiu: LGTM.
And we've already had USE_HDD variable in the jenkins pipeline. We use it to specify this LONGHORN_DISK_TYPE value.

manager/integration/tests/common.py
```diff
@@ -233,6 +233,12 @@
 NODE_UPDATE_RETRY_COUNT = 30
 disk_being_syncing = "being syncing and please retry later"
 
+# customize the timeout for HDD
+disktype = os.environ['LONGHORN_DISK_TYPE']
+if disktype == "hdd":
+    RETRY_COUNTS = 300
+    DEFAULT_DEPLOYMENT_TIMEOUT = 300
```
yangchiu, line 240: From the previous experiments, the timeout should be 32x longer to get rid of most of (not all of) the timeout test cases.
And there is another timeout value STREAM_EXEC_TIMEOUT should be adjusted too.

manager/integration/deploy/test.yaml
PhanLe1010: Thanks

manager/integration/tests/common.py
PhanLe1010: Thank you. Fixed

manager/integration/tests/common.py
```diff
@@ -4027,7 +4034,9 @@ def wait_for_volume_restoration_start(client, volume_name, backup_name,
 @pytest.fixture
 def make_deployment_with_pvc(request):
     def _generate_deployment_with_pvc_manifest(deployment_name, pvc_name, replicas=1): # NOQA
-        make_deployment_with_pvc.deployment_manifest = {
+        if not hasattr(make_deployment_with_pvc, 'deployment_manifests'):
+            make_deployment_with_pvc.deployment_manifests = []
+        make_deployment_with_pvc.deployment_manifests.append({
```
shuo-wu, line 4039: Why do you use a list here?

PhanLe1010: It is in the finalizer so it get clean up: https://github.com/longhorn/longhorn-tests/blob/aa28f5d19238e0a5c483c7a5cbb9673c520ec651/manager/integration/tests/common.py#L4089-L4097

shuo-wu: Should this function just clean up the deployment it generates?

PhanLe1010: > Should this function just clean up the deployment it generates?

Yeah, it should. With the fix, it will clean up the deployments that it created. Without the fix, it will only clean up the last deployment that it created

shuo-wu (approved):

yangchiu (approved):

innobead (approved):
LGTM

manager/integration/tests/common.py
innobead: This is a good catch!

innobead:
@PhanLe1010 Please backport this to 1.3.

@yangchiu We need to backport this https://github.com/longhorn/longhorn-tests/pull/1013 as well.

PhanLe1010:
backported: https://github.com/longhorn/longhorn-tests/pull/1069

PR #1058, opened 2022-08-05
Fix flaky case test_cloning_interrupted

cchien816:

In step7 the volume `faulted` state can be observed but not by script because the state transition were too fast 
Modified `wait_for_volume_status`, add option to disable check `wait_for_volume_creation` let script can quick check volume status

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1685,9 +1685,11 @@ def wait_for_volume_degraded(client, name):
 
 
 def wait_for_volume_faulted(client, name):
-    wait_for_volume_status(client, name,
-                           VOLUME_FIELD_STATE,
-                           VOLUME_STATE_DETACHED)
+    # Comment out detach status check because status transition
+    # were too fast recently
+    # wait_for_volume_status(client, name,
+    #                       VOLUME_FIELD_STATE,
+    #                       VOLUME_STATE_DETACHED)
     return wait_for_volume_status(client, name,
                                   VOLUME_FIELD_ROBUSTNESS,
                                   VOLUME_ROBUSTNESS_FAULTED)


manager/integration/tests/test_cloning.py
```diff
@@ -392,7 +393,10 @@ def test_cloning_interrupted(client, core_api, pvc, pod, clone_pvc, clone_pod):
     crash_replica_processes(client, core_api, source_volume_name)
 
     # Step-7
-    wait_for_volume_faulted(client, source_volume_name)
+    wait_for_volume_status(client, source_volume_name,
```
innobead, line 396: Could you check other cases also flaky as well when using `crash_replica_processes` + `wait_for_volume_faulted` together?

I think there are other places using these as well.

manager/integration/tests/common.py
```diff
@@ -1694,8 +1694,10 @@ def wait_for_volume_faulted(client, name):
 
 
 def wait_for_volume_status(client, name, key, value,
-                           retry_count=RETRY_COUNTS):
-    wait_for_volume_creation(client, name)
+                           retry_count=RETRY_COUNTS,
+                           check_volume_creation=True):
+    if check_volume_creation is True:
+        wait_for_volume_creation(client, name)
```
innobead, line 1700: Why we need to do this? `wait_for_volume_creation` just checked if the volume exists or not, so why we have an option to ignore this check?

manager/integration/tests/test_cloning.py
cchien816: Yes in `test_ha.py`, test case `test_ha_salvage` was [flaky](https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/209/testReport/tests/test_ha/test_ha_salvage/) due to waiting volume faulted status.
I will submit another PR for the flaky case

manager/integration/tests/common.py
cchien816: Origin idea was thinking to minimize process time before catching volume faulted status, after did some test on pipeline for run test_cloning_interrupted 10 times below

- Do not have [wait_for_volume_creation](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1816/) : 8 Pass, 2 Fail 
- Have [wait_for_volume_creation](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1817/)            : 7 Pass, 3 Fail

Test result were almost the same no matter this part of code changed or not, I modified PR keep this part no changed, thank you

innobead: I see. This modification is not really meaningful. We should just keep this instead of making a special case for this kind of tunning.

manager/integration/tests/test_cloning.py
```diff
@@ -392,7 +393,9 @@ def test_cloning_interrupted(client, core_api, pvc, pod, clone_pvc, clone_pod):
...
```
innobead, line 396: I feel we should update `wait_for_volume_faulted` instead, but only focus on checking the robust become faulted. 

This should impact all tests using wait_for_volume_faulted, right?

cchien816: Yes will impact test case in `test_ha.py` and `test_cloning.py`

Below are related cases
- test_ha_salvage
- test_all_replica_restore_failure
- test_salvage_auto_crash_all_replicas
- test_volume_reattach_after_engine_sigkill
- test_cloning_interrupted

innobead: ok, so I suggest updating `wait_for_volume_faulted` and commenting out the detached check w/ a reason.

innobead (approved):

innobead:
@cchien816 Please backport this to 1.3. and 1.2. Thanks.

PR #1065, opened 2022-08-09
Modify manual case Longhorn with engine is not deployed on all the nodes

cchien816:

According original issue https://github.com/longhorn/longhorn/issues/2081#issuecomment-783747541 to modify case 2 step 9

### DIFF
--- a/docs/content/manual/pre-release/ha/partial-engine-deployment.md
+++ b/docs/content/manual/pre-release/ha/partial-engine-deployment.md
@@ -22,7 +22,7 @@ https://github.com/longhorn/longhorn/issues/2081
 1. Detach `vol-1`
 1. Verify that Longhorn cannot attach `vol-1` to `node-1` since there is no engine image on `node-1`
 1. Check that all functions (attach to other nodes, detach, snapshot, backup, expand, restore, creating DR volume, ... ) are working ok for `vol-1`
-1. Verify that `vol-2` can be attached to any nodes except for `node-1`
+1. Verify that `vol-2` cannot be attached to any nodes because one of its replica is sitting on the `node-1` which doesn't have the engine image
 
 ### Case 3: Test engine upgrade when some of the engine image DaemonSet pods are not fully deployed
 1. Continue from case 2


docs/content/manual/pre-release/ha/partial-engine-deployment.md
```diff
@@ -22,7 +22,7 @@ https://github.com/longhorn/longhorn/issues/2081
 1. Detach `vol-1`
 1. Verify that Longhorn cannot attach `vol-1` to `node-1` since there is no engine image on `node-1`
 1. Check that all functions (attach to other nodes, detach, snapshot, backup, expand, restore, creating DR volume, ... ) are working ok for `vol-1`
-1. Verify that `vol-2` can be attached to any nodes except for `node-1`
+1. Verify that `vol-2` cannot be attached to any nodes because one of its replica is sitting on the `node-1` which doesn't have the engine image
```
innobead, line 25: I feel this is not right and like an issue instead. 

I would assume the replica on node-1 will go to error, and the other two still running. So, the volume should be able to attach to schedulable nodes w/ degraded status.

shuo-wu, line 25: IIRC, after the fix of longhorn/longhorn#4197, the volume can be attached even if there are failed replicas (caused by lacking the engine image for example) during attachment. 
Will the engine image get stuck in state `deploying` and block the attachment?

cchien816: Yes you are right, replica on tainted become in failed status, status of remained 2 replicas on other nodes were in stopped status, I logged https://github.com/longhorn/longhorn/issues/4386 for tracking

cchien816: After command `kubectl get pods -n longhorn-system`, from the response it can not find stuck pods ( all in running status)

innobead: I think @shuo-wu was asking about the engine image CR in deploying state. I think it should be in deploying state because there is no engine image pod on the trained node, but the longhorn manager pod is still there? @cchien816

innobead: Why the other running replicas will become stopped? cc @shuo-wu @c3y1huang @PhanLe1010

cchien816: Yes by `kubectl get engineimages.longhorn.io` can see the engine image CR is in deploying status
```
NAME          STATE       IMAGE                                   REFCOUNT   BUILDDATE   AGE
ei-1b463ad1   deploying   longhornio/longhorn-engine:v1.3.1-rc2   5          24h         41m
```

shuo-wu: Then will this `deployment` engine image block the volume attachment?

innobead:
Ok, this is the context https://github.com/longhorn/longhorn/issues/2935#issue-979741053 

When attaching to the tainted node which does not has the engine image, it will be detached and faulted. <-- this is expected
Did you attach to other non-tainted nodes having engine images? This should be expected to work.

cchien816:
> Ok, this is the context [longhorn/longhorn#2935 (comment)](https://github.com/longhorn/longhorn/issues/2935#issue-979741053)
> 
> @cchien816
> 
> * When attaching the volume to the tainted node which does not has the engine image, it will be detached and faulted. <-- this should be expected
> * Did you attach the volume to other non-tainted nodes having engine images? This should be expected to work.

Yes I attached volume to non-tainted nodes, and test result were the same, volume become in faulted status

innobead:
It's good to clarify and actually found a real issue 👍

PR #1071, opened 2022-08-15
Fix Ubuntu 22.04 SSH authentication failed

yangchiu:
Fix Ubuntu 22.04 SSH authentication failed

For https://github.com/longhorn/longhorn/issues/4320


innobead:
Probably need to update to other distros as well, because this will possibly happen to them in the future.

innobead (approved):

yangchiu:
> Probably need to update to other distros as well, because this will possibly happen to them in the future.

The correct solution should be updating terraform version to let it support newer ssh key algorithms, then we don't have to modify distros to make them support old ssh algorithms. I've created PR https://github.com/longhorn/longhorn-tests/pull/1072 for this.

PR #1072, opened 2022-08-15
Update terraform version to support newer ssh key algorithms.

yangchiu:
Update terraform version to support newer ssh key algorithms.

For https://github.com/longhorn/longhorn/issues/4320


innobead (approved):

innobead:
LGTM, but the one thing I am curious about is that the weak hashing function used in the SSH connection is rejected by the connected host, but why changing the terraform version can actually fix the problem? IIRC, the SSH keys are generated by us instead of any terraform plugins.

yangchiu:
> LGTM, but the one thing I am curious about is that the weak hashing function used in the SSH connection is rejected by the connected host, but why changing the terraform version can actually fix the problem? IIRC, the SSH keys are generated by us instead of any terraform plugins.

It's because the rejection is due to the host has disabled the weak algorithm and the client (terraform) still wants to use this weak algorithm to connect the host (So this error would only happen in terraform. It works fine when we try to ssh into the host through our laptops), and terraform already fix this issue in [v1.2.0](https://github.com/hashicorp/terraform/releases/tag/v1.2.0) (The last enhancement https://github.com/hashicorp/terraform/pull/30962). After this version, terraform wouldn't try to use weak algorithm so it won't be rejected.

PR #1074, opened 2022-08-17
Add test case test_default_storage_class_syncup

chrisc1981:

Add test case test_default_storage_class_syncup

https://github.com/longhorn/longhorn/issues/4417

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import time
 import random
+import yaml
 
 import common
 from common import client, random_labels, volume_name  # NOQA
@@ -4018,9 +4019,8 @@ def test_restore_basic(set_random_backupstore, client, core_api, volume_name, po
     assert output == ''
 
 
-@pytest.mark.skip(reason="TODO") # NOQA
 @pytest.mark.coretest   # NOQA
-def test_default_storage_class_syncup():  # NOQA
+def test_default_storage_class_syncup(core_api, request):  # NOQA
     """
     Steps:
     1. Record the current Longhorn-StorageClass-related ConfigMap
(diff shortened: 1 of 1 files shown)


khushboo-rancher (approved):

manager/integration/tests/test_basic.py
```diff
@@ -4038,6 +4038,82 @@ def test_default_storage_class_syncup():  # NOQA
...
+    longhorn_storage_class = storage_api.read_storage_class("longhorn")
+    longhorn_storage_class.reclaim_policy = "Retain"
+
+    # step 3
+    try:
+        not_allow_modify_reclaim_policy = False
+        storage_api.patch_storage_class("longhorn", longhorn_storage_class)
+    except Exception as e:
+        print(e)
+        not_allow_modify_reclaim_policy = True
```
innobead, line 4090: This part would not take effect, right? 

Because the patch operation will still work (there is no reason the API server will reject this, and we even don't have validation admission for storageclass resource), however, the storage class will be just reverted later to the content of the `longhorn-storageclass` configmap by longhorn controller.

see: https://github.com/longhorn/longhorn-manager/blob/c3e9b1c8c7d1043134bfbf2c8ceb25c789935bc0/controller/kubernetes_configmap_controller.go#L163-L201

innobead (changes requested):
In general, LGTM except the step 3.

manager/integration/tests/test_basic.py
chrisc1981: Hi @innobead ,
Yes this part is not take effect.

When I use API  `storage_api.patch_storage_class("longhorn", longhorn_storage_class)` I will get 
```
kubernetes.client.rest.ApiException: (422)
E           Reason: Unprocessable Entity
E           HTTP response headers: HTTPHeaderDict({'Audit-Id': 'd0cb4f21-3e7c-4f90-b677-04816c7423ef', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': '9f5a14a6-7287-4b6e-8d74-e90cd4f009f5', 'X-Kubernetes-Pf-Prioritylevel-Uid': '15a86490-9398-463a-85cc-db72981ccb11', 'Date': 'Thu, 18 Aug 2022 09:15:14 GMT', 'Content-Length': '438'})
E           HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"StorageClass.storage.k8s.io \"longhorn\" is invalid: reclaimPolicy: Forbidden: updates to reclaimPolicy are forbidden.","reason":"Invalid","details":{"name":"longhorn","group":"storage.k8s.io","kind":"StorageClass","causes":[{"reason":"FieldValueForbidden","message":"Forbidden: updates to reclaimPolicy are forbidden.","field":"reclaimPolicy"}]},"code":422}
```
If I manually run `kubectl edit storageclass longhorn` and change  `reclaimPolicy: Delete` to
`reclaimPolicy: Retain`, console will reply 
```
longhorn-test:/integration/tests # kubectl edit storageclass longhorn
error: storageclasses.storage.k8s.io "longhorn" is invalid
A copy of your changes has been stored to "/tmp/kubectl-edit-gz60e.yaml"
error: Edit cancelled, no valid changes were saved.
```
So I think except in code is OK, I will keep try if I am in wrong usage of k8s API. thank you

chrisc1981: Hi @innobead ,
After discuss with developers, I figured out that step 3 means to edit reclaim policy in storage class's annotation `longhorn.io/last-applied-configmap`, not directly edit value of key `reclaimPolicy`
I updated code for review, thank you.

innobead (approved):
LGTM

innobead:
@chrisc1981 Please help backport to 1.3.x. Thanks.

PR #1082, opened 2022-08-25
ci: fix manifest air gap installation pipeline failed in v1.2.x

yangchiu:
ci: fix manifest air gap installation pipeline failed in v1.2.x by avoiding duplicated document separator in yaml file causing yq parsing error.

For https://github.com/longhorn/longhorn/issues/4479

### DIFF
--- a/test_framework/scripts/longhorn-setup.sh
+++ b/test_framework/scripts/longhorn-setup.sh
@@ -114,6 +114,7 @@ wait_longhorn_status_running(){
 
 get_longhorn_manifest(){
   wget ${LONGHORN_MANIFEST_URL} -P ${TF_VAR_tf_workspace}
+  sed -i ':a;N;$!ba;s/---\n---/---/g' "${TF_VAR_tf_workspace}/longhorn.yaml"
 }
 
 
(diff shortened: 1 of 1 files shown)


test_framework/scripts/longhorn-setup.sh
```diff
@@ -114,6 +114,7 @@ wait_longhorn_status_running(){
 
 get_longhorn_manifest(){
   wget ${LONGHORN_MANIFEST_URL} -P ${TF_VAR_tf_workspace}
+  sed -i ':a;N;$!ba;s/---\n---/---/g' "${TF_VAR_tf_workspace}/longhorn.yaml"
```
innobead, line 117: Why we have some chars before the replacement command? Is this intended?

innobead (commented):
LGTM except one comment.

test_framework/scripts/longhorn-setup.sh
yangchiu: sed does its job line by line so if we'd like to make it do this kind of cross-line replacement, we need this trick to let it read all lines of this document and then start the replacement.

ref: https://stackoverflow.com/a/1252191

innobead (approved):

test_framework/scripts/longhorn-setup.sh
innobead: I see. Thanks.

PR #1075, opened 2022-08-18
Fix flaky test_ha_salvage

chrisc1981:
Signed-off-by: Chris Chien [chris.chien@suse.com](mailto:chris.chien@suse.com)

Fix flaky test_ha_salvage https://github.com/longhorn/longhorn/issues/4381#issuecomment-1212704361

In case the volume faulted status was not easily to catch by script. modify script to detect volume `attaching` state is more stable
 Test PR code 20 times on pipeline were [passed](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1873/)

PR #1083, opened 2022-08-26
Modify test_cloning_interrupted

chrisc1981:

https://github.com/longhorn/longhorn/issues/4487

Somehow `test_cloning_interrupted` still [flaky](https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/227/testReport/junit/tests/test_cloning/test_cloning_interrupted/) because step 7 `wait_for_volume_faulted` not every time can catch by script0
I tried to use workaround the of test case [test_ha_salvage](https://github.com/longhorn/longhorn-tests/pull/1075) and run this test case on pipleline 20 times. the result were all [passed](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1936/)  for reference.

### DIFF
--- a/manager/integration/tests/test_cloning.py
+++ b/manager/integration/tests/test_cloning.py
@@ -9,13 +9,14 @@ from common import wait_for_volume_clone_status, VOLUME_FIELD_STATE
 from common import VOLUME_FIELD_CLONE_COMPLETED, wait_for_pvc_phase
 from common import get_self_host_id, write_volume_random_data
 from common import wait_for_volume_detached, check_volume_data
-from common import crash_replica_processes, wait_for_volume_faulted
+from common import crash_replica_processes
 from common import delete_and_wait_pvc, wait_for_volume_attached
 from common import generate_random_suffix, wait_for_volume_endpoint
 from common import wait_for_snapshot_count, DATA_SIZE_IN_MB_3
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_cloning.py
```diff
@@ -392,7 +393,11 @@ def test_cloning_interrupted(client, core_api, pvc, pod, clone_pvc, clone_pod):
     crash_replica_processes(client, core_api, source_volume_name)
 
     # Step-7
-    wait_for_volume_faulted(client, source_volume_name)
+    # This is a workaround, since in some case it's hard to
+    # catch faulted volume status
+    wait_for_volume_status(client, source_volume_name,
```
khushboo-rancher, line 398: Does the volume starts to attach immediately after becoming faulted?

Also a qq, why are we waiting for volume to become degraded before we are crashing all the replicas in step 6?

chrisc1981: Hi @khushboo-rancher ,

Yes, in master-head volume will start to attach after faulted.
In step6, I think it detect new created clone volume, clone volume's replica are not all ready so it will be in degraded status

innobead (approved):

PR #1084, opened 2022-08-26
ci: add code flow to create new secret in air gap installation

yangchiu:
ci: add code flow to create new secret in air gap installation

For https://github.com/longhorn/longhorn/issues/4480


test_framework/Jenkinsfile
```diff
@@ -12,6 +12,7 @@ def AIR_GAP_INSTALLATION = params.AIR_GAP_INSTALLATION ? params.AIR_GAP_INSTALLA
 def LONGHORN_INSTALL_VERSION = params.LONGHORN_INSTALL_VERSION ? params.LONGHORN_INSTALL_VERSION : "master"
 def LONGHORN_INSTALL_METHOD = params.LONGHORN_INSTALL_METHOD ? params.LONGHORN_INSTALL_METHOD : "custom"
 def CREATE_LOAD_BALANCER = params.LONGHORN_INSTALL_METHOD == "rancher" ? true : false
+def CREATE_PRIVATE_REGISTRY_SECRET = params.CREATE_PRIVATE_REGISTRY_SECRET ? params.CREATE_PRIVATE_REGISTRY_SECRET : ""
```
innobead, line 15: groovy allows you to use `?:` ternary operator to simplify the code.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -393,11 +396,16 @@ main(){
       customize_longhorn_manifest_for_airgap
       install_longhorn_by_manifest "${TF_VAR_tf_workspace}/longhorn.yaml"
     elif [[ "${LONGHORN_INSTALL_METHOD}" == "helm-chart" ]]; then
-      create_registry_secret
+      if [[ "${CREATE_PRIVATE_REGISTRY_SECRET}" == false ]]; then
+        create_registry_secret
+      fi
```
innobead, line 401: Do we still need this if clause? I think the secret will be handled in `customize_longhorn_chart_for_airgap` already?

test_framework/scripts/longhorn-setup.sh
```diff
@@ -393,11 +396,16 @@ main(){
...
       get_longhorn_chart
       customize_longhorn_chart_for_airgap
       install_longhorn_by_chart
     elif [[ "${LONGHORN_INSTALL_METHOD}" == "rancher" ]]; then
+      if [[ "${CREATE_PRIVATE_REGISTRY_SECRET}" == false ]]; then
```
innobead, line 406: Ok, this just leaves to the chart installation which will use the default value `true` to create the secret instead. So when running for 1.2, we should select `false`. I feel this is quite complicated to understand. Can we just make it simple to implicitly make it take effect.

Can we just leave the secret only created during the chart installation instead of creating before that? so we will not need to differentiate different cases here? WDYT?

innobead (changes requested):

PR #1085, opened 2022-08-26
Fix test_rebuild_with_restoration on master head

chrisc1981:

Increase volume/data size make volume restoration time longer
20 times test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1940/)

https://github.com/longhorn/longhorn/issues/4488


innobead (approved):

innobead:
Please backport this to 1.2.x to ensure the test case consistent. 

1.1.x will not be maintained, so no need to backport there.

PR #1091, opened 2022-09-02
[Backport v1.2.x] Modify test_rebuild_with_restoration

chriscchien:

Backport https://github.com/longhorn/longhorn-tests/pull/1085 to v1.2.x for fix flaky test_rebuild_with_restoration


innobead (approved):
LGTM

innobead:
Probably we should a larger volume size as the default for all test cases as well to avoid this kind of timing issue happening again. cc @longhorn/qa

PR #1095, opened 2022-09-06
Add test case test_migration_with_unscheduled_replica

chriscchien:

Add test case `test_migration_with_unscheduled_replica`

### DIFF
--- a/manager/integration/tests/test_migration.py
+++ b/manager/integration/tests/test_migration.py
@@ -5,6 +5,7 @@ from common import clients, volume_name  # NOQA
 from common import get_random_client
 from common import SIZE
 from common import wait_for_volume_delete
+from common import set_node_scheduling
 
 REPLICA_COUNT = 2
 
@@ -100,7 +101,6 @@ def migration_rollback_test(clients, volume_name, backing_image=""):  # NOQA
 
 @pytest.mark.coretest  # NOQA
 @pytest.mark.migration # NOQA
-@pytest.mark.skip(reason="TODO") # NOQA
 def test_migration_with_unscheduled_replica(clients, volume_name):  # NOQA
     """
     Test that a degraded migratable RWX volume that contain an unscheduled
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_migration.py
```diff
@@ -123,7 +123,64 @@ def test_migration_with_unscheduled_replica(clients, volume_name):  # NOQA
     11. Enable the scheduling for the node and wait for rebuilding complete.
     12. Validate initially written test data.
     """
-    return
+    # Step 1
+    client = get_random_client(clients)
+
+    # test_nodes[0] : node which running script
+    # test_nodes[1] : migrate target node
+    # test_nodes[2] : node to schedule on/off
```
khushboo-rancher, line 131: According to your test steps 6-9, the volume is being migrated from node[1] to node[0]. Can we update the description here, target becomes the node[0] and destination becomes node[1].

manager/integration/tests/test_migration.py
```diff
@@ -123,7 +123,64 @@ def test_migration_with_unscheduled_replica(clients, volume_name):  # NOQA
...
+    test_nodes = list()
+    nodes = client.list_node()
+    for node in nodes:
+        if node.name == common.get_self_host_id():
+            test_nodes.append(node)
+    for node in nodes:
+        if node.name != common.get_self_host_id():
+            test_nodes.append(node)
```
khushboo-rancher, line 139: Are these two iterations on nodes for getting the Longhorn node first, and then other nodes in the list?

khushboo-rancher (commented):
LGTM in general, some minor comments.

manager/integration/tests/test_migration.py
chriscchien: Yes, currently I can not find a more efficient way to do that so I just use 2 loop to come up then I can clearly know every node's role in this test case

manager/integration/tests/test_migration.py
chriscchien: Hello @khushboo-rancher, I updated the description, thank you

khushboo-rancher (approved):

manager/integration/tests/test_migration.py
innobead: I think you can make the test simpler. WDYT? 

- Use nodes from `get_hosts_for_migration_test` to do the migration testings
- Enable/disable scheduling of any of them
- Just use the local node for the first attachment and write the data

You don't need to prepare a list to store 3 nodes' info, but just reuse an existing function to get the migration nodes to start the testing as above.

innobead (changes requested):

manager/integration/tests/test_migration.py
chriscchien: Yes this is more simpler, I updated code by using `get_hosts_for_migration_test` to get nodes  for review

manager/integration/tests/test_migration.py
```diff
@@ -123,7 +123,59 @@ def test_migration_with_unscheduled_replica(clients, volume_name):  # NOQA
...
+
+    # Step 6
+    volume.attach(hostId=hosts[0])
+    volume = common.wait_for_volume_degraded(client, volume_name)
+
+    # Step 7
+    volume.attach(hostId=local_node)
+
+    # Step 8
+    volume = common.wait_for_volume_migration_ready(client, volume_name)
```
innobead, line 162: You should verify the volume is still degraded and the number of new migrated replicas should the same as the original replicas.

innobead (changes requested):

manager/integration/tests/test_migration.py
chriscchien: HI @innobead , I update code for checking replica count after migrate, thank you

innobead (approved):
LGTM

innobead:
@chriscchien Please backport to 1.3 and 1.2. Thanks.

PR #1096, opened 2022-09-07
ci: prevent corrupt regression test from stuck in download support bundle forever

yangchiu:
ci: prevent corrupt regression test from stuck in download support bundle forever

For https://github.com/longhorn/longhorn/issues/4072

### DIFF
--- a/test_framework/scripts/download-support-bundle.sh
+++ b/test_framework/scripts/download-support-bundle.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -eo pipefail
+
 SUPPORT_BUNDLE_FILE_NAME=${1:-"lh-support-bundle.zip"}
 SUPPORT_BUNDLE_ISSUE_URL=${2:-""}
 SUPPORT_BUNDLE_ISSUE_DESC=${3:-"Auto-generated support buundle"}


test_framework/scripts/download-support-bundle.sh
```diff
@@ -35,7 +37,8 @@ SUPPORT_BUNDLE_URL=`kubectl exec -n longhorn-system svc/longhorn-frontend -- bas
 
 SUPPORT_BUNDLE_READY=false
 while [[ ${SUPPORT_BUNDLE_READY} == false ]]; do
-    PERCENT=`kubectl exec -n longhorn-system svc/longhorn-frontend -- curl -H 'Accept: application/json' ${SUPPORT_BUNDLE_URL} | jq -r '.progressPercentage'`
+    RES=`kubectl exec -n longhorn-system svc/longhorn-frontend -- curl -H 'Accept: application/json' ${SUPPORT_BUNDLE_URL}`
+    PERCENT=`echo "${RES}" | jq -r '.progressPercentage'`
```
innobead, line 41: @yangchiu could you explain a bit why the previous pipe will probably not work?

yangchiu: `set -e` only works when there is an error in the last command of a pipe.
In our case, the corrupt environment can't even get the longhorn-frontend url, so the error would happen in the first command of the pipe (`kubectl exec`) but `set -e` would not notice this error because of the pipe.
Separating commands instead of using pipe can fix this, but I find `set -o pipefail` is more graceful. Modified this PR to use `set -o pipefail`

test_framework/scripts/download-support-bundle.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -eo pipefail
```
innobead, line 3:
```suggestion
set -o pipefail
```

innobead (changes requested):

test_framework/scripts/download-support-bundle.sh
innobead: I see, but my question is why this change will fix the issue of downloading bundle forever? Previously, the script did not have `set -e`.

yangchiu: Yes, previously the script didn't have `set -e ` so if there is an error it would stuck in while loop forever.
Add `set -e` won't fix the problem because the error happened in the pipe.
Add `set -eo pipefail` fix the problem because `set -o pipefail` would let the whole pipe fail if one of command failed in the pipe.

innobead: Ah u are right!

innobead (approved):

PR #1099, opened 2022-09-08
Add test case test_migration_with_rebuilding_replica

chriscchien:

Add test case test_migration_with_rebuilding_replica

### DIFF
--- a/manager/integration/tests/test_migration.py
+++ b/manager/integration/tests/test_migration.py
@@ -1,11 +1,16 @@
 import pytest
 import subprocess
 import common
-from common import clients, volume_name  # NOQA
+from common import clients, volume_name, wait_for_volume_healthy  # NOQA
 from common import get_random_client
 from common import SIZE
 from common import wait_for_volume_delete
 from common import set_node_scheduling
+from common import get_volume_endpoint, write_volume_dev_random_mb_data
+from common import get_device_checksum
+from common import wait_for_rebuild_start, wait_for_rebuild_complete
+from common import Gi
+from test_scheduling import get_host_replica
 
 REPLICA_COUNT = 2
 
@@ -266,7 +271,6 @@ def test_migration_with_failed_replica(clients, request, volume_name):  # NOQA
 
 @pytest.mark.coretest  # NOQA
 @pytest.mark.migration # NOQA
-@pytest.mark.skip(reason="TODO") # NOQA
 def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
     """
     Test that a degraded migratable RWX volume that contain a rebuilding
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_migration.py
```diff
@@ -220,7 +223,70 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
...
+                                  numberOfReplicas=3,
+                                  backingImage="",
+                                  accessMode="rwx", migratable=True)
+    volume = common.wait_for_volume_detached(client, volume_name)
+    volume.attach(hostId=current_host)
+    volume = common.wait_for_volume_healthy(client, volume_name)
+
+    volume_endpoint = get_volume_endpoint(volume)
+    write_volume_dev_random_mb_data(volume_endpoint,
+                                    1, 1500)
```
khushboo-rancher, line 242: Is 1M data enough for replica to rebuild while the volume got attached to another node and we can assert 'there should be no replica created before rebuilding complete' ?

manager/integration/tests/test_migration.py
```diff
@@ -220,7 +223,70 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
...
+            if r.name != replica0.name and \
+                    r.name != replica1.name and \
+                    r.name != replica2.name:
+                new_replica_found = True
+                break
+        if new_replica_found:
+            break
+        time.sleep(RETRY_INTERVAL)
+
+    assert new_replica_found is True
```
khushboo-rancher, line 265: In the step 3, a replica has to be removed only. the new replicas for migration will get rebuild only after this volume is attached to another node.

khushboo-rancher (changes requested):

manager/integration/tests/test_migration.py
chriscchien: Hello @khushboo-rancher , 
I use 1500M to make sure the volume attached on origin node is in rebuilding status and then we can do migration to target node. If use small data size, rebuild on origin node will finish soon and hard to match the scenario test case wanted.

manager/integration/tests/test_migration.py
chriscchien: Hi @khushboo-rancher ,

Above code is to make sure replica rebuilding happened on volume which not migrate yet, if I missed something, I will update soon, thank you

innobead:
@chriscchien Please resolve the conflicts from the branch first.

manager/integration/tests/test_migration.py
```diff
@@ -220,7 +223,70 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
     7. Observe volume migrated to node 2 (single active engine).
     8. Validate initially written test data.
     """
-    return
+    # Step 1
+    client = get_random_client(clients)  # NOQA
```
innobead, line 227: Why we need to add `# NOQA`, also this commit looks weird and not sure the purpose `sdf`?.

manager/integration/tests/test_migration.py
```diff
@@ -220,7 +223,70 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
     7. Observe volume migrated to node 2 (single active engine).
     8. Validate initially written test data.
     """
-    return
+    # Step 1
+    client = get_random_client(clients)
+    current_host = common.get_self_host_id()
+    host1, host2 = get_hosts_for_migration_test(clients) # NOQA
```
innobead, line 229: Why keep adding `# NOQA`? any purpose?

manager/integration/tests/test_migration.py
```diff
@@ -220,7 +223,70 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
...
+                new_replica_found = True
+                break
+        if new_replica_found:
+            break
+        time.sleep(RETRY_INTERVAL)
+
+    assert new_replica_found is True
+
+    # Step 4
+    volume.attach(hostId=host1)
```
innobead, line 268: As per the step 4, need to see new replicas created after the removed replica gets rebuilt successfully first.

manager/integration/tests/test_migration.py
```diff
@@ -220,7 +223,70 @@ def test_migration_with_rebuilding_replica(clients, volume_name):  # NOQA
...
+
+    # Step 5
+    volume = common.wait_for_volume_migration_ready(client, volume_name)
```
innobead, line 271: Need to check the number of replicas from the first volume the same as the second migrated.

innobead (changes requested):
Please ensure the implementation based on the test steps.

manager/integration/tests/test_migration.py
chriscchien: I update this part of code by using `wait_for_rebuild_start` to reuse exist function

manager/integration/tests/test_migration.py
chriscchien: Sorry, I removed the unnecessary comment

manager/integration/tests/test_migration.py
chriscchien: Hi @innobead , I added `wait_for_rebuild_start` and `wait_for_rebuild_complete` to check the delete replica rebuild process

manager/integration/tests/test_migration.py
chriscchien: I update code to check the new replica numbers for review, thank you.

manager/integration/tests/test_migration.py
khushboo-rancher: @chriscchien I'm not sure, I understood. I think we need to check the replica count as `old_replicas` until the rebuilding is completed.

khushboo-rancher (commented):
LGTM in general, one last comment.

manager/integration/tests/test_migration.py
chriscchien: Hi @khushboo-rancher , you are right, I add check code at line 323 to check replica count. but it's not to easy checkout the count during rebuild dynamically, so I checked once, thank you

innobead (approved):

innobead:
@chriscchien Please backport to 1.3 and 1.2. Thanks.

PR #1100, opened 2022-09-12
ci: get rid of longhorn-manger dependencies in longhorn-tests repo

yangchiu:
ci: get rid of longhorn-manger dependencies in longhorn-tests repo

For https://github.com/longhorn/longhorn/issues/4476


innobead (approved):
LGTM

PR #1101, opened 2022-09-13
Add test case: check restored volume same as original volume

weizhe0422:
[Longhorn 4531](https://github.com/longhorn/longhorn/issues/4531)

## Purpose
Add test cases to check whether the backup access mode is stored and the correctness of restore.

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -7,7 +7,7 @@ import random
 import yaml
 
 import common
-from common import client, random_labels, volume_name  # NOQA
+from common import client, random_labels, volume_name, clients  # NOQA
 from common import core_api, apps_api, pod, statefulset   # NOQA
 from common import SIZE, EXPAND_SIZE
 from common import check_device_data, write_device_random_data
@@ -81,6 +81,7 @@ from common import write_volume_dev_random_mb_data
 from common import VOLUME_HEAD_NAME
 from common import set_node_scheduling
 from common import SETTING_FAILED_BACKUP_TTL
+from common import wait_for_volume_creation
 
 from backupstore import backupstore_delete_volume_cfg_file
 from backupstore import backupstore_cleanup
@@ -4558,3 +4559,59 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
                       + volume_name + " is deleted"
     except AssertionError:
         pass
+
+
+@pytest.mark.parametrize(
+    "access_mode,overridden_restored_access_mode",
+    [
+        pytest.param("rwx", "rwo"),
+        pytest.param("rwo", "rwx")
+    ],
+)
+def test_backup_volume_restore_with_access_mode(set_random_backupstore, # NOQA
+                                                client, # NOQA
+                                                access_mode, # NOQA
+                                       overridden_restored_access_mode): # NOQA
+    """
+    Test the backup w/ the volume access mode, then restore a volume w/ the
+     original access mode or being overridden.
+
+    1. Prepare a healthy volume
+    2. Create a backup for the volume
+    3. Restore a volume from the backup w/o specifying the access mode
+       => Validate the access mode should be the same the volume
+    4. Restore a volume from the backup w/ specifying the access mode
+       => Validate the access mode should be the same as the specified
+    """
+    # Step 1
+    test_volume_name = generate_volume_name()
+    client.create_volume(name=test_volume_name,
+                         size=str(DEFAULT_VOLUME_SIZE * Gi),
+                         numberOfReplicas=2,
+                         accessMode=access_mode)
+    wait_for_volume_creation(client, test_volume_name)
+    volume = wait_for_volume_detached(client, test_volume_name)
+    volume.attach(hostId=common.get_self_host_id())
+    volume = common.wait_for_volume_healthy(client, test_volume_name)
+
+    # Step 2
+    _, b, _, _ = create_backup(client, test_volume_name)
+
+    # Step 3
+    volume_name_ori_access_mode = test_volume_name + '-default-access-mode'
+    client.create_volume(name=volume_name_ori_access_mode,
+                         size=str(DEFAULT_VOLUME_SIZE * Gi),
+                         numberOfReplicas=2,
+                         fromBackup=b.url)
+    volume_ori_access_mode = client.by_id_volume(volume_name_ori_access_mode)
+    assert volume_ori_access_mode.accessMode == access_mode
+
+    # Step 4
+    volume_name_sp_access_mode = test_volume_name + '-specified-access-mode'
+    client.create_volume(name=volume_name_sp_access_mode,
+                         size=str(DEFAULT_VOLUME_SIZE * Gi),
+                         numberOfReplicas=2,
+                         accessMode=overridden_restored_access_mode,
+                         fromBackup=b.url)
+    volume_sp_access_mode = client.by_id_volume(volume_name_sp_access_mode)
+    assert volume_sp_access_mode.accessMode == overridden_restored_access_mode


manager/integration/tests/test_basic.py
```diff
@@ -4558,3 +4561,66 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
                       + volume_name + " is deleted"
     except AssertionError:
         pass
+
+
+def test_backup_save_access_mode(client, core_api, apps_api,  # NOQA
```
innobead, line 4566: Suggest renaming to `test_backup_volume_restore_with_access_mode`

Suggest having two test cases for rwx and rwo instead of mixing together. Probably you can leverage https://docs.pytest.org/en/7.1.x/example/parametrize.html to avoid duplicated code.

manager/integration/tests/test_basic.py
```diff
@@ -4558,3 +4561,66 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
...
+    5. Restoring the backup using the `readwriteonce` access mode
+    6. Check the restored volume's access mode should be 'readwriteonce'
+    """
+
+    test_volume_name = "test-rwx-volume"
+    test_pv_name = test_volume_name + "-pv"
+    test_pvc_name = test_volume_name + "-pvc"
+    volume_access_mode = "rwx"
+
+    # create a volume
```
innobead, line 4584: Suggest adding the corresponding step number above, so it will be easier to understand which part is.

manager/integration/tests/test_basic.py
```diff
@@ -4558,3 +4561,66 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
...
+    # create the deployment with specific PV & PVC
+    create_pv_for_volume(client, core_api, volume, test_pv_name)
+    create_pvc_for_volume(client, core_api, volume, test_pvc_name)
+    deployment = make_deployment_with_pvc('deployment-rwx-volume-test',
+                                          test_pvc_name, replicas=2)
+
+    # deploy and get a rwx volume
+    apps_api = get_apps_api_client()
+    create_and_wait_deployment(apps_api, deployment)
+    volume.attach(hostId=get_self_host_id())
```
innobead, line 4599: You don't need to create a deployment for this test, but rather have a volume w/ access mode, then restore to check the access mode correct.

Also, don't need to call `volume.attach` if you have a deployment w/ the volume, because it will be automatically attached after the deployment deployed.

innobead (changes requested):

manager/integration/tests/test_basic.py
weizhe0422: Got it, I will add the step number to mark the section and the description will be in the head section.

manager/integration/tests/test_basic.py
weizhe0422: I'm going to rename the function name and parameterize the parameters to have multiple scenarios.

https://github.com/longhorn/longhorn-tests/blob/c0f5ac9ca5589e641830fe791497c56494412de1/manager/integration/tests/test_basic.py#L4565-L4573

manager/integration/tests/test_basic.py
weizhe0422: After consult with @chriscchien and the other test case, I will user `create_volume()` function with access mode argument.

https://github.com/longhorn/longhorn-tests/blob/c0f5ac9ca5589e641830fe791497c56494412de1/manager/integration/tests/test_basic.py#L4583-L4593

manager/integration/tests/test_basic.py
```diff
@@ -4558,3 +4560,60 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
...
+
+@pytest.mark.parametrize(
+    "access_mode,restored_access_mode",
+    [
+        pytest.param("rwx", "rwo"),
+        pytest.param("rwo", "rwx")
+    ],
+)
+def test_backup_volume_restore_with_access_mode(client, access_mode, # NOQA
+                                                restored_access_mode): # NOQA
```
innobead, line 4573: Suggest reanming `restored_access_mode` to `overridden_restored_access_mode`

innobead (changes requested):
In general, LGTM, but some minor requests needed to change. 

Also, for some knowledge about migratble, you can get more understanding about it which has been used for volume migration purpose between nodes.

manager/integration/tests/test_basic.py
```diff
@@ -4558,3 +4560,60 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
...
+    """
+    Test the backup should be save access mode and restore based on the value
+    1. Prepare a healthy volume
+    2. Create a backup for the volume generated in Step 1
+    3. Restore from backup w/o specify the access mode
+       => Validate the access mode should be the same as the Step 1.
+    4. Restore from backup w/ specify the access mode
+       => Validate the access mode should be the same as the specified.
```
innobead, line 4581:
```suggestion
    Test the backup should be save access mode and restore based on the value
    1. Prepare a healthy volume
    2. Create a backup for the volume
    3. Restore from backup w/o specifying the access mode
       => Validate the access mode should be the same the volume
    4. Restore from backup w/ specifying the access mode
       => Validate the access mode should be the same as the specified
```

manager/integration/tests/test_basic.py
```diff
@@ -4558,3 +4560,60 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
...
+                         numberOfReplicas=2,
+                         accessMode=access_mode,
+                         migratable=True if access_mode == "rwx" else False)
+    wait_for_volume_creation(client, test_volume_name)
+    volume = wait_for_volume_detached(client, test_volume_name)
+    volume.attach(hostId=common.get_self_host_id())
+    volume = common.wait_for_volume_healthy(client, test_volume_name)
+
+    # Step 2
+    set_backupstore_nfs(client)
```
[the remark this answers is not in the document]
weizhe0422: Reference the other test function to add `set_random_backupstore` fixture.

https://github.com/longhorn/longhorn-tests/blob/4d01c58adc04fd265fce65b9bcf848e0d7472e69/manager/integration/tests/test_basic.py#L4571-L4574

manager/integration/tests/test_basic.py
```diff
@@ -4558,3 +4560,60 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
...
+    4. Restore from backup w/ specify the access mode
+       => Validate the access mode should be the same as the specified.
+    """
+    # Step 1
+    test_volume_name = generate_volume_name()
+    client.create_volume(name=test_volume_name,
+                         size=str(DEFAULT_VOLUME_SIZE * Gi),
+                         numberOfReplicas=2,
+                         accessMode=access_mode,
+                         migratable=True if access_mode == "rwx" else False)
```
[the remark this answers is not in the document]
weizhe0422: Remove `migratable` argument since this case w/o migratable property support.

https://github.com/longhorn/longhorn-tests/blob/4d01c58adc04fd265fce65b9bcf848e0d7472e69/manager/integration/tests/test_basic.py#L4588-L4591
https://github.com/longhorn/longhorn-tests/blob/4d01c58adc04fd265fce65b9bcf848e0d7472e69/manager/integration/tests/test_basic.py#L4602-L4605
https://github.com/longhorn/longhorn-tests/blob/4d01c58adc04fd265fce65b9bcf848e0d7472e69/manager/integration/tests/test_basic.py#L4602-L4605

innobead (approved):
LGTM.

Not sure if it's possible to test resotring a backup which does not exist in the cluster yet, so we can catch this issue, https://github.com/longhorn/longhorn/issues/4637#issuecomment-1261784478.

innobead:
@weizhe0422 Please backport to 1.3. Thanks.

PR #1111, opened 2022-09-16
ci: ensure all testing resources have proper tag to be identified

yangchiu:
ci: ensure all testing resources have proper tag to be identified

(1) ensure aws ec2 instances have Owner="longhorn-infra" to be identified as longhorn testing instances.
(2) ensure all aws resources have Name contains random suffix to be associated to aws ec2 instances, then if there is a long-running aws ec2 instance, the random suffix can be used to track and delete all the aws resources associated to this aws ec2 instance.
(3) add Owner="longhorn-infra" to all aws resources to reverse lookup orphaned resources with no aws ec2 instance.

For https://github.com/longhorn/longhorn/issues/4395

### DIFF
--- a/airgap/terraform/main.tf
+++ b/airgap/terraform/main.tf
@@ -53,6 +53,7 @@ resource "aws_vpc" "lh_registry_aws_vpc" {
 
   tags = {
     Name = "lh-registry-vpc-${random_string.random_suffix.id}"
+    Owner = "longhorn-infra"
   }
 }
 
@@ -62,6 +63,7 @@ resource "aws_internet_gateway" "lh_registry_aws_igw" {
 
   tags = {
     Name = "lh-registry-igw-${random_string.random_suffix.id}"
+    Owner = "longhorn-infra"
   }
 }
 
@@ -120,6 +122,7 @@ resource "aws_security_group" "lh_registry_aws_secgrp" {
 
   tags = {
     Name = "lh-registry-secgrp-${random_string.random_suffix.id}"
+    Owner = "longhorn-infra"
   }
 }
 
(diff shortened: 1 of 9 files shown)


airgap/terraform/main.tf
```diff
@@ -240,6 +252,8 @@ resource "aws_instance" "lh_registry_aws_instance" {
 
   tags = {
     Name = "lh-registry-${random_string.random_suffix.id}"
+    DoNotDelete = "true"
```
innobead, line 255: Do we need to have `DoNotDelete` to all resourecs as well? Is this tag really taking effect?

yangchiu: As far as I know, `DoNotDelete` prevent ec2 instances from being deleted by rancher IT, but it's not applicable for other resources like vpc, subnet, etc. Other resources will live forever without being periodically deleted even though they don't have any tag.

yangchiu: So there would be bunch of orphaned resources without associated to any ec2 instances, since sometimes the ec2 instance is deleted manually but other resources associated to this ec2 instance aren't cleaned up.
Currently https://github.com/longhorn/longhorn-tests/pull/1116 only deletes long-running ec2 instances with their associated resources, but if ec2 instance is manually deleted, the remaining orphaned resources are still unable to be cleaned up. The next step of https://github.com/longhorn/longhorn/issues/3913 should be reverse lookup these resources to find the associated ec2 instances, if the associated ec2 instances not found, it means the resource is orphaned and should be cleaned up.

innobead: yeah, sounds good to me.

innobead (approved):

PR #1123, opened 2022-09-27
ci: update rockylinux ami filter to get 8.6 image and refine init script to disable nm-cloud-setup

yangchiu:
ci: update rockylinux ami filter to get 8.6 image and refine init script to disable nm-cloud-setup

For https://github.com/longhorn/longhorn/issues/4620


innobead (approved):
LGTM

innobead:
@yangchiu We need to fix this in 1.3 and 1.2 but keep the original OS version instead of bumping to 8.6.

PR #1124, opened 2022-09-27
test(replica-auto-balance): unschedulable node

c3y1huang:
Implement the case to reproduce and validate the fix for https://github.com/longhorn/longhorn/issues/4502

The replica auto-balance should not go into a rebuilding loop when the replica is already running on an un-schedulable node.

### DIFF
--- a/manager/integration/tests/test_zone.py
+++ b/manager/integration/tests/test_zone.py
@@ -28,6 +28,8 @@ from common import SETTING_DEFAULT_DATA_LOCALITY
 
 from test_scheduling import wait_new_replica_ready
 
+from test_node import set_node_cordon
+
 ZONE1 = "lh-zone1"
 ZONE2 = "lh-zone2"
 ZONE3 = "lh-zone3"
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_zone.py
```diff
@@ -492,6 +494,91 @@ def test_replica_auto_balance_zone_best_effort(client, core_api, volume_name):
...
+        z1_r_count = get_zone_replica_count(
+            client, volume_name, ZONE1, chk_running=True)
+        z2_r_count = get_zone_replica_count(
+            client, volume_name, ZONE2, chk_running=True)
+        z3_r_count = get_zone_replica_count(
+            client, volume_name, ZONE3, chk_running=True)
+
+        assert z1_r_count == 0 and (z2_r_count and z3_r_count == 1)
+        time.sleep(RETRY_INTERVAL)
+
```
khushboo-rancher, line 580: Should we also assert that the volume status is not isRebuilding?

khushboo-rancher (commented):
LGTM in general, left a comment

manager/integration/tests/test_zone.py
c3y1huang: Not necessary, because replica auto-balance will first create a new replica and then remove the old replica. This means checking on the replica number change should be sufficient.

But it's nice to have, updated!

khushboo-rancher (approved):
LGTM

manager/integration/tests/test_zone.py
```diff
@@ -492,6 +494,95 @@ def test_replica_auto_balance_zone_best_effort(client, core_api, volume_name):
...
 
+def test_replica_auto_balance_when_replica_on_unschedulable_node(client, core_api, volume_name, request):  # NOQA
+    """
+    Scenario: replica auto-balance when replica already running on
+              an unschedulable node.
+
+    Issue: https://github.com/longhorn/longhorn/issues/4502
+
+    Given set `replica-soft-anti-affinity` to `true`.
+    And set `replica-zone-soft-anti-affinity` to `true`.
```
innobead, line 505: Why we need to make both `soft-anti-affinity` to `true`?

That's because replica auto-rebalancing will only take effect when node/zone or both soft-anti-affinity, because replicas could be possible to stay on the same node/zone, then they will be rebalanced based on the `replicaAutoBalance` setting.

innobead (approved):
LGTM

PR #1125, opened 2022-09-27
test: initialize negative testing

yangchiu:
test: initialize negative testing


yangchiu:
@longhorn/qa 

After investigation, the test cases in `1. Longhorn operations running continuously` part of Negative Testing should be able to be automated, since the test pod has the ability to `directly manipulate the filesystem on the host node` (e.g. execute `mkfs`, `dd` or other shell commands). If the Longhorn volume we'd like to manipulate is attached to the same node as the test pod is running on, there should be no problems.

Also I think we can introduce a test framework with higher abstraction than pytest. I use `robot framework` here but just for demonstration. As we can see in https://github.com/longhorn/longhorn-tests/pull/1125/files, there is just a 1-to-1 mapping between each test step and python function, it would be more easier to implement and maintain test code, also it supports `FOR ... IN ... ` syntax and it's what we need in the negative test cases.

WDYT?

innobead:
Awesome to have robot!

manager/integration/negative_tests/lib/Operations.py
```diff
@@ -0,0 +1,62 @@
+import socket
+import time
+
+from kubernetes import config as k8sconfig
+from common import get_mgr_ips, get_client, PORT, Gi, create_volume as _create_volume, get_self_host_id, prepare_host_disk, \
```
c3y1huang, line 5: Suggest splitting the imports for readability.

manager/integration/negative_tests/negative.robot
```diff
@@ -0,0 +1,24 @@
...
+    ${MOUNT_PATH}=    Format Volume    volume-1    -b 1024
+    Write Random Data To Volume    ${MOUNT_PATH}/data
+    ${CHECKSUM}=    Get Volume Data Checksum    ${MOUNT_PATH}/data
+    FOR    ${i}    IN    RANGE    ${LOOP_COUNT}
+        Delete Replica To Trigger Rebuilding    volume-1    ${NODE_IDS}[0]
+        Delete Replica To Trigger Rebuilding    volume-1    ${NODE_IDS}[1]
+        Delete Replica To Trigger Rebuilding    volume-1    ${NODE_IDS}[2]
+        ${CHECKSUM_PRIME}=    Get Volume Data Checksum    ${MOUNT_PATH}/data
+        Should Be Equal    ${CHECKSUM}    ${CHECKSUM_PRIME}
+    END
```
innobead, line 23: For the upper layer of each test case, suggest making it human-readable, and the detailed steps can be implemented in separate keyword in the same robot file. Later, we can use the test case to generate the test cases for user reference.

For example here can be. Also, if keywords can be reused, you can place them in a separate dir to make them reusable for other test suites/cases.

```robot
*** Test Cases ***
Replica Rebuilding With Replica Deletion
    Given create Volume
    and write random data to volume
    When delete replica to trigger replica rebuilding
    Then after replica rebuilt, the volume data is intact
 
 *** Keywords ***
  Create Volume
  ..
  
  Write Random Data to Volume
  ...
```

BTW, this is not a negative test case but a normal replica rebuilding.

innobead: cc @khushboo-rancher, she is planning negative tests, so we can base on that to write down all robot test cases w/o implementation first.

yangchiu: Normally our test case will involve creating multiple volumes and dealing with different objects of different volumes, like: https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_basic.py#L1584.

Is there a good way to write BDD style test step to deal with this kind of object referencing? like:
```
snapshot1 of volume1
backup2 of volume2
data1 in volume3
data2 in volume3
replica3 of volume3
node of replica3 of volume3 
... and so on
```

For the existing test cases, each test case is just a big function, and every object is just a local variable, so we don't need to worry about this. But for BDD, each test step (and its corresponding function) is independent, I don't have a clear idea about this yet. Can anyone shed some light on this?

yangchiu: I think we can first pick some most complex test cases from the existing test code (like `test_listing_backup_volume`, `test_salvage_auto_crash_all_replicas`) or the negative test cases (like `Replica Rebuilding with replica deletion`) and try to translate them into robot framework test case definitions, then we know whether there is any obstacle.

yangchiu: Take this test case definition for example:
> 
> ```
> *** Test Cases ***
> Replica Rebuilding With Replica Deletion
>     Given create Volume
>     and write random data to volume
>     When delete replica to trigger replica rebuilding
>     Then after replica rebuilt, the volume data is intact
>  
The original negative test case is:
```
Create a volume 5 GB with 3 replicas. r1, r2 and r3
Write in the volume.
Delete one replica to trigger the rebuilding. Let's say r1 is rebuilding.
While rebuilding is happening, delete other replica r2.
Wait for replica rebuilding r1 get finished. Delete replica r3.
Check the data
```
This requires delete different replica every time.

But for the step definition:
>     When delete replica to trigger replica rebuilding
It doesn't indicate which replica to be deleted, and if it doesn't store the state somewhere neither, probably it will always delete the same replica.

innobead: How about below? Besides, you should be able to use some argument to make the below test run with different cases like different size fo volume or number of replica.

```robot
*** Test Cases ***
Replica Rebuilding With Replica Deletion
	Given create a volume 5 GB with 3 replicas # create 3 replicas with a convention name implicitly, like r1, r2, r3
	And write data to the volume
	
	When delete replica 1 to trigger the replica rebuilding # "replica 1" means r1, you should be able to save this variable (rebuilding_replica) via "Set Test Variable" builtin function provided by robot
	And during the rebuilding, delete replica 2 # you can check "rebuilding_replica" is rebuilding, then delete r2
	And wait for replica 1 rebuilt # just wait "rebuilding_replica" is ready
	And delete replica 3 # check r3
	
	Then check the data is intact
```

manager/integration/negative_tests/lib/Operations.py
yangchiu: most of functions in common.py use longhorn python REST API client, but using longhorn custom resources is preferred in the future. Probably will not reuse/import functions in common.py.

yangchiu:
The test steps definition has been refined, but for the test steps implementation, since it involves refactoring to use crd instead of REST python client, could take more time.

manager/integration/negative_tests/lib/Operations.py
c3y1huang: Suggest having the Longhorn HTTP API client as an option. This will be useful to test user-facing UI behaviors.

yangchiu:
@longhorn/qa @innobead @c3y1huang 

The very first negative test case [`Replica Rebuilding with replica deletion`](https://confluence.suse.com/pages/viewpage.action?spaceKey=LON&title=Negative+Testing) has been implemented in this PR.

- Basically it will [use CRD to operate Longhorn](https://github.com/longhorn/longhorn-tests/pull/1125/files#diff-95a12b8453103344ec76112b1e3db6e1078f6bb4c956101cd8bce8c63eea5576R18) instead of using Longhorn Python client
- If unable to use CRD to operate Longhorn, [fall back to use Longhorn Python client](https://github.com/longhorn/longhorn-tests/pull/1125/files#diff-95a12b8453103344ec76112b1e3db6e1078f6bb4c956101cd8bce8c63eea5576R116)
- [Keep using Longhorn Python client as an option](https://github.com/longhorn/longhorn-tests/pull/1125/files#diff-24f81d2bc84d5216c44b199a3d0bdb9c4220e810097dd376b49a238e753c79d3R15), as discussed in https://github.com/longhorn/longhorn-tests/pull/1125#discussion_r1071743101
- Briefly summarize the `Volume` package structure: `AbstractVolume` works as an interface, `CRDVolume` and `RestVolume` are CRD implementation and Python client implementation of `AbstractVolume`, and `Volume` works as a factory to return a `CRDVolume` or `RestVolume` according to the [strategy](https://github.com/longhorn/longhorn-tests/pull/1125/files#diff-c0bf92c63b32ed17df64511774efafd38f12e857e003f738a4932a9e315ce7a3R3)
- This PR is just a prototype, more error handling and debugging message should be added
- It takes more than **8** hours to run this single test case (since it's needed to be run for 200 times according to the test case definition), so it could be impossible to run negative cases one by one, some parallelism should be implemented

innobead:
> * It takes more than **8** hours to run this single test case (since it's needed to be run for 200 times according to the test case definition), so it could be impossible to run negative cases one by one, some parallelism should be implemented

This is a good result for the initial work. If taking a long execution time matters or impacts our productivity, we should make the parallel execution generally in the future.

c3y1huang:
> * [Keep using Longhorn Python client as an option](https://github.com/longhorn/longhorn-tests/pull/1125/files#diff-24f81d2bc84d5216c44b199a3d0bdb9c4220e810097dd376b49a238e753c79d3R15), as discussed in [test: initialize negative testing #1125 (comment)](https://github.com/longhorn/longhorn-tests/pull/1125#discussion_r1071743101)

Do you mean there will be an option to choose to test by the python HTTP API client with the robot test cases?

yangchiu:
> > * [Keep using Longhorn Python client as an option](https://github.com/longhorn/longhorn-tests/pull/1125/files#diff-24f81d2bc84d5216c44b199a3d0bdb9c4220e810097dd376b49a238e753c79d3R15), as discussed in [test: initialize negative testing #1125 (comment)](https://github.com/longhorn/longhorn-tests/pull/1125#discussion_r1071743101)
> 
> Do you mean there will be an option to choose to test by the python HTTP API client with the robot test cases?

Yes, [Volume._strategy](https://github.com/longhorn/longhorn-tests/pull/1125/files#diff-2d2f3f8bceb9c163a5041dcddbab88ff468489611e9950c09551a581541c05a2R10) should be configurable and determine the test should use HTTP API client or CRD

manager/integration/negative_tests/negative.robot
```diff
@@ -0,0 +1,50 @@
...
+Documentation    Negative Test Cases
+
+Library    ../framework/steps.py
+
+Suite Setup    suite_setup
+
+Test Teardown    cleanup_resources
+
+*** Test Cases ***
+Replica Rebuilding With Replica Deletion
```
innobead, line 11: The test name should be `Replica Deletion While Replica Rebuilding`, which can express  the `timing`.

manager/integration/negative_tests/negative.robot
```diff
@@ -0,0 +1,39 @@
...
+Documentation    Negative Test Cases
+
+Library    ../libs/common.py
+
+*** Test Cases ***
+Replica Rebuilding With Replica Deletion
+    Create a volume 5 GB with 3 replicas
+    Write data to the volume
+    FOR    ${i}    IN RANGE    200
+        Delete replica 0 to trigger replica rebuilding
```
innobead, line 15: `Delete replica 0 to trigger replica 0 rebuild`

manager/integration/negative_tests/negative.robot
```diff
@@ -0,0 +1,39 @@
...
+        Check data is intact
+    END
+
+*** Keywords ***
+Create a volume ${size} GB with ${replica_count} replicas
+    ${volume_name} =    create_volume   ${size}    ${replica_count}
+    Set Test Variable    ${volume_name}
+
+Write data to the volume
+    ${checksum} =     write_volume_random_data   ${volume_name}    2048
```
innobead, line 29: `${volume_data_checksum}`

manager/integration/framework/node_exec/node_exec.py
```diff
@@ -0,0 +1,116 @@
...
+    _instance = None
+    node_exec_pod = {}
+
+    def __new__(cls, *args, **kwargs):
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
+
+    def __init__(cls):
+        cls.core_api = client.CoreV1Api()
```
innobead, line 20: Suggest having a super class to have these common constructor for instance or class, so no need to repeat this kind of similar code in each subclass.

manager/integration/framework/node_exec/node_exec.py
```diff
@@ -0,0 +1,116 @@
...
+            'default',
+            command=exec_command,
+            stderr=True,
+            stdin=False,
+            stdout=True,
+            tty=False
+        )
+        return res
+
+    def launch_pod(cls, node_name):
```
innobead, line 53: It seems you would like to reuse a singleton pod to run commands instead of starting several pods. For the parallel execution in the future, please do concurrent handle for this method like mutex.

manager/integration/framework/strategy/strategy.py
```diff
@@ -0,0 +1,5 @@
+from enum import Enum
+
+class Strategy(Enum):
+    CRD = 'crd'
+    REST = 'rest'
```
innobead, line 5: Suggest making the name meaningful like `LonghornOperationStrategy`

manager/integration/framework/strategy/__init__.py
```diff
@@ -0,0 +1 @@
+from .strategy import Strategy
```
innobead, line 1: I feel this package is not really necessary. You probably can just have strategy.py in the root package folder, but you can refactor anytime in the future. KISS first.

manager/integration/framework/longhorn/__init__.py
```diff
@@ -0,0 +1 @@
+from .longhorn import from_env
```
innobead, line 1: The same as strategy. I feel you can just have longhorn.py under the root package, or make it specific like longhorn_api.py.

innobead (changes requested):
The idea is OK, but the folder structure I would suggest. For each module, if need to be more granualr, we can refactor it anytime.

In generaly, for import statement, suggest using `absolute path` instead of relative path to avoid unnecessary refactoring efforts.

libs/
    - __init__.py
    - longhorn.py
    - node_exec.py
    - strategy.py
    - util.py
    - resources
        - volume
        - ...
    - steps
        - <by resource type or test step catagory>

manager/integration/negative_tests/negative.robot
yangchiu: The test name is following https://confluence.suse.com/pages/viewpage.action?spaceKey=LON&title=Negative+Testing. So probably the document needs to be refined? @khushboo-rancher

manager/integration/negative_tests/negative.robot
yangchiu: Anyone can write a new test case skeleton. Should we need a rule about how to describe a test step?

innobead: Sounds good to me. Initially, this can be based on review and then figuring out the pattern we would like.

manager/integration/framework/node_exec/node_exec.py
yangchiu: I make it namespaced and every test case holds it own node_exec, volume, etc class instances to make it possible to run parallel, and these classes don't need to be singleton anymore.

manager/integration/framework/node_exec/node_exec.py
yangchiu: as per https://github.com/longhorn/longhorn-tests/pull/1125#discussion_r1136475552, these classes don't need to be singleton anymore.

khushboo-rancher (changes requested):
As @innobead suggested, let's make steps as folder and categorize them.
```
- steps
        - <by resource type or test step category>
```
The `negative.robot` can be renamed to something meaningful like `replica_resiliency_test.robot`

manager/integration/negative_tests/negative.robot
```diff
@@ -0,0 +1,50 @@
...
+
+Library    ../framework/steps.py
+
+Test setup    set_test_name    ${TEST NAME}
+
+Test Teardown    cleanup_resources
+
+*** Test Cases ***
+Replica Rebuilding While Replica Deletion
+    Create a volume 5 GB with 3 replicas
```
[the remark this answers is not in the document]
yangchiu: refined!

manager/integration/negative_tests/negative.robot
```diff
@@ -0,0 +1,50 @@
...
+    Write data to the volume
+    FOR    ${i}    IN RANGE    20
```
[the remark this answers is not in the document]
yangchiu: refined!

yangchiu:
> As @innobead suggested, let's make steps as folder and categorize them.
> 
> ```
> - steps
>         - <by resource type or test step category>
> ```
> 
> The `negative.robot` can be renamed to something meaningful like `replica_resiliency_test.robot`

refined!

khushboo-rancher (dismissed):

innobead (changes requested):
Please  help take this time to re-orgainze the folder strucutre of longhorn-tests as below.

- pipelines (airgap, benchmark_test, ...)
- docs (qa_docs)
- testbed_setup (test_framework)
- e2e (new framework)
  - libs (your framework folder in this PR. I already mentioned [here](https://github.com/longhorn/longhorn-tests/pull/1125#pullrequestreview-1297734547))
  - tests (negative_tests in this PR)
- manager (don't touch)

In README.md, explain the purpose of each folder, and mention `manager` folder will be deprecated after `e2e` (new framework) is ready.

manager/integration/negative_tests/replica_resiliency_test.robot
```diff
@@ -0,0 +1,53 @@
...
+    Write data to the volume
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        Delete replica 0 to trigger replica 0 rebuilding
+        During replica 0 rebuilding, delete replica 1
+        Wait until replica 0 rebuilt, delete replica 2
+        Check data is intact
+        Wait until all replicas rebuilt
+    END
+
+*** Keywords ***
```
innobead, line 25: These keywords can be placed in different keyword files, so they can be reused in other test cases later. 

How to decide what keyword files should be created. You can base on types of Longhorn resources. For example, volume related, you can add those keywords to `keywords/volume.robot`. Replica related, can have `replica.robot`.

About folder structure, it can be:
- libs (python functions for programmable keywords)
- tests (robot test cases)
- keywords (robot keywords)

manager/integration/framework/steps/volume_steps.py
```diff
@@ -0,0 +1,75 @@
+from pathlib import Path
+import sys
+sys.path.append(str(Path(__file__).parents[1]))
+from volume.volume import Volume
+from utility import Utility
+from nodes import Nodes
+from node_exec import NodeExec
+
+class volume_steps:
```
innobead, line 9: let's rename steps pkg folder to keywords to align with robot terminology. 

`keywords/volume.py` instead of `steps/volume_steps.py` (no need to have _steps, which is redundant.)

manager/integration/framework/volume/__init__.py
```diff
@@ -0,0 +1 @@
+from volume.volume import Volume
```
innobead, line 1: let's simplify module naming.

volume/base.py (add abstract and default volume class here)
volume/crd.py
volume/rest.py

then, we use the same pattern for other resources in the future.

innobead (changes requested):
In general, LGTM.

However, some commentes need to fix. We need to clarify some naming convention. (base on this PR, we need to document it later)

manager/integration/framework/volume/__init__.py
yangchiu: > (add abstract and default volume class here)

Put abstract and implementation class in the same file causes import error: `failed: ImportError: cannot import name 'AbstractVolume' from partially initialized module 'volumes.base' (most likely due to a circular import) `

Others refined.

manager/integration/framework/steps/volume_steps.py
yangchiu: refined.

manager/integration/negative_tests/replica_resiliency_test.robot
yangchiu: refined.

yangchiu:
> Please help take this time to re-organize the folder structure of longhorn-tests as below.
> 
>     * pipelines (airgap, benchmark_test, ...)
> 
>     * docs (rename qa_docs)
> 
>     * testbed_setup (rename test_framework)
> 
>     * e2e (new framework)
>       
>       * libs (your framework folder in this PR. I already mentioned [here](https://github.com/longhorn/longhorn-tests/pull/1125#pullrequestreview-1297734547))
>       * tests (negative_tests in this PR)
> 
>     * manager (don't touch)
> 
> 
> In README.md, explain the purpose of each folder, and mention `manager` folder will be deprecated after `e2e` (new framework) is ready.
> 
> cc @longhorn/qa

Opened another ticket to track https://github.com/longhorn/longhorn/issues/5936, since it's not related to negative testing PR.

manager/integration/framework/volume/__init__.py
innobead: Interesting, not sure why having this limitation.

e2e/libs/keywords/volume.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def set_test_name(self, test_name):
+        self.namespace = test_name.lower().replace(' ', '-')
+        self.node_exec = NodeExec(self.namespace)
+        self.volume = Volume(self.node_exec)
+
+
+    def create_volume(self, size, replica_count):
+        print('create_volume')
+        volume_name = Utility().generate_volume_name()
```
innobead, line 25: ddd

e2e/libs/keywords/volume_keywords.py
```diff
@@ -0,0 +1,73 @@
+from pathlib import Path
+from volume import Volume
+from utility import Utility
+from node import Node
+from node_exec import NodeExec
+
+class volume_keywords:
+
```
innobead, line 8: This naming convention is specific for robotframework. It needs to have the file name consistent with the internal library class.

Also, it's not always to have a class inside, because sometimes function keyword is good enough instead of method keywoard.

PR #1145, opened 2022-10-11
test: fix busybox version to 1.34.0  to avoid the CrashLoopBackOff issue in its latest version

yangchiu:
test: fix busybox version to 1.34.0  to avoid the CrashLoopBackOff issue in its latest version

For https://github.com/longhorn/longhorn/issues/4685


innobead (approved):
Probably we can use a variable for image:tag, so don't need to repeat the same update again if using a newer version.

PR #1147, opened 2022-10-13
bug: fix to create rwx volume with PV & PVC deployment

weizhe0422:
[Longhorn 4653](https://github.com/longhorn/longhorn/issues/4653)

Description:  As RWX volumes can only be attached in maintenance mode and cannot be attached using the API. Therefore, volumes require additional deployment PVs and PVCs.

Self test: Pass with parameter matrix with 4 condition
image

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -82,6 +82,8 @@ from common import VOLUME_HEAD_NAME
 from common import set_node_scheduling
 from common import SETTING_FAILED_BACKUP_TTL
 from common import wait_for_volume_creation
+from common import get_apps_api_client, create_and_wait_deployment
+from common import make_deployment_with_pvc # NOQA
 
 from backupstore import backupstore_delete_volume_cfg_file
 from backupstore import backupstore_cleanup
@@ -4572,10 +4574,13 @@ def test_backup_failed_disable_auto_cleanup(set_random_backupstore,  # NOQA
         pytest.param("rwo", "rwx")
     ],
 )
-def test_backup_volume_restore_with_access_mode(set_random_backupstore, # NOQA
+def test_backup_volume_restore_with_access_mode(core_api, # NOQA
+                                                set_random_backupstore, # NOQA
                                                 client, # NOQA
                                                 access_mode, # NOQA
-                                       overridden_restored_access_mode): # NOQA
+                                       overridden_restored_access_mode, # NOQA
+                                       make_deployment_with_pvc): # NOQA
+
     """
     Test the backup w/ the volume access mode, then restore a volume w/ the
      original access mode or being overridden.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -4588,21 +4593,34 @@ def test_backup_volume_restore_with_access_mode(set_random_backupstore, # NOQA
...
+    #  the volume was created by PV and PVC deployed.
+    if access_mode == "rwx":
+        test_pv_name = test_volume_name + "-pv"
+        test_pvc_name = test_volume_name + "-pvc"
+
+        create_pv_for_volume(client, core_api, volume, test_pv_name)
+        create_pvc_for_volume(client, core_api, volume, test_pvc_name)
+        deployment = make_deployment_with_pvc(
+            'deployment-multi-pods-test', test_pvc_name, replicas=2)
+        create_and_wait_deployment(get_apps_api_client(), deployment)
```
Lichiehyao, line 4614: Regarding `create_and_wait_deployment(get_apps_api_client(), deployment)`
Do we need a standard code format like [this](https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_ha.py#L2757-L2759)?

manager/integration/tests/test_basic.py
```diff
@@ -4588,21 +4593,34 @@ def test_backup_volume_restore_with_access_mode(set_random_backupstore, # NOQA
...
                          size=str(DEFAULT_VOLUME_SIZE * Gi),
                          numberOfReplicas=2,
                          accessMode=access_mode)
     wait_for_volume_creation(client, test_volume_name)
     volume = wait_for_volume_detached(client, test_volume_name)
-    volume.attach(hostId=common.get_self_host_id())
-    volume = common.wait_for_volume_healthy(client, test_volume_name)
+
+    # Since RWX volumes can only be manually attached in maintenance mode,
+    #  the volume was created by PV and PVC deployed.
```
innobead, line 4605: # Since RWX volumes can only be manually attached in maintenance mode, 
    # the volume should be created by PVC and PV instead to ensure the volume attach/deattach successfully.

manager/integration/tests/test_basic.py
```diff
@@ -4588,21 +4593,34 @@ def test_backup_volume_restore_with_access_mode(set_random_backupstore, # NOQA
...
+    if access_mode == "rwx":
+        test_pv_name = test_volume_name + "-pv"
+        test_pvc_name = test_volume_name + "-pvc"
```
innobead, line 4608: Suggest referring to other cases by using `pv_name = get_volume_name(core_api, pvc_name)` for using the similar convention.

manager/integration/tests/test_basic.py
```diff
@@ -4588,21 +4593,34 @@ def test_backup_volume_restore_with_access_mode(set_random_backupstore, # NOQA
        => Validate the access mode should be the same as the specified
     """
     # Step 1
-    test_volume_name = generate_volume_name()
+    test_volume_name = "test-volume-" + str(random.randint(0, 9))
```
innobead, line 4596: Any problem about using `generate_volume_name`? I thought all other cases use it w/o problems.

manager/integration/tests/test_basic.py
```diff
@@ -4588,21 +4593,34 @@ def test_backup_volume_restore_with_access_mode(set_random_backupstore, # NOQA
...
+    # Since RWX volumes can only be manually attached in maintenance mode,
+    #  the volume was created by PV and PVC deployed.
+    if access_mode == "rwx":
+        test_pv_name = test_volume_name + "-pv"
+        test_pvc_name = test_volume_name + "-pvc"
+
+        create_pv_for_volume(client, core_api, volume, test_pv_name)
+        create_pvc_for_volume(client, core_api, volume, test_pvc_name)
+        deployment = make_deployment_with_pvc(
+            'deployment-multi-pods-test', test_pvc_name, replicas=2)
```
innobead, line 4613: You probably should make the name unique instead of `deployment-multi-pods-test` to keep chances for future concurrent testing. 

Suggest referring to other cases by using volume + `-dep` like below.

```python
    deployment_name = volume_name + "-dep"
    deployment = make_deployment_with_pvc(deployment_name, pvc_name)
```

manager/integration/tests/test_basic.py
innobead: Basically, we should try to use pytest fixture instead (the code in longhorn-test here is not quite good formatted, but I think it should be fine here.

Need a further effort in the future when doing refactoring.

innobead (changes requested):
In general, LGTM. Some parts needed to resolve

manager/integration/tests/test_basic.py
weizhe0422: I use it for debug and forget to revert as `generate_volume_name()`, and I will fix it.

manager/integration/tests/test_basic.py
weizhe0422: Since the `create_pvc_for_volume()` need a PV which was already exists or it will return error message. 

Therefore, I think pv_name can't be created through pvc_name, but individually constructed PVC and PV

innobead: It's not right. PV is created after the PVC.

innobead: oh, you are talking about the function calls instead. Yeah, then it's OK here!

weizhe0422:
Revise and follow the to fix, and the self test result is pass with 4 test cases.
image

innobead (approved):

innobead:
@weizhe0422 Please backport to 1.3. Thanks.

PR #1148, opened 2022-10-18
Fix test_autosalvage_with_data_locality_enabled

chriscchien:

https://github.com/longhorn/longhorn/issues/4382

In step7 restarted pod is in running phase but immediately exec command into pod will hit I/O error
Add 3 seconds retry loop and tested in local environment
```
7. In a 3-min retry loop, verify that Longhorn salvage the volume
       and the workload pod is restarted. Exec into the workload pod.
       Verify that read/write to the volume is ok
```


innobead (changes requested):

chriscchien:
Sorry this PR's content is not correct, I forget add  data `assert` outside of for loop so test result will always pass. 

Actually I can reproduce IO error in master-head every time https://github.com/longhorn/longhorn/issues/4382#issuecomment-1283655312, I will temporary close this PR, thank you

PR #1156, opened 2022-10-21
test(replica-auto-balance): with dataLocality

c3y1huang:
https://github.com/longhorn/longhorn/issues/4763
https://github.com/longhorn/longhorn/issues/4761


manager/integration/tests/test_scheduling.py
```diff
@@ -694,6 +694,91 @@ def test_replica_auto_balance_node_best_effort(client, volume_name):  # NOQA
...
+
+    for _ in range(30):
+        volume = client.by_id_volume(volume_name)
+        if len(volume.replicas) == number_of_replicas and \
+                volume.replicas[0]['hostId'] == self_node:
+            break
+        try:
+            if len(volume.replicas) == number_of_replicas + 1:
+                is_rebuilded = True
+            assert len(volume.replicas) == number_of_replicas
```
khushboo-rancher, line 753: I think this assert can make the test flaky sometimes. If the replica on the self node was just built and the replica from another node is not yet removed?
Do you have result from our CI `longhorn-tests-regression`?

khushboo-rancher (dismissed):
LGTM in general. Left one comment.

manager/integration/tests/test_scheduling.py
c3y1huang: > I think this assert can make the test flaky sometimes. If the replica on the self node was just built and the replica from another node is not yet removed? 

The `except AssertionError` should catch the assert and retry in the for loop statement.

Do you have result from our CI `longhorn-tests-regression`?

https://github.com/longhorn/longhorn-manager/pull/1526#issuecomment-1286686066

manager/integration/tests/test_scheduling.py
```diff
@@ -694,6 +694,91 @@ def test_replica_auto_balance_node_best_effort(client, volume_name):  # NOQA
...
+            # breaking this for loop asserts the result of a single rebuild.
+            if is_rebuilded and len(volume.replicas) == number_of_replicas:
+                break
+            time.sleep(RETRY_INTERVAL)
+
+    assert len(volume.replicas) == number_of_replicas, \
+        f"Unexpected replica count for volume {volume_name}.\n"
+    assert volume.replicas[0]['hostId'] == self_node, \
+        f"Unexpected replica host ID for volume {volume_name}.\n"
+    assert volume[VOLUME_FIELD_ROBUSTNESS] == VOLUME_ROBUSTNESS_HEALTHY
```
innobead, line 768: Why not just check these 10 times? I don't really understand the repeated check above? Could you explain more for range(30) and range(15)? I feel nothing different.

innobead (changes requested):

manager/integration/tests/test_scheduling.py
c3y1huang: The range(30) interaction allows time for the data locality to rebuild the replica.
The range(15) interaction asserts there is no more rebuilding because after the first rebuilding the replica should be on the correct node.
The caller range(10) repeats the tests to ensure the original issue can be reproduced on every run of this e2e case. This is because though this case has high reproducibility, there is a possibility we miss it on a single run.

c3y1huang: After discussing with @innobead , we will skip this case from the daily run.

innobead (approved):

PR #1162, opened 2022-10-27
Fix the flaky issue in test_allow_volume_creation_with_degraded_avail…

roger-ryao:
…ability

After all nodes are enabled and replicas are scheduled, the replicas might be rebuilt before wait_for_volume_healthy(), the volume becomes healthy.

Thus, we wait for the volume healthy rather than degraded, and then check the data.

Longhorn 3220

(cherry picked from commit f3b65c6d5585910ef07a4f09e913f2f7e34b70b7)

issue https://github.com/longhorn/longhorn/issues/4787

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -3172,7 +3172,7 @@ def test_allow_volume_creation_with_degraded_availability(client, volume_name): 
     volume = common.wait_for_volume_detached(client, volume_name)
 
     volume.attach(hostId=self_host)
-    volume = common.wait_for_volume_degraded(client, volume_name)
+    volume = common.wait_for_volume_healthy(client, volume_name)
     check_volume_data(volume, data)
 
 


khushboo-rancher (approved):
LGTM, 
I'm okay with checking healthy state rather than checking degraded state as we are writing 512 Mi in the volume and rebuilt can happen so fast that there are chances of missing that degraded state and volume becoming healthy immediately.

Lichiehyao:
> LGTM, I'm okay with checking healthy state rather than checking degraded state as we are writing 512 Mi in the volume and rebuilt can happen so fast that there are chances of missing that degraded state and volume becoming healthy immediately.

Hi @khushboo-rancher :
     After I had expanded the volume size to 5 Gi and written 2 Gi data in the volume, detach and re-attach the volume again.
     The volume's `robustness` status changed from `unknown` to `healthy`. It look like we could not get `VOLUME_ROBUSTNESS_DEGRADED = "degraded"` anymore.
     
Screenshot_20221028_130119
     
[longhorn-support-bundle_7c433a98-b640-4c9f-be74-98e37db12bb6_2022-10-28T05-01-28Z.zip](https://github.com/longhorn/longhorn-tests/files/9885385/longhorn-support-bundle_7c433a98-b640-4c9f-be74-98e37db12bb6_2022-10-28T05-01-28Z.zip)

Lichiehyao:
I found that we sometimes could get volume status `degraded`
Screenshot_20221103_145245
https://ci.longhorn.io/job/public/job/v1.2.x/job/v1.2.x-longhorn-upgrade-tests-sles-arm64/270/

innobead (approved):
You can backport from https://github.com/longhorn/longhorn-tests/commit/f3b65c6d instead.

~In the future, you can check the commit from the original fix, then just do the backport. This way can make the context clear as well.~ My fault, you already backported this, then it's good! :+1:

PR #1164, opened 2022-11-01
Unscheduled replica clean-up

mantissahz:
Unscheduled replica would be deleted when volume was detached, so there would not be failed replicas when degraded volume was detached.

Merged after PR https://github.com/longhorn/longhorn-manager/pull/1534

longhorn/longhorn#4791

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -3395,8 +3395,7 @@ def test_allow_volume_creation_with_degraded_availability_restore(set_random_bac
                          numberOfReplicas=3, fromBackup=backup.url)
     common.wait_for_volume_replica_count(client, dst_vol_name, 3)
     common.wait_for_volume_restoration_start(client, dst_vol_name, backup.name)
-    wait_for_volume_condition_scheduled(client, dst_vol_name,
-                                        "status", "True")
+    common.wait_for_volume_degraded(client, dst_vol_name)
 
     # check only 1 replica scheduled successfully
     common.wait_for_replica_scheduled(client, dst_vol_name,
--- a/manager/integration/tests/test_migration.py
+++ b/manager/integration/tests/test_migration.py
@@ -159,13 +159,15 @@ def test_migration_with_unscheduled_replica(clients, volume_name):  # NOQA
     data = common.write_volume_random_data(volume)
     common.check_volume_data(volume, data)
 
-    volume.detach(hostId="")
-    volume = common.wait_for_volume_detached(client, volume_name)
+    # collect replicas name before detaching,
+    # unscheduled replica would be deleted.
     old_replicas = []
     v = client.by_id_volume(volume_name)
     replicas = v.replicas
     for r in replicas:
         old_replicas.append(r.name)
+    volume.detach(hostId="")
+    volume = common.wait_for_volume_detached(client, volume_name)
 
     # Step 6
     volume.attach(hostId=hosts[0])


khushboo-rancher (commented):
@mantissahz Could you please point me to the ticket where the behavior was changed? 
I have few qq:
1. When a volume is created using the Longhorn UI, the volume remains in detached state, does it mean that volume will have only replica equal to available disks in a degraded availability situations.
2. What if some disks were removed momentarily and no other disks are available for scheduling and volume got detached. Will the failed replicas be removed? If yes, we won't be able to reuse the existing replicas if the older disks come back.
3. I not sure if we need to revisit all the `degraded_availability` test cases.

mantissahz:
Hi @khushboo-rancher, 
After PR https://github.com/longhorn/longhorn-manager/pull/1534 merged, the degraded volume that was restored from a backup would not have a failed replica (that would be deleted) after the restoration, so this ticket removes the `expected_fail`.
>1. When a volume is created using the Longhorn UI, the volume remains in detached state, does it mean that volume will have only replica equal to available disks in a degraded availability situations.

There would be replicas that user requested when user created the volume but only schedule the number of replicas equal to the number of available disks
For example, 3 replicas asked when volume created, two available nodes or two available disks on different nodes and two replicas will be scheduled, one would be N/A.

>2. What if some disks were removed momentarily and no other disks are available for scheduling and volume got detached. Will the failed replicas be removed? If yes, we won't be able to reuse the existing replicas if the older disks come back.

No, usable failed replica would not be deleted.

>3. I not sure if we need to revisit all the degraded_availability test cases.

I have run tests on degraded_availability test cases, it looks fine.

manager/integration/tests/test_basic.py
```diff
@@ -3402,7 +3402,6 @@ def test_allow_volume_creation_with_degraded_availability_restore(set_random_bac
     common.wait_for_replica_scheduled(client, dst_vol_name,
```
innobead, line 3402: This step is while the volume is attached, so it should be able to verify `expect_fail=2`, right?

innobead (changes requested):
Exccept one comment, LGTM>

manager/integration/tests/test_basic.py
mantissahz: Found, I think the problem is at `wait_for_volume_condition_scheduled`.
`status` would only be `true` after the volume are detached because there are two replicas unscheduled.

innobead: > Found, I think the problem is at `wait_for_volume_condition_scheduled`. `status` would only be `true` after the volume are detached because there are two replicas unscheduled.

@mantissahz could you elaborate this a bit? I assume the status should be true when the volume is degraded when the replicas are scheduled during the attachment, right?

innobead: ref: https://github.com/longhorn/longhorn-manager/blob/5a75273b50210cc5e652efc7597d821aac65702a/controller/volume_controller.go#L1169-L1169

Not sure about the previous context, but I feel this condition is a little confusing here because even if the volume is degraded, the condition of `scheduled` should be true as well but w/ a meaningful reason to explain that only some replicas are actually scheduled. 

Probably you can check `Robustness` of volume to see if it's `degraded`, so can you add `wait_for_volume_degraded` here instead of just removing the check?

innobead (changes requested):

innobead (approved):

innobead:
@mantissahz Please backport.

PR #1165, opened 2022-11-02
ci: refine terraform k3s cloud-init scripts to support k8s v1.25

yangchiu:
ci: refine terraform k3s cloud-init scripts to support k8s v1.25

For https://github.com/longhorn/longhorn/issues/4807

### DIFF
--- a/test_framework/terraform/aws/centos/user-data-scripts/provision_k3s_agent.sh.tpl
+++ b/test_framework/terraform/aws/centos/user-data-scripts/provision_k3s_agent.sh.tpl
@@ -24,7 +24,7 @@ if [ -b "/dev/xvdh" ]; then
   mount /dev/xvdh /var/lib/longhorn
 fi
 
-until (curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="agent" K3S_URL="${k3s_server_url}" INSTALL_K3S_VERSION="${k3s_version}" K3S_CLUSTER_SECRET="${k3s_cluster_secret}" sh -); do
+until (curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="agent --token ${k3s_cluster_secret}" K3S_URL="${k3s_server_url}" INSTALL_K3S_VERSION="${k3s_version}" sh -); do
   echo 'k3s agent did not install correctly'
   sleep 2
 done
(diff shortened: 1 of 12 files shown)


test_framework/terraform/aws/centos/user-data-scripts/provision_k3s_agent.sh.tpl
```diff
@@ -24,7 +24,7 @@ if [ -b "/dev/xvdh" ]; then
   mount /dev/xvdh /var/lib/longhorn
 fi
 
-until (curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="agent" K3S_URL="${k3s_server_url}" INSTALL_K3S_VERSION="${k3s_version}" K3S_CLUSTER_SECRET="${k3s_cluster_secret}" sh -); do
+until (curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="agent --token ${k3s_cluster_secret}" K3S_URL="${k3s_server_url}" INSTALL_K3S_VERSION="${k3s_version}" sh -); do
```
innobead, line 27: Does this mean the installation way is incompatible with before? For example, using `--token` option with the agent command instead of an environment variable anymore `K3S_CLUSTER_SECRET`. Or, actually `K3S_CLUSTER_SECRET` is still working?

innobead: image

The environment var should be still valid, so I am curious why you would like to change it?

innobead (commented):
LGTM, but

test_framework/terraform/aws/centos/user-data-scripts/provision_k3s_agent.sh.tpl
yangchiu: > Does this mean the installation way is incompatible with before? 

Yes!

> Or, actually K3S_CLUSTER_SECRET is still working?

No, if using `K3S_CLUSTER_SECRET` in `v1.25.3+k3s1`, the k3s service will return error message `k3s[1724]: time="2022-11-02T09:23:46Z" level=fatal msg="cluster-secret is deprecated. Use --token instead."`, and the k3s service will not start.

For `--write-kubeconfig-mode 644`, I add this just for debugging purpose, otherwise it always needs `sudo` to execute `kubectl` on the server instances.

innobead: Then, I think k3s doc should have better clarification :slightly_frowning_face:

innobead (approved):

innobead:
This change should only apply to the master/1.4 w/o any backports.

PR #1167, opened 2022-11-02
test: fix flaky test case test_autosalvage_with_data_locality_enabled

yangchiu:
test: fix flaky test case test_autosalvage_with_data_locality_enabled

(1) add mechanism to wait for a stable running pod when deployment restarts its workload
(2) use dd to write large size test data

For https://github.com/longhorn/longhorn/issues/4814

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -107,6 +107,7 @@ DEFAULT_STATEFULSET_TIMEOUT = 180
 
 DEFAULT_DEPLOYMENT_INTERVAL = 1
 DEFAULT_DEPLOYMENT_TIMEOUT = 120
+WAIT_FOR_POD_STABLE_MAX_RETRY = 30
 
 
 DEFAULT_VOLUME_SIZE = 3  # In Gi
@@ -4138,13 +4139,31 @@ def create_and_wait_deployment(apps_api, deployment_manifest):
 
 def wait_and_get_any_deployment_pod(core_api, deployment_name,
                                     is_phase="Running"):
+    """
+    Add mechanism to wait for a stable running pod when deployment restarts its
+    workload, since Longhorn manager could create/delete the new workload pod
+    multiple times, it's possible that we get an unstable pod which will be
+    deleted immediately, so add a wait mechanism to get a stable running pod.
+    ref: https://github.com/longhorn/longhorn/issues/4814
+    """
+    stable_pod = None
+    wait_for_stable_retry = 0
+
     for _ in range(DEFAULT_DEPLOYMENT_TIMEOUT):
         label_selector = "name=" + deployment_name
         pods = core_api.list_namespaced_pod(namespace="default",
                                             label_selector=label_selector)
         for pod in pods.items:
             if pod.status.phase == is_phase:
-                return pod
+                if stable_pod is None or \
+                        stable_pod.metadata.name != pod.metadata.name:
+                    stable_pod = pod
+                    wait_for_stable_retry = 0
+                    break
+                else:
+                    wait_for_stable_retry += 1
+                    if wait_for_stable_retry == WAIT_FOR_POD_STABLE_MAX_RETRY:
+                        return stable_pod
 
         time.sleep(DEFAULT_DEPLOYMENT_INTERVAL)
     assert False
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -4138,13 +4139,23 @@ def create_and_wait_deployment(apps_api, deployment_manifest):
 
 def wait_and_get_any_deployment_pod(core_api, deployment_name,
                                     is_phase="Running"):
+    stable_pod = None
```
innobead, line 4142: Suggest adding a comment to describe why we do this way to checking `WAIT_FOR_POD_STABLE_MAX_RETRY` times to ensure the pod is really ready (still best efforts).

Also, create a github issue for resolving the problem about the pod is somehow sometimes deleted by longhorn manager a few times to cause this test flaky. (ref: https://github.com/longhorn/longhorn/issues/4382#issuecomment-1297639493)

manager/integration/tests/test_ha.py
```diff
@@ -2759,8 +2759,9 @@ def test_autosalvage_with_data_locality_enabled(client, core_api, make_deploymen
     create_and_wait_deployment(apps_api, deployment)
 
     pod_names = common.get_deployment_pod_names(core_api, deployment)
-    test_data = generate_random_data(VOLUME_RWTEST_SIZE)
-    write_pod_volume_data(core_api, pod_names[0], test_data, filename='test')
+    data_path = '/data/test'
+    write_pod_volume_random_data(core_api, pod_names[0], data_path, DATA_SIZE_IN_MB_1)
```
innobead, line 2763: Will dd really ensure to flush data during writes by default? Or, should we need any extra option to make it work?

manager/integration/tests/test_ha.py
```diff
@@ -2779,14 +2780,11 @@ def test_autosalvage_with_data_locality_enabled(client, core_api, make_deploymen
...
-    wait_pod(pod_names[0])
+    pod = common.wait_and_get_any_deployment_pod(core_api, deployment_name)
 
-    command = 'cat /data/test'
-    verified_data = exec_command_in_pod(
-        core_api, command, pod_names[0], 'default')
+    _md5sum = get_pod_data_md5sum(core_api, pod.metadata.name, data_path)
 
-    assert test_data == verified_data
+    assert md5sum == _md5sum
```
innobead, line 2787: Suggest using meaningful variables like `expected_test_data_checksum` and `test_data_checksum`.

innobead (changes requested):
In general, LGTM, but some questions needed to answer/resolve first.

innobead:
Also, fix the PR check failure.

manager/integration/tests/test_ha.py
innobead: cc @derekbit

manager/integration/tests/common.py
yangchiu: issue created (https://github.com/longhorn/longhorn/issues/4814) and comment added.

manager/integration/tests/test_ha.py
yangchiu: Thanks @derekbit for the information:
```
we can add `direct`:
`dd if=/dev/urandom of=1M bs=1M count=1 oflag=direct`
Then, the page cache will be bypassed

and adding `sync` will flush the data to dev before finishing the command:
`dd if=/dev/urandom of=1M bs=1M count=1 oflag=direct,sync`
```

But in our test code we use none of these flags, so actually it's possible that there are some cache issues. I think we can further investigate this and see if any failed test cases could probably suffer from this issue.

innobead (approved):
LGTM

innobead:
@yangchiu Could you create a test issue to apply https://github.com/longhorn/longhorn-tests/pull/1167#discussion_r1012440178 to related common libraries? I think we should apply this change to avoid any unstable results happening later.

yangchiu:
> @yangchiu Could you create a test issue to apply [#1167 (comment)](https://github.com/longhorn/longhorn-tests/pull/1167#discussion_r1012440178) to related common libraries? I think we should apply this change to avoid any unstable results happening later.

created https://github.com/longhorn/longhorn/issues/4844 for this.

yangchiu:
@Mergify backport v1.3.x

PR #1173, opened 2022-11-07
test: update kubernetes client version to 25.3 and replace batch/v1beta1 with batch/v1

yangchiu:
test: update kubernetes client version to 25.3 and replace batch/v1beta1 with batch/v1 to fix recurring job test cases failed on master pipeline

For https://github.com/longhorn/longhorn/issues/4830

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1087,17 +1087,17 @@ def apps_api(request):
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


innobead (approved):
LGTM

innobead:
Also, this means we have not updated K8s client lib for a long while.. not a good practice. We should update along with each release. cc @longhorn/qa

PR #1175, opened 2022-11-07

derekbit:
longhorn/longhorn#4210
longhorn/longhorn#3198

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -182,6 +182,16 @@ SETTING_CONCURRENT_AUTO_ENGINE_UPGRADE_NODE_LIMIT = \
     "concurrent-automatic-engine-upgrade-per-node-limit"
 SETTING_SUPPORT_BUNDLE_FAILED_LIMIT = "support-bundle-failed-history-limit"
 SETTING_RESTORE_RECURRING_JOBS = "restore-volume-recurring-jobs"
+SETTING_SNAPSHOT_DATA_INTEGRITY = "snapshot-data-integrity"
+SETTING_SNAPSHOT_DATA_INTEGRITY_IMMEDIATE_CHECK_AFTER_SNAPSHOT_CREATION = \
+    "snapshot-data-integrity-immediate-check-after-snapshot-creation"
+SETTING_SNAPSHOT_DATA_INTEGRITY_CRONJOB = "snapshot-data-integrity-cronjob"
+SETTING_SNAPSHOT_FAST_REPLICA_REBUILD_ENABLED = "fast-replica-rebuild-enabled"
+
+SNAPSHOT_DATA_INTEGRITY_IGNORED = "ignored"
+SNAPSHOT_DATA_INTEGRITY_DISABLED = "disabled"
+SNAPSHOT_DATA_INTEGRITY_ENABLED = "enabled"
+SNAPSHOT_DATA_INTEGRITY_FAST_CHECK = "fast-check"
 
 CSI_UNKNOWN = 0
 CSI_TRUE = 1
@@ -421,7 +431,8 @@ def delete_backup_volume(client, volume_name):
 
 def create_and_check_volume(client, volume_name,
                             num_of_replicas=3, size=SIZE, backing_image="",
-                            frontend=VOLUME_FRONTEND_BLOCKDEV):
+                            frontend=VOLUME_FRONTEND_BLOCKDEV,
+                            snapshot_data_integrity=SNAPSHOT_DATA_INTEGRITY_IGNORED): # NOQA
     """
     Create a new volume with the specified parameters. Assert that the new
     volume is detached and that all of the requested parameters match.
@@ -439,7 +450,8 @@ def create_and_check_volume(client, volume_name,
         backing_image = None
     client.create_volume(name=volume_name, size=size,
                          numberOfReplicas=num_of_replicas,
-                         backingImage=backing_image, frontend=frontend)
+                         backingImage=backing_image, frontend=frontend,
+                         snapshotDataIntegrity=snapshot_data_integrity)
     volume = wait_for_volume_detached(client, volume_name)
     assert volume.name == volume_name
     assert volume.size == size
@@ -4025,9 +4037,11 @@ def wait_for_rebuild_complete(client, volume_name, retry_count=RETRY_COUNTS):
     assert completed == len(rebuild_statuses)
 
 
-def wait_for_rebuild_start(client, volume_name):
+def wait_for_rebuild_start(client, volume_name,
+                           retry_count=RETRY_COUNTS,
+                           retry_interval=RETRY_INTERVAL):
     started = False
-    for i in range(RETRY_COUNTS):
+    for i in range(retry_count):
         v = client.by_id_volume(volume_name)
         rebuild_statuses = v.rebuildStatus
         for status in rebuild_statuses:
@@ -4036,7 +4050,7 @@ def wait_for_rebuild_start(client, volume_name):
                 break
         if started:
             break
-        time.sleep(RETRY_INTERVAL)
+        time.sleep(retry_interval)
     assert started
     return status.fromReplica, status.replica
 
--- a/manager/integration/tests/test_snapshot_check.py
+++ b/manager/integration/tests/test_snapshot_check.py
@@ -1,0 +1,491 @@
+import os
+import subprocess
+import time
+import datetime
+import json
+import pandas as pd
+
+from common import client, volume_name  # NOQA
+from common import update_setting
+from common import get_self_host_id
+from common import create_and_check_volume, wait_for_volume_healthy
+from common import wait_for_rebuild_start, wait_for_rebuild_complete
+from common import create_snapshot
+from common import settings_reset # NOQA
+
+from test_basic import get_volume_endpoint
+
+from common import Gi, Mi
+from common import RETRY_COUNTS, RETRY_INTERVAL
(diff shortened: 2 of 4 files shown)


manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+def check_hash_calculated_and_with_immediate_hash(client, volume_name, snapshot_data_integrity):  # NOQA
+    """
+    1. Create and attach a volume
+    2. Create snapshots
+    3. Check snapshots' checksums are calculated and set and also
+       check the checksums of the snapshot disk files are matched
+       with the values from the api
+    4. Check snapshots' checksums are calculated by periodic snapshot
```
khushboo-rancher, line 326: This step is not implemented in the test. Should we first clean up the volume's `checksum` manually and see if the periodic snapshot verification set the `checksum` again?

manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,425 @@
...
+    check_hash_calculated_and_with_immediate_hash(client, volume_name, SNAPSHOT_DATA_INTEGRITY_IGNORED)   # NOQA
+
+
+def test_snapshot_hash_global_fast_check_without_immediate_hash(client, volume_name, settings_reset):  # NOQA
+    update_settings(client, SNAPSHOT_DATA_INTEGRITY_FAST_CHECK, "false")
+    check_hash_calculated_and_without_immediate_hash(client, volume_name,
+                                                     SNAPSHOT_DATA_INTEGRITY_IGNORED)  # NOQA
+
+
+def test_snapshot_hash_global_disabled_with_immediate_hash(client, volume_name, settings_reset):  # NOQA
```
khushboo-rancher, line 52: Does it mean that the immediate_hash depends on the `snapshot_data_integrity`? If `snapshot_data_integrity` is disabled, immediate_hash will not take effect?

khushboo-rancher (changes requested):
@derekbit Thanks for adding the multiple test cases. 
Few comments and a doubt, if we change the volume object and remove the checksum manually (No modification in the replica on the disk), will the periodic snapshot Data Integrity update volume object again?

manager/integration/tests/common.py
```diff
@@ -177,6 +177,15 @@
 SETTING_DISABLE_REVISION_COUNTER = "disable-revision-counter"
 SETTING_ORPHAN_AUTO_DELETION = "orphan-auto-deletion"
 SETTING_FAILED_BACKUP_TTL = "failed-backup-ttl"
+SETTING_SNAPSHOT_DATA_INTEGRITY = "snapshot-data-integrity"
+SETTING_SNAPSHOT_DATA_INTEGRITY_IMMEDIATE_CHECK_AFTER_SNAPSHOT_CREATION = \
+    "snapshot-data-integrity-immediate-check-after-snapshot-creation"
+SETTING_SNAPSHOT_DATA_INTEGRITY_CRONJOB = "snapshot-data-integrity-cronjob"
+
+SNAPSHOT_DATA_INTEGRITY_IGNORED = "ignored"
```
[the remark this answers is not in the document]
derekbit: Yes, for the existing volumes, the `volume.spec.SnapshotDataIntegrity` is set to `ignored`. It means the volumes the `SnapshotDataIntegrity` following the global setting `snapshot-data-integrity`.

manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,425 @@
...
+    assert snapMap[snap1.name].removed is False
+    assert snapMap[snap2.name].name == snap2.name
+    assert snapMap[snap2.name].parent == snap1.name
+    assert snapMap[snap2.name].removed is False
+    assert snapMap[snap3.name].name == snap3.name
+    assert snapMap[snap3.name].parent == snap2.name
+    assert snapMap[snap3.name].removed is False
+
+
+def write_device_random_data(dev, size):
```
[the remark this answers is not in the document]
derekbit: Yes, but the function's behavior is different from that in the common. Should I rename this function?

manager/integration/tests/test_snapshot_check.py
derekbit: Yes. The `snapshot_data_integrity` is like a main option. The `immediate_hash` and `cronjob` are like suboptions.

manager/integration/tests/test_snapshot_check.py
derekbit: Ah, my bad. The 4. is actually done in the step 3. 

> hould we first clean up the volume's checksum manually and see if the periodic snapshot verification set the checksum again?

If the checksum is somehow cleaned up, the checksum will be re-calculated by the periodic snapshot verification.

manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,425 @@
...
+    wait_for_rebuild_complete(client, volume_name)
+
+    # Step 6
+    # Wait for 5 (+ 2 for tolerating some delay)  minutes, the checksum in
+    # getxattr should be updated by the periodic check
+    time.sleep(420)
+    assert check_snapshot_checksums(volume)
+
+    # Step 7
+    cleanup_volume(client, volume)
```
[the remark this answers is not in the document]
derekbit: Removed. Thank you,

manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,425 @@
...
+    """
+    1. Create and attach a volume
+    2. Create snapshots
+    3. Check snapshots' checksums are calculated and set and also
+       check the checksums of the snapshot disk files are matched
+       with the values from the api
+    4. Check snapshots' checksums are calculated by periodic snapshot
+       verification mechanism and set and also check the checksums of
+       the snapshot disk files are matched with the values from the api
+    5. Tear down volume
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

manager/integration/tests/test_snapshot_check.py
derekbit: Updated the steps. Thank you.

manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,409 @@
...
+    time.sleep(120)
+    assert check_snapshot_checksums_not_set(volume)
+
+    # Step 4
+    # After 5 minutes, the periodic snapshot check should
+    # calculate the checksum and set it in the xattr
+    time.sleep(240)
+    assert check_snapshot_checksums(volume)
+
+
```
roger-ryao, line 380: Hi @derekbit :
     According to https://github.com/longhorn/longhorn/issues/4210#issuecomment-1286534849, I think we could add the function `wait_for_snapshot_checksums_gernerate`.
     This function that will print snapshot checksum generates time on the console. We also could restrict the range of periods instead of `time.sleep`. What do you think?
Screenshot_20221126_002523
Screenshot_20221126_001659

```suggestion
    # Step 3
    # Step 4
    # After 5 minutes, the periodic snapshot check should
    # calculate the checksum and set it in the xattr
    checksums_set, generate_checksums_time = wait_for_snapshot_checksums_gernerate(volume)
    time_range = range(120,360)
    assert checksums_set
    assert generate_checksums_time in time_range

def wait_for_snapshot_checksums_gernerate(volume_name):
    for i in range(RETRY_COUNTS*4):
        snapshot_checksums_not_set = check_snapshot_checksums_not_set(volume_name)
        if snapshot_checksums_not_set:
            time.sleep(RETRY_INTERVAL)
            continue
        else:
            print(f'checksums set in {i} sec')
            generate_checksums_time = i
            snapshot_checksums_generate = True
            break
    return snapshot_checksums_generate, generate_checksums_time
```

manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,409 @@
...
+    volume = volume.attach(hostId=lht_hostId)
+    volume = wait_for_volume_healthy(client, volume_name)
+
+    volume = client.by_id_volume(volume_name)
+
+    # Step 2
+    create_snapshots(client, volume, 1)
+
+    # Step 3
+    assert check_snapshot_checksums(volume)
```
roger-ryao, line 342:
```suggestion
    checksums_set, generate_checksums_time = wait_for_snapshot_checksums_gernerate(volume)
    time_range = range(0,30)
    assert checksums_set
    assert generate_checksums_time in time_range
```

manager/integration/tests/test_snapshot_check.py
```diff
@@ -0,0 +1,425 @@
...
+    value = volume.snapshotList()
+    snapshots = value.data
+    for s in snapshots:
+        if s.checksum != "":
+            return False
+
+    return True
+
+
+def get_local_host_replica_data_path(volume):
```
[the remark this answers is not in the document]
derekbit: This function is only used by test_snapshot_check.py, so I put it here.
We can move it common.py later if other test cases need this function in the future. WDYT?

manager/integration/tests/test_snapshot_check.py
derekbit: LGTM. Thank you.

manager/integration/tests/test_snapshot_check.py
derekbit: Addressed. Thank you.

manager/integration/tests/test_snapshot_check.py
derekbit: Addressed. Thank you.

PR #1177, opened 2022-11-08
Add test test_auto_upgrade_engine_to_default_version

chriscchien:

https://github.com/longhorn/longhorn/issues/4737
Add test case `test_auto_upgrade_engine_to_default_version`
Run PR code 10 time on both [AMD64](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2385/testReport/tests/test_engine_upgrade/) / [ARM64](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2386/testReport/tests/test_engine_upgrade/) were  all passed

For the removed test step in PR code
```
6. In a retry loop, verify that the number of volumes who
       is upgrading engine is always smaller or equal to 3
```
I tried to compare value `volume.currentImage` between `volume.engineImage`,  if they were different and `volume.engineImage` is target image, I can consider that the volume is upgrading engine image, and total volume in upgrade status can not greater than 3

However due to timing issue(upgrade too fast), some times this step will become flaky so I removed this test step, thank you

### DIFF
--- a/manager/integration/tests/test_engine_upgrade.py
+++ b/manager/integration/tests/test_engine_upgrade.py
@@ -2,7 +2,7 @@ import pytest
 
 import common
 import time
-from common import client, core_api, volume_name  # NOQA
+from common import client, core_api, volume_name # NOQA
 from common import SIZE
 from common import check_volume_data, get_self_host_id
 from common import wait_for_volume_current_image, wait_for_volume_delete
@@ -25,7 +25,10 @@ from common import update_setting, wait_for_volume_healthy
 from common import get_volume_endpoint, write_volume_dev_random_mb_data
 from common import wait_for_rebuild_complete, RETRY_COUNTS, RETRY_INTERVAL
 from common import wait_for_rebuild_start
+from common import create_backup, wait_for_backup_restore_completed
+from common import SETTING_CONCURRENT_AUTO_ENGINE_UPGRADE_NODE_LIMIT
 from test_settings import delete_replica_on_test_node
+from backupstore import set_random_backupstore # NOQA
 
 REPLICA_COUNT = 2
 ENGINE_IMAGE_TEST_REPEAT_COUNT = 5
@@ -747,8 +750,51 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
     assert volume_file_md5sum2 == original_md5sum2
 
 
-@pytest.mark.skip(reason="TODO") # NOQA
-def test_auto_upgrade_engine_to_default_version():
+def prepare_auto_upgrade_engine_to_default_version(client): # NOQA
+    default_img = common.get_default_engine_image(client)
+    default_img_name = default_img.name
+    default_img = wait_for_engine_image_ref_count(client, default_img_name, 0)
+    cli_v = default_img.cliAPIVersion
+    cli_minv = default_img.cliAPIMinVersion
+    ctl_v = default_img.controllerAPIVersion
+    ctl_minv = default_img.controllerAPIMinVersion
+    data_v = default_img.dataFormatVersion
+    data_minv = default_img.dataFormatMinVersion
+    engine_upgrade_image = common.get_upgrade_test_image(cli_v, cli_minv,
+                                                         ctl_v, ctl_minv,
+                                                         data_v, data_minv)
+
+    compatible_img = client.create_engine_image(image=engine_upgrade_image)
+    compatible_img_name = compatible_img.name
+    ei_status_value = get_engine_image_status_value(client,
+                                                    compatible_img_name)
+    compatible_img = wait_for_engine_image_state(client,
+                                                 compatible_img_name,
+                                                 ei_status_value)
+    assert compatible_img.refCount == 0
+    assert compatible_img.noRefSince != ""
+
+    return default_img, default_img_name, engine_upgrade_image, \
+        compatible_img, compatible_img_name
+
+
+def check_replica_engine(volume, engineimage):
+
+    for replica in volume.replicas:
+        if volume["state"] == "attached":
+            if volume["robustness"] != "degraded":
+                assert replica.running is True
+
+            if replica.running is True:
+                assert replica.engineImage == engineimage
+                assert replica.currentImage == engineimage
+        elif volume["state"] == "detached":
+            assert replica.running is False
+            assert replica.engineImage == engineimage
+            assert replica.currentImage == ""
+
+
+def test_auto_upgrade_engine_to_default_version(client): # NOQA
     """
     Steps:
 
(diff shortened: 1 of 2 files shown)


c3y1huang (commented):
Suggest breaking up the test cases.

chriscchien:
> Suggest breaking up the test cases.

Hi @c3y1huang , I broke up test case into separate test cases, thank you

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -747,61 +749,261 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+    assert compatible_img.noRefSince != ""
+
+    return default_img, default_img_name, engine_upgrade_image, \
+        compatible_img, compatible_img_name
+
+
+def check_replica_engine(volume, engineimage):
+    # Chekout attached volume's replica engine
+    for replica in volume.replicas:
+        if replica.running is True:
```
c3y1huang, line 783: Should we also assert replica is running?

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -747,61 +749,261 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+    wait_for_engine_image_ref_count(client, compatible_img_name, 0)
+
+    wait_for_volume_healthy(client, volume1.name)
+    check_volume_data(volume1, data)
+
+    volume1 = client.by_id_volume(volume1.name)
+    volume2 = client.by_id_volume(volume2.name)
+    assert volume1.engineImage == default_img.image
+    assert volume2.engineImage == default_img.image
+    check_replica_engine(volume1, default_img.image)
```
c3y1huang, line 852: Why are we skipping the check for `volume2`?

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -747,61 +749,261 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+    dr_volume = client.create_volume(name=dr_volume_name, size=SIZE,
+                                     numberOfReplicas=3, fromBackup=b.url,
+                                     frontend="", standby=True)
+
+    wait_for_backup_restore_completed(client, dr_volume_name, b.name)
+
+    dr_volume = common.wait_for_volume_healthy_no_frontend(client,
+                                                           dr_volume_name)
+    update_setting(client,
+                   "concurrent-automatic-engine-upgrade-per-node-limit",
```
c3y1huang, line 905: nit: suggest putting this in the [common.py](https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/common.py#L146-L180).

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -747,61 +749,261 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+                   "3")
+
+    try:
+        dr_volume.engineUpgrade(image=engine_upgrade_image)
+        raise ("Engine upgrade should not accected when \
```
c3y1huang, line 910:
```suggestion
        raise ("Engine upgrade should fail when \
```

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -747,61 +749,261 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+    volume1 = common.wait_for_volume_detached(client, volume1.name)
+    volume1.engineUpgrade(image=engine_upgrade_image)
+    wait_for_engine_image_ref_count(client, default_img_name, 0)
+    volume1.expand(size=str(5 * Gi))
+
+    update_setting(client,
+                   "concurrent-automatic-engine-upgrade-per-node-limit",
+                   "3")
+
+    # check volume is expaning
```
c3y1huang, line 982:
```suggestion
    # check volume is expanding
```

c3y1huang (changes requested):

manager/integration/tests/test_engine_upgrade.py
chriscchien: Updated `check_replica_engine` to check attached, degraded and detached volume replica engine image

manager/integration/tests/test_engine_upgrade.py
chriscchien: Hi @c3y1huang , I added volume2 replica engine check

manager/integration/tests/test_engine_upgrade.py
chriscchien: Hi @c3y1huang , please to updated code, thank you

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -747,61 +750,274 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+                                                 compatible_img_name,
+                                                 ei_status_value)
+    assert compatible_img.refCount == 0
+    assert compatible_img.noRefSince != ""
+
+    return default_img, default_img_name, engine_upgrade_image, \
+        compatible_img, compatible_img_name
+
+
+def check_replica_engine(volume, engineimage):
```
c3y1huang, line 781: This function is a bit hard to read. Suggest having the for loop `for replica in volume.replicas` first, then check the conditions inside.

c3y1huang (changes requested):

manager/integration/tests/test_engine_upgrade.py
chriscchien: Hi @c3y1huang , please see updated code, thank you

manager/integration/tests/test_engine_upgrade.py
```diff
@@ -747,61 +750,272 @@ def test_engine_live_upgrade_with_intensive_data_writing(client, core_api, volum
...
+    for replica in volume.replicas:
+        if volume["state"] == "attached":
+            if volume["robustness"] == "degraded":
+                if replica.running is True:
+                    assert replica.engineImage == engineimage
+                    assert replica.currentImage == engineimage
+            else:
+                assert replica.running is True
+                assert replica.engineImage == engineimage
+                assert replica.currentImage == engineimage
```
c3y1huang, line 792: How about this?
```suggestion
            if volume["robustness"] != "degraded":
                assert replica.running is True

            if replica.running is True:
                assert replica.engineImage == engineimage
                assert replica.currentImage == engineimage
```

chriscchien: Hi @c3y1huang , thanks for the suggestion, it's very helpful to me

c3y1huang (approved):

PR #1179, opened 2022-11-14
test: implement 2 stage upgrade

yangchiu:
test: implement 2 stage upgrade

For https://github.com/longhorn/longhorn/issues/4817


khushboo-rancher (dismissed):

manager/integration/scripts/upgrade-longhorn.sh
```diff
@@ -9,6 +11,8 @@ CUSTOM_LONGHORN_SHARE_MANAGER_IMAGE="${6}"
 CUSTOM_LONGHORN_BACKING_IMAGE_MANAGER_IMAGE="${7}"
 
 LONGHORN_REPO_DIR="/tmp/longhorn"
+mkdir -p "${LONGHORN_REPO_DIR}"
+rm -rf "${LONGHORN_REPO_DIR}/"
```
c3y1huang, line 15: What is the purpose of `mkdir` then `rm` immediately?

yangchiu: remove existing things in /tmp/longhorn since upgrade-longhorn.sh will be executed more than once.

c3y1huang: Got it. Thanks for clarifying. Suggest putting a comment here.

c3y1huang: btw, wouldn't this also remove the subdirectory `longhorn`? why not use `rm -rf "${LONGHORN_REPO_DIR}/*"`?

yangchiu: Yes if `longhorn` folder not removed, `git clone` will complain `fatal: destination path 'tmp/longhorn' already exists and is not an empty directory.`, so probably the correct way is to remove the `mkdir` part and let `git clone` to create the path for us, and we just need to ensure the existing things are all removed.

I've removed the `mkdir` part and still kept the `rm -rf` part.

manager/integration/tests/test_upgrade.py
```diff
@@ -1,3 +1,5 @@
+import os
+
```
c3y1huang, line 2: Do we need the blank line?

test_framework/scripts/longhorn-setup.sh
```diff
@@ -399,16 +398,37 @@ main(){
...
+    LONGHORN_UPGRADE_TYPE="from_stable"
+    LONGHORN_UPGRADE_TEST_POD_NAME="longhorn-test-upgrade-from-stable"
+    if [[ -n "${LONGHORN_TRANSIENT_VERSION}" ]]; then
+      UPGRADE_LH_REPO_URL="${LONGHORN_REPO_URI}"
+      UPGRADE_LH_REPO_BRANCH="${LONGHORN_TRANSIENT_VERSION}"
+      UPGRADE_LH_MANAGER_IMAGE=""
+      UPGRADE_LH_ENGINE_IMAGE="longhornio/longhorn-engine:${LONGHORN_TRANSIENT_VERSION}"
+      UPGRADE_LH_INSTANCE_MANAGER_IMAGE=""
+      UPGRADE_LH_SHARE_MANAGER_IMAGE=""
+      UPGRADE_LH_BACKING_IMAGE_MANAGER_IMAGE=""
```
c3y1huang, line 414: Do we need to fill in the empty value here?
```bash
      UPGRADE_LH_MANAGER_IMAGE=""
      UPGRADE_LH_INSTANCE_MANAGER_IMAGE=""
      UPGRADE_LH_SHARE_MANAGER_IMAGE=""
      UPGRADE_LH_BACKING_IMAGE_MANAGER_IMAGE=""
```

manager/integration/tests/test_upgrade.py
yangchiu: modified!

test_framework/scripts/longhorn-setup.sh
c3y1huang: IMO keeping only `${LONGHORN_TRANSIENT_VERSION}` specific variable assignments is easier to read. But that is just my 2 cents.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -254,27 +254,25 @@ create_aws_secret(){
 
 
 run_longhorn_upgrade_test(){
-  LONGHORH_TESTS_REPO_BASEDIR=${1}
```
innobead, line 257: Unnecessary empty line.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -399,16 +398,37 @@ main(){
       get_rancher_api_key
       install_longhorn_by_rancher
     fi
-    run_longhorn_tests ${WORKSPACE}
+    run_longhorn_tests
   elif [[ "${LONGHORN_UPGRADE_TEST}" == true || "${LONGHORN_UPGRADE_TEST}" == True ]]; then
```
innobead, line 402: Why there are two same conditions? `elif [[ "${LONGHORN_UPGRADE_TEST}" == true || "${LONGHORN_UPGRADE_TEST}" == True ]]; then`

test_framework/scripts/longhorn-setup.sh
innobead, line 414: Assume we will just use versions from the corresponding branch?

innobead (changes requested):
In general, LGTM. Just minor things needed to resolve.

test_framework/scripts/longhorn-setup.sh
yangchiu: refined!

test_framework/scripts/longhorn-setup.sh
yangchiu: removed!

test_framework/scripts/longhorn-setup.sh
yangchiu: removed the duplicated one.

test_framework/scripts/longhorn-setup.sh
yangchiu: Yes, for a 2 stage upgrade, the upgrade path will always be release version A -> release version B -> latest or custom version.
Release version B will always be directly from a branch (tag).

Unless we want to upgrade from release version A, to a custom version, and then to another custom version.

innobead (approved):

PR #1182, opened 2022-11-16
test: running different versions of longhorn in the same benchmark test build for comparison

yangchiu:
test: running different versions of longhorn in the same benchmark test build for comparison

For https://github.com/longhorn/longhorn/issues/4259


innobead (approved):
LGTM

PR #1183, opened 2022-11-18
Add test case test_backing_image_cleanup

chriscchien:

Add test case `test_backing_image_cleanup`


innobead (approved):
LGTM

PR #1187, opened 2022-11-21
test(auto): support bundle enhancement

c3y1huang:
https://github.com/longhorn/longhorn/issues/4908
https://github.com/longhorn/longhorn/issues/2759

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -180,6 +180,7 @@ SETTING_ORPHAN_AUTO_DELETION = "orphan-auto-deletion"
 SETTING_FAILED_BACKUP_TTL = "failed-backup-ttl"
 SETTING_CONCURRENT_AUTO_ENGINE_UPGRADE_NODE_LIMIT = \
     "concurrent-automatic-engine-upgrade-per-node-limit"
+SETTING_SUPPORT_BUNDLE_FAILED_LIMIT = "support-bundle-failed-history-limit"
 
 CSI_UNKNOWN = 0
 CSI_TRUE = 1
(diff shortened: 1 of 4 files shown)


manager/integration/tests/test_support_bundle.py
```diff
@@ -0,0 +1,103 @@
...
+
+    When create support bundle
+    Then should fail to create
+
+    """
+    update_setting(client, SETTING_SUPPORT_BUNDLE_FAILED_LIMIT, "3")
+    create_failed_support_bundles(client, apps_api, number=2)
+
+    try:
+        create_failed_support_bundles(client, apps_api, number=1)
```
innobead, line 84: Why not just make `create_failed_support_bundles` return `True`, then assert it w/ message instead of having another `assert False, ....`.

innobead (changes requested):
In general, LGTM. Just a few comments needed to resolve.

manager/integration/tests/test_support_bundle.py
c3y1huang: Not sure if I understand the question correctly. The `assert False` is to assert Longhorn is prohibiting the creation when the limit reaches the upper bound. The test is expecting `create_failed_support_bundles` to fail by its own assertions and be caught by the `except Expection`.

manager/integration/tests/common.py
```diff
@@ -3094,6 +3097,9 @@ def reset_settings(client):
         if setting_name == "storage-network":
             continue
 
+        if setting_name == "support-bundle-manager-image":
```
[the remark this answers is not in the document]
c3y1huang: Updated.

manager/integration/tests/test_support_bundle.py
innobead: You can do below, if the function just returns bool instead of having another statement.

assert create_failed_support_bundles, message

innobead: Ah u are right. All good.

innobead (approved):

PR #1189, opened 2022-11-23
Enhance test_offline_node

roger-ryao:
verify aws node status, before checking k8s node status

add retry logic for checking aws and k8s node status

https://github.com/longhorn/longhorn/issues/4921

### DIFF
--- a/manager/integration/tests/test_infra.py
+++ b/manager/integration/tests/test_infra.py
@@ -59,7 +59,7 @@ def wait_for_node_down_k8s(node_name, k8s_api_client):
             node_down = True
             break
         else:
-            time.sleep(RETRY_INTERVAL)
+            time.sleep(RETRY_INTERVAL_LONG)
             continue
     return node_down
 
@@ -109,7 +109,7 @@ def wait_for_node_down_aws(cloudprovider, node):
             aws_node_down = True
             break
         else:
-            time.sleep(RETRY_INTERVAL)
+            time.sleep(RETRY_INTERVAL_LONG)
             continue
 
     return aws_node_down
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_infra.py
```diff
@@ -239,9 +239,22 @@ def test_offline_node(reset_cluster_ready_status):
...
     cloudprovider.instance_stop(node)
-    wait_for_node_down_aws(cloudprovider, node)
-    k8s_node_down = wait_for_node_down_k8s(node_name, k8s_api_client)
+    aws_node_down = wait_for_node_down_aws(cloudprovider, node)
+    if aws_node_down:
+        print(f'AWS node is down in {RETRY_COUNTS} sec')
+    else:
+        print(f'AWS node not down in {RETRY_COUNTS} sec')
+        print(f'Check AWS node status again in {RETRY_COUNTS} sec')
+        aws_node_down = wait_for_node_down_aws(cloudprovider, node)
```
c3y1huang, line 248: Seems the instance just needs more time to stop? Why not just change `RETRY_INTERVAL` to `RETRY_INTERVAL_LONG`?

roger-ryao: I was set to print some information on console output, I didn’t realize that we have `RETRY_INTERVAL_LONG` in the common.py. I would modify my PR.

roger-ryao:
@c3y1huang 
I modify my PR, please review it again.
Thanks.

c3y1huang:
Does the fix also apply for master branch?

roger-ryao:
> Does the fix also apply for master branch?
Hi @c3y1huang 
   Sure. Before applying for the master branch,  I want to track the 1.2.x status first around 1-2 weeks.

innobead (approved):

innobead:
This is quite a general change, so suggest just backporting to the master and 1.3 in case this commit was missed.

PR #1190, opened 2022-11-23
Add skeleton of test_upgrade_with_auto_upgrade_latest_engine_enabled

chriscchien:

Ref:
    https://github.com/longhorn/longhorn/issues/4726
    https://github.com/longhorn/longhorn/issues/4730


c3y1huang:
Suggest removing the `()` in the commit message and adding a reference to the issue so its easier to track.

khushboo-rancher (commented):
Is it not possible to incorporate this in the existing upgrade test?

chriscchien:
Hi @khushboo-rancher ,
Yes I will add those steps into the exist case, skeleton were separated was due to concern about easily reading

c3y1huang (approved):

PR #1198, opened 2022-11-25
Add engine auto upgrade step into upgrade test

chriscchien:
Ref
- [Longhorn-4726](https://github.com/longhorn/longhorn/issues/4726)
- [Longhorn-4730
](https://github.com/longhorn/longhorn/issues/4730)

Verified in below upgrade path and the upgrade/check engine part passed 
- v1.3.2 to master-head
- v1.2.6 to v1.3.2 then master-head

### DIFF
--- a/manager/integration/tests/test_upgrade.py
+++ b/manager/integration/tests/test_upgrade.py
@@ -34,6 +34,9 @@ from common import write_pod_volume_data
 from common import read_volume_data
 from common import settings_reset # NOQA
 from common import client, core_api # NOQA
+from common import update_setting, get_engine_image_status_value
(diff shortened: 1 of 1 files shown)


yangchiu:
this PR overwrites the original upgrade test case, so the normal manual upgrade path will not be tested after this PR, right?

chriscchien:
Yes it override the normal upgrade case

c3y1huang:
> Yes it override the normal upgrade case

Currently, we are using the same case to `test_upgrade`, this makes it impossible to adapt to different scenarios.
/cc @innobead

PR #1204, opened 2022-11-29
test(manual): engine binary recovery

c3y1huang:
https://github.com/longhorn/longhorn/issues/4380


c3y1huang:
This is tested by @longhorn/qa , ref: https://github.com/longhorn/longhorn/issues/4380#issuecomment-1334689982

innobead (approved):
Both are the same cases because deleting the binary from the pod which is actually just a host mount.

PR #1207, opened 2022-11-29
Update single replica node down test case

roger-ryao:
Make test steps more clear

ref: https://github.com/longhorn/longhorn/issues/4793

### DIFF
--- a/docs/content/manual/pre-release/ha/single-replica-node-down.md
+++ b/docs/content/manual/pre-release/ha/single-replica-node-down.md
@@ -13,7 +13,7 @@ https://github.com/longhorn/longhorn/issues/2309
 1. Create RWO|RWX volume with replica count = 1 & data locality = enabled|disabled.
 2. Create deployment|statefulset for volume.
 3. Power down node of volume/replica.
-4. The workload pod will get stuck in the `unknown` state.
+4. The workload pod will get stuck in the `terminating` state.
 5. Volume will fail to attach since volume is not ready (i.e remains faulted, since single replica is on downed node).
 6. Power up node or delete the workload pod so that kubernetes will recreate pod on another node.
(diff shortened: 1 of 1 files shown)


docs/content/manual/pre-release/ha/single-replica-node-down.md
```diff
@@ -24,7 +24,9 @@ https://github.com/longhorn/longhorn/issues/2309
 2. Create deployment|statefulset for volume.
 3. Power down node of volume/replica.
 4. Volume will become faulted.
-5. Wait for pod deletion & recreation on another node. For RWX volume power on the node. 
-6. Verify the auto salvage finishes for volumes.
-7. Verify volume attached & accessible by pod (i.e test data is available).
+5. Wait for 7 min for pod deletion & recreation on another node.
```
c3y1huang, line 27: Curious, why 7 minutes?

roger-ryao: After turning off the node of volume/replica, I observed that 7 minutes later the pod deletion & recreation would start. I just update this information on the test steps.

c3y1huang: Got it, thank you for the clarification.

7 minutes is very specific and a little misleading. Could you add a comment mentioning that the pod recreation will not happen immediately?

innobead: This should be related to `--pod-eviction-timeout` of kube control manager. You can mention that, but no need to have a specific number here. As Chinya said, it will be a little confusing. 

Also, why removing `For RWX volume power on the node`?

innobead: Sounds good :+1:

innobead (approved):

PR #1193, opened 2022-11-24
Reorder cleanup_client

chriscchien:
Ref: longhorn-test PR 1181

Avoid teardown error when a volume attached to another disk volume

Signed-off-by: Chris Chien <chris.chien@suse.com>

PR #1209, opened 2022-11-29
Modify test_replica_schedule_to_disk_with_most_usable_storage

chriscchien:
To avoid volume not clean up clearly after longhorn-test [PR1193](https://github.com/longhorn/longhorn-tests/pull/1193) merged

Ref: [4957](https://github.com/longhorn/longhorn/issues/4957)

Signed-off-by: Chris Chien chris.chien@suse.com

### DIFF
--- a/manager/integration/tests/test_scheduling.py
+++ b/manager/integration/tests/test_scheduling.py
@@ -1445,11 +1445,6 @@ def test_replica_schedule_to_disk_with_most_usable_storage(client, volume_name, 
     Then volume replica     on the current node scheduled to disk-1.
          volume replica not on the current node scheduled to default disk.
     """
-    # Remove volumes let no volume mounted to extra disk
-    def finalizer():
-        common.cleanup_all_volumes(client)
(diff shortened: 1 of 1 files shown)


chriscchien:
Here is test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2586/) by command `pytest -k disk` , no teardown error observed

chriscchien:
[Here](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2587/) is full run of regression, it still inprogress

innobead (approved):
LGTM

As discussed, let's backport to 1.2 and 1.3 as well, because this is quite common. Also, create an issue to do the following refactoring, because just relying on the order to delete volumes to avoid this race happen, this is very implicit, and people will have no ideas why we need to use `vol-...` volume name..

chriscchien:
I will create  backport PR after local test on v1.2.x and v1.3.x complete, thank you

innobead:
Thank you @chriscchien :+1:

PR #1212, opened 2022-11-30
test: fix upgrade test failed due to list_support_bundle api error

yangchiu:
test: fix upgrade test failed due to list_support_bundle api error

For https://github.com/longhorn/longhorn/issues/4970

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4992,6 +4992,12 @@ def cleanup_all_support_bundles(client):
     Clean up all support bundles
     :param client: The Longhorn client to use in the request.
     """
+    longhorn_version = client.by_id_setting('current-longhorn-version').value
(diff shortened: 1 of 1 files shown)


c3y1huang (approved):

manager/integration/tests/common.py
```diff
@@ -4992,6 +4992,17 @@ def cleanup_all_support_bundles(client):
     Clean up all support bundles
     :param client: The Longhorn client to use in the request.
     """
+    ret = get_core_api_client().list_pod_for_all_namespaces(
```
innobead, line 4995: Can you get `current-longhorn-version` setting instead of figuring out the version from LH pods?

yangchiu: modified!

innobead (approved):

PR #1215, opened 2022-12-02
fix flaky test_setting_concurrent_rebuild_limit

chriscchien:
Ref: longhorn-[4977](https://github.com/longhorn/longhorn/issues/4977)

1. Extend rebuild wait to 10 minutes or test case sometimes faced rebuild timeout
2. Remove test step 1-10. Because by test scenario `ConcurrentReplicaRebuildPerNodeLimit`=1 volume1 is rebuilding and volume2 degraded, crash replica process in volume 1 can not guarantee volume2 will be the next one to rebuild, sometime volume1 will start rebuild again and script expect the rebuild 1 was volume2, so test case failed(flaky). For ConcurrentReplicaRebuildPerNodeLimit`=1 related test already tested in step 1-4, so I removed the flaky part for review, thank you.

10 times test on pipeline all [passed](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2630/)

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -755,61 +755,38 @@ def test_setting_backing_image_auto_cleanup(client, core_api, volume_name):  # N
     check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 1, "ready")
 
 
-def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQA
+@pytest.mark.skip(reason="TODO")  # NOQA
+def test_setting_concurrent_rebuild_by_crashing_replica():
     """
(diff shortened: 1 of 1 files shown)


innobead (changes requested):
I think the test case is still valid, but what we should do is to improve test case implementation instead.  

Basically, the test case wants the below two statements to run concurrently, but right now they run sequentially because `crash_replica_processes` is blocking. So, suggest using `wait_to_fail=False` when calling the below function to make `crash_replica_processes` run asynchronous instead of waiting for the replica process to fail first. 

`def crash_replica_processes(client, api, volname, replicas=None wait_to_fail=True):`

chriscchien:
This test case has passed in continues 7 days on test pipeline, looks like it is not flaky anymore, close this PR.

PR #1217, opened 2022-12-05
doc : Update manual test Single replica node down for strict-local volume

chriscchien:
Update manual test Single replica node down for strict-local volume in section Node restart/down scenario with `Pod Deletion Policy When Node is Down` set to `delete-both-statefulset-and-deployment-pod`

Ref - longhorn-[4967](https://github.com/longhorn/longhorn/issues/4967)

### DIFF
--- a/docs/content/manual/pre-release/ha/single-replica-node-down.md
+++ b/docs/content/manual/pre-release/ha/single-replica-node-down.md
@@ -5,12 +5,14 @@ title: Single replica node down
 ## Related Issues
 https://github.com/longhorn/longhorn/issues/2329
 https://github.com/longhorn/longhorn/issues/2309
+https://github.com/longhorn/longhorn/issues/3957
 
 ## Default Setting
 `Automatic salvage` is enabled.
 
 ## Node restart/down scenario with `Pod Deletion Policy When Node is Down` set to default value `do-nothing`.
-1. Create RWO|RWX volume with replica count = 1 & data locality = enabled|disabled.
+1. Create RWO|RWX volume with replica count = 1 & data locality = enabled|disabled|strict-local.
+   - For data locality = strict-local, use RWO volume to do test.
 2. Create deployment|statefulset for volume.
 3. Power down node of volume/replica.
 4. The workload pod will get stuck in the `terminating` state.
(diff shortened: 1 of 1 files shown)


docs/content/manual/pre-release/ha/single-replica-node-down.md
```diff
@@ -20,7 +21,7 @@ https://github.com/longhorn/longhorn/issues/2309
 8. Verify volume attached & accessible by pod (i.e test data is available).
 
 ## Node restart/down scenario with `Pod Deletion Policy When Node is Down` set to `delete-both-statefulset-and-deployment-pod`
-1. Create RWO|RWX volume with replica count = 1 & data locality = enabled|disabled.
+1. Create RWO|RWX volume with replica count = 1 & data locality = enabled|disabled|strict-local.
```
innobead, line 24: @derekbit does local volume support RWX? It seems we don't have a specific limitation, but however that mode is specifically for RWO.

innobead: Need to add strict-local to the first test case as well.

docs/content/manual/pre-release/ha/single-replica-node-down.md
```diff
@@ -30,4 +31,4 @@ https://github.com/longhorn/longhorn/issues/2309
 7. Power on node of volume/replica.
 8. Verify the auto salvage finishes for volumes.
 9. Verify volume attached & accessible by pod (i.e test data is available).
- 
\ No newline at end of file
+   - For volume which data locality = strict-local, verify that volume/replica/pod still work on original node.
```
innobead, line 34: If data locality is not strict-local, the volume/replica will run on another node, not the original node?

innobead (changes requested):

docs/content/manual/pre-release/ha/single-replica-node-down.md
chriscchien: Sorry this part is redundant, I removed that.

docs/content/manual/pre-release/ha/single-replica-node-down.md
derekbit: > does local volume support RWX? It seems we don't have a specific limitation, but however that mode is specifically for RWO.

Yes. It is supported.

khushboo-rancher (approved):
LGTM

docs/content/manual/pre-release/ha/single-replica-node-down.md
innobead: The reason we only test for RWO, because testing the volume with data-locality-strict does not make sense for RWX, because the volume will be just next to the share manager pod instead of the user workload pod.

PR #1218, opened 2022-12-06
ci: remove rke related logic from setup scripts

yangchiu:
ci: remove rke related logic from setup scripts

For https://github.com/longhorn/longhorn/issues/4990


innobead:
Let's close this first since rke1 is still supported. Continue this when it's needed.

PR #1222, opened 2022-12-12
ci: run CI for release and master branches only

innobead:
Also, exclude s390x PR check

ref: https://github.com/longhorn/longhorn/issues/5040

Signed-off-by: David Ko <dko@suse.com>


yangchiu:
@Mergify backport v1.3.x

PR #1223, opened 2022-12-12
Update tests for online expansion

shuo-wu:
longhorn/longhorn#1674

### DIFF
--- a/docs/content/manual/release-specific/v1.4.0/test-online-expansion.md
+++ b/docs/content/manual/release-specific/v1.4.0/test-online-expansion.md
@@ -1,0 +1,95 @@
+---
+title: Test Online Expansion
+---
+
+## Related issue
+https://github.com/longhorn/longhorn/issues/1674
+
+## Test online expansion with continuous reading/writing
+
+**Given** Prepare a relatively large file (5Gi for example) with the checksum calculated.
+
+*And* Create and attach a volume.
+
+*And* Monitor the instance manager pod logs.
+
+**When** Use `dd` to copy data from the file to the Longhorn block device.
+
+```
+dd if=/mnt/data of=/dev/longhorn/vol bs=1M
+```
+
+*And* Do online expansion for the volume during the copying.
+
+**Then** The expansion should success. The corresponding block device on the attached node is expanded.
+
+*And* Should not see any io or tgt related error logs in the instance manager pods.
+
+**When** Use `dd` to calculate the data checksum for the Longhorn block device..
+
+```
+dd if=/dev/longhorn/vol bs=1M count=5000 status=none | md5sum
+```
+
+*And* Do online expansion for the volume during the calculation.
+
+**Then** The expansion should success. The corresponding block device on the attached node is expanded.
+
+*And* Should not see any io or tgt related error logs in the instance manager pods.
+
+*And* The checksum matches.
+
+
+## Test online expansion with upgrade
+
+**Given** Deploy an old version Longhorn, v1.3.2 for example.
+
+*And* Create and attach 2 volumes.
+
+*And* Upgrade Longhorn to the latest version.
+
+**When** Do online expansion for volume 1.
+
+**Then** The expansion of volume 1 should get stuck. (old engine image + old instance manager)
+
+*And* The engine CR of volume 1 should contain the expansion error. e.g.,
+```yaml
+status:
+  lastExpansionError: cannot do online expansion for the old engine vol-e-b2d9b924
+  with cli API version 6
+  lastExpansionFailedAt: 2022-12-29 02:44:15.481809613 +0000 UTC m=+146155.850227034
+```
+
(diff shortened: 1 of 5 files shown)


shuo-wu:
Updated with manual tests

innobead (commented):
The manula test is good, and alos we shuould mention this note to the doc of online volume expansion about that it will only take effect after the running volume gets reattached.

manager/integration/tests/test_basic.py
```diff
@@ -2476,52 +2525,83 @@ def test_expansion_canceling(client, core_api, volume_name, pod):  # NOQA
...
+    resp = read_volume_data(core_api, pod_name, "test_file1")
+    assert resp == test_data1
+    resp = read_volume_data(core_api, pod_name, "test_file2")
+    assert resp == test_data2
+
+    # Without a short wait, the volume replicas consider that the directory
+    # created in func fix_replica_expansion_failure still exists then fail
+    # the following expansion request.
+    # Not sure if it's caused by the caching.
+    time.sleep(5)
```
c3y1huang, line 2586: Looks like an improvement in stability, should we create an issue for this?

c3y1huang (changes requested):

docs/content/manual/release-specific/v1.4.0/test-online-expansion.md
```diff
@@ -0,0 +1,89 @@
...
+
+**When** Use `dd` to calculate the data checksum for the Longhorn block device..
+
+```
+dd if=/dev/longhorn/vol bs=1M count=5000 status=none | md5sum
+```
+
+*And* Do online expansion for the volume during the calculation.
+
+**Then** The expansion should success.
```
khushboo-rancher, line 36: Can we elaborate this point as something like this `The expansion should succeed and the size of the volume needs to be verified on the attached node.`

manager/integration/tests/test_csi.py
```diff
@@ -376,6 +371,54 @@ def test_csi_offline_expansion(client, core_api, storage_class, pvc, pod_manifes
...
+       and Check the PVC & PV size
+    7. Wait for the copy complete.
+    8. Calculate the checksum for the copied data inside the block volume
+       asynchronously.
+    9. During the calculation, update pvc.spec.resources to expand the volume
+       again.
+    10. Wait for the calculation complete, then compare the checksum.
+    11. Do cleanup: Remove the original `test_data`as well as the pod and PVC.
+    """
+    pass
```
khushboo-rancher, line 395: Can we skip the test instead of passing it?

khushboo-rancher (changes requested):

docs/content/manual/release-specific/v1.4.0/test-online-expansion.md
shuo-wu: Sure. Let me add one more statement: `The corresponding block device on the attached node is expanded.`

manager/integration/tests/test_basic.py
shuo-wu: Not sure. I spent almost one day resolving this but no luck.

shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2817/

manager/integration/tests/test_basic.py
shuo-wu: The culprit of this wait: https://github.com/longhorn/longhorn/issues/5169#issuecomment-1367202468
Will fix the issue and remove this wait

docs/content/manual/release-specific/v1.4.0/test-online-expansion.md
```diff
@@ -0,0 +1,95 @@
...
+
+**When** Directly detach and reattach volume 2.
+
+*And* Do online expansion for volume 2.
+
+**Then** The expansion should get stuck. (old engine image + new instance manager)
+
+*And* The engine CR of volume 2 should contain the expansion error.
+
+*And* Then expansion can be cancelled.
```
c3y1huang, line 89:
```suggestion
*And* expansion can be canceled.
```

docs/content/manual/release-specific/v1.4.0/test-online-expansion.md
```diff
@@ -0,0 +1,95 @@
...
+
+**When** Do live upgrade for volume 1.
+
+*And* Retry online expansion.
+
+**Then** The expansion should get stuck. (new engine image + old instance manager)
+
+*And* The engine CR should contain the expansion error.
+
+*And* Then expansion can be cancelled.
```
c3y1huang, line 73:
```suggestion
*And* expansion can be canceled.
```

c3y1huang (changes requested):

khushboo-rancher (approved):
LGTM, Thanks @shuo-wu

c3y1huang (approved):

PR #1224, opened 2022-12-13

derekbit:
https://github.com/longhorn/longhorn/issues/4210
https://github.com/longhorn/longhorn/issues/3198

Signed-off-by: Derek Su [derek.su@suse.com](mailto:derek.su@suse.com)


derekbit:
Will create a new PR later.

PR #1228, opened 2022-12-23
test(Manual): Stopped replicas on deleted nodes should not be counted as healthy replicas when draining nodes

weizhe0422:
[Longhorn 2237](https://github.com/longhorn/longhorn/issues/2237)

### DIFF
--- a/docs/content/manual/pre-release/node/node-drain-deletion.md
+++ b/docs/content/manual/pre-release/node/node-drain-deletion.md
@@ -1,7 +1,7 @@
 ---
 title: Node drain and deletion test
 ---
-# Drain with force
+## Drain with force
 Make sure the volumes on the drained/removed node can be detached or recovered correctly. The related issue: https://github.com/longhorn/longhorn/issues/1214
 1. Deploy a cluster contains 3 worker nodes N1, N2, N3.
 2. Deploy Longhorn.
@@ -20,7 +20,7 @@ Make sure the volumes on the drained/removed node can be detached or recovered c
 9. Wait for the volume detaching then being recovered. Will get attached to the workload/node.
 10. Validate the volume content. The data is intact.
 
-# Drain without force
+## Drain without force
 1. Cordon the node. Longhorn will automatically disable the node scheduling when a Kubernetes node is cordoned.
 2. Evict all the replicas from the node.
 3. Run the following command to drain the node with ```force``` flag set to false.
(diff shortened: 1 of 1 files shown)


docs/content/manual/pre-release/node/node-drain-deletion.md
```diff
@@ -65,3 +65,39 @@ NOTE: We might still 1 or 2 times of the pod's disruption budget message, but wi
...
+4. The node become cordoned. All pods on the cordoned node are evicted except the replica instance manager pod. And the message like below keeps appearing.
+    ```
+    evicting pod longhorn-system/instance-manager-r-xxxxxxxx
+    error when evicting pods/"instance-manager-r-xxxxxxxx" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
+    ```
+5. Enable "Allow Node Drain with the Last Healthy Replica" and check return **true** with following command:
+    ```shell
+    kubectl get settings.longhorn.io/allow-node-drain-with-last-healthy-replica -n longhorn-system
+    ```
+6. The pod `longhorn-system/instance-manager-r-xxxxxxxx` will be evicted successfully and use the following command to ensure only daemonset pods such as `engine-image`, `longhorn-csi-plugin` and `longhorn-manager` are running on the nodee:
```
c3y1huang, line 100:
```suggestion
6. The pod `longhorn-system/instance-manager-r-xxxxxxxx` will be evicted successfully and use the following command to ensure only `engine-image`, `longhorn-csi-plugin` and `longhorn-manager` daemonset pods are running on Node_2:
```

docs/content/manual/pre-release/node/node-drain-deletion.md
```diff
@@ -65,3 +65,39 @@ NOTE: We might still 1 or 2 times of the pod's disruption budget message, but wi
...
+ - Start with 2 nodes cluster: Node_1, Node_2
+ - Double check that "Allow Node Drain with the Last Healthy Replica" is unchecked or return **false** with following command:
+    - `kubectl get settings.longhorn.io/allow-node-drain-with-last-healthy-replica -n longhorn-system`
+
+##### Steps:
+1. Create a 5Gi detached volume with 2 replicas.
+2. Stop Node_1 that contains one of the replicas.
+3. Attempts to drain Node_2 that contains remaining replica.
+    ```shell
+    kubectl drain <Node name> --delete-emptydir-data=true --force=true --grace-period=-1 --ignore-daemonsets=true --timeout=<Desired timeout in secs>
```
c3y1huang, line 89: Can you clarify why we need `--timeout` ? Does it make a difference with and without setting this flag?

docs/content/manual/pre-release/node/node-drain-deletion.md
```diff
@@ -65,3 +65,39 @@ NOTE: We might still 1 or 2 times of the pod's disruption budget message, but wi
...
+    ```
+4. The node become cordoned. All pods on the cordoned node are evicted except the replica instance manager pod. And the message like below keeps appearing.
+    ```
+    evicting pod longhorn-system/instance-manager-r-xxxxxxxx
+    error when evicting pods/"instance-manager-r-xxxxxxxx" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
+    ```
+5. Enable "Allow Node Drain with the Last Healthy Replica" and check return **true** with following command:
```
c3y1huang, line 96: nit: looks like there are 2 cases you are testing. Suggest breaking them up.
- Stopped replicas on deleted nodes should not be counted as healthy replicas when draining nodes
- Setting `Allow Node Drain with the Last Healthy Replica`

c3y1huang (changes requested):

c3y1huang (commented):
Suggest practicing writing test cases in BDD style to make the case easier to understand of the preparation, testing, and expectation.

docs/content/manual/pre-release/node/node-drain-deletion.md
weizhe0422: No, there is no different result between with and without parameter `timeout`.

Since after the node_2 is drained in step 3, it keeps giving the message that replica cannot be evicted as mentioned in step 4, I pass the `timeout` to indicate length of waiting time.

I figured I'd remove the argument to avoid confusion.

weizhe0422: No, there is no different result between with and without parameter `timeout`.

Since after the node_2 is drained in step 3, it keeps giving the message that replica cannot be evicted as mentioned in step 4, I pass the `timeout` to indicate length of waiting time.

I figured I'd remove the argument to avoid confusion.

weizhe0422:
Hi @c3y1huang, thanks for your suggestion. 

I have tried to rewrite the manual test case in BDD style and split the test case into two scenarios.

khushboo-rancher (commented):
LGTM in general, left some minor comments.

docs/content/manual/pre-release/node/node-drain-deletion.md
```diff
@@ -65,3 +65,79 @@ NOTE: We might still 1 or 2 times of the pod's disruption budget message, but wi
...
+```shell
+kubectl get pods --field-selector spec.nodeName=<Node_2 name> -o wide -n longhorn-system
+```
+
+*And* The message like below keeps appearing.
+```
+evicting pod longhorn-system/instance-manager-r-xxxxxxxx
+error when evicting pods/"instance-manager-r-xxxxxxxx" -n "longhorn-system" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
+```
+
```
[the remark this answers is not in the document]
weizhe0422: Updated by adding statement description. Thank you.

c3y1huang (approved):

PR #1230, opened 2022-12-27

derekbit:
Longhorn/longhorn#4210
Longhorn/longhorn#3198

### DIFF
--- a/manager/integration/tests/test_snapshot.py
+++ b/manager/integration/tests/test_snapshot.py
@@ -1,0 +1,594 @@
+import os
+import subprocess
+import time
+import datetime
+import json
+import pandas as pd
+
+from common import client, volume_name  # NOQA
+from common import update_setting
+from common import get_self_host_id
+from common import create_and_check_volume, wait_for_volume_healthy
+from common import wait_for_rebuild_start, wait_for_rebuild_complete
+from common import create_snapshot
+from common import settings_reset # NOQA
+
+from test_basic import get_volume_endpoint
+
+from test_scheduling import get_host_replica, wait_new_replica_ready  # NOQA
+
+from common import Gi, Mi
+from common import RETRY_COUNTS, RETRY_INTERVAL
+from common import SETTING_SNAPSHOT_DATA_INTEGRITY
+from common import SETTING_SNAPSHOT_DATA_INTEGRITY_IMMEDIATE_CHECK_AFTER_SNAPSHOT_CREATION  # NOQA
+from common import SETTING_SNAPSHOT_DATA_INTEGRITY_CRONJOB
+from common import SETTING_SNAPSHOT_FAST_REPLICA_REBUILD_ENABLED
+from common import SNAPSHOT_DATA_INTEGRITY_IGNORED
+from common import SNAPSHOT_DATA_INTEGRITY_ENABLED
+from common import SNAPSHOT_DATA_INTEGRITY_FAST_CHECK
+from common import SNAPSHOT_DATA_INTEGRITY_DISABLED
+
+RETRY_WAIT_CHECKSUM_COUNTS = 600
+SNAPSHOT_CHECK_PERIOD = 300
+
+# For the checksum calculated by periodic check mechanism
+# We cannot precisely know when the event happens,
+# so put a toleration here and check if the time elapsed is expected.
+SNAPSHOT_CHECK_TOLERATION_DELAY = 60
+
+def test_snapshot_hash_global_enabled_with_immediate_hash(client, volume_name, settings_reset):  # NOQA
+    """
+    Check snapshots' checksums are immediately calculated when the snapshots
+    are created
+    - global data-integrity is set to enabled
+    - global immediate_hash is enabled
+    """
+    prepare_settings_for_snapshot_test(client,
+                                       SNAPSHOT_DATA_INTEGRITY_ENABLED,
+                                       "true", "true")
+    check_hashed_and_with_immediate_hash(client, volume_name,
+                                         SNAPSHOT_DATA_INTEGRITY_IGNORED)
+
+
+def test_snapshot_hash_global_enabled_without_immediate_hash(client, volume_name, settings_reset):  # NOQA
+    """
+    Check snapshots' checksums are calculated by the periodic checksum check
+    - global data-integrity is set to enabled
+    - global immediate_hash is disabled
+    """
+    prepare_settings_for_snapshot_test(client,
+                                       SNAPSHOT_DATA_INTEGRITY_ENABLED,
+                                       "false", "true")
+    check_hashed_and_without_immediate_hash(client, volume_name,
+                                            SNAPSHOT_DATA_INTEGRITY_IGNORED)
+
+
+def test_snapshot_hash_global_fast_check_with_immediate_hash(client, volume_name, settings_reset):  # NOQA
+    """
+    Check snapshots' checksums are immediately calculated when the snapshots
+    are created
+    - global data-integrity is set to fast-check
+    - global immediate_hash is enabled
+    """
+    prepare_settings_for_snapshot_test(client,
+                                       SNAPSHOT_DATA_INTEGRITY_FAST_CHECK,
+                                       "true", "true")
+    check_hashed_and_with_immediate_hash(client,
+                                         volume_name,
+                                         SNAPSHOT_DATA_INTEGRITY_IGNORED)
+
+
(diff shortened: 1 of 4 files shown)


derekbit:
e2e
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2793/

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,491 @@
...
+from common import SNAPSHOT_DATA_INTEGRITY_FAST_CHECK
+from common import SNAPSHOT_DATA_INTEGRITY_DISABLED
+
+RETRY_WAIT_CHECKSUM_COUNTS = 600
+SNAPSHOT_CHECK_PERIOD = 300
+
+def test_snapshot_hash_global_enabled_with_immediate_hash(client, volume_name, settings_reset):  # NOQA
+    update_settings(client, SNAPSHOT_DATA_INTEGRITY_ENABLED, "true", "true")
+    check_hash_calculated_and_with_immediate_hash(client, volume_name,
+                                                  SNAPSHOT_DATA_INTEGRITY_IGNORED)  # NOQA
```
c3y1huang, line 37: Why do we need the `# NOQA` here?

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,491 @@
...
+
+
+def test_snapshot_hash_detect_corruption_in_global_fast_check_mode(client, volume_name, settings_reset):  # NOQA
+    update_settings(client, SNAPSHOT_DATA_INTEGRITY_FAST_CHECK, "true",
+                    "false")
+    detect_and_repair_corrupted_replica(client, volume_name,
+                                        SNAPSHOT_DATA_INTEGRITY_FAST_CHECK)
+
+
+def update_settings(client, data_integrity, immediate_check, fast_replica_rebuild, period_in_second=SNAPSHOT_CHECK_PERIOD):  # NOQA
```
c3y1huang, line 109: Suggest renaming this to something more descriptive and less general, ex: `prepare_setting_for_snapshot_test`.

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,491 @@
...
+    count = 0
+    for count in range(RETRY_WAIT_CHECKSUM_COUNTS):
+        if check_snapshot_checksums_set(volume_name):
+            print(f'All checksums are set in {count} sec')
+            snapshot_checksums_generate = True
+            break
+        else:
+            time.sleep(RETRY_INTERVAL)
+
+    return snapshot_checksums_generate, count
```
c3y1huang, line 491: Why not just `assert snapshot_checksums_generate` here instead of returning it?

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,491 @@
...
+    volume = volume.attach(hostId=lht_hostId)
+    volume = wait_for_volume_healthy(client, volume_name)
+
+    volume = client.by_id_volume(volume_name)
+
+    # Step 2
+    create_snapshots(client, volume, 1)
+
+    # Step 3
+    # +60 is a toleration for some delay
```
c3y1huang, line 474: Can you clarify why the case needs the toleration for the delay?

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,491 @@
...
+from common import SETTING_SNAPSHOT_FAST_REPLICA_REBUILD_ENABLED
+from common import SNAPSHOT_DATA_INTEGRITY_IGNORED
+from common import SNAPSHOT_DATA_INTEGRITY_ENABLED
+from common import SNAPSHOT_DATA_INTEGRITY_FAST_CHECK
+from common import SNAPSHOT_DATA_INTEGRITY_DISABLED
+
+RETRY_WAIT_CHECKSUM_COUNTS = 600
+SNAPSHOT_CHECK_PERIOD = 300
+
+def test_snapshot_hash_global_enabled_with_immediate_hash(client, volume_name, settings_reset):  # NOQA
```
c3y1huang, line 34: Suggest adding some descriptions or context to the test cases.

c3y1huang (changes requested):

manager/integration/tests/test_snapshot.py
derekbit: Removed.

manager/integration/tests/test_snapshot.py
derekbit: Updated. Thank you.

manager/integration/tests/test_snapshot.py
derekbit: Updated. Thank you.

manager/integration/tests/test_snapshot.py
derekbit: Because the checksum can be calculated by `perioc check mechanism` or `immedidate calculation after the snapshot is created`. For the first case, we cannot precisely known when the event happens, so I put a toleration here and check both the time elapsed and the checksum value.

manager/integration/tests/test_snapshot.py
derekbit: Updated. Thank you.

manager/integration/tests/test_snapshot.py
c3y1huang: Suggest having a `SNAPSHOT_CHECK_TOLERATION_DELAY=60` variable and add the comment to it.

derekbit: Updated. Thank you.

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,601 @@
...
+    create_snapshots(client, volume, 1)
+
+    # Step 3
+    sleep_time = 120
+    time.sleep(sleep_time)
+    assert not check_snapshot_checksums_set(volume)
+
+    # Step 4
+    generate_checksums_time = wait_for_snapshot_checksums_generate(volume) # NOQA
+    # +-SNAPSHOT_CHECK_TOLERATION_DELAY is a toleration for some delays
```
c3y1huang, line 549: Redundant?

derekbit: Yeah, removed. Thank you.

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,598 @@
...
+    try:
+        output = subprocess.check_output(exec_cmd)
+    except subprocess.CalledProcessError as e:
+        print(e.output)
+        return ""
+
+    return str(output, 'utf-8').rstrip('\n') + " UTC"
+
+
+def create_snapshots(client, volume, data_size):  # NOQA
```
c3y1huang, line 447: Can we make this function definition more generic? ex: `get_ctime_from_snapshot_disk_file(data_path, snapshot_name, num_snapshots)`

c3y1huang (changes requested):

manager/integration/tests/test_snapshot.py
derekbit: Do you mean get multiple snapshots' ctimes at the same time?

c3y1huang: Apology for the confusion, that is not what I meant and I've updated my previous comment example.

derekbit: Updated. Thank you.

khushboo-rancher (approved):

manager/integration/tests/test_snapshot.py
c3y1huang: Where is this updated?

derekbit: Sorry...forgot to push it...

manager/integration/tests/test_snapshot.py
```diff
@@ -0,0 +1,590 @@
...
+
+    snapshots = volume.snapshotList()
+    snapMap = {}
+    for snap in snapshots:
+        snapMap[snap.name] = snap
+
+    for i in range(num_snapshots):
+        snap = snaps[i]
+        assert snapMap[snap.name].name == snap.name
+        assert snapMap[snap.name].removed is False
```
c3y1huang, line 463: Noticed the assert for `snapMap[snap.name].parent` is removed. Is this intended?

derekbit: Removing it won't influence the test cases. But, I added them back.

c3y1huang (approved):

roger-ryao:
@Mergifyio backport v1.4.x

PR #1239, opened 2023-01-07
Add check for system backup feature support

roger-ryao:


innobead:
@roger-ryao you can mention `longhorn/longhorn#5232` in your git commit as well.

innobead (approved):
LGTM.

However, the better solution is to use the right version of the longhorn test image when testing the corresponding Longhorn version.

innobead:
@Mergifyio backport v1.4.x

PR #1249, opened 2023-01-12
Implement test case `test_delete_backup_during_restoring_volume`

roger-ryao:
https://github.com/longhorn/longhorn/issues/5032


manager/integration/tests/test_basic.py
```diff
@@ -5047,7 +5048,41 @@ def test_delete_backup_during_restoring_volume():  # NOQA
...
+
+    #  And delete the backup immediately
+    delete_backup(client, bv.name, b.name)
+    volume = wait_for_volume_status(client, vol_v1_name,
+                                    "lastBackup", "")
+    assert volume.lastBackupAt == ""
+
+    #  Then volume v2 "robustness" should be "faulted"
+    #   And `status` of volume restore condition should be "False",
+    #   And `reason` of volume restore condition should be "RestoreFailure"
```
chriscchien, line 5080: Just wondering if those extra spaces in comments were need

roger-ryao: I just want to implement the test step with the BDD description. 
There is 2 spaces and 3 spaces after #.  That is not the rule from the testing method. 
I am used to using this format.

c3y1huang: Suggest also updating the test steps to BDD style. 

I think it's fine keeping the spacing for alignment in the test steps. But maybe we don't need the spacing in the comments for tracking the steps. WDYT @roger-ryao @chriscchien ?

roger-ryao: > Suggest also updating the test steps to BDD style.
> 
> I think it's fine keeping the spacing for alignment in the test steps. But maybe we don't need the spacing in the comments for tracking the steps. WDYT @roger-ryao @chriscchien ?

Hi @c3y1huang 
I could refer to [your previous PR](https://github.com/longhorn/longhorn-tests/pull/1124/files#diff-a8f0fde79c3b8e81457325c6c5f55d814da023c4116cacbfc5a82f42158911d4R497-R523). I think I should remove the spacing in the comments and update the test steps to BDD style.

cc @chriscchien

roger-ryao: After discussed with @longhorn/qa and @c3y1huang , I update the [PR](https://github.com/longhorn/longhorn-tests/compare/e48efa15cca8d5d3c1cc6276f42623115bf66b2b..2928ae8c5e1e2691ab6cb021a563efc17b7d1b71)

manager/integration/tests/test_basic.py
```diff
@@ -5021,8 +5023,7 @@ def test_backup_volume_restore_with_access_mode(core_api, # NOQA
     assert volume_sp_access_mode.accessMode == overridden_restored_access_mode
 
 
-@pytest.mark.skip(reason="TODO")  # NOQA
-def test_delete_backup_during_restoring_volume():  # NOQA
+def test_delete_backup_during_restoring_volume(set_random_backupstore, client, core_api, volume_name):  # NOQA
```
c3y1huang, line 5026: The `core_api` and `volume_name` seems redundant.

If want to keep the volume_name you can use `vol_v1_name = volume_name + "-v1"`.

c3y1huang (changes requested):

manager/integration/tests/test_basic.py
roger-ryao: > The `core_api` and `volume_name` seems redundant.
> 
> If want to keep the volume_name you can use `vol_v1_name = volume_name + "-v1"`.

Remove the `core_api` and `volume_name` and the test result is passed.
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2982/
Screenshot_20230210_101921

khushboo-rancher (commented):
LGTM

This case (Deleting the backup immediately after restoring of volume triggered) looks fine but I have another case also in my mind as below. In the implemented case, I believe by the time volume starts restore, the backup is deleted.

1. How do we make sure that the volume is in middle of the restoring when the backup was deleted?

manager/integration/tests/test_basic.py
```diff
@@ -5031,23 +5032,45 @@ def test_delete_backup_during_restoring_volume():  # NOQA
...
-       the status of volume restore condition is "False",
-       the reason of volume restore condition is "RestoreFailure"
-    """
-    pass
+    vol_v2_name = "vol-v2"
+    client.create_volume(name=vol_v2_name,
+                         size=str(512 * Mi),
+                         numberOfReplicas=3,
+                         fromBackup=b.url)
+
```
roger-ryao, line 5062: > LGTM
> 
> This case (Deleting the backup immediately after restoring of volume triggered) looks fine but I have another case also in my mind as below. In the implemented case, I believe by the time volume starts restore, the backup is deleted.
> 
> 1. How do we make sure that the volume is in middle of the restoring when the backup was deleted?

Hi @khushboo-rancher :
    I implemented the test case base on https://github.com/longhorn/longhorn/issues/1867#issuecomment-1325880709, but **Scenario 2** couldn't be reproduced. That's why, I don't add `wait_for_volume_restoration_start` before `delete_backup`. If we need to make sure that the volume is in the middle of the restoring when the backup was deleted, I need to create another test case for it.
**WDYT**  ?  
    
    
    
Scenario 1:
  - [X] Restore volume from backup then immediately delete the backup - volume become faulted immediately
  
Scenario 2:
  - [ ] Restore volume from backup then delete the backup while restoring was in progress - volume become faulted after awhile

mantissahz: @roger-ryao could you confirm that we could delete the backup that is used to be the source data of the restoration and restoring is in progress? I could not delete the backup in master-head environment.

roger-ryao: > @roger-ryao could you confirm that we could delete the backup that is used to be the source data of the restoration and restoring is in progress? I could not delete the backup in master-head environment.

Hi @mantissahz :
     We have two scenarios. One is **Deleting the backup immediately after restoring of volume triggered**, we could delete the backup. This [PR-1249](https://github.com/longhorn/longhorn-tests/pull/1249) would cover it.
     Another one is **Deleting the backup  after restoring is in progress**, we couldn't delete the backup. 
     `test_basic.py::test_backup_lock_deletion_during_restoration`  have covered this scenario.

mantissahz: @roger-ryao 

so Scenario 2:

>Restore volume from backup then delete the backup while restoring was in progress - volume become faulted after awhile

It will not happen, right?

chriscchien: I can not reproduce scenario 2 in master-head and v1.4.x-head, maybe my test environment was  not stable when I was testing.

roger-ryao: > It will not happen, right

@mantissahz 
   No, scenario 2 won't happen. We should refer to `test_basic.py::test_backup_lock_deletion_during_restoration`  when testing scenario 2.

roger-ryao:
According to the information in the https://github.com/longhorn/longhorn/issues/1867#issuecomment-1453139606, it seems that we need to review our The https://github.com/longhorn/longhorn-tests/pull/1249 test story.

cc @longhorn/qa  @mantissahz @derekbit

roger-ryao:
We do not need to update the test case as per the information provided in https://github.com/longhorn/longhorn/issues/1867#issuecomment-1457572250.

khushboo-rancher (approved):

innobead (approved):
LGTM. Thanks @roger-ryao

innobead:
@Mergifyio backport v1.4.x v1.3.x

PR #1255, opened 2023-02-21
ci: add eks provisioning

yangchiu:
ci: add eks provisioning

For https://github.com/longhorn/longhorn/issues/4621
    

yangchiu:
* Sometimes `terraform destroy` could encounter errors and fail like mentioned in https://aws.amazon.com/premiumsupport/knowledge-center/eks-delete-cluster-issues/ `Complete an Amazon EKS automatic platform version upgrade` section and it's inevitable. Probably need `terraform destroy` retry mechanism.
* `longhorn-setup.sh` needs further refactoring, too many functions for different purposes in the same file.

test_framework/scripts/longhorn-setup.sh
```diff
@@ -250,6 +275,7 @@ create_aws_secret(){
   yq e -i '.data.AWS_DEFAULT_REGION |= "'${AWS_DEFAULT_REGION_BASE64}'"' "${TF_VAR_tf_workspace}/templates/aws_cred_secrets.yml"
 
   kubectl apply -f "${TF_VAR_tf_workspace}/templates/aws_cred_secrets.yml"
+  kubectl apply -f "${TF_VAR_tf_workspace}/templates/aws_cred_secrets.yml" -n kube-system
```
innobead, line 278: Why need to apply the same manifest twice in different namespaces?

test_framework/scripts/longhorn-setup.sh
```diff
@@ -373,13 +399,20 @@ run_longhorn_tests(){
...
+
   # set debugging mode off to avoid leaking aws secrets to the logs.
   # DON'T REMOVE!
   set +x
   create_aws_secret
   set -x
+
+  if [[ ${TF_VAR_k8s_distro_name} == "eks" ]]; then
+    create_admin_service_account
+    install_iscsi
```
innobead, line 411: Do we need to install nfs/cifs client as well?

test_framework/scripts/longhorn-setup.sh
```diff
@@ -25,20 +25,45 @@ set_kubeconfig_envvar(){
...
+  yq -i ".users[0].user.token=\"${TOKEN}\""  "${TF_VAR_tf_workspace}/eks.yml"
+}
+
+
+install_iscsi(){
+  kubectl apply -f "https://raw.githubusercontent.com/longhorn/longhorn/master/deploy/prerequisite/longhorn-iscsi-installation.yaml"
+}
+
+
+install_cluster_autoscaler(){
```
innobead, line 60: Suggest just installing from the manifest curl -O https://raw.githubusercontent.com/kubernetes/autoscaler/master/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml instead of creating cluster_autoscaler.yaml maintained by our own. 

The same for other public cloud distros.

innobead (changes requested):
In general, LGTM, except some comments needed to answer/resolve first.

test_framework/scripts/longhorn-setup.sh
yangchiu: cluster-autoscaler is in kube-system, otherwise cluster-autoscaler is unable to see aws credentials.

test_framework/scripts/longhorn-setup.sh
yangchiu: Seems not necessary. Longhorn successfully installed without apply these two.

test_framework/scripts/longhorn-setup.sh
yangchiu: refined.

yangchiu:
> This pull request is now in conflicts. Could you fix it @yangchiu? pray

Conflicts resolved.

test_framework/scripts/longhorn-setup.sh
innobead: Well explained. The answer was in the example upstream manifests.

innobead (approved):
In general, LGTM.

PR #1277, opened 2023-03-06
test: consolidate instance managers

c3y1huang:
https://github.com/longhorn/longhorn/issues/5208


manager/integration/tests/common.py
```diff
@@ -3251,18 +3252,24 @@ def wait_for_all_instance_manager_running(client):
...
         try:
             for im in instance_managers:
                 if im.managerType == "engine" and im.currentState == "running":
                     node_to_engine_manager_map[im.nodeID] = im
                 elif im.managerType == "replica" and \
                         im.currentState == "running":
                     node_to_replica_manager_map[im.nodeID] = im
+                elif im.managerType == "aio" and \
+                        im.currentState == "running":
+                    node_to_instance_manager_map[im.nodeID] = im
```
derekbit, line 3267: Do we need to check `engine` and `replica` after introducing `aio`?

c3y1huang: IIRC we still need them for upgrade tests.

c3y1huang: Regard about the concern on the time of wait:
The flow is:
1. putting the instance_managers to `node_to_engine_manager_map`, `node_to_replica_manager_map` and `node_to_instance_manager_map` according the the manager type.
2. check if any of the `len(node_to_<type>_manager_map) == len(node)`.
3. Loop through each `node_to_<type>_manager_map` to check for desire state.

In case when all instance managers are `aio` type, `node_to_engine_manager_map` and `node_to_replica_manager_map` will contain an empty array, hence should not check for  `wait_for_instance_manager_desire_state`.

derekbit: Sounds good! Thank you.

manager/integration/tests/common.py
```diff
@@ -2204,7 +2205,14 @@ def wait_for_replica_failed(client, volname, replica_name,
                 break
             if r['instanceManagerName'] != "":
                 im = client.by_id_instance_manager(r['instanceManagerName'])
-                if r['name'] in im['instances']:
+
+                instance_dict = {}
+                if im['instances'] is not None:
```
innobead, line 2210: Suggest adding a comment to mention this is for backward-compatibility.

manager/integration/tests/common.py
```diff
@@ -3279,18 +3294,24 @@ def wait_for_all_instance_manager_running(client):
...
                     node_to_replica_manager_map[im.nodeID] = im
+                elif im.managerType == "aio" and \
+                        im.currentState == "running":
+                    node_to_instance_manager_map[im.nodeID] = im
                 else:
                     print("\nFound unknown instance manager:", im)
             if len(node_to_engine_manager_map) != len(nodes) or \
-                    len(node_to_replica_manager_map) != len(nodes):
+                    len(node_to_replica_manager_map) != len(nodes) or \
+                    len(node_to_instance_manager_map) != len(nodes):
```
innobead, line 3314: In fresh install, the number of engine/replica instance manager should be 0.
In upgrade, the number of existing engine/replica should depend on how many attached volumes are still running. 

Why we expect the number of each instance manager type should be equal to the node number?

innobead (changes requested):
In general, LGTM, but some comments left.

manager/integration/tests/common.py
c3y1huang: Good question. After checking the code again, the `wait_for_all_instance_manager_running` is used in `cleanup_client` for the client fixture. Longhorn will get upgraded during the `test_upgrade` and cleaning up shouldn't require checking on the instance-manager-e/r.

Removed.

innobead (approved):

PR #1279, opened 2023-03-08
test: update manual case test node deletion

chriscchien:
ref: [5508](https://github.com/longhorn/longhorn/issues/5508)
After https://github.com/longhorn/longhorn-manager/pull/1691 merged, replica on not ready node won't marked as failed.

Will have below test scenario with 1 replica volume after node delete
- Volume engine and replica both on the delete node -> volume become `Unknown` because replica not failed
- Volume engine on the delete node, replica on not delete node -> volume become `Unknown` because replica not failed
- Replica on the delete node, volume engine on not delete node -> volume become `Faulted` because replica failed

### DIFF
--- a/docs/content/manual/pre-release/node/node-deletion.md
+++ b/docs/content/manual/pre-release/node/node-deletion.md
@@ -9,24 +9,28 @@ title: Test node deletion
 
 
 ### Case 2: Delete nodes when there are running volumes:
+We need to update step 4.1.1 related steps after https://github.com/longhorn/longhorn/issues/5542 resolved.
+
 1. Deploy Longhorn.
 2. For each node that will be deleted later: 
-    1. create and attach 4 volumes:
+    1. Create and attach 4 volumes:
         1. The 1st volume contains 1 replica only. Both the engines and the replicas are on the pre-delete node. (Attached to the pre-delete node.)
         2. The 2nd volume contains 1 replica only. The engine is on the pre-delete node and the replica is on another node. (Attached to the pre-delete node.)
-        3. The 3rd and the 4th volume contain 3 replicas. Both volumes are attached to a node except for the pre-delete node.
+        3. The 3rd volume contains 1 replica only. The replica is on the pre-delete node and the engine is on another node. (Attached to a node except for the pre-delete node.)
+        4. The 4th volume contain 3 replicas. Attached to a node except for the pre-delete node.
     2. Write some data to all volumes.
 3. Delete multiple nodes in the cluster simultaneously. Make sure the corresponding Kubernetes node object is deleted. --> The related Longhorn node objects will be cleaned up immediately, too.
 4. For each deleted node:
     1. Verify the volume Health state  -->
-        1. The 1st volume should become `Faulted`.
+        1. The 1st volume should keep `Unknown`.
         2. The 2nd volume should keep `Unknown`.
-        3. The 3rd and the 4th volume should become `Degraded`.
-    2. Delete the 1st volume. --> The volume can be deleted.
+        3. The 3rd volume should become `Faulted`.
+        4. The 4th and 5th volume should become `Degraded`.
+    2. Detach then reattach the 1st volume. --> The volume in detached state and can not be attached to another node.
     3. Detach then reattach the 2nd volume. --> The volume works fine and the data is correct.
-    4. Crash all replicas of the 3rd volume and trigger the auto salvage. --> The auto salvage should work. The volume works fine and the data is correct after the salvage.
+    4. Delete the 3rd volume. --> The volume can be deleted.
+    5. Deleted the replica on the deleted node for the 4th volume. --> The replica can be deleted.
+    6. Crash all replicas of the 4th volume and trigger the auto salvage. --> The auto salvage should work. The volume works fine and the data is correct after the salvage.
         * An example to crash every volume replica instance manager pods with kubectl:
             - `kubectl delete pods -l longhorn.io/instance-manager-type=replica -n longhorn-system --wait`
-    5. Disabled the auto salvage setting. Then crash all replicas of the 3rd volume again. --> The replica on the deleted node cannot be salvaged manually, too. The salvage feature still works fine.
-    6. Deleted the replica on the deleted node for the 4th volume. --> The replica can be deleted.
-
+    7. Disabled the auto salvage setting. Then crash all replicas of the 4th volume again. --> The replica on the deleted node cannot be salvaged manually, too. The salvage feature still works fine.


docs/content/manual/pre-release/node/node-deletion.md
```diff
@@ -11,22 +11,26 @@ title: Test node deletion
...
 1. Deploy Longhorn.
 2. For each node that will be deleted later: 
-    1. create and attach 4 volumes:
+    1. Create and attach 5 volumes:
         1. The 1st volume contains 1 replica only. Both the engines and the replicas are on the pre-delete node. (Attached to the pre-delete node.)
         2. The 2nd volume contains 1 replica only. The engine is on the pre-delete node and the replica is on another node. (Attached to the pre-delete node.)
-        3. The 3rd and the 4th volume contain 3 replicas. Both volumes are attached to a node except for the pre-delete node.
+        3. The 3rd volume contains 1 replica only. The replica is on the pre-delete node and the engine is on another node. (Attached to a node except for the pre-delete node.)
+        4. The 4th volume contain 3 replicas. Attached to a node except for the pre-delete node.
+        5. The 5th volume contain 3 replicas. Attached to a node except for the pre-delete node.
```
innobead, line 19: Why need to have 4 & 5, they look identical?

docs/content/manual/pre-release/node/node-deletion.md
```diff
@@ -11,22 +11,26 @@ title: Test node deletion
...
     1. Verify the volume Health state  -->
-        1. The 1st volume should become `Faulted`.
+        1. The 1st volume should keep `Unknown`.
         2. The 2nd volume should keep `Unknown`.
-        3. The 3rd and the 4th volume should become `Degraded`.
-    2. Delete the 1st volume. --> The volume can be deleted.
+        3. The 3rd volume should become `Faulted`.
+        4. The 4th and 5th volume should become `Degraded`.
+    2. Detach then reattach the 1st volume. --> The volume can not be attached.
     3. Detach then reattach the 2nd volume. --> The volume works fine and the data is correct.
```
innobead, line 29: What's the status for the 2? the volume will be in `attaching`?

innobead:
> ref: [5508](https://github.com/longhorn/longhorn/issues/5508) After [longhorn/longhorn-manager#1691](https://github.com/longhorn/longhorn-manager/pull/1691) merged, replica on not ready node won't marked as failed.
> 
> Will have below test scenario with 1 replica volume after node delete
> 
> * Volume engine and replica both on the delete node -> volume become `Unknown` because replica not failed

This is not completely right.

If the node becomes not ready and the IM pod is not evicted yet, the engine/replica status should be unknown; otherwise, it should become error. 

@derekbit @shuo-wu please correct me if anything is wrong. 

[/controller/instance_handler.go#L56-L88](https://github.com/longhorn/longhorn-manager/blob/423513ad5838695c0b481a1c5309a0ecb73abbe9/controller/instance_handler.go#L56-L88)
```go
	if im == nil || im.Status.CurrentState == longhorn.InstanceManagerStateUnknown {
		if status.Started {
			if status.CurrentState != longhorn.InstanceStateUnknown {
				logrus.Warnf("The related node %v of instance %v is down or deleted, will mark the instance as state UNKNOWN", spec.NodeID, instanceName)
			}
			status.CurrentState = longhorn.InstanceStateUnknown
		} else {
			status.CurrentState = longhorn.InstanceStateStopped
			status.CurrentImage = ""
		}
		status.IP = ""
		status.StorageIP = ""
		status.Port = 0
		return
	}

	if im.Status.CurrentState == longhorn.InstanceManagerStateStopped ||
		im.Status.CurrentState == longhorn.InstanceManagerStateError ||
		im.DeletionTimestamp != nil {
		if status.Started {
			if status.CurrentState != longhorn.InstanceStateError {
				logrus.Warnf("Cannot find the instance manager for the running instance %v, will mark the instance as state ERROR", instanceName)
			}
			status.CurrentState = longhorn.InstanceStateError
		} else {
			status.CurrentState = longhorn.InstanceStateStopped
		}
		status.CurrentImage = ""
		status.IP = ""
		status.StorageIP = ""
		status.Port = 0
		return
	}
```

> * Volume engine on the delete node, replica on not delete node -> volume become `Unknown` because replica not failed

I assume the engine should become unknow and then failed in the end? because the IM-E has been gone? @shuo-wu @derekbit can you correct if anything wrong here?

innobead (changes requested):

docs/content/manual/pre-release/node/node-deletion.md
shuo-wu: These 2 volumes has identical init state but later on they will be used for different tests

shuo-wu:
> If the node becomes not ready and the IM pod is not evicted yet, the engine/replica status should be unknown; otherwise, it should become error.

The default pod eviction timeout after node down is 5 min IIRC. Hence the volume won't become Faulted immediately. Do we have another test for the node deletion + a long wait case (IM pod eviction will be triggered)?

innobead:
> > If the node becomes not ready and the IM pod is not evicted yet, the engine/replica status should be unknown; otherwise, it should become error.
> 
> The default pod eviction timeout after node down is 5 min IIRC. Hence the volume won't become Faulted immediately. Do we have another test for the node deletion + a long wait case (IM pod eviction will be triggered)?

YES, we need to differentiate these cases. It will help us better understand the status transition. @chriscchien Could you update this by adding different cases considering the `pod eviction timeout` factor if we don't have that test case yet? Thanks.

cc @longhorn/qa

chriscchien:
> > If the node becomes not ready and the IM pod is not evicted yet, the engine/replica status should be unknown; otherwise, it should become error.
> 
> The default pod eviction timeout after node down is 5 min IIRC. Hence the volume won't become Faulted immediately. Do we have another test for the node deletion + a long wait case (IM pod eviction will be triggered)?

Hi @shuo-wu ,

After 20 minutest from I shoutdown the node which volume and it's only replica attached to, the volume kept in unknown state. From kubectl I see the engine is still in running status on the shutdown node (ip-172-31-95-182)
```
engine-image-ei-aacc37ad-5kdfk                        1/1     Running       0          25m   10.42.3.7    ip-172-31-95-182   <none>           <none>
```
When I try exec into the running engine will get  error
```
Error from server: error dialing backend: proxy error from 127.0.0.1:6443 while dialing 172.31.95.182:10250, code 503: 503 Service Unavailable
```
And here is the cluster and supportbundle
https://44.202.23.17/#/volume
[supportbundle_1bde7a91-e22e-494a-9ed1-8525e74eba50_2023-03-13T05-17-07Z.zip](https://github.com/longhorn/longhorn-tests/files/10953784/supportbundle_1bde7a91-e22e-494a-9ed1-8525e74eba50_2023-03-13T05-17-07Z.zip)

cc @innobead

chriscchien:
> > > If the node becomes not ready and the IM pod is not evicted yet, the engine/replica status should be unknown; otherwise, it should become error.
> > 
> > 
> > The default pod eviction timeout after node down is 5 min IIRC. Hence the volume won't become Faulted immediately. Do we have another test for the node deletion + a long wait case (IM pod eviction will be triggered)?
> 
> Hi @shuo-wu ,
> 
> After 20 minutest from I shoutdown the node which volume and it's only replica attached to, the volume kept in unknown state. From kubectl I see the engine is still in running status on the shutdown node (ip-172-31-95-182)
> 
> ```
> engine-image-ei-aacc37ad-5kdfk                        1/1     Running       0          25m   10.42.3.7    ip-172-31-95-182   <none>           <none>
> ```
> 
> When I try exec into the running engine will get error
> 
> ```
> Error from server: error dialing backend: proxy error from 127.0.0.1:6443 while dialing 172.31.95.182:10250, code 503: 503 Service Unavailable
> ```
> 
> And here is the cluster and supportbundle https://44.202.23.17/#/volume [supportbundle_1bde7a91-e22e-494a-9ed1-8525e74eba50_2023-03-13T05-17-07Z.zip](https://github.com/longhorn/longhorn-tests/files/10953784/supportbundle_1bde7a91-e22e-494a-9ed1-8525e74eba50_2023-03-13T05-17-07Z.zip)
> 
> cc @innobead

Hi @shuo-wu Please ignore this comment. It's a false alarm, I will update test case for review later, thank you

docs/content/manual/pre-release/node/node-deletion.md
chriscchien: This volume can detach success, after that if I try to attach this volume to another node,  attach will faie and volume kept in detach state
```
unable to attach volume vol1 to ip-172-31-82-34: cannot attach volume vol1 because the engine image longhornio/longhorn-engine:v1.4.1-rc2 is not deployed on at least one of the the replicas' nodes or the node that the volume is going to attach to
```

innobead: I see, then this is expected because the replica was on the deleted node. However, the message looks misleading, because it should be the only replica is failed on the deleted node, so there is no available replica instead.

[/datastore/longhorn.go#L1489-L1494](https://github.com/longhorn/longhorn-manager/blob/423513ad5838695c0b481a1c5309a0ecb73abbe9/datastore/longhorn.go#L1489-L1494)
```go
	for _, r := range replicas {
		isReady, err := s.CheckEngineImageReadiness(image, r.Spec.NodeID)
		if err != nil || isReady {
			return isReady, err
		}
	}
```

derekbit:
`unknown` is the current behavior. We should distinguish between two cases: node not found and node ready. For the node not found case, we can directly mark the replica as failed. Open a ticket https://github.com/longhorn/longhorn/issues/5542 for improvement.

innobead:
@chriscchien Please confirm the status you see below for the node deletion. 

### Node CR is deleted (ex: kubectl delete node) <-- this PR

- Node and IM CR are deleted, so the engine and replica on the deleted node become unknown forever. (volume state becomes unknown as well)

### Node CR is around (ex: kubelet restart)

- IM, engine, replica on the node should become `unknown`, volume state will become `unknown` as well.
- Resource IM, running engine/replica on the node should become `error`, after IM pod is evicted (graceful pod eviction).
- If the engine is alive but somehow the replica is unable to reach, the replica will become `error` instead

innobead:
> ### Node CR is deleted (ex: kubectl delete node) <-- this PR
> * Given node and IM CR are deleted, so the engine and replica on the deleted node become `unknown` forever. (volume state becomes unknown as well)

If this is right, I feel the node has been deleted, why not just make the engine on the node/replica `error` instead of `unknown`? @shuo-wu @derekbit WDYT?

chriscchien:
Hi @innobead,

> @chriscchien Please confirm the status you see below for the node deletion.
> 
> ### Node CR is deleted (ex: kubectl delete node) <-- this PR
> * Given node and IM CR are deleted, so the engine and replica on the deleted node become unknown forever. (volume state becomes unknown as well)
> * Given node and IM CR are deleted, if the engine is alive on another node, because the replica on the deleted node is unable to reach by the engine, the replica will become `error`.

Yes in v1.4.1-rc2, for test case **Node CR is deleted**, volume behavior acted as expected in those 2 scenarios.


> 
> ### Node CR is around (ex: kubelet restart) <-- we should have another test already cover this?
> * IM, engine, replica on the node should become `unknown`, volume state will become `unknown` as well.
> * Resource IM, running engine/replica on the node should become `error`, after IM pod is evicted (graceful pod eviction).
> * If the engine is alive but somehow the replica is unable to reach, the replica will become `error` instead

And for test case **Node CR is around**, covered in test case [Single replica node down](https://longhorn.github.io/longhorn-tests/manual/pre-release/ha/single-replica-node-down/)

innobead:
Thanks @chriscchien , so let's just base on the current behavior to update the test cases for node deletion. 

cc @longhorn/qa

docs/content/manual/pre-release/node/node-deletion.md
chriscchien: Hi @shuo-wu , updated! thank you.

shuo-wu:
> > ### Node CR is around (ex: kubelet restart) <-- we should have another test already cover this?
> > 
> > * IM, engine, replica on the node should become `unknown`, volume state will become `unknown` as well.
> > * Resource IM, running engine/replica on the node should become `error`, after IM pod is evicted (graceful pod eviction).
> > * If the engine is alive but somehow the replica is unable to reach, the replica will become `error` instead
> 
> And for test case **Node CR is around**, covered in test case [Single replica node down](https://longhorn.github.io/longhorn-tests/manual/pre-release/ha/single-replica-node-down/)

Will kubelet restart lead to IM pod restarting? I think there are 2 case at least:
1. kubelet is restarted/back within 5 min, then the IM pod eviction probably is not triggered. This is different from the node down/restart case.
2. kubelet is restarted/back after 5 min, then the IM pod eviction can be triggered if the cluster applies the default config. Then this is a little bit similar to the node restart case.

chriscchien:
> 



> > > ### Node CR is around (ex: kubelet restart) <-- we should have another test already cover this?
> > > 
> > > * IM, engine, replica on the node should become `unknown`, volume state will become `unknown` as well.
> > > * Resource IM, running engine/replica on the node should become `error`, after IM pod is evicted (graceful pod eviction).
> > > * If the engine is alive but somehow the replica is unable to reach, the replica will become `error` instead
> > 
> > 
> > And for test case **Node CR is around**, covered in test case [Single replica node down](https://longhorn.github.io/longhorn-tests/manual/pre-release/ha/single-replica-node-down/)
> 
> Will kubelet restart lead to IM pod restarting? I think there are 2 case at least:
> 
> 1. kubelet is restarted/back within 5 min, then the IM pod eviction probably is not triggered. This is different from the node down/restart case.
> 2. kubelet is restarted/back after 5 min, then the IM pod eviction can be triggered if the cluster applies the default config. Then this is a little bit similar to the node restart case.

Hi @shuo-wu , I created a ticket for the test case need be updated [here](https://github.com/longhorn/longhorn/issues/5555) https://github.com/longhorn/longhorn/issues/5555 for track

chriscchien:
Close this PR first because node down behavior below will change after https://github.com/longhorn/longhorn/issues/5542 ready
- Volume engine and replica both on the delete node -> volume become Unknown because replica not failed

PR #1280, opened 2023-03-09
ci: refine sles ami rule to avoid error from ami being deleted

yangchiu:
ci: refine sles ami rule to avoid error from ami being deleted


innobead (approved):

innobead:
Do we need to apply this to other distros?

yangchiu:
@mergify backport v1.4.x v1.3.x

PR #1283, opened 2023-03-10
test(recurring-job): filesystem-trim

c3y1huang:
https://github.com/longhorn/longhorn/issues/5528
https://github.com/longhorn/longhorn/issues/5186

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4988,7 +4988,8 @@ def check_recurring_jobs(client, recurring_jobs):
         assert recurring_job.cron == spec["cron"]
 
         expect_retain = spec["retain"]
-        if recurring_job.task == "snapshot-cleanup":
+        if recurring_job.task == "snapshot-cleanup" or \
+                recurring_job.task == "filesystem-trim":
             expect_retain = 0
         assert recurring_job.retain == expect_retain
 
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -11,6 +11,9 @@ from backupstore import set_random_backupstore  # NOQA
 
 import common
 from common import client, core_api, apps_api, batch_v1_api  # NOQA
+from common import csi_pv  # NOQA
+from common import pod_make  # NOQA
+from common import pvc  # NOQA
 from common import random_labels, volume_name  # NOQA
 from common import storage_class, statefulset, pvc  # NOQA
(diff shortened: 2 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -4988,7 +4988,8 @@ def check_recurring_jobs(client, recurring_jobs):
         assert recurring_job.cron == spec["cron"]
 
         expect_retain = spec["retain"]
-        if recurring_job.task == "snapshot-cleanup":
+        if recurring_job.task == "snapshot-cleanup" or \
+                recurring_job.task == "filesystem-trim":
```
innobead, line 4992: nit: If having more than one check, probably can use `in` to check the task type.

manager/integration/tests/test_recurring_job.py
```diff
@@ -2069,3 +2075,96 @@ def test_recurring_job_restored_from_backup_target(set_random_backupstore, clien
...
+    size_change = 0
+    for _ in range(retry_counts):
+        time.sleep(5)
+
+        volume = client.by_id_volume(vol_name)
+        new_size = int(int(volume.controllers[0].actualSize) / Mi)
+
+        if wait_stablize and new_size != size_change:
+            size_change = new_size
+            continue
```
innobead, line 2166: I would assume `actualSize` potentially has some late update, so it could make this wait function unstable somehow.

However, we will have 5 seconds sleep to check it, so it could be fine. Need to follow up to see if this test case becomes flaky.

innobead (approved):
LGTM.

manager/integration/tests/test_recurring_job.py
innobead: cc @longhorn/qa

manager/integration/tests/test_recurring_job.py
```diff
@@ -2069,3 +2075,96 @@ def test_recurring_job_restored_from_backup_target(set_random_backupstore, clien
...
+    And actual size of the volume should not decrease.
+
+    When create a recurring job with:
+         - task: filesystem-trim
+         - retain: 1
+    Then recurring job retain mutated to 0.
+
+    When assign the recurring job to volume.
+    And wait for the cron job scheduled time.
+    Then volume actual size should decrease 50mb.
```
innobead, line 2098: This is a good example of BDD. Of course, some parts need to adjust for a better keyword statement. 

cc @longhorn/qa

PR #1284, opened 2023-03-14
Update CLI steps for verifying backing image sync failure test case

roger-ryao:
issue https://github.com/longhorn/longhorn/issues/5539

### DIFF
--- a/docs/content/manual/pre-release/ha/backing-image-error-reporting-and-retry.md
+++ b/docs/content/manual/pre-release/ha/backing-image-error-reporting-and-retry.md
@@ -10,12 +10,19 @@ title: Backing Image Error Reporting and Retry
 
 ## Backing image with sync failure
 1. Create a backing image. Then create and attach a volume using this backing image.
-2. Exec into the host of a worker node. remove the content of the backing image work directory then set the directory as immutable. This operation should fail one file of the backing image as well as keeping failing the following sync retries.
+2. Exec into one of the worker node, remove the files in that backing image directory and set the directory as immutable.
+   The removal of the files will trigger the sync process to sync backing image file from another worker node. Setting immutable to the directory will make the sync process and the following retry failed.
+   Option 1. Run the following command to remove the content of the backing image work directory and set the directory as immutable, which should cause the sync to fail.
     ```shell
-    rm -r /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/
+    rm /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/backing*
     chattr +i /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/
     ```
-3. Monitor the backing image manager pod log. Verify the backoff works for the sync retry as well.
+   Option 2: Run the following command to remove the content of the backing image work directory and wait for the backing.tmp file to be generated, indicating that the sync has started and the file is transferring. Set the directory as immutable during the transfer should still fail the sync since it can't rename the tmp file to remove the .tmp extent in the end.
+    ```shell
+    rm /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/backing* && while true; do if ls -l /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/backing.tmp >/dev/null 2>&1; then ls -l /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>; break; fi; done
+    chattr +i /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/
+    ```
+3. Monitor the backing-image-manager pod log. Verify the backoff works for the sync retry as well.
 4. Unset the immutable flag for the backing image directory. Then the retry should succeed, and the volume should become `healthy` again after the backing image re-sync complete.
     ```shell
     chattr -i /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/


docs/content/manual/pre-release/ha/backing-image-error-reporting-and-retry.md
```diff
@@ -10,12 +10,19 @@ title: Backing Image Error Reporting and Retry
 
 ## Backing image with sync failure
 1. Create a backing image. Then create and attach a volume using this backing image.
-2. Exec into the host of a worker node. remove the content of the backing image work directory then set the directory as immutable. This operation should fail one file of the backing image as well as keeping failing the following sync retries.
+2. Exec into the host of a worker node. remove the content of the backing image work directory then set the directory as immutable. 
+   This operation should fail one file of the backing image as well as keeping failing the following sync retries.
```
innobead, line 14: What does failing one file mean here?

innobead (commented):
Why we need 2 options here? I feel they are almost the same. The goal should just verify the file should be resynced if somehow it's able to sync and save the file to the destination node.

roger-ryao:
> Why we need 2 options here? I feel they are almost the same. The goal should just verify the file should be resynced if somehow it's able to sync and save the file to the destination node.

Hi @innobead 
I added option 2 based on the discussion from the [Slack channel](https://suse.slack.com/archives/C02HBK5AVC5/p1677835571602969?thread_ts=1677829738.913769&cid=C02HBK5AVC5). Perhaps some of our colleagues might find it useful, which is why I included it. If you think we don't need option 2, I can remove it later.

docs/content/manual/pre-release/ha/backing-image-error-reporting-and-retry.md
roger-ryao: Hi @innobead :
That's the origin test case description. I think I could modify it to say "**This should result in the failure of the sync operation and subsequent retries, as Longhorn will not be able to sync the data due to the immutable directory.**" What do you think?

innobead:
> Hi @innobead I added option 2 based on the discussion from the [Slack channel](https://suse.slack.com/archives/C02HBK5AVC5/p1677835571602969?thread_ts=1677829738.913769&cid=C02HBK5AVC5). Perhaps some of our colleagues might find it useful, which is why I included it. If you think we don't need option 2, I can remove it later.

@roger-ryao can you elaborate on this a bit? why might it be helpful?

roger-ryao:
> > Hi @innobead I added option 2 based on the discussion from the [Slack channel](https://suse.slack.com/archives/C02HBK5AVC5/p1677835571602969?thread_ts=1677829738.913769&cid=C02HBK5AVC5). Perhaps some of our colleagues might find it useful, which is why I included it. If you think we don't need option 2, I can remove it later.
> 
> @roger-ryao can you elaborate on this a bit? why might it be helpful?

`Option 1` deletes the content of the backing image folder and sets the directory as immutable.
`Option 2` not only deletes the content of the backing image folder but also waits for the `backing.tmp` file to be generated before setting the directory as immutable.  

I learned about the `option 2` test method from @ChanYiLin in the Slack channel. We were able to reproduce the issue in longhorn/longhorn#5481 using option 1, but we couldn't reproduce the issue with option 2. 
Using option 2 may be useful for testing purposes or for observing the behavior of Longhorn during sync retries, which is why I included it.

innobead:
Sounds good. I remembered that now. Thanks!

docs/content/manual/pre-release/ha/backing-image-error-reporting-and-retry.md
ChanYiLin: I think we can modify it into

2. Exec into one of the worker node, remove the files in that backing image directory and set the directory as immutable. 
   The removal of the files will trigger the sync process to sync backing image file from another worker node. Setting immutable to the directory will make the sync process and the following retry failed.

docs/content/manual/pre-release/ha/backing-image-error-reporting-and-retry.md
```diff
@@ -10,12 +10,19 @@ title: Backing Image Error Reporting and Retry
...
+   Option 1. Run the following command to remove the content of the backing image work directory and set the directory as immutable, which should cause the sync to fail.
     ```shell
-    rm -r /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/
+    rm /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/backing*
     chattr +i /var/lib/longhorn/backing-images/<backing image name>-<backing image UUID>/
     ```
-3. Monitor the backing image manager pod log. Verify the backoff works for the sync retry as well.
+   Option 2: Run the following command to remove the content of the backing image work directory and wait for the backing.tmp file to be generated, indicating that the sync is being retried. Once the retry succeeds, set the directory as immutable, which should cause the sync to fail
```
ChanYiLin, line 20: I don't think we need to add the option2 

Before the fix of https://github.com/longhorn/longhorn/issues/5481
option1 would stuck the process and there would be no retry triggered

But after the fix , both option1 and option2 will keep triggering the retry until it succeed.

ChanYiLin:
> > > Hi @innobead I added option 2 based on the discussion from the [Slack channel](https://suse.slack.com/archives/C02HBK5AVC5/p1677835571602969?thread_ts=1677829738.913769&cid=C02HBK5AVC5). Perhaps some of our colleagues might find it useful, which is why I included it. If you think we don't need option 2, I can remove it later.
> > 
> > 
> > @roger-ryao can you elaborate on this a bit? why might it be helpful?
> 
> `Option 1` deletes the content of the backing image folder and sets the directory as immutable. `Option 2` not only deletes the content of the backing image folder but also waits for the `backing.tmp` file to be generated before setting the directory as immutable.
> 
> I learned about the `option 2` test method from @ChanYiLin in the Slack channel. We were able to reproduce the issue in [longhorn/longhorn#5481](https://github.com/longhorn/longhorn/issues/5481) using option 1, but we couldn't reproduce the issue with option 2. Using option 2 may be useful for testing purposes or for observing the behavior of Longhorn during sync retries, which is why I included it.

The issue is unexpected and has been fixed, so both option1 and option2 should have the same behavior now.
Both will fail the sync process and keep triggering the retry until it succeed.

To be clear
option1:
Will fail since it can't create tmp file at the beginning.
Before the fix, it stuck and couldn't trigger retry cause we didn't catch the error.
After the fix, it will trigger retry as expected

option2:
Will fail since it cant rename tmp to final file (immutable).
It acts as expected that it will trigger retry cause we catch the error.

docs/content/manual/pre-release/ha/backing-image-error-reporting-and-retry.md
ChanYiLin: Or we can still add it ~
but there is something need to fix in the description

Option 2: Run the following command to remove the content of the backing image work directory and wait for the backing.tmp file to be generated, **indicating that the sync has started and the file is transferring**. Set the directory as immutable during the transferring should still fail the sync **since it can't rename the tmp file to remove the `.tmp` extent in the end.**

ChanYiLin:
LGTM

ChanYiLin (approved):

innobead (commented):
The update is good.

cc @longhorn/qa

PR #1295, opened 2023-03-27
Add Rancher integration test

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/5505

### DIFF
--- a/docs/content/manual/rancher-integration/drain-using-rancher-ui.md
+++ b/docs/content/manual/rancher-integration/drain-using-rancher-ui.md
@@ -1,0 +1,240 @@
+---
+title: Drain using Rancher UI
+---
+
+**Note:** Enabling `Delete Empty Dir Data` is mandatory to drain a node if a pod is associated with any storage.
+
+
+Test with Longhorn default setting of 'Node Drain Policy': `block-if-contains-last-replica`
+----
+
+### 1. Drain operation on single node using Rancher UI
+
+**Given**  Single node (1 Worker) cluster with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND 1 RWO and 1 RWX volumes unattached
+
+**When** Drain the node with default values of Rancher UI
+
+**Then** Drain should be blocked, user can see Rancher trying to drain the node but stuck in `drainig` and `cordoned`
+
+AND Volumes should be healthy and data should be intact once user stops `drain` and `uncordoned` the node
+
+AND The volumes which were attached initially should be attached and the ones which is unattached should be unattached
+
+AND Logs should not have any unexpected errors
+
+### 2. Drain operation on single node using Rancher UI with `Delete Empty Dir Data`
+
+**Given**  Single node (1 Worker) cluster with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND 1 RWO and 1 RWX volumes unattached
+
+**When** Drain the worker node one by one with `Delete Empty Dir Data` set to `Yes` and other values to default using Rancher UI
+
+**Then** Drain should be blocked. Longhorn's pdb should prevent draining the IM from the node.
+
+AND Volumes should be healthy and data should be intact once user stops `drain` and `uncordoned` the node
+
+AND The volumes which were attached initially should be attached and the ones which is unattached should be unattached
+
+AND Logs should not have any unexpected errors
+
+
+### 3. Drain operation on a multiple node using Rancher UI - default value
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND few RWO and RWX volumes unattached
+
+**When** Drain the worker node one by one with `Delete Empty Dir Data` set to `Yes` and other values to default using Rancher UI
+
+**Then** Drain should be successful
+ 
+AND Volumes should be healthy and data should be intact after drain completes
+
+AND The volumes which were attached initially should be attached and the ones which is unattached should be unattached
+
+AND Logs should not have any unexpected errors
+
+
+### 4. Drain operation on a multiple node using Rancher UI - Force
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND few RWO and RWX volumes unattached
+
+**When** Drain the worker node one by one with `Force` and `Delete Empty Dir Data` option enabled using Rancher UI.
+
+**Then** Drain should be successful
+
+AND Volumes should be healthy and data should be intact after drain completes
+
+AND The volumes which were attached initially should be attached and the ones which is unattached should be unattached
+
+AND Logs should not have any unexpected errors.
+
+
+### 5. Drain operation on a multiple node using Rancher UI - Single replica of volumes
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND few RWO and RWX volumes with 1 replica attached with node/pod exists
+
+AND few RWO and RWX volumes with 1 replica unattached
+
+**When** Drain the worker node one by one, start with the node which has volume's replica
+
+**Then** Drain should be blocked. Longhorn's pdb should prevent draining the IM from the node.
+
+AND Volumes should be healthy and data should be intact after drain completes
+
+AND The volumes which were attached initially should be attached and the ones which is unattached should be unattached
+
+AND Logs should not have any unexpected errors
+
+
+### 6. Longhorn volume with 50/100 Gi data
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND 1 volume with 50/100 Gi of data attached to a node
+
+**When** Drain the node using Rancher UI with default value and `Delete Empty Dir Data` enabled
+
+**Then** Drain should be successful for 2 node but should fail for last node
+
+AND Volume data should be intact
+
+AND Once there are 2 healthy replica rebuilt, then retry drain should be successful
+
+
+### 7. Drain the attached node when backup in Longhorn is in progress
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND 1 RWO volume (3 replicas) attached with 10 Gi data
+
+AND Backup is in progress
+
+**When** Drain the attached node using Rancher UI with default value and `Delete Empty Dir Data` enabled
+
+**Then** Drain should be successful
+
+AND Backup should be interrupted
+
+AND Volume's data should be intact
+
+AND Volume should be attached once the drain finishes
+
+AND Next backup should be successful
+
+### 8. Drain the replica node when backup in Longhorn is in progress
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND 1 RWO volume (3 replicas) attached with 10 Gi data
+
+AND Backup is in progress
+
+**When** Drain the node on which replica responsible for backup exists using Rancher UI
+
+**Then** Drain should be successful
+
+AND Backup should be interrupted
+
+AND Volume's data should be intact
+
+AND Volume should be attached once the drain finishes
+
+AND Next backup should be successful
+
+
+### 9. Drain operation with short timeout
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exist
+
+AND few RWO and RWX volumes unattached
+
+**When** Drain the worker node one by one with timeout 5 sec
+
+**Then** Drain should be interrupted due to timeout
+
+AND No volumes should stuck in attaching/detaching state
+
+**When** Uncordon all the nodes one by one
+
+**Then** All volumes should become healthy eventually
+
+AND There shouldn't be any data loss in the volumes
+
+
+### 10. Interrupt drain operation
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exist
+
+AND few RWO and RWX volumes unattached
+
+**When** Trigger the drain and cancel it when drain is in progress 
+
+**Then** Drain should be interrupted without causing data loss of volumes.
+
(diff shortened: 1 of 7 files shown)


docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,211 @@
...
+
+AND 1 volume with 50/100 Gi of data attached to a node
+
+**When** Drain the node using Rancher UI with default value 
+
+**Then** Drain should be successful for 2 node but should fail for last node
+
+AND Volume data should be intact
+
+AND Once there are 2 healthy replica rebuilt, then retry drain should be successful
```
yangchiu, line 96: Why there will be 2 healthy replicas after we've drained 2 nodes and only 1 node left?

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Given** Single node cluster with Longhorn installed
+
+AND 1 RWO and 1 RWX volumes attached with node
+
+**When** Drain the node using Rancher UI default value
+
+**Then** Drain should be successful and replica should fail.
+
+AND Volumes should be become faulted and auto salvage should trigger which results in healthy replica and volume.
```
yangchiu, line 207: After the only node is drained, is it possible to still have a healthy replica?

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
khushboo-rancher: I mean once there are 2 healthy replicas eventually on the nodes which got drained successfully, then retry drain should be successful on the last node.

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
khushboo-rancher: The replicas are not deleted from the disk, so Longhorn auto salvage should recover them after the drain is complete and make them healthy.

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Given**  Single node cluster with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND 1 RWO and 1 RWX volumes unattached
+
+**When** Drain the node with default values of Rancher UI
+
+**Then** Drain should be successful
```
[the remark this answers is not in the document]
khushboo-rancher: Fixed. thanks

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Given**  Single node cluster with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND 1 RWO and 1 RWX volumes unattached
+
+**When** Drain the node with default values of Rancher UI
+
+**Then** Drain should be blocked
```
roger-ryao, line 18: Regarding the test step `Drain should be blocked`, I have observed that the node status can change from `Active` to `Draining` and back to `Cordoned` in a loop. Is this the expected behavior?

chriscchien: I imported k3s into Rancher v2.7.1 in below scenario:

- 1 node (control-plane), the drain button hideen hence I can not execute drain operation
- 2 nodes (1 control and 1 worker) which controlplane tainted NoSchedule, LH only deploy on worker node, when drain the worker node, I also faced draining -> cordoned loop, because no other available node for pod can move on, so the drain not finish I think, should we consider this scenario into test steps?

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,211 @@
...
+
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND few RWO and RWX volumes unattached
+
+**When** Drain the worker node one by one with default values of Rancher UI
+
+**Then** Drain should be successful
```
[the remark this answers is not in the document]
chriscchien: I encounter the same problem in Rancher v2.7.1 + Longhorn v1.3.3-rc2
Node i`p-172-31-90-161` not drain success and some pod still running on that node
Screenshot from 2023-03-31 14-05-09

Screenshot from 2023-03-31 14-09-29

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,211 @@
...
+**When** Drain the worker node one by one with Force option enabled using Rancher UI.
+
+**Then** Drain should be successful
+
+AND Volumes should be healthy and data should be intact after drain completes
+
+AND The volumes which were attached initially should be attached and the ones which is unattached should be unattached
+
+AND Logs should not have any unexpected errors.
+
```
chriscchien, line 63: In Rancher v2.7.1 + Longhorn v1.3.3-rc2, I need enable both `Delete Empty Dir Data` and `Force` so the drain node will complete, if only enable `Force` checkbox, the drain will never complete, should we update steps?

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,211 @@
...
+**When** Drain the worker node one by one, start with the node which has volume's replica
+
+**Then** Drain should be successful
+
+AND Volumes should be healthy and data should be intact after drain completes
+
+AND The volumes which were attached initially should be attached and the ones which is unattached should be unattached
+
+AND Logs should not have any unexpected errors
+
```
chriscchien, line 82: In this case, I enabled both `Delete Empty Dir Data` and `Force` when drain from Rancher UI, I can see volume move from drain node to another node, but replica still running on drain node and from Rancher UI the node state was `cordoned` not `drained`.

In addition, if i change Node Drain Policy to `always allow`, the drain success.

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
chriscchien: I need to enable both `Delete Empty Dir Data` and `Force` to drain node which volume not attached to so the node drain success, when I try drain node which volume attached to, the drain stuck ( there still exist another available node in cluster)

docs/content/manual/rancher-integration/lh-hardend-rancher.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Longhorn in an hardened cluster
+---
+
+**Given** Hardened Downstream (RKE2/RKE1/K3s) cluster in Rancher with CIS 2.6
```
roger-ryao, line 5: In the case of Rancher 2.6.9 and 2.7.1, the CIS benchmark version used may differ, with Rancher 2.6.9 using CIS benchmark version 2.1.0/2.1.1 and Rancher 2.7.1 using CIS benchmark version 3.0.0.
Do we need to use a specific CIS benchmark version for testing?

Screenshot_20230404_200441
Screenshot_20230404_194945

khushboo-rancher: This app in the attached screenshot is for scanning.

As per https://ranchermanager.docs.rancher.com/v2.6/pages-for-subheaders/rancher-v2.6-hardening-guides CIS version is 1.6 for Rancher v2.6.x.
For v2.7 its CIS 1.6 and 1.20 & v1.23 as per https://ranchermanager.docs.rancher.com/pages-for-subheaders/rancher-v2.7-hardening-guides
I'll update the version. Thanks

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
khushboo-rancher: The draining should success on the node where the volume is not attached as far I know. Let's discuss this offline.

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,232 @@
...
+**Given**  Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exist
+
+AND few RWO and RWX volumes unattached
+
+**When** Drain the worker node one by one with timeout 5 sec
+
+**Then** Drain should be interrupted due to timeout
```
roger-ryao, line 171: Regarding the test step `Drain the worker node one by one with timeout 5 sec,` I have observed that the node status can change from `Active` to `Cordoned.` After observing that `Drain should be interrupted due to timeout,` I suggest adding a step to `Uncordon the worker node one by one` to make the node status `Active` again before verifying the volume status. What do you think?
cc @longhorn/qa

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,232 @@
...
+**Given** Single node cluster with Longhorn installed
+
+AND 1 RWO and 1 RWX volumes unattached
+
+
+**When** Drain the node using Rancher UI default value and `Delete Empty Dir Data` enabled
+
+**Then** Drain should be successful
+
+AND Volumes should be unattached and data should be intact after drain completes
```
roger-ryao, line 207: Then, after the node drain, I think we should uncordon the node using the Rancher UI.  Afterwards, we could verify that the volumes data is intact by checking that the data is still accessible. Therefore, we should modify the test step as follows:

Then Uncordon the worker node using the Rancher UI
AND volumes data is intact after uncordon completes

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,232 @@
...
+
+AND No volumes should stuck in attaching/detaching state
+
+
+Test with Longhorn setting of 'Node Drain Policy': `allow-if-replica-is-stopped`
+--------
+
+### 1. Drain using Rancher UI - Single replica:
+
+**Given** Single node cluster with Longhorn installed
```
roger-ryao, line 198: Hi @khushboo-rancher 
I think we could also modify `Given Single node cluster with Longhorn installed` to `Given Cluster with 1 etcd/control plane and 1 worker nodes with Longhorn installed`

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,232 @@
...
+AND 1 RWO and 1 RWX volumes attached with node
+
+**When** Drain the node using Rancher UI default value and `Delete Empty Dir Data` enabled
+
+**Then** Drain should be successful and replica should fail.
+
+AND Volumes should be become faulted and auto salvage should trigger which results in healthy replica and volume.
+
+### 2. Repeat the below test cases:
+1. Drain operation on a multiple node using Rancher UI - default value
```
roger-ryao, line 231: My test steps as below.

```
Given Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
AND 1 RWO and 1 RWX volumes attached with node
When Drain one of nodes using Rancher UI default value enabled
Then Drain should be successful and replica should fail.
AND Volumes should be become Degraded and auto salvage should trigger which results in healthy replica and volume.
```

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
```diff
@@ -0,0 +1,232 @@
...
+2. Drain operation on a multiple node using Rancher UI - Force
```
roger-ryao, line 232: My test steps as below.
```
Given Cluster with 1 etcd/control plane and 3 worker nodes with Longhorn installed
AND 1 RWO and 1 RWX volumes attached with node
When Drain one of nodes using Rancher UI default value and `Force` enabled
Then Drain should be successful and replica should fail.
AND Volumes should be become faulted and auto salvage should trigger which results in healthy replica and volume.
```

innobead: @roger-ryao why does draining one node caused the volume to become faulted? I thought the volume had more than one replica, it should become degraded instead.

roger-ryao: > @roger-ryao why does draining one node caused the volume to become faulted? I thought the volume had more than one replica, it should become degraded instead.

Hi @innobead :
      Thank you for pointing out my mistake. I will correct my previous comment.

docs/content/manual/rancher-integration/upgrade-using-rancher-ui.md
```diff
@@ -0,0 +1,142 @@
...
+
+**Given**  Multi node (1 master, 3 workers) imported K3s cluster with K3s prior version with Longhorn installed
+
+AND few RWO and RWX volumes attached with node/pod exists
+
+AND 1 RWO and 1 RWX volumes unattached
+
+**When** Upgrade the Kubernetes version to latest version, while upgrade is in progress kill the kubelet process on nodes
+
+**Then** Upgrade shouldn't be successful
```
roger-ryao, line 125: Hi @khushboo-rancher 
    However, the upgrade was still successful because the `kubelet` process was relaunched. I executed the script on one of the nodes and kept it running to continuously kill the kubelet process. Despite this, no volume data was lost after the successful upgrade.
 
 ```
 #!/bin/bash

while true; do
  # Find the process ID of the k3s kubelet
  pid=$(ps -ef | grep kubelet | grep -v grep | awk '{print $2}')

  if [ -z "$pid" ]; then
    echo "kubelet process not found"
  else
    # Kill the k3s kubelet process
    sudo kill $pid
    echo "kubelet process killed"
  fi

  # Wait for 5 seconds before checking again
  sleep 5
done
 ```

docs/content/manual/rancher-integration/drain-using-rancher-ui.md
chriscchien: Hi @khushboo-rancher ,
For this test case, enable `Delete Empty Dir Data` is enough and test result [here](https://github.com/longhorn/longhorn/issues/5708#issuecomment-1498770558) for your reference.

chriscchien (dismissed):
We tested this PR and LGTM

khushboo-rancher:
@chriscchien There was two commits, I fixed that. Could you please approve this again?

chriscchien (approved):

PR #1306, opened 2023-04-12
fix(expand): expansion behavior changed

mantissahz:
Check if new volume size is smaller than old one in the validator and PVC size is expanded before expanding the volume.

Expansion is not allowed if storage class does not exist.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4072,11 +4072,16 @@ def wait_and_get_pv_for_pvc(api, pvc_name):
     return pv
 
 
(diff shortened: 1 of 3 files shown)


mantissahz:
Behavior changed for https://github.com/longhorn/longhorn-manager/pull/1847

yangchiu (approved):
LGTM

mantissahz:
@Mergifyio backport v1.4.x

innobead (approved):
LGTM. 

Changing to use dyanmic volume provision is because the static volume w/ storage class will not be supported for volume expansion.

PR #1316, opened 2023-04-19
test: add case test_provisioner_fs_format

chriscchien:
Add case `test_provisioner_fs_format`

Longhorn [5428](https://github.com/longhorn/longhorn/issues/5428)

test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3505/testReport/tests/test_provisioner/)

### DIFF
--- a/manager/integration/tests/test_provisioner.py
+++ b/manager/integration/tests/test_provisioner.py
@@ -8,6 +8,8 @@ from common import generate_random_data, get_storage_api_client
 from common import get_volume_name, read_volume_data, size_to_string
 from common import write_pod_volume_data, check_volume_replicas
 from common import SETTING_REPLICA_NODE_SOFT_ANTI_AFFINITY
+from common import get_self_host_id
+import subprocess
 
 DEFAULT_STORAGECLASS_NAME = "longhorn-provisioner"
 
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_provisioner.py
```diff
@@ -196,16 +198,44 @@ def test_provisioner_tags(client, core_api, node_default_tags, storage_class, pv
...
+    storage_class['parameters']['mkfsParams'] = \
+        "-I {} -b {} -O ^metadata_csum,^64bit".format(inode_size, block_size)
+
+    create_storage(core_api, storage_class, pvc)
+    create_and_wait_pod(core_api, pod)
+    pvc_volume_name = get_volume_name(core_api, pvc['metadata']['name'])
+
+    command = "tune2fs -l /dev/longhorn/{} | grep -i 'block size:'"\
+        .format(pvc_volume_name)
+    assert block_size in str(subprocess.check_output(command, shell=True))
```
khushboo-rancher, line 237: Just to understand, why are we using `shell=True` here?

chriscchien: Let pipe command can execute in 1 line command or If I remember correctly it will need using extra codes to fetch out the output of `grep`

khushboo-rancher (changes requested):
Should we try to write into the volume at the end to ensure that the volume is behaving normally?

khushboo-rancher (approved):
Thanks @chriscchien

khushboo-rancher:
@Mergifyio backport v1.4.x

PR #1318, opened 2023-04-21
ci: add aks provisioning

yangchiu:
ci: add aks provisioning

For https://github.com/longhorn/longhorn/issues/4621

### DIFF
--- a/test_framework/terraform/azure/aks/main.tf
+++ b/test_framework/terraform/azure/aks/main.tf
@@ -1,0 +1,60 @@
+terraform {
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "=3.0.0"
+    }
+  }
+}
+
+# Configure the Microsoft Azure Provider
+provider "azurerm" {
+  features {}
+
+  client_id                   = var.azure_client_id
+  client_certificate_path     = var.azure_crt_path
+  client_certificate_password = var.azure_crt_password
+  tenant_id                   = var.azure_tenant_id
+  subscription_id             = var.azure_subscription_id
+}
+
+resource "random_string" "random_suffix" {
+  length           = 8
+  special          = false
+  lower            = true
+  upper            = false
+}
+
+resource "azurerm_resource_group" "resource_group" {
+  name     = "${var.test_name}-${random_string.random_suffix.id}-resources"
+  location = "Southeast Asia"
+}
+
+resource "azurerm_kubernetes_cluster" "cluster" {
+  name                = "${var.test_name}-${random_string.random_suffix.id}-cluster"
+  location            = azurerm_resource_group.resource_group.location
+  resource_group_name = azurerm_resource_group.resource_group.name
+  dns_prefix          = "${var.test_name}-${random_string.random_suffix.id}"
+
+  default_node_pool {
+    name       = "default"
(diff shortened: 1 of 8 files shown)


manager/integration/deploy/test.yaml
```diff
@@ -29,7 +29,7 @@ rules:
   resources: ["customresourcedefinitions"]
   verbs: ["*"]
 - apiGroups: ["policy"]
-  resources: ["podsecuritypolicies"]
+  resources: ["podsecuritypolicies", "poddisruptionbudgets"]
```
innobead, line 32: Same need as GKE, but why?

test_framework/terraform/azure/aks/main.tf
```diff
@@ -0,0 +1,60 @@
...
+resource "random_string" "random_suffix" {
+  length           = 8
+  special          = false
+  lower            = true
+  upper            = false
+}
+
+resource "azurerm_resource_group" "resource_group" {
+  name     = "${var.test_name}-${random_string.random_suffix.id}-resources"
+  location = "Southeast Asia"
```
innobead, line 30: nit: can be variable instead of hard code for region.

innobead (changes requested):
LGTM. The same as other distro, there are some questions needed to answer first.

test_framework/terraform/azure/aks/main.tf
yangchiu: Set the region/location as a variable for each managed k8s services means we have to provide options/parameters on the Jenkins pipeline, but currently all types of regression pipelines share the same `Jenkinsfile`, if these parameters have to added, they will be all added in the same `Jenkinsfile`, and it would cause there are redundant parameters only used in some specific pipeline, the parameters initialization could be tedious too. Suggest if we really need it running on different region/location, refactor the `Jenkinsfile` and the setup scripts to make each pipeline more independent.

manager/integration/deploy/test.yaml
yangchiu: This follows the CA test case PR: https://github.com/longhorn/longhorn-tests/pull/936/files#diff-deb01c8f2cd2d469bfeed0826ebac624022e4b668a4f7d1d2f04d4d9c22d5f82R32.

innobead: I see. This seems related to PDB by IM. However, I did not see any PDB operations in our test case.

c3y1huang: I don't remember. We can try to remove it to see if it blocks.

test_framework/terraform/azure/aks/main.tf
innobead: @yangchiu let's create an issue for this to improve this later. I think hard code is not the right way.

manager/integration/deploy/test.yaml
innobead: @yangchiu when doing another round, we can see if the resource authorization is needed or not. Not a priority for now.

test_framework/terraform/azure/aks/main.tf
yangchiu: Created https://github.com/longhorn/longhorn/issues/5960 for this.

innobead (approved):

PR #1321, opened 2023-04-25
test: fix flaky test case test_cloning_with_detached_source_volume

yangchiu:
test: fix flaky test case test_cloning_with_detached_source_volume by deferring the snapshot creation check

For https://github.com/longhorn/longhorn/issues/4848

The source volume will be automatically attached and detached during the clone process, since it will be detached immediately after the clone is completed, it's nearly impossible to always catch the moment of the source volume is still attached and get the snapshot list. Defer the snapshot creation check to the end of the test case by manually attaching the source volume.

### DIFF
--- a/manager/integration/tests/test_cloning.py
+++ b/manager/integration/tests/test_cloning.py
@@ -198,13 +198,13 @@ def test_cloning_with_detached_source_volume(client, core_api, pvc, clone_pvc): 
                   storage: 10Gi
             ```
         6. Wait for `source-pvc` to be attached
-        7. Wait for a new snapshot created in `source-pvc` volume created
-        8. Wait for the `CloneStatus.State` in `cloned-pvc` to be `completed`
-        9. Wait for `source-pvc` to be detached
-        10. Attach the cloned volume to a node
-        11. Verify the data in `cloned-pvc` is the same as in `source-pvc`.
-        12. In 2-min retry loop, verify the volume of the `clone-pvc`
+        7. Wait for the `CloneStatus.State` in `cloned-pvc` to be `completed`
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_cloning.py
```diff
@@ -244,29 +244,32 @@ def test_cloning_with_detached_source_volume(client, core_api, pvc, clone_pvc):
...
     check_volume_data(clone_volume, data)
 
-    # Step-12
+    # Step-11
     wait_for_volume_healthy(client, clone_volume_name)
 
+    # Step-12
+    source_volume = client.by_id_volume(source_volume_name)
+    source_volume.attach(hostId=lht_host_id)
+    source_volume = wait_for_volume_healthy(client, source_volume_name)
```
khushboo-rancher, line 270: I think the appropriate check would be to wait for volume to get attached instead of healthy.

khushboo-rancher (changes requested):
In general LGTM, one minor comment.

manager/integration/tests/test_cloning.py
yangchiu: Can you elaborate more about the difference?

khushboo-rancher: `wait_for_volume_healthy` has more checks than `wait_for_volume_attached`. It will first check whether volume is attached and then checks about the robustness of the volume. So if you just intend to check the snapshot count, after the volume is attached, you should be able to get that. I don't think the intention here to check the robustness.

yangchiu: Thank you for the explanation! Refined!

PR #1324, opened 2023-04-26
Add case test_csi_block_volume_online_expansion

chriscchien:
Test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3600/)

Ref [5820](https://github.com/longhorn/longhorn/issues/5820)

Signed-off-by: Chris <chris.chien@suse.com>

### DIFF
--- a/manager/integration/tests/test_csi.py
+++ b/manager/integration/tests/test_csi.py
@@ -31,7 +31,11 @@ from common import create_pv_for_volume, create_pvc_for_volume
 from common import get_self_host_id, get_volume_endpoint
 from common import wait_for_volume_healthy
 from common import fail_replica_expansion
+from common import get_volume_name, get_volume_dev_mb_data_md5sum # NOQA
 from backupstore import set_random_backupstore  # NOQA
+from kubernetes.stream import stream
+from kubernetes import client as k8sclient
+import threading
 
 # Using a StorageClass because GKE is using the default StorageClass if not
 # specified. Volumes are still being manually created and not provisioned.
@@ -371,7 +375,6 @@ def test_csi_offline_expansion(client, core_api, storage_class, pvc, pod_manifes
     assert volume.size == engine.size
 
 
-@pytest.mark.skip(reason="TODO")  # NOQA
 @pytest.mark.csi  # NOQA
 @pytest.mark.csi_expansion  # NOQA
 def test_csi_block_volume_online_expansion(client, core_api, storage_class, pvc, pod_manifest):  # NOQA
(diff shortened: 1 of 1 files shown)


khushboo-rancher:
@chriscchien Can you please link the ticket this is associated with?
@yangchiu Could you please help with the review?

yangchiu (commented):
According to the test skeleton:
```
3. Generate `test_data` in a directory the volume is not mounted on.
4. Copy the `test data` to the volume block device asynchronously.
```
Do we actually do this in this implementation?

chriscchien:
> According to the test skeleton:
> 
> ```
> 3. Generate `test_data` in a directory the volume is not mounted on.
> 4. Copy the `test data` to the volume block device asynchronously.
> ```
> 
> Do we actually do this in this implementation?

I merged step 3 and 4 by using `dd` command with suprogress.pipe(), while the pipe is still executing, do volume expansion mentioned in step 5

It is because if I first generate test-data by `dd if=/dev/urandom of=/tmp/test-data bs=1M count=2500` which mentioned in step 3 and then immediately copy test-data into volume device by `dd if=/tmp/test-data of=/dev/<volume>`, the data in `/dev/<volume>` is depend on the time when I start to copy, not the whole test-data.

yangchiu:
> > I merged step 3 and 4 by using `dd` command with suprogress.pipe(), while the pipe is still executing, do volume expansion mentioned in step 5
> 
> > It is because if I first generate test-data by `dd if=/dev/urandom of=/tmp/test-data bs=1M count=2500` which mentioned in step 3 and then immediately copy test-data into volume device by `dd if=/tmp/test-data of=/dev/<volume>`, the data in `/dev/<volume>` is depend on the time when I start to copy, not the whole test-data.


If the implementation is different from the skeleton, can we modify the skeleton to make them consistent?

chriscchien:
> > > I merged step 3 and 4 by using `dd` command with suprogress.pipe(), while the pipe is still executing, do volume expansion mentioned in step 5
> > 
> > 
> > > It is because if I first generate test-data by `dd if=/dev/urandom of=/tmp/test-data bs=1M count=2500` which mentioned in step 3 and then immediately copy test-data into volume device by `dd if=/tmp/test-data of=/dev/<volume>`, the data in `/dev/<volume>` is depend on the time when I start to copy, not the whole test-data.
> 
> If the implementation is different from the skeleton, can we modify the skeleton to make them consistent?

Hi @yangchiu,updated.

yangchiu (approved):
LGTM

khushboo-rancher (approved):

PR #1350, opened 2023-05-10
ci: refine sles ami filter to avoid using ecs optimized image

yangchiu:
ci: refine sles ami filter to avoid using ecs optimized image

it's possible that unable to run any pods on the ecs optimized instance and then causes longhorn installation failed.


innobead (approved):

yangchiu:
@Mergifyio backport v1.4.x v1.3.x

PR #1357, opened 2023-05-12
test: fix upgrade test failed by checking instance manager with backward compatibility

yangchiu:
test: fix upgrade test failed by checking instance manager with backward compatibility

For https://github.com/longhorn/longhorn/issues/5907


yangchiu:
If fresh install v1.4.x, there will be instance-manager-e/r only.
If fresh install v1.5.0-dev, there will be instance-manager only.
If upgrade from v1.4.x to v1.5.0-dev, there will be instance-manager-e/r and instance-manager at the same time.

manager/integration/tests/common.py
```diff
@@ -3377,20 +3377,42 @@ def wait_for_all_instance_manager_running(client):
...
+                elif im.managerType == "aio" and im.currentState == "running":
                     node_to_instance_manager_map[im.nodeID] = im
                 else:
                     print("\nFound unknown instance manager:", im)
-            if len(node_to_instance_manager_map) != len(nodes):
+            if len(node_to_instance_manager_map) + \
+                len(node_to_engine_manager_map) + \
+                    len(node_to_replica_manager_map) < len(nodes):
+                time.sleep(RETRY_INTERVAL)
+                continue
```
innobead, line 3397: What does this mean? Why we need to check this condition?

manager/integration/tests/common.py
```diff
@@ -3377,20 +3377,42 @@ def wait_for_all_instance_manager_running(client):
...
+            elif 0 < len(node_to_instance_manager_map) < len(nodes) or \
+                0 < len(node_to_engine_manager_map) < len(nodes) or \
+                    0 < len(node_to_replica_manager_map) < len(nodes):
                 time.sleep(RETRY_INTERVAL)
                 continue
```
innobead, line 3402: This will always sleep and retry util finishing the all rounds.

innobead (commented):
From the https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-upgrade-tests-sles-amd64/458/execution/node/28/log/?consoleFull, it seems failure happened in the upgrade_longhorn.sh, but what this fix can help with that?

Can you expalin a bit?

```shell
+ [[ 9 -eq 10 ]]
++ kubectl get pods -n longhorn-system --no-headers
++ grep -v Running
++ awk '{print $3}'
+ [[ -n Init:0/1
CrashLoopBackOff
CrashLoopBackOff
CrashLoopBackOff ]]
+ echo 'Longhorn is still installing ... re-checking in 1m'
+ sleep 1m
Longhorn is still installing ... re-checking in 1m
Error: longhorn installation timeout
+ RETRIES=10
+ [[ 10 -eq 10 ]]
+ echo 'Error: longhorn installation timeout'
+ exit 1
```

innobead:
wait_for_all_instance_manager_running seems best effort, it should not be able to break any test cases? or anything I missed cc @c3y1huang

yangchiu:
> From the https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-upgrade-tests-sles-amd64/458/execution/node/28/log/?consoleFull, it seems failure happened in the upgrade_longhorn.sh, but what this fix can help with that?
> 
> Can you expalin a bit?
> 
> ```shell
> + [[ 9 -eq 10 ]]
> ++ kubectl get pods -n longhorn-system --no-headers
> ++ grep -v Running
> ++ awk '{print $3}'
> + [[ -n Init:0/1
> CrashLoopBackOff
> CrashLoopBackOff
> CrashLoopBackOff ]]
> + echo 'Longhorn is still installing ... re-checking in 1m'
> + sleep 1m
> Longhorn is still installing ... re-checking in 1m
> Error: longhorn installation timeout
> + RETRIES=10
> + [[ 10 -eq 10 ]]
> + echo 'Error: longhorn installation timeout'
> + exit 1
> ```

Yes, you are right!

This PR only suppresses the overwhelming warning messages and doesn't affect the result.

The `CrashLoopBackOff` is due to upgrade path enforcement and needs to update the upgrade path of daily regression pipelines to v1.4.x to fix the error.

And after updated the upgrade path, the upgrade test still fails, more investigation needed.

Sorry for the confusing!

innobead:
cc @mantissahz this is about upgrade path enforcement. Please work with @yangchiu to figure out why the upgrade test failed.

yangchiu:
Close this one and updated the [ticket](https://github.com/longhorn/longhorn/issues/5907).

PR #1361, opened 2023-05-15
test: fix flaky test case test_snapshot_hash_detect_corruption

yangchiu:
test: fix flaky test cases test_snapshot_hash_detect_corruption_in_global_enabled_mode and test_snapshot_hash_detect_corruption_in_global_fast_check_mode

(1) cronjob setting: in the current implementation, it's possible that the cronjob is set to something like: 59/5 6 * * *. This cronjob will only be executed once per day on 6:59 AM, and if the test case missed this chance to trigger the integrity check, it has to wait until tomorrow, so the test case will timeout and fail. This PR changes the cronjob to something like: 59,4,9,14,19,24,29 6,7 * * * (every 5 minutes in the upcoming 30 minutes), so if 6:59 AM is missed, it can still be triggered on 7:04 AM.  

(2) wait toleration: in the current implementation, it has a wait toleration 60s to handle that we're unable to precisely predict when will the rebuilding caused by corrupted replica is triggered, but even though there is a 60s toleration, this test case is still flaky, this test case will still fail due to timeout waiting for the rebuilding. Since the cronjob is set to every 5 minutes, if the 1st timing is somehow missed, we have to wait another 5 minutes, so this PR change to the wait timeout to 5 minutes * 2 instead of 5 minutes + 60s.

For https://github.com/longhorn/longhorn/issues/5394

### DIFF
--- a/manager/integration/tests/test_snapshot.py
+++ b/manager/integration/tests/test_snapshot.py
@@ -261,7 +261,8 @@ def test_snapshot_hash_detect_corruption_in_global_enabled_mode(client, volume_n
                                        "true",
                                        "false")
     detect_and_repair_corrupted_replica(client, volume_name,
-                                        SNAPSHOT_DATA_INTEGRITY_ENABLED)
+                                        SNAPSHOT_DATA_INTEGRITY_ENABLED,
+                                        retry_count=SNAPSHOT_CHECK_PERIOD * 2)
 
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead:
@mergify backport v1.4.x

PR #1369, opened 2023-05-17
feat(backup test): add storage class name from backupvolume e2e skeleton

ChanYiLin:
ref: [longhorn/longhorn#3506](https://github.com/longhorn/longhorn/issues/3506)


manager/integration/tests/test_basic.py
```diff
@@ -1775,6 +1775,63 @@ def test_volume_multinode(client, volume_name):  # NOQA
...
+    Test the storageClasName of the restored volume's PV/PVC 
+    should be from the backup volume
+
+    Given
+    - Create a new StorageClass
+      ```
+      kind: StorageClass
+      apiVersion: storage.k8s.io/v1
+      metadata:
+        name: longhorn-backing-image-sc
```
khushboo-rancher, line 1790: We can change the name related to the test.

manager/integration/tests/test_basic.py
```diff
@@ -1775,6 +1775,63 @@ def test_volume_multinode(client, volume_name):  # NOQA
...
+      provisioner: driver.longhorn.io
+      allowVolumeExpansion: true
+      reclaimPolicy: Delete
+      volumeBindingMode: Immediate
+      parameters:
+        numberOfReplicas: "2"
+        staleReplicaTimeout: "2880"
+        backingImage: "bi-download"
+        backingImageDataSourceType: "download"
+        backingImageDataSourceParameters: '{"url": "https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.raw"}' # NOQA
```
khushboo-rancher, line 1800: Are these required for this test?

manager/integration/tests/test_basic.py
```diff
@@ -1775,6 +1775,63 @@ def test_volume_multinode(client, volume_name):  # NOQA
...
+      ```
+    - Create a PVC to use this SC
+      ```
+      apiVersion: v1
+      kind: PersistentVolumeClaim
+      metadata:
+        name: longhorn-backing-image-pvc
```
khushboo-rancher, line 1807: Name can be changed to something related to the test

manager/integration/tests/test_basic.py
```diff
@@ -1775,6 +1775,63 @@ def test_volume_multinode(client, volume_name):  # NOQA
...
+    Then
+    - the backupvolume's status.storageClassName should be longhorn-backing-image-sc
+
+    When
+    - Restore the backup to a new volume
+    - Create PV/PVC from the new volume with create new PVC option
+
+    Then
+    - The new PVC's storageClassName should still be longhorn-backing-image-sc
+    """
```
khushboo-rancher, line 1831: Can you please add a step to verify the data

manager/integration/tests/test_basic.py
```diff
@@ -1775,6 +1775,63 @@ def test_volume_multinode(client, volume_name):  # NOQA
...
+        name: longhorn-backing-image-pvc
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        storageClassName: longhorn-backing-image-sc
+        resources:
+          requests:
+            storage: 5Gi
+      ```
+    - A BackingImage `bi-download` and a Volume will be created
```
khushboo-rancher, line 1816: We can avoid having additional step of creating backing image, WDYT?

khushboo-rancher (changes requested):
Can we add more steps in the test to make `status.storageClassName` empty in backup and on restoring it, the storage class should pick the storage class from the default in setting?

manager/integration/tests/test_basic.py
ChanYiLin: You are right, we don't need backing image involved.

manager/integration/tests/test_basic.py
ChanYiLin: Yes I agree. Thanks

ChanYiLin:
Hi @khushboo-rancher sorry for the late reply
for storageClassName empty case, this can only happen in upgrade after v1.5.0
for example, having a backup in v1.4.x and then upgrade to v1.5.0 then restore the volume and create the PVC

khushboo-rancher (approved):

khushboo-rancher:
@mergify backport v1.5.x

PR #1372, opened 2023-05-18
test(recurring-job): PVC label

c3y1huang:
https://github.com/longhorn/longhorn/issues/5791

### DIFF
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -2152,6 +2152,134 @@ def test_recurring_job_filesystem_trim(client, core_api, batch_v1_api, volume_na
     assert size_trimmed == test_size
 
 
+@pytest.mark.recurring_job
+def test_recurring_job_label_on_pvc(client, core_api, volume_name):  # NOQA
+    """
+    Scenario: test recurring job label on PVC
+
+    Given volume created.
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_recurring_job.py
```diff
@@ -2152,6 +2152,236 @@ def test_recurring_job_filesystem_trim(client, core_api, batch_v1_api, volume_na
...
+    unexpected_count = 0
+    for label in claim.metadata.labels:
+        if not label.startswith("recurring-job"):
+            continue
+
+        if label == "recurring-job.longhorn.io/source":
+            continue
+
+        unexpected_count += 1
+    assert unexpected_count == 0
```
innobead, line 2266: Why need this check? the labels of the PVC are deleted and volume labels also, so what do you want to check?

innobead (changes requested):
In general, LGTM. Just one comment.

manager/integration/tests/test_recurring_job.py
c3y1huang: The purpose of this check is to ensure when the Volume gets updated with the default recurring job (or any recurring job), the changes do not get reflected back to the PVC.

innobead (approved):

PR #1371, opened 2023-05-18
Fix e2e tests

PhanLe1010:
- [x] Fix test test_backing_image.py::test_csi_mount_with_backing_image
        If `volume.spec.fromBackup` is empty, do not wait for volume restoration finished because the volume doesn't require restoration
- [x] test_backing_image.py::test_csi_io_with_backing_image
- [x] test_backing_image.py::test_csi_backup_with_backing_image
- [x] test_csi.py::test_csi_mount
- [x] test_csi.py::test_csi_io
- [x] test_csi.py::test_csi_backup
- [x] test_csi_snapshotter.py::test_csi_snapshot_snap_create_csi_snapshot
- [x] test_csi_snapshotter.py::test_csi_snapshot_snap_create_volume_from_snapshot
- [x] test_csi_snapshotter.py::test_csi_snapshot_snap_delete_csi_snapshot_volume_detached
- [x] test_node.py::test_replica_scheduler_too_large_volume_fit_any_disks
- [x] test_ha.py::test_inc_restoration_with_multiple_rebuild_and_expansion (Flaky test. Existed before https://github.com/longhorn/longhorn/issues/5460)
- [x] test_ha.py::test_auto_remount_with_subpath (No fix but retested 3 times and all passed. Will wait for next daily test run)
- [x] test_ha.py::test_engine_image_not_fully_deployed_perform_auto_upgrade_engine (No fix but retested 5 times and all passed. Will wait for next daily test run)
- [x] test_settings.py::test_setting_concurrent_volume_backup_restore_limit_should_not_effect_dr_volumes (No fix but passed manual test run. Will wait for next daily test run)
- [x] test_cloning.py::test_cloning_with_detached_source_volume (root cause is same as https://github.com/longhorn/longhorn/issues/5835)
- [x] test_ha.py::test_autosalvage_with_data_locality_enabled (No fix but retested 3 times and all passed. Will wait for next daily test run)

PR #1373, opened 2023-05-19
test: update default storage over provisioning percentage

c3y1huang:
https://github.com/longhorn/longhorn-tests/pull/1371#discussion_r1198502881
https://github.com/longhorn/longhorn/issues/2694


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3859/

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3859/

manager/integration/tests/test_node.py
```diff
@@ -498,12 +498,6 @@ def test_replica_scheduler_too_large_volume_fit_any_disks(client):  # NOQA
     6. Attach the volume.
     7. Make sure every replica landed on different nodes's default disk.
     """
-
-    over_provisioning_setting = client.by_id_setting(
-        SETTING_STORAGE_OVER_PROVISIONING_PERCENTAGE)
-    client.update(over_provisioning_setting,
-                  value=DEFAULT_STORAGE_OVER_PROVISIONING_PERCENTAGE)
-
```
PhanLe1010, line 506: If we remove this setting, I think we need to add some logic to pytest.fixture client to reset the over_provisioning_setting to DEFAULT_STORAGE_OVER_PROVISIONING_PERCENTAGE before each test. Otherwise, this setting might have value from previous tests, which makes the tests like this one flaky. WDYT?

manager/integration/tests/test_node.py
```diff
@@ -652,18 +646,11 @@ def test_replica_scheduler_exceed_over_provisioning(client):  # NOQA
...
     3. Try to schedule a volume of 2G. Volume scheduled condition should be
     false
     """
-    over_provisioning_setting = client.by_id_setting(
-        SETTING_STORAGE_OVER_PROVISIONING_PERCENTAGE)
-    old_provisioning_setting = over_provisioning_setting.value
-    # set storage over provisioning percentage to 100
-    over_provisioning_setting = client.update(over_provisioning_setting,
-                                              value="100")
-
```
PhanLe1010, line 666: Same as above

manager/integration/tests/test_node.py
```diff
@@ -689,27 +676,19 @@ def test_replica_scheduler_exceed_over_provisioning(client):  # NOQA
...
     4. Volume scheduled condition should be true.
     5. Make sure every replica landed on different nodes's default disk.
     """
-    over_provisioning_setting = client.by_id_setting(
-        SETTING_STORAGE_OVER_PROVISIONING_PERCENTAGE)
-    old_provisioning_setting = over_provisioning_setting.value
-    # set storage over provisioning percentage to 100
-    over_provisioning_setting = client.update(over_provisioning_setting,
-                                              value="100")
-
```
PhanLe1010, line 712: Same as above

manager/integration/tests/test_node.py
c3y1huang: Can you elaborate on how this setting would have value from previous tests?

IIRC, the client fixture already includes the [cleanup_client](https://github.com/longhorn/longhorn-tests/blob/83d3cf9178696830587792cd081541fc89cb5ad9/manager/integration/tests/common.py#L1548-L1550) which [resets the settings](https://github.com/longhorn/longhorn-tests/blob/83d3cf9178696830587792cd081541fc89cb5ad9/manager/integration/tests/common.py#L1599) to their [default values](https://github.com/longhorn/longhorn-tests/blob/83d3cf9178696830587792cd081541fc89cb5ad9/manager/integration/tests/common.py#L3240). Currently, the default value is set to [100](https://github.com/longhorn/longhorn-manager/blob/4410d5c1a807928de4c9049a9578a074d6a70813/types/setting.go#L458) which is the same as the DEFAULT_STORAGE_OVER_PROVISIONING_PERCENTAGE.

innobead (approved):
LGTM

PR #1374, opened 2023-05-22
test_engine_image_not_fully_deployed_perform_dr_restoring_expanding_volume

chriscchien:
Add test case `test_engine_image_not_fully_deployed_perform_dr_restoring_expanding_volume`

ref: [5974](https://github.com/longhorn/longhorn/issues/5974)

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -3028,8 +3028,7 @@ def test_engine_image_not_fully_deployed_perform_auto_upgrade_engine(client, cor
     assert volume2.engineImage == default_img.image
 
 
-@pytest.mark.skip(reason="TODO") # NOQA
-def test_engine_image_not_fully_deployed_perform_dr_restoring_expanding_volume(): # NOQA
+def test_engine_image_not_fully_deployed_perform_dr_restoring_expanding_volume(client, core_api, set_random_backupstore): # NOQA
     """
     Test DR, restoring, expanding volumes when engine image DaemonSet
     is not fully deployed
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -3060,7 +3059,134 @@ def test_engine_image_not_fully_deployed_perform_dr_restoring_expanding_volume()
...
+    wait_for_backup_restore_completed(client, dr_volume_name, b1.name)
+    dr_volume = client.by_id_volume(dr_volume_name)
+
+    # step 4
+    on_node2 = False
+    on_node3 = False
+    for replica in dr_volume.replicas:
+
+        print()
+        if replica.hostId == node2.id:
```
innobead, line 3109: Print this dummy print.

Also plz remove unnecessary empty line. Usually we only use empty lines to group the related code for better understanding.

innobead (changes requested):
In general, LGTM.

Some comments need to resolve, and also ensure the test cass run well.

manager/integration/tests/test_ha.py
chriscchien: Sorry for the wrong code format, I will pay more attention in the future

chriscchien:
Updated code and here is the test result on [pipeline](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3917/testReport/tests/test_ha/), thank you

innobead (approved):

innobead:
@mergify backport v1.4.x v1.3.x

PR #1378, opened 2023-05-23
test: add try/except for generate support bundle function to avoid internal error

yangchiu:
test: add try/except for generate support bundle function to avoid internal error

In gke environment, the credential will keep expire and automatically renew during the test running. If the test tries to make requests when the credntial expired and without this try/except, it would cause internal error and fail the whole test process instead of just failing the current test case.


innobead:
It's about K8s access credentials.

innobead (approved):
LGTM.

This is a workaround. We can create a ticket for the later investigation which is low priority.

PR #1379, opened 2023-05-24
test: enable cluster autoscaler test and add option for it

yangchiu:
test: enable cluster autoscaler test and add option for it


manager/integration/tests/conftest.py
```diff
@@ -174,6 +179,15 @@ def pytest_collection_modifyitems(config, items):
             if "upgrade" in item.keywords:
                 item.add_marker(skip_upgrade)
 
+    if not config.getoption(INCLUDE_CA_OPT):
+        skip_upgrade = pytest.mark.skip(reason="include " +
```
innobead, line 183: Should use right var name for each skipped test cases instead of `skip_upgrade`. Please fix this in `pytest_collection_modifyitems` generally later.

innobead (approved):

PR #1380, opened 2023-05-24
test: adjust cluster autoscaler parameters in eks

yangchiu:
test: adjust cluster autoscaler parameters in eks

adjust cluster autoscaler parameters in eks to make it possible to scale down before timeout in the test cases


innobead (approved):
LGTM.

Good job.

PR #1383, opened 2023-05-24
test: refine pipeline to let cluster autoscaler test pass

yangchiu:
test: refine pipeline to let cluster autoscaler test pass

(1) Continue and retry the support bundle downloading process even though encounter errors. When it tries to access longhorn frontend and get downloading progress, it's possible that the service is temporarily unavailable due to container restarting (it could happen in cluster autoscaler test). So ignore the error to let the downloading process can continue.

(2) Do not install backup store in cluster autoscaler test, otherwise the test_cluster_autoscaler_all_nodes_with_volume_replicas and test_cluster_autoscaler_backing_image test case will fail because autoscaler tries to scale down the node which backup store is running on.

(3) Adjust aks to use machine type with lower cpu capacity. Whether cluster autoscaler test cases will pass or not highly depends on the machine type (cpu capacity). It implicitly expects the cluster scales up/down 2 nodes. If the cluster eventually scales up/down either 1 node or 3 nodes, the test cases will fail. On the other hand, cluster autoscaler will scale up a node only if there are unschedulable pods, and once there's no unschedulable pod, cluster autoscaler will stop.
For aks, if using Standard_D3_v2 (4 cpus), it only needs to scale up 1 node to let all pods running, so the test cases will fail. If using Standard_D2_v2 (2 cpus), it needs to scale up 2 node to let all pods running, so the test cases can pass.
For gke, currently using e2-standard-2 (2 cpus), and it needs to scale up 2 nodes to let all pods running, so the test cases can pass.
For eks, no matter using t2.xlarge (4 cpus) or t2.large (2 cpus), it scales up 1 node or 3 nodes and doesn't meet the test cases' expectation, so the test cases will all fail. Currently unable to figure out a setting to let them pass on eks.


innobead (approved):
LGTM.

Ideally, the backup store should run outside the cluster. Let's create a ticket for the later improvement.  @yangchiu

PR #1388, opened 2023-05-29
test(system-backup): volume backup policies

c3y1huang:
https://github.com/longhorn/longhorn/issues/5011

### DIFF
--- a/manager/integration/tests/test_system_backup_restore.py
+++ b/manager/integration/tests/test_system_backup_restore.py
@@ -8,6 +8,7 @@ from common import cleanup_volume
 from common import create_and_check_volume
 from common import create_backup
 from common import get_self_host_id
+from common import system_backups_cleanup
 from common import system_backup_random_name
 from common import system_backup_wait_for_state
 from common import system_restore_random_name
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_system_backup_restore.py
```diff
@@ -100,3 +106,118 @@ def test_system_backup_and_restore_volume_with_data(client, volume_name, set_ran
...
+    And system backup (system-backup-1) created.
+    Then system backup has volume backup policy (if-not-present).
+    And system backup is in state (Ready).
+    And volume has backup count (1).
+
+    When system backup (system-backup-2) has volume backup policy
+         (if-not-present).
+    And system backup (system-backup-2) created.
+    Then system backup is in state (Ready).
+    And volume has backup count (1).
```
khushboo-rancher, line 132: Can we extend the test case to verify  - if the existing backup doesn't have the latest data, the system backup should trigger a volumes backup?

Next steps:
When Data is written in the volume (No backup taken for the volume)
AND system backup (system-backup-3) created.
Then system backup is in state (Ready).
And volume has backup count (2).

khushboo-rancher (changes requested):
Thanks @c3y1huang, This looks good. Just one comment about adding more checks.

manager/integration/tests/test_system_backup_restore.py
c3y1huang: The if-not-present policy only checks if the volume has an existing backup, it doesn't check if the latest backup reflects the most recent data.

khushboo-rancher: oh okay, do you think we should file a ticket to improve this one in future?
cc: @innobead

innobead: We should update the doc as well to explain it well if needed.  Sounds good to me to create a ticket for the improvement later.

c3y1huang: https://github.com/longhorn/longhorn/issues/6027

khushboo-rancher (approved):

PR #1398, opened 2023-05-31
ci: refactor to have separate gke pipeline and setup scripts

yangchiu:
ci: refactor to have separate gke pipeline and setup scripts

To be able to run gke test, we need to install gcloud in a test build, but it causes each test build occupies significant amount of disk space (> 4GB) and causes infra node disk pressure and affects the services if lots of test builds are triggered at the same time. 
So refactor to separate gke pipeline first to avoid installing gcloud in every build.

For https://github.com/longhorn/longhorn/issues/6031, https://github.com/longhorn/longhorn/issues/5936


innobead (approved):
LGTM.

This is a special case.

PR #1407, opened 2023-06-09
doc(manual test): update the setting name for drain node policy

ChanYiLin:
Update the manual test plan since the setting has been changed


innobead (approved):
LGMT.

It's ok to replace the `allow node drain w/ the last healthy replica` with the new node drain policy, because the first one was deprecated and removed from 1.5.

innobead:
cc @longhorn/qa

PR #1408, opened 2023-06-09
fix: change the command for killing processes

mantissahz:
It needs to change the command instead `awk` because it is not installed in default for the `registry.suse.com/bci/bci-base:15.4`.

Use `pgrep -f` to get the process id.

Regression test for `test_ha.py`:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4055

Local tests
```text
platform linux -- Python 3.9.16, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/bin/python3.9
cachedir: .pytest_cache
plugins: order-1.0.1, repeat-0.9.1
collecting ... collected 11 items

test_recurring_job.py::test_recurring_jobs_when_volume_detached_unexpectedly[s3] PASSED [  9%]
test_recurring_job.py::test_recurring_jobs_when_volume_detached_unexpectedly[nfs] FAILED [ 18%]
test_scheduling.py::test_data_locality_basic PASSED                      [ 27%]
test_basic.py::test_running_volume_with_scheduling_failure PASSED        [ 36%]
test_basic.py::test_expansion_with_scheduling_failure PASSED             [ 45%]
test_basic.py::test_backup_failed_enable_auto_cleanup[s3] PASSED         [ 54%]
test_basic.py::test_backup_failed_enable_auto_cleanup[nfs] PASSED        [ 63%]
test_basic.py::test_backup_failed_disable_auto_cleanup[s3] PASSED        [ 72%]
test_basic.py::test_backup_failed_disable_auto_cleanup[nfs] PASSED       [ 81%]
test_cloning.py::test_cloning_interrupted PASSED                         [ 90%]
test_settings.py::test_setting_concurrent_rebuild_limit FAILED           [100%]

```

Ref: longhorn/longhorn#5460

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2194,8 +2194,7 @@ def crash_replica_processes(client, api, volname, replicas=None,
 
     for r in replicas:
         assert r.instanceManagerName != ""
-        kill_command = "kill `ps aux | grep '" + r['dataPath'] +\
-                       "' | grep -v grep | awk '{print $2}'`"
+        kill_command = "kill `pgrep -f " + r['dataPath'] + "`"
         exec_instance_manager(api, r.instanceManagerName, kill_command)
 
         if wait_to_fail is True:
@@ -4772,8 +4771,7 @@ def crash_engine_process_with_sigkill(client, core_api, volume_name):
 
     kill_command = [
             '/bin/sh', '-c',
-            "kill -9 `ps aux | grep -i \"controller " +
-            volume_name + "\" | grep -v grep | awk '{print $2}'`"]
+            "kill `pgrep -f \"controller " + volume_name + "\"`",]
 
     with timeout(seconds=STREAM_EXEC_TIMEOUT,
                  error_message='Timeout on executing stream read'):


manager/integration/tests/common.py
```diff
@@ -2194,8 +2194,7 @@ def crash_replica_processes(client, api, volname, replicas=None,
 
     for r in replicas:
         assert r.instanceManagerName != ""
-        kill_command = "kill `ps aux | grep '" + r['dataPath'] +\
-                       "' | grep -v grep | awk '{print $2}'`"
+        kill_command = "kill `pgrep -f " + r['dataPath'] +" | grep -v grep`"
```
innobead, line 2197: Do we still need `grep -v grep`? since you already use pgrep.

innobead (changes requested):

yangchiu:
Just for reference, the problematic functions in https://github.com/longhorn/longhorn/issues/5460#issuecomment-1583924424 are:

- crash_engine_process_with_sigkill
- crash_replica_processes
- get_pod_data_md5sum

`crash_engine_process_with_sigkill` and `crash_replica_processes` are fixed in this PR.

`get_pod_data_md5sum` currently doesn't need to be refined, because it executes `awk` in workload pods or share-manager pod which has `awk` installed: https://github.com/longhorn/longhorn-share-manager/blob/master/package/Dockerfile#L48. And basically we will always check the output of `get_pod_data_md5sum` to get the calculated checksum, so we're aware if it goes wrong.

But for `crash_engine_process_with_sigkill` and `crash_replica_processes`, we never check output of the executed command. That's why we don't noticed it even though it's broken since long time ago.

We can refine this later by checking executed command output to avoid this from happening again.

innobead:
Sounds good. @yangchiu Please create another issue for the improvements later.

yangchiu:
> Sounds good. @yangchiu Please create another issue for the improvements later.

Created https://github.com/longhorn/longhorn/issues/6095 for this.

manager/integration/tests/common.py
mantissahz: Yes, we don't need it.
Removed, thanks.

innobead (approved):

innobead:
@mergify backport v1.5.x v1.4.x v1.3.x

PR #1427, opened 2023-06-16
test(upgrade-responder): collected tags and fields

c3y1huang:
https://github.com/longhorn/longhorn/issues/5235


innobead:
@roger-ryao also review this PR when testing the source issue. Thanks.

PR #1426, opened 2023-06-15
Explicitly enable SELinux if it is desired for k3s Rocky testing

ejweber:
longhorn/longhorn#6108

The behavior of Longhorn with SELinux "enabled" on RKE2 is different than on k3s. This is because the current k3s deployment scripts don't go far enough to actually enable it.

PR #1442, opened 2023-07-03
ci: explicitly enable selinux if it's desired for k3s rhel/oracle testing

yangchiu:
ci: explicitly enable selinux if it's desired for k3s rhel/oracle testing

Sync https://github.com/longhorn/longhorn-tests/pull/1426 to rhel and oracle

### DIFF
--- a/test_framework/terraform/aws/oracle/user-data-scripts/provision_k3s_server.sh.tpl
+++ b/test_framework/terraform/aws/oracle/user-data-scripts/provision_k3s_server.sh.tpl
@@ -11,7 +11,7 @@ sudo yum install -y iscsi-initiator-utils nfs-utils nfs4-acl-tools jq rsync
 sudo systemctl -q enable iscsid
 sudo systemctl start iscsid
 
-until (curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="server --node-taint "node-role.kubernetes.io/master=true:NoExecute" --node-taint "node-role.kubernetes.io/master=true:NoSchedule" --tls-san ${k3s_server_public_ip} --write-kubeconfig-mode 644 --token ${k3s_cluster_secret}" INSTALL_K3S_VERSION="${k3s_version}" sh -); do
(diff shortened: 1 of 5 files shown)


test_framework/terraform/aws/oracle/user-data-scripts/provision_k3s_server.sh.tpl
```diff
@@ -11,7 +11,7 @@ sudo yum install -y iscsi-initiator-utils nfs-utils nfs4-acl-tools jq rsync
 sudo systemctl -q enable iscsid
 sudo systemctl start iscsid
 
-until (curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="server --node-taint "node-role.kubernetes.io/master=true:NoExecute" --node-taint "node-role.kubernetes.io/master=true:NoSchedule" --tls-san ${k3s_server_public_ip} --write-kubeconfig-mode 644 --token ${k3s_cluster_secret}" INSTALL_K3S_VERSION="${k3s_version}" sh -); do
+until (curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="server --node-taint "node-role.kubernetes.io/master=true:NoExecute" --node-taint "node-role.kubernetes.io/master=true:NoSchedule" --tls-san ${k3s_server_public_ip} --write-kubeconfig-mode 644 --token ${k3s_cluster_secret} --selinux=true" INSTALL_K3S_VERSION="${k3s_version}" sh -); do
```
innobead, line 14: Why for oracle, it's always enabled by default?

yangchiu: Because selinux is enabled for oracle by default, and currently there is no option in our oracle terraform pipeline script to adjust the value.

But of course we can provide the option to set it to permissive/enforcing just like other selinux distro. Do we need to add this?

innobead: I feel no need since enabling selinux is a superset of disabling.

innobead (approved):

PR #1447, opened 2023-07-07
test: fix flaky test case test_orphan_with_same_orphaned_dir_name_in_another_disk

yangchiu:
test: fix flaky test case test_orphan_with_same_orphaned_dir_name_in_another_disk by retry orphan deletion

The client.delete(orphan) function is somehow not reliable. It's possible we called this api but it takes no effect and the orphan still exists. Also the api doesn't return anything so there's no way to know it succeeds or not by checking the response.

Add retry mechanism to ensure the orphan can be deleted as expected.

For https://github.com/longhorn/longhorn/issues/6244

### DIFF
--- a/manager/integration/tests/test_orphan.py
+++ b/manager/integration/tests/test_orphan.py
@@ -83,24 +83,24 @@ def create_orphaned_directories_on_host(volume, disk_paths, num_orphans):  # NOQ
     return paths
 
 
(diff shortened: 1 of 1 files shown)


khushboo-rancher (approved):

khushboo-rancher:

PR #1450, opened 2023-07-10
fix: wait for snapshot counts of a DR volume

mantissahz:
There is a low chance that previous restored snapshot is not deleted and counted with restored snapshot this time at step 16 of the test case `test_inc_restoration_with_multiple_rebuild_and_expansion`

Modified to be that wait for snapshot counts as expected.

Ref: longhorn/longhorn#5496


yangchiu (approved):
LGTM

khushboo-rancher (approved):

khushboo-rancher:

PR #1457, opened 2023-07-11
Update CSI external-snapshotter version to v6.2.1

chriscchien:
Update CSI external-snapshotter version to `v6.2.1`

ref: [6283](https://github.com/longhorn/longhorn/issues/6283)

After upgrade CSI external-snapshotter version,  use `snapshot.storage.k8s.io/v1beta1` to create VolumesnapshotClass, VolumeSnapshotContent, VolumeSnapshot despertched.
Change `snapshot.storage.k8s.io/v1beta1` to `snapshot.storage.k8s.io/v1` in test script.

### DIFF
--- a/manager/integration/tests/test_csi_snapshotter.py
+++ b/manager/integration/tests/test_csi_snapshotter.py
@@ -41,7 +41,7 @@ def volumesnapshotclass(request):
         def create_volumesnapshotclass(name, deletepolicy, snapshot_type=None):
             manifest = {
                 'kind': 'VolumeSnapshotClass',
-                'apiVersion': 'snapshot.storage.k8s.io/v1beta1',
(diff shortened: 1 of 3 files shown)


innobead (approved):
LGTM. 

I think we only need to backport to 1.5.x, right?

chriscchien:
> LGTM.
> 
> I think we only need to backport to 1.5.x, right?

Yes v1.4.2 [documentation](https://longhorn.io/docs/1.4.2/snapshots-and-backups/csi-snapshot-support/enable-csi-snapshot-support/) specified use CSI external-snapshotter `v5.0.1` which is the current value on v1.4.x.

yangchiu:
@Mergify backport v1.5.x

PR #1474, opened 2023-07-21
ci: implement helm install/upgrade regression test pipeline

yangchiu:
ci: implement helm install/upgrade regression test pipeline

(1) add a separate independent pipeline for helm to avoid heavy Jenkinsfile and longhorn-setup.sh
(2) implement refresh install longhorn helm chart and run regression test
(3) implement (2 stage) upgrade longhorn helm chart for test_upgrade
(4) for setup scripts, separate common functions into utility functions for code reuse

For https://github.com/longhorn/longhorn/issues/3987 https://github.com/longhorn/longhorn/issues/5936

### DIFF
--- a/manager/integration/Dockerfile
+++ b/manager/integration/Dockerfile
@@ -1,22 +1,32 @@
 FROM registry.suse.com/bci/python:3.9
 
 ARG KUBECTL_VERSION=v1.17.0
+ARG YQ_VERSION=v4.24.2
 ARG ARCH=amd64
 
 RUN zypper ref -f
-RUN zypper in -y vim-small nfs-client xfsprogs e2fsprogs util-linux-systemd gcc python39-devel gawk java-11-openjdk && \
+RUN zypper in -y vim-small nfs-client xfsprogs e2fsprogs util-linux-systemd gcc python39-devel gawk java-11-openjdk tar awk gzip wget && \
     rm -rf /var/cache/zypp/*
 
 RUN curl -sO https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_VERSION/bin/linux/${ARCH}/kubectl && \
     mv kubectl /usr/local/bin/kubectl && \
-    chmod +x /usr/local/bin/kubectl
+    chmod +x /usr/local/bin/kubectl && \
+    curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && \
+    chmod 700 get_helm.sh && \
+    ./get_helm.sh && \
+    wget -q "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_${ARCH}" && \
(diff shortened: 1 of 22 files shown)


manager/integration/Dockerfile
```diff
@@ -1,22 +1,32 @@
 FROM registry.suse.com/bci/python:3.9
 
 ARG KUBECTL_VERSION=v1.17.0
```
khushboo-rancher, line 3: nit: Can we make the kubectl version same across the test infra? Like at other places, we are using 1.20.2 or we can use the latest stable version.

manager/integration/tests/conftest.py
```diff
@@ -50,6 +51,12 @@ def pytest_addoption(parser):
                      default=False,
                      help="include cluster autoscaler tests (default: False)")
 
+    parser.addoption(LH_INSTALL_METHOD, action="store",
```
khushboo-rancher, line 54: So, this will be applicable for all the pipeline that we can choose if want to use manifest or helm to upgrade the Longhorn?
Generally, we use the same way of upgrading Longhorn the way we installed it, isn't it?

khushboo-rancher (commented):
Thanks @yangchiu, In general LGTM, just few doubts

manager/integration/Dockerfile
yangchiu: kubectl binary doesn't provide a download url with `latest` tag. We have to manually do `$(curl -L -s https://dl.k8s.io/release/stable.txt)` to know the latest version, and then use `curl -LO https://dl.k8s.io/release/${LATEST_VERSION}/bin/linux/amd64/kubectl` to download the latest one. 

But I think we can update the version and make it consistent across the 2 Dockerfiles in longhorn-tests.

yangchiu: Created https://github.com/longhorn/longhorn/issues/6485 for this.

manager/integration/tests/conftest.py
yangchiu: We will have separate pipelines for each install method (like manifest, helm, rancher, flux, argocd).

And for each pipeline, the install method is fixed in code (unable to and no need to set this value as a Jenkins option), so the pipeline knows how to install Longhorn. 

But for upgrading Longhorn, it's done by `pytest` (`test container`) run by the `test build container`. The `test build container` knows the install method (which written in code), but `pytest` (`test container`) didn't. So we need `LH_INSTALL_METHOD` to pass the install method value into `pytest` (`test container`).

khushboo-rancher: Okay, so we should set default `LH_INSTALL_METHOD` as per the pipelines. e.g. for helm pipeline default `LH_INSTALL_METHOD` will be helm

yangchiu: Yes, just like what I mentioned, it's already hard coded in each pipeline.

khushboo-rancher (approved):

PR #1477, opened 2023-07-27
Fix flaky test_setting_concurrent_rebuild_limit Extend Volume Rebuild Timeout

roger-ryao:
https://github.com/longhorn/longhorn/issues/6382

The [Volume Replica Rebuilding Speedup](https://github.com/longhorn/longhorn/issues/4783) feature was introduced as an enhancement in `1.4.0`

The current timeout value for checking the rebuild status may not be sufficient for `v1.3.x`.

To address this, we propose extending the timeout for checking the rebuild volume status on `v1.3.x` branch.

The execution time for each test case to approximately 15-20 minutes.

Screenshot_20230802_172123

Private image test result 
- [x] coretest : https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4727/
- [x] test case test : https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4725/

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -29,7 +29,7 @@ from common import (  # NOQA
     SETTING_PRIORITY_CLASS,
     SETTING_DEFAULT_REPLICA_COUNT,
     SETTING_BACKUP_TARGET,
-    SIZE, RETRY_COUNTS, RETRY_EXEC_COUNTS,
+    SIZE, RETRY_COUNTS, RETRY_COUNTS_LONG, RETRY_EXEC_COUNTS,
     RETRY_INTERVAL, RETRY_INTERVAL_LONG,
     update_setting, BACKING_IMAGE_QCOW2_URL, BACKING_IMAGE_NAME,
     create_backing_image_with_matching_url, BACKING_IMAGE_EXT4_SIZE,
@@ -835,9 +835,9 @@ def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQ
 
         time.sleep(RETRY_INTERVAL)
 
-    wait_for_rebuild_complete(client, volume1_name)
+    wait_for_rebuild_complete(client, volume1_name, RETRY_COUNTS_LONG)
     wait_for_rebuild_start(client, volume2_name)
-    wait_for_rebuild_complete(client, volume2_name)
+    wait_for_rebuild_complete(client, volume2_name, RETRY_COUNTS_LONG)
 
     # Step 1-6
     wait_for_volume_healthy(client, volume1_name)
@@ -867,8 +867,8 @@ def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQ
     assert concourent_build is True
 
     # Step 1-8
-    wait_for_rebuild_complete(client, volume1_name)
-    wait_for_rebuild_complete(client, volume2_name)
+    wait_for_rebuild_complete(client, volume1_name, RETRY_COUNTS_LONG)
+    wait_for_rebuild_complete(client, volume2_name, RETRY_COUNTS_LONG)
 
     # Step 1-9
     update_setting(client,
@@ -891,28 +891,30 @@ def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQ
 
     # While one volume is rebuilding, verify another volume is not
     # rebuilding and stuck in degrading state
-    rebuild_started = False
-    for i in range(RETRY_COUNTS):
+    v1_rebuild_started = False
+    v2_rebuild_started = False
+    for i in range(RETRY_COUNTS_LONG):
         volume1 = client.by_id_volume(volume1_name)
         volume2 = client.by_id_volume(volume2_name)
 
-        if volume1.rebuildStatus == [] and \
-                volume2.rebuildStatus == [] and \
-                rebuild_started is False:
-            continue
-        elif volume1.rebuildStatus == [] and \
-                volume2.rebuildStatus == [] and \
-                rebuild_started is True:
+        if volume1.rebuildStatus == [] and volume2.rebuildStatus == [] \
+                and v1_rebuild_started and v2_rebuild_started:
             break
         elif volume2.rebuildStatus == []:
-            assert volume1.rebuildStatus[0].state == "in_progress"
-            rebuild_started = True
+            if volume1.rebuildStatus and \
+                    volume1.rebuildStatus[0].state == "in_progress" and \
+                    len(volume1.replicas) == 3:
+                v1_rebuild_started = True
         elif volume1.rebuildStatus == []:
-            assert volume2.rebuildStatus[0].state == "in_progress"
-            rebuild_started = True
+            if volume2.rebuildStatus and \
+                    volume2.rebuildStatus[0].state == "in_progress" and \
+                    len(volume2.replicas) == 3:
+                v2_rebuild_started = True
 
-        time.sleep(RETRY_INTERVAL)
+        time.sleep(RETRY_INTERVAL_LONG)
 
+    assert v2_rebuild_started
+    assert v1_rebuild_started
     wait_for_rebuild_complete(client, volume2_name)
     wait_for_rebuild_complete(client, volume1_name)
 
@@ -951,6 +953,7 @@ def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQ
         time.sleep(RETRY_INTERVAL)
     assert expect_case is True
 
+    wait_for_rebuild_complete(client, volume1_name)
     wait_for_volume_healthy(client, volume1_name)
     volume2.detach(hostId=lht_host_id)
     wait_for_volume_detached(client, volume2_name)
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -62,6 +62,7 @@ PORT = ":9500"
 RETRY_COMMAND_COUNT = 3
 RETRY_COUNTS = 150
 RETRY_COUNTS_SHORT = 30
+RETRY_COUNTS_LONG = 360
 RETRY_INTERVAL = 1
 RETRY_INTERVAL_LONG = 2
 RETRY_BACKUP_COUNTS = 300
@@ -4061,10 +4062,10 @@ def wait_for_expansion_failure(client, volume_name, last_failed_at=""):
     assert failed
 
 
-def wait_for_rebuild_complete(client, volume_name):
+def wait_for_rebuild_complete(client, volume_name, retry_count=RETRY_COUNTS):
     completed = 0
     rebuild_statuses = {}
-    for i in range(RETRY_COUNTS):
+    for i in range(retry_count):
         completed = 0
         v = client.by_id_volume(volume_name)
         rebuild_statuses = v.rebuildStatus


yangchiu:
According to the description https://github.com/longhorn/longhorn/issues/6382

```python
        wait_for_rebuild_complete(client, volume2_name)
        wait_for_rebuild_complete(client, volume1_name)
    
        # Step 2-1
        # Step 2-2
>       wait_for_volume_healthy(client, volume1_name)
```

If `wait_for_rebuild_complete` already passes, doesn't it mean the rebuilding has completed? Why does it still have rebuild timeout issue here?

roger-ryao:
> According to the description [longhorn/longhorn#6382](https://github.com/longhorn/longhorn/issues/6382)
> 
> ```python
>         wait_for_rebuild_complete(client, volume2_name)
>         wait_for_rebuild_complete(client, volume1_name)
>     
>         # Step 2-1
>         # Step 2-2
> >       wait_for_volume_healthy(client, volume1_name)
> ```
> 
> If `wait_for_rebuild_complete` already passes, doesn't it mean the rebuilding has completed? Why does it still have rebuild timeout issue here?

Hi @yangchiu :
      Yes, you are right. When I execute the test case and encounter a failure, the replica is still rebuilding. I added a rebuild timeout for that scenario, but I didn't investigate why "wait_for_rebuild_complete(client, volume1_name)" passed without any error. It seems that the function didn't correctly detect the rebuilding status in some cases, which is why I added the timeout as a temporary solution.

yangchiu:
>Yes, you are right. When I execute the test case and encounter a failure, the replica is still rebuilding. I added a rebuild timeout for that scenario, but I didn't investigate why "wait_for_rebuild_complete(client, volume1_name)" passed without any error. It seems that the function didn't correctly detect the rebuilding status in some cases, which is why I added the timeout as a temporary solution.

Can you help to do more experiments to investigate why? Just insert some logs and you're able to know the whole status at that moment.

roger-ryao:
> Can you help to do more experiments to investigate why? Just insert some logs and you're able to know the whole status at that moment.

I will take some time to investigate why the replica is still rebuilding even after wait_for_rebuild_complete(client, volume1_name) has passed without any error.

roger-ryao:
> > Can you help to do more experiments to investigate why? Just insert some logs and you're able to know the whole status at that moment.
> 
> I will take some time to investigate why the replica is still rebuilding even after wait_for_rebuild_complete(client, volume1_name) has passed without any error.

Hi @yangchiu 
The existing code at https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_settings.py#L856-L878 does not guarantee that the volume has actually started rebuilding. To ensure more rigor, I have added a check to verify that the rebuild status must be greater than 50% before considering `rebuild_started` as True. Before calling wait_for_rebuild_complete, I think it's essential to ensure that wait_for_rebuild_start is executed. This way, if any issues arise during wait_for_rebuild_start, we can better understand the root cause of the problem.

yangchiu:
> Hi @yangchiu The existing code at https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_settings.py#L856-L878 does not guarantee that the volume has actually started rebuilding. To ensure more rigor, I have added a check to verify that the rebuild status must be greater than 50% before considering `rebuild_started` as True. Before calling wait_for_rebuild_complete, I think it's essential to ensure that wait_for_rebuild_start is executed. This way, if any issues arise during wait_for_rebuild_start, we can better understand the root cause of the problem.

Why doesn't the original `assert volume1.rebuildStatus[0].state == "in_progress"` ensure the volume has started rebuilding?
Isn't the same checking used in `wait_for_rebuild_start` later also being used by you to ensure the same thing?

manager/integration/tests/test_settings.py
```diff
@@ -896,25 +896,28 @@ def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQ
...
-            rebuild_started = True
+
+        # Check volume1 if volume2 is not rebuilding
+        if not volume2.rebuildStatus:
+            if volume1.rebuildStatus:
+                assert volume1.rebuildStatus[0].state == "in_progress"
+                rebuild_started = volume1.rebuildStatus[0].progress >= 50
+
+        # Check volume2 if volume1 is not rebuilding
+        if not volume1.rebuildStatus:
```
yangchiu, line 910: Will a volume doesn't have `rebuildStatus`?

manager/integration/tests/test_settings.py
```diff
@@ -896,25 +896,28 @@ def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQ
...
+                assert volume2.rebuildStatus[0].state == "in_progress"
+                rebuild_started = volume2.rebuildStatus[0].progress >= 50
 
         time.sleep(RETRY_INTERVAL)
 
-    wait_for_rebuild_complete(client, volume2_name)
-    wait_for_rebuild_complete(client, volume1_name)
+    assert rebuild_started
+    wait_for_rebuild_complete(client, volume2_name, 300)
+    wait_for_rebuild_start(client, volume1_name)
```
yangchiu, line 919: This implies `volume2` started rebuilding first, and then `volume1`.

If `volume1` started rebuilding first, this wait will fail, right?

yangchiu (changes requested):

manager/integration/tests/test_settings.py
roger-ryao: Yes, according to [the distribution scenario mentioned](https://github.com/longhorn/longhorn-tests/blob/v1.3.x/manager/integration/tests/test_settings.py#L783-L786), one of the volumes does not have a `rebuildStatus`.

manager/integration/tests/test_settings.py
roger-ryao: Yes, you are correct. To enhance and test this situation, I will modify the wait condition to check for the status of both `volume1` and `volume2`.

roger-ryao:
> Why doesn't the original `assert volume1.rebuildStatus[0].state == "in_progress"` ensure the volume has started rebuilding? Isn't the same checking used in `wait_for_rebuild_start` later also being used by you to ensure the same thing?


Initially, the rebuild process starts for `Volume1` and progresses up to `13%`. However, after a certain period of time, the rebuild process for `volume1` is canceled. Subsequently, the rebuild process for `volume2` begins. This is why I added a check status process for both volumes, ensuring the rebuild progresses beyond 50%, and set the `rebuild_start` flag to `True`.

yangchiu:
> Initially, the rebuild process starts for `Volume1` and progresses up to `13%`. However, after a certain period of time, the rebuild process for `volume1` is canceled. Subsequently, the rebuild process for `volume2` begins. This is why I added a check status process for both volumes, ensuring the rebuild progresses beyond 50%, and set the `rebuild_start` flag to `True`.

When `volume1` rebuilding is canceled, what's the `rebuildStatus` of `volume1`? Will it return to an empty array `[]`? Or will any message present in the volume status?

roger-ryao:
> > Initially, the rebuild process starts for `Volume1` and progresses up to `13%`. However, after a certain period of time, the rebuild process for `volume1` is canceled. Subsequently, the rebuild process for `volume2` begins. This is why I added a check status process for both volumes, ensuring the rebuild progresses beyond 50%, and set the `rebuild_start` flag to `True`.
> 
> When `volume1` rebuilding is canceled, what's the `rebuildStatus` of `volume1`? Will it return to an empty array `[]`? Or will any message present in the volume status?

When the rebuilding of `volume1` is canceled, it will return to an empty array [].

Screenshot_20230729_223914

roger-ryao:
> > When the rebuilding of `volume1` is canceled, it will return to an empty array [].
> 
> Only the rebuilding of `volume1` will be canceled, right?
> 
> Will it happen on `volume2`?

Yes, you are correct. Based on the scenario mentioned in https://github.com/longhorn/longhorn-tests/blob/v1.3.x/manager/integration/tests/test_settings.py#L783-L786, we observed that the rebuilding process of `volume1` was canceled. However, after a certain period of time, `volume2` started its rebuilding process, and unlike `volume1`, its rebuilding process was not canceled.

yangchiu:
> > > When the rebuilding of `volume1` is canceled, it will return to an empty array [].
> > 
> > 
> > Only the rebuilding of `volume1` will be canceled, right?
> > Will it happen on `volume2`?
> 
> Yes, you are correct. Based on the scenario mentioned in https://github.com/longhorn/longhorn-tests/blob/v1.3.x/manager/integration/tests/test_settings.py#L783-L786, we observed that the rebuilding process of `volume1` was canceled. However, after a certain period of time, `volume2` started its rebuilding process, and unlike `volume1`, its rebuilding process was not canceled.

So does this test case fail only because it doesn't make sure `volume1`'s rebuilding is canceled, so in the `for loop`, it incorrectly thought rebuilding has been done on both of 2 volumes, and then causes the `wait_for_volume_healthy` for `volume1` failed?

roger-ryao:
> So does this test case fail only because it doesn't make sure `volume1`'s rebuilding is canceled, so in the `for loop`, it incorrectly thought rebuilding has been done on both of 2 volumes, and then causes the `wait_for_volume_healthy` for `volume1` failed?

You are correct in your analysis. The [Volume Replica Rebuilding Speedup feature introduced](https://github.com/longhorn/longhorn/issues/4783) in version `1.4.0`, which is why the test case doesn't fail in versions `1.4`, `1.5`, and the `master` branch. it's still possible to enhance and test the case on the master branch for further validation and improvement.

yangchiu:
> > So does this test case fail only because it doesn't make sure `volume1`'s rebuilding is canceled, so in the `for loop`, it incorrectly thought rebuilding has been done on both of 2 volumes, and then causes the `wait_for_volume_healthy` for `volume1` failed?
> 
> You are correct in your analysis. The [Volume Replica Rebuilding Speedup feature introduced](https://github.com/longhorn/longhorn/issues/4783) in version `1.4.0`, which is why the test case doesn't fail in versions `1.4`, `1.5`, and the `master` branch. it's still possible to enhance and test the case on the master branch for further validation and improvement.

Can't make sure `volume`'s rebuilding is canceled before enter the `for loop` solve this problem?

roger-ryao:
> Can't make sure `volume`'s rebuilding is canceled before enter the `for loop` solve this problem?

After discussing with @yangchiu, I will try to enhance the solution by ensuring that the volume's rebuilding is canceled before entering the for loop.

roger-ryao:
> > Can't make sure `volume`'s rebuilding is canceled before enter the `for loop` solve this problem?
> 
> After discussing with @yangchiu, I will try to enhance the solution by ensuring that the volume's rebuilding is canceled before entering the for loop.

Hi @yangchiu :
     I observed that when we retrieve the rebuilding status of `volume1`, the replica count remains two. Therefore, in addition to checking `volume.rebuildStatus[0].state`, I suggest confirming that `volume replicas count` is [also equal to three](https://github.com/longhorn/longhorn-tests/pull/1477/files#diff-8416f1cda57e975c3c1d0238a26790d74d71dbb3ae4b6e3cb468209d56bf1a12R904-R917). This precaution will prevent acquiring incorrect information about `volume.rebuildStatus` within the for loop.

yangchiu:
> 
> Hi @yangchiu : I observed that when we retrieve the rebuilding status of `volume1`, the replica count remains two. Therefore, in addition to checking `volume.rebuildStatus[0].state`, I suggest confirming that `volume replicas count` is [also equal to three](https://github.com/longhorn/longhorn-tests/pull/1477/files#diff-8416f1cda57e975c3c1d0238a26790d74d71dbb3ae4b6e3cb468209d56bf1a12R904-R917). This precaution will prevent acquiring incorrect information about `volume.rebuildStatus` within the for loop.

`replica count = 2` is because we remove one of replicas in previous step. Once the rebuilding started, the 3rd replica will be created automatically. Why we need to explicitly check this? And why does it result in _incorrect information_?

roger-ryao:
> > Hi @yangchiu : I observed that when we retrieve the rebuilding status of `volume1`, the replica count remains two. Therefore, in addition to checking `volume.rebuildStatus[0].state`, I suggest confirming that `volume replicas count` is [also equal to three](https://github.com/longhorn/longhorn-tests/pull/1477/files#diff-8416f1cda57e975c3c1d0238a26790d74d71dbb3ae4b6e3cb468209d56bf1a12R904-R917). This precaution will prevent acquiring incorrect information about `volume.rebuildStatus` within the for loop.
> 
> `replica count = 2` is because we remove one of replicas in previous step. Once the rebuilding started, the 3rd replica will be created automatically. Why do we need to explicitly check this? And why does it result in _incorrect information_?

Because I discovered that when we observe the rebuild process for `volume1` being canceled, the replica count for `volume1` remains at `2`. Our expectation is that when we retrieve volume information during an ongoing rebuild process, the `replica count` for the volume should be `3`.

yangchiu:
@roger-ryao What's the reproducibility of the original issue https://github.com/longhorn/longhorn/issues/6382 in longhorn `v1.3.x-head` tested with not modified longhorn-tests `v1.3.x-head`?

roger-ryao:
> @roger-ryao What's the reproducibility of the original issue [longhorn/longhorn#6382](https://github.com/longhorn/longhorn/issues/6382) in longhorn `v1.3.x-head` tested with not modified longhorn-tests `v1.3.x-head`?

Hi @yangchiu :
     The reproduction rate was 80%.

roger-ryao:
> @roger-ryao What's the reproducibility of the original issue [longhorn/longhorn#6382](https://github.com/longhorn/longhorn/issues/6382) in longhorn `v1.3.x-head` tested with not modified longhorn-tests `v1.3.x-head`?

Hi @yangchiu 
I attempted to rerun the test case 20 times, and it only passed once. You can find the results here:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4751/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4752/

When I executed the test case to address the original issue longhorn/longhorn#6382, I also discovered that we have three additional steps that are prone to failure due to timeouts during the wait for replica rebuild completion. This is why you can see that I extended the timeout for some of the steps.

yangchiu:
@roger-ryao Is it possible to have the test result of the latest version of this PR?

roger-ryao:
> @roger-ryao Is it possible to have the test result of the latest version of this PR?

Hi @yangchiu:

I have tested the private image before submitting the code, and the test results are as follows:
Test case test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4725/testReport/tests/test_settings/

yangchiu (approved):

manager/integration/tests/test_settings.py
```diff
@@ -891,28 +891,30 @@ def test_setting_concurrent_rebuild_limit(client, core_api, volume_name):  # NOQ
...
+            if volume2.rebuildStatus and \
+                    volume2.rebuildStatus[0].state == "in_progress" and \
+                    len(volume1.replicas) == 3:
+                v2_rebuild_started = True
 
-        time.sleep(RETRY_INTERVAL)
+        time.sleep(RETRY_INTERVAL_LONG)
 
+    assert v2_rebuild_started
+    assert v1_rebuild_started
```
innobead, line 917: I am a little confused. What we want to test is volume 2 rebuilding wil run first, then after volume 1. However, here, we just check both are started to rebuild. It seems not what we want to test?

innobead (changes requested):
Just one feedback.

manager/integration/tests/test_settings.py
roger-ryao: Hi @innobead :
     After discussing with James previously, there were concerns about the order of rebuilding [between volume 1 and volume 2](https://github.com/longhorn/longhorn-tests/blob/c9e603cf6815030305d8a0e8e504251b9431243d/manager/integration/tests/test_settings.py#L785C1-L786C62.). 
     
It was noted that there are two possible scenarios:
Volume 2's rebuilding starts first, and volume 1's rebuilding is delayed until volume 2 becomes healthy.
Volume 1's rebuilding starts first, and volume 2's rebuilding waits until volume 1 becomes healthy.

The exact order of rebuilding initiation may vary, our enhancement ensures that both volumes have been successfully under rebuilding.

innobead: Then, this verification does not achieve what we want to check. Probably you can juts check there is only one rebuilding, then then all rebuilt successfully eventually. This can make the test meaningful and complete. WDYT?

yangchiu: @roger-ryao Maybe you can try to print out all the rebuild status throughout the whole test case process, so we can have a clear picture of the rebuilding orders, which rebuilding is canceled/restarted, or any behavior that is unexpected.

innobead (approved):

PR #1486, opened 2023-08-01
ci: implement flux install/upgrade regression test pipeline

yangchiu:
ci: implement flux install/upgrade regression test pipeline

For https://github.com/longhorn/longhorn/issues/3987 https://github.com/longhorn/longhorn/issues/6343 https://github.com/longhorn/longhorn/issues/5936

### DIFF
--- a/manager/integration/Dockerfile
+++ b/manager/integration/Dockerfile
@@ -21,7 +21,10 @@ RUN curl -sO https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_
     wget -q https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_${ARCH}.zip && \
     unzip terraform_${TERRAFORM_VERSION}_linux_${ARCH}.zip && rm terraform_${TERRAFORM_VERSION}_linux_${ARCH}.zip && \
     mv terraform /usr/bin/terraform && \
-    chmod +x /usr/bin/terraform
+    chmod +x /usr/bin/terraform && \
+    curl -s -o flux_install.sh https://fluxcd.io/install.sh && \
+    chmod 700 flux_install.sh && \
+    ./flux_install.sh
 
 RUN curl -L https://github.com/jonelo/jacksum/releases/download/v3.4.0/jacksum-3.4.0.jar --output /jacksum.jar
 
@@ -34,5 +37,6 @@ WORKDIR /integration/tests
 ADD pipelines/utilities ./pipelines/utilities
 ADD pipelines/helm/scripts/upgrade-longhorn.sh ./pipelines/helm/scripts/upgrade-longhorn.sh
 ADD pipelines/rancher/scripts/upgrade-longhorn.sh ./pipelines/rancher/scripts/upgrade-longhorn.sh
+ADD pipelines/flux/scripts/upgrade-longhorn.sh ./pipelines/flux/scripts/upgrade-longhorn.sh
 
 ENTRYPOINT ["./run.sh"]
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -5,50 +5,13 @@ metadata:
   namespace: default
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: longhorn-test-role
-rules:
-- apiGroups: [""]
-  resources: ["nodes", "nodes/status", "pods", "pods/exec", "persistentvolumes", "persistentvolumeclaims",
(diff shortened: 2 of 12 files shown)


yangchiu:
> This pull request is now in conflicts. Could you fix it @yangchiu? 🙏

conflicts resolved and ready for review. cc @longhorn/qa

manager/integration/deploy/test.yaml
```diff
@@ -40,6 +40,12 @@ rules:
 - apiGroups: ["longhorn.io"]
   resources: ["supportbundles", "supportbundles/status"]
   verbs: ["*"]
+- apiGroups: ["source.toolkit.fluxcd.io"]
+  resources: ["helmrepositories"]
+  verbs: ["*"]
+- apiGroups: ["helm.toolkit.fluxcd.io"]
+  resources: ["helmreleases"]
+  verbs: ["*"]
```
innobead, line 48: We can just use cluster admin role, so we don't need to worry about the permission later for any kind of tests.

This is another topic, so can be tackled in another ticket.

manager/integration/tests/test_upgrade.py
```diff
@@ -243,6 +270,8 @@ def test_upgrade(longhorn_upgrade_type,
     rancher_secret_key = rancher_secret_key
     rancher_chart_install_version = rancher_chart_install_version
     longhorn_repo = longhorn_repo
+    flux_helm_chart_url = flux_helm_chart_url
+    flux_helm_chart_version = flux_helm_chart_version
```
innobead, line 274: Why we need this kind of self assignment? it seems no any effect?

manager/integration/Dockerfile
```diff
@@ -21,7 +21,10 @@ RUN curl -sO https://storage.googleapis.com/kubernetes-release/release/$KUBECTL_
     wget -q https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_${ARCH}.zip && \
     unzip terraform_${TERRAFORM_VERSION}_linux_${ARCH}.zip && rm terraform_${TERRAFORM_VERSION}_linux_${ARCH}.zip && \
     mv terraform /usr/bin/terraform && \
-    chmod +x /usr/bin/terraform
+    chmod +x /usr/bin/terraform && \
+    curl -s -o flux_install.sh https://fluxcd.io/install.sh && \
```
innobead, line 25: I assume this is always the latest and stable version of flux?

pipelines/flux/scripts/longhorn-setup.sh
```diff
@@ -0,0 +1,68 @@
...
+      UPGRADE_LH_ENGINE_IMAGE="longhornio/longhorn-engine:${LONGHORN_TRANSIENT_VERSION}"
+      run_longhorn_upgrade_test
+      LONGHORN_UPGRADE_TYPE="from_transient"
+      LONGHORN_UPGRADE_TEST_POD_NAME="longhorn-test-upgrade-from-transient"
+    fi
+    FLUX_HELM_CHART_URL="${HELM_CHART_URL}"
+    FLUX_HELM_CHART_VERSION="${LONGHORN_INSTALL_VERSION}"
+    UPGRADE_LH_ENGINE_IMAGE="longhornio/longhorn-engine:${LONGHORN_INSTALL_VERSION}"
+    run_longhorn_upgrade_test
+    run_longhorn_test
```
innobead, line 60: Does this mean we don't use flux to do the upgrade testing but rather relying on `run_longhorn_upgrade_test`?

innobead (changes requested):
Lood good for the flux part, but some questions need to resolve/answer first.

manager/integration/deploy/test.yaml
yangchiu: Refined to use `ClusterRole cluster-admin`.

manager/integration/tests/test_upgrade.py
yangchiu: Removed.

manager/integration/Dockerfile
yangchiu: Yes, it always gets the latest release version.

pipelines/flux/scripts/longhorn-setup.sh
yangchiu: `run_longhorn_upgrade_test` is just a function, it applies the `test.yaml` with desired configuration and starts the test pod, and the test pod is who performing the upgrade test in `pytest`, and `pytest`'s upgrade test case will determine which method should be used to perform the upgrade according to the configuration. In this case, it's the configuration for `flux`, so `pytest`'s upgrade test case will call the `flux` related shell script to perform some `flux` operations.

Other upgrade methods do the same thing. The only difference is the configuration.

innobead (approved):
LGTM

PR #1499, opened 2023-08-09
test: add debug message for cleanup function reset disk

yangchiu:
test: add debug message for cleanup function reset disk

If wait_for_disk_status(key ="storageScheduled", value=0) failed, it means there is a volume replica scheduled to this disk and causes some storage being occupied, so the value is non-zero and fail the assertion. 

But since we have deleted all volumes before, what causes this problem is still unknown. Add more debug messages to see if it
can provide more information.

For https://github.com/longhorn/longhorn/issues/5120


chriscchien (approved):
LGTM

khushboo-rancher:

PR #1500, opened 2023-08-10
ci: add terraform setup retry mechanism

yangchiu:
ci: add terraform setup retry mechanism

(1) add 15 minutes timeout to waiting for k8s cluster up, instead of a while loop running forever
(2) if the created cluster doesn't work normally, delete it and re-create a new one up to 3 times

For https://github.com/longhorn/longhorn/issues/6463


khushboo-rancher (dismissed):

test_framework/terraform/aws/sles/k3s_instances.tf
```diff
@@ -117,7 +117,7 @@ resource "null_resource" "rsync_kubeconfig_file" {
     inline = [
       "cloud-init status --wait",
       "if [ \"`cloud-init status | grep error`\" ]; then sudo cat /var/log/cloud-init-output.log; fi",
-      "until([ -f /etc/rancher/k3s/k3s.yaml ] && [ `sudo /usr/local/bin/kubectl get node -o jsonpath='{.items[*].status.conditions}'  | jq '.[] | select(.type  == \"Ready\").status' | grep -ci true` -eq $((${var.lh_aws_instance_count_controlplane} + ${var.lh_aws_instance_count_worker})) ]); do echo \"waiting for k3s cluster nodes to be running\"; sleep 2; done"
+      "RETRY=0; MAX_RETRY=450; until([ -f /etc/rancher/k3s/k3s.yaml ] && [ `sudo /usr/local/bin/kubectl get node -o jsonpath='{.items[*].status.conditions}'  | jq '.[] | select(.type  == \"Ready\").status' | grep -ci true` -eq $((${var.lh_aws_instance_count_controlplane} + ${var.lh_aws_instance_count_worker})) ]); do echo \"waiting for k3s cluster nodes to be running\"; sleep 2; if [ $RETRY -eq $MAX_RETRY ]; then break; fi; RETRY=$((RETRY+1)); done"
```
innobead, line 120: If failed to retry, have a log dumped.

innobead (changes requested):
In general, LGTM.

Do we need to update this change to rke2_instances.tf as well?

test_framework/terraform/aws/sles/k3s_instances.tf
yangchiu: Create https://github.com/longhorn/longhorn/issues/6601 for this.

yangchiu:
> In general, LGTM.
> 
> Do we need to update this change to rke2_instances.tf as well?

Daily regression is running on k3s, so it would be rare that rke2 encounters this issue.

innobead:
> > In general, LGTM.
> > Do we need to update this change to rke2_instances.tf as well?
> 
> Daily regression is running on k3s, so it would be rare that rke2 encounters this issue.

I see. It seems rke2 is based on k3s, so probably make the change consistent to prevent any confusion later. Or, at least have a comment to explain why we only improve this in K3s. WDYT?

yangchiu:
> > > In general, LGTM.
> > > Do we need to update this change to rke2_instances.tf as well?
> > 
> > 
> > Daily regression is running on k3s, so it would be rare that rke2 encounters this issue.
> 
> I see. It seems rke2 is based on k3s, so probably make the change consistent to prevent any confusion later. Or, at least have a comment to explain why we only improve this in K3s. WDYT?

Yes, this can be added. I'll refine this after the initial PR of https://github.com/longhorn/longhorn/issues/6499 is ready.

yangchiu:
> > > In general, LGTM.
> > > Do we need to update this change to rke2_instances.tf as well?
> > 
> > 
> > Daily regression is running on k3s, so it would be rare that rke2 encounters this issue.
> 
> I see. It seems rke2 is based on k3s, so probably make the change consistent to prevent any confusion later. Or, at least have a comment to explain why we only improve this in K3s. WDYT?

Same logic is added to `rke2` too.

innobead (approved):

PR #1507, opened 2023-08-18
test(support-bundle): syslog paths

c3y1huang:
https://github.com/longhorn/longhorn/issues/6544


innobead (approved):

innobead:
@longhorn/qa When verifying new features, need to see if there are manual test cases that should be added.

PR #1511, opened 2023-08-21
Add Storage network test in pre-release suit

khushboo-rancher:
Add the Storage network test in the pre release test suit till https://github.com/longhorn/longhorn/issues/4218 is handled. So that we don't miss executing it before major releases


innobead (approved):
LGTM.

We should make sure the pre-release keeps updated after each release.

innobead:
cc @longhorn/qa

PR #1514, opened 2023-08-23
test case test_rwx_offline_expansion

chriscchien:
Add test case test case `test_rwx_offline_expansion`

ref: [2484](https://github.com/longhorn/longhorn/issues/2484)

test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4842/)


khushboo-rancher (commented):
LGTM, just a doubt.

manager/integration/tests/test_rwx.py
```diff
@@ -572,4 +576,85 @@ def test_rwx_offline_expansion(): # NOQA
...
+        volume = client.by_id_volume(pv_name)
+        actual_size = int(volume.controllers[0].actualSize)
+        actual_size_in_mb = actual_size/1024/1024
+        if actual_size_in_mb > data_size_in_mb:
+            break
+        time.sleep(RETRY_EXEC_INTERVAL)
+    assert actual_size_in_mb > data_size_in_mb
+
+    def finalizer():
+        delete_and_wait_deployment(
```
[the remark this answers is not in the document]
chriscchien: Updated! sorry I missed that.

manager/integration/tests/test_rwx.py
```diff
@@ -572,4 +575,78 @@ def test_rwx_offline_expansion(): # NOQA
...
+    pv_name = get_volume_name(core_api, pvc_name)
+    share_manager_name = 'share-manager-' + pv_name
+    deployment_label_selector = "name=" + \
+                                deployment["metadata"]["labels"]["name"]
+
+    deployment_pod_list = \
+        core_api.list_namespaced_pod(namespace="default",
+                                     label_selector=deployment_label_selector)
+
+    assert deployment_pod_list.items.__len__() == 1
```
innobead, line 601: What's the purpose to check the number of the deployment pod? Isn't it quite straightforward? why need to validate it? This is a trivial assertion to check here, so I would suggest just assert what we need to assert (related to the test case) and just let it fail if there is no pod created for the deployment (it's about K8s not Longhorn).  

Below is good instead of verifying the number of pod in the deployment..

```python
    deployment_pod_list = \
        core_api.list_namespaced_pod(namespace="default",
                                     label_selector=deployment_label_selector)
    pod_name = deployment_pod_list.items[0].metadata.name
```

manager/integration/tests/test_rwx.py
```diff
@@ -572,4 +575,78 @@ def test_rwx_offline_expansion(): # NOQA
...
+    for i in range(RETRY_COUNTS):
+        pods = core_api.list_namespaced_pod(namespace=LONGHORN_NAMESPACE)
+        pod_names = []
+        for item in pods.items:
+            pod_names.append(item.metadata.name)
+        if share_manager_name in pod_names:
+            found = True
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert found
```
innobead, line 634: Probably we don't need to check if share manager pod exists, but you already check if the volume is healthy that implicitly means the share manager pod is running.

manager/integration/tests/test_rwx.py
```diff
@@ -572,4 +575,78 @@ def test_rwx_offline_expansion(): # NOQA
...
+                                 "/data/test2",
+                                 data_size_in_mb)
+
+    for i in range(RETRY_EXEC_COUNTS):
+        volume = client.by_id_volume(pv_name)
+        actual_size = int(volume.controllers[0].actualSize)
+        actual_size_in_mb = actual_size/1024/1024
+        if actual_size_in_mb > data_size_in_mb:
+            break
+        time.sleep(RETRY_EXEC_INTERVAL)
```
innobead, line 651: Why need retry? to ensure the status of update?

manager/integration/tests/test_rwx.py
```diff
@@ -572,4 +575,78 @@ def test_rwx_offline_expansion(): # NOQA
...
+    assert actual_size_in_mb > data_size_in_mb
```
innobead, line 652: Why this purpose? don't we just need to check if the nominal size is 4GB for the volume?

Probably you can mount the volume and check the nominal size.

innobead (changes requested):

manager/integration/tests/test_rwx.py
chriscchien: I removed redundant code, thank you

manager/integration/tests/test_rwx.py
chriscchien: Yes I found in SLES the actual size status change will slow and make test case flaky.
But I removed this part in the updated PR code, thank you

manager/integration/tests/test_rwx.py
chriscchien: In updated code, I use `exec_command_in_pod` into deployment pod to get actual file size and it should the same with the write size. I added `assert` to compare the read data size and write data size for review, thank you.

khushboo-rancher:
Thanks @chriscchien for doing the changes, could you run the updated test on our CI with `--count 5`?

chriscchien:
> Thanks @chriscchien for doing the changes, could you run the updated test on our CI with `--count 5`?

Hi @khushboo-rancher  here is the test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4891/testReport/tests/test_rwx/)

khushboo-rancher (approved):

innobead:
@khushboo-rancher @yangchiu as mentioned, if a PR is fine with your review and doesn't need an extra review, feel free to merge it.

manager/integration/tests/test_rwx.py
```diff
@@ -572,4 +574,63 @@ def test_rwx_offline_expansion(): # NOQA
...
+    assert resp == test_data
+
+    data_size_in_mb = 1536
+    write_pod_volume_random_data(core_api, pod_name, "/data/test2",
+                                 data_size_in_mb)
+    command = 'stat -c \"%s\" /data/test2'
+    data_size_in_pod = exec_command_in_pod(core_api, command,
+                                           pod_name,
+                                           'default')
+    assert int(data_size_in_pod)/1024/1024 == data_size_in_mb
```
innobead, line 636: There is some implicit info here. I would suggest making them clear in the future. 
- `write_pod_volume_random_data` is overwrite instead of resuming writing.
- the second write size is larger than 1G

innobead (approved):

innobead:
@chriscchien feel free to decide if this should be backported. cc @longhorn/qa

chriscchien:

PR #1518, opened 2023-08-24
negative test : Attaching and Detaching of Volume with replica deletion

nitendra-suse:
negative test : Attaching and Detaching of Volume with replica deletion
Item number 6 from https://confluence.suse.com/display/LON/Negative+Testing 

**Steps to run tests :**
1. enter into directory e2e
2. run command : ./run.sh -s ./tests/attach_detach_volume_with_replica_deletion.robot

**Results :** 
```
./run.sh -s ./tests/attach_detach_volume_with_replica_deletion.robot 
==============================================================================
Attach Detach Volume With Replica Deletion :: Negative Test Cases             
==============================================================================
Write Volume And Attach and DeAttach Volume With Replica Deletion     | PASS |
------------------------------------------------------------------------------
Attach Detach Volume With Replica Deletion :: Negative Test Cases     | PASS |
1 test, 1 passed, 0 failed
```
**For local setup use below settings.ini file :** 

```
[DEFAULT]
RETRY_INTERVAL = 5
SSH_CONFIG_PATH = ~/.ssh/config
CLOUD_PROVIDER = local
K8S_DISTRO = rke2
LONGHORN_CLIENT_URL = http://127.0.0.1:7000/
KUBECONFIG_PATH = ~/.kube/local
```

### DIFF
--- a/e2e/libs/node/aws.py
+++ b/e2e/libs/node/aws.py
@@ -5,7 +5,7 @@ from botocore.exceptions import ClientError
 
 class EC2:
 
-    _ec2_instance = boto3.resource('ec2')
+    _ec2_instance = boto3.resource('ec2',region_name='us-west-1')
     _all_nodes = None
 
     def __init__(self, all_nodes) -> None:
--- a/e2e/libs/utils/common_utils.py
+++ b/e2e/libs/utils/common_utils.py
@@ -18,11 +18,17 @@ RETRY_EXEC_COUNTS = 150
 RETRY_INTERVAL = 1
 RETRY_INTERVAL_LONG = 2
 
+def try_loading_config():
+    if os.getenv('LONGHORN_CLIENT_URL') or config_utils.Config.get("LONGHORN_CLIENT_URL"):
+        config.load_kube_config()
+    else:
+        config.load_incluster_config()
+
 def k8s_core_api():
     c = Configuration()
     c.assert_hostname = False
     Configuration.set_default(c)
-    config.load_incluster_config()
+    try_loading_config()
     core_api = client.CoreV1Api()
     return core_api
 
@@ -30,27 +36,32 @@ def k8s_cr_api():
     c = Configuration()
     c.assert_hostname = False
     Configuration.set_default(c)
-    config.load_incluster_config()
+    try_loading_config()
     cr_api = client.CustomObjectsApi()
     return cr_api
 
 def get_longhorn_api_client():
-    for i in range(RETRY_EXEC_COUNTS):
-        try:
-            config.load_incluster_config()
-            ips = get_mgr_ips()
+    if os.getenv('LONGHORN_CLIENT_URL'):
+        return os.getenv('LONGHORN_CLIENT_URL')
+    elif config_utils.Config.get("LONGHORN_CLIENT_URL"):
+        return config_utils.Config.get("LONGHORN_CLIENT_URL")
+    else:
+        for i in range(RETRY_EXEC_COUNTS):
+            try:
+                config.load_incluster_config()
+                ips = get_mgr_ips()
 
-            # check if longhorn manager port is open before calling get_client
-            for ip in ips:
-                sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-                mgr_port_open = sock.connect_ex((ip, 9500))
+                # check if longhorn manager port is open before calling get_client
+                for ip in ips:
+                    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+                    mgr_port_open = sock.connect_ex((ip, 9500))
 
-                if mgr_port_open == 0:
-                    client = 'http://' + ip + PORT + '/' #get_client(ip + PORT)
-                    break
-            return client
-        except Exception:
-            time.sleep(RETRY_INTERVAL)
+                    if mgr_port_open == 0:
+                        client = 'http://' + ip + PORT + '/' #get_client(ip + PORT)
+                        break
+                return client
+            except Exception:
+                time.sleep(RETRY_INTERVAL)
 
 def get_mgr_ips():
     ret = k8s_core_api().list_pod_for_all_namespaces(
--- a/e2e/libs/volume/crd.py
+++ b/e2e/libs/volume/crd.py
@@ -2,6 +2,8 @@ import time
 import logging
 
 from utils.common_utils import k8s_cr_api
+from utils.common_utils import k8s_core_api
+from utils.common_utils import get_longhorn_client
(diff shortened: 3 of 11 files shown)


yangchiu:
@nitendra-suse I'm planning to remove the `config`, `cloud_provider` and `aws` related code. So if you're struggling to understand/deal with these code, just feel free to modify and have your temporary solution. I'll refine this and have a more clear picture of how it should work.

e2e/libs/volume/crd.py
```diff
@@ -174,3 +177,40 @@ def check_data(self, volume_name, checksum):
...
+        eim_name = vol.controllers[0].instanceManagerName
+
+        target_pod = None
+        api = k8s_core_api()
+        namespace = 'longhorn-system'
+        try:
+            target_pod = api.read_namespaced_pod(name=eim_name,
+                                                namespace=namespace)
+        except ApiException as e:
+            assert e.status == 404
```
yangchiu, line 206: When will this happen? What cause this happen?

e2e/tests/attach_detach_volume_with_replica_deletion.robot
```diff
@@ -0,0 +1,42 @@
...
+    Set Test Variable    ${number_of_replicas}
+
+    ${volume_type}=    Evaluate    "${volume_type}".lower()
+
+    ${field1}=    Convert To String    {"spec": {"size": "${${volume_size_gb} * ${Gi}}"}}
+    ${field2}=    Convert To String    {"spec": {"numberOfReplicas": ${number_of_replicas}}}
+    ${field3}=    Convert To String    {"spec": {"accessMode": "${volume_type}"}}
+    @{list_of_fields}=    Create List    ${field1}    ${field2}    ${field3}
+
+    Given Create Volume With Fields    ${list_of_fields}
```
yangchiu, line 32: This volume creation step will be refined later since it's tedious. FYI

e2e/libs/utils/common_utils.py
```diff
@@ -18,39 +18,50 @@
...
-
-                if mgr_port_open == 0:
-                    client = 'http://' + ip + PORT + '/' #get_client(ip + PORT)
-                    break
-            return client
-        except Exception:
-            time.sleep(RETRY_INTERVAL)
+    if os.getenv('LONGHORN_CLIENT_URL'):
+        return os.getenv('LONGHORN_CLIENT_URL')
+    elif config_utils.Config.get("LONGHORN_CLIENT_URL"):
```
yangchiu, line 46: This could be refined later after checking if `settings.ini` is necessary or not. FYI

e2e/libs/node/aws.py
```diff
@@ -5,7 +5,7 @@
 
 class EC2:
 
-    _ec2_instance = boto3.resource('ec2')
+    _ec2_instance = boto3.resource('ec2',region_name='us-west-1')
```
yangchiu, line 8: `aws` and `cloud_provider` related code could be removed later if `litmus chaos` can be integrated into negative test. FYI

e2e/tests/attach_detach_volume_with_replica_deletion.robot
```diff
@@ -0,0 +1,42 @@
...
+
+*** Variables ***
+${Gi}=                  2**30
+${LOOP_COUNT}           1
+${volume_size_gb}=      5
+${volume_type}=         RWO
+${data_size_gb}=        2560
+
+*** Test Cases ***
+Write Volume And Attach and DeAttach Volume With Replica Deletion
```
yangchiu, line 21: Is the test case name correct? Seems it has nothing to do with `volume attach/detach`? cc @khushboo-rancher

e2e/tests/attach_detach_volume_with_replica_deletion.robot
```diff
@@ -0,0 +1,42 @@
...
+    @{list_of_fields}=    Create List    ${field1}    ${field2}    ${field3}
+
+    Given Create Volume With Fields    ${list_of_fields}
+    And Attach volume to node 1
+    And Write data into mount point ${data_size_gb}
+
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        And Delete the replica on node 1
+        And Delete im engine and wait
+        And Wait for replica on node 1 complete rebuilding
```
yangchiu, line 39: It's great to see we can start implementing more negative test cases like this, but I think some details of the original test case definition are missing.

The original definition is:
```
1. Create a 5 GI volume.
2. Write into the volume - ~5 Gi
3. Delete a replica, this will start rebuilding.
4. While rebuilding is in progress, crash the IM-engine.
5. Wait for a new IM pod spin up.
6. Wait for rebuilding to complete.
7. Repeat the steps from 3 - 6 200 times.
```

In `step 4`, we should check rebuilding already started before crash the im engine.
In `step 5`, we should check new im engine pod being spun up after the old one crashed.

e2e/tests/attach_detach_volume_with_replica_deletion.robot
```diff
@@ -0,0 +1,42 @@
...
+    ${field3}=    Convert To String    {"spec": {"accessMode": "${volume_type}"}}
+    @{list_of_fields}=    Create List    ${field1}    ${field2}    ${field3}
+
+    Given Create Volume With Fields    ${list_of_fields}
+    And Attach volume to node 1
+    And Write data into mount point ${data_size_gb}
+
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        And Delete the replica on node 1
+        And Delete im engine and wait
```
yangchiu, line 38: `Delete im engine and wait` is tedious. 

`Delete im engine` should already imply we wait until the pod being deleted before the function return. 

So `Delete im engine` should be good enough.

But the next question is this test step doesn't indicate we'd like to _delete im engine on which node_.

Should we delete the im engine on the same node as the deleted replica? @khushboo-rancher 

And should we delete different replica on different node in each loop?

yangchiu (changes requested):

e2e/libs/volume/crd.py
nitendra-suse: Reused code from https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/common.py#L574 ,with the hope, it's well tested.

innobead:
@nitendra-suse suggests moving to draft and restart this when the ground work is ready by @yangchiu . Please follow up on the status with @yangchiu .

innobead:
@yangchiu @nitendra-suse do we still need this? Probably @nitendra-suse need to reupdate this based on the current implementation. 

cc @khushboo-rancher

PR #1519, opened 2023-08-30
test(scheduling): add skeleton for allow empty selector settings

ChanYiLin:
ref: [longhorn/longhorn#4826](https://github.com/longhorn/longhorn/issues/4826)


roger-ryao (dismissed):
LGTM

manager/integration/tests/test_scheduling.py
```diff
@@ -1722,3 +1722,67 @@ def finalizer():
...
+    - Set allow-empty-node-selector-volume to `false`
+
+    When
+    - Create a Volume with 3 replicas without tag
+
+    Then
+    - All replicas can not be scheduled to the nodes
+
+    When
+    - Set allow-empty-node-selector-volume to `true`
```
khushboo-rancher, line 1750: Before setting `allow-empty-node-selector-volume` to `true`,  can we add one more step of removing the `AVAIL` tag from one of the node and as a result volume replica should get scheduled on that node without tag. 

Purpose of adding the test step is to ensure `allow-empty-node-selector-volume` setting is respecting tags correctly and just not blindly overruling the schedule criteria.

manager/integration/tests/test_scheduling.py
```diff
@@ -1722,3 +1722,67 @@ def finalizer():
...
+    - Set allow-empty-disk-selector-volume to `false`
+
+    When
+    - Create a Volume with 3 replicas without tag
+
+    Then
+    - All replicas can not be scheduled to the disks on the nodes
+
+    When
+    - Set allow-empty-disk-selector-volume to `true`
```
khushboo-rancher, line 1782: Same as above comment, can we add a step to remove tag from one of the disks and replica should get scheduled on that?

khushboo-rancher (changes requested):

manager/integration/tests/test_scheduling.py
ChanYiLin: Fixed, thanks!

manager/integration/tests/test_scheduling.py
ChanYiLin: Fixed, thanks!

ChanYiLin:
> This pull request is now in conflicts. Could you fix it @ChanYiLin? 🙏

Conflicts resolved

khushboo-rancher (approved):

PR #1522, opened 2023-09-05
Add manual tests and automated skeletons for Replica Disk Soft Anti-Affinity

ejweber:
longhorn/longhorn#3823

### DIFF
--- a/docs/content/manual/release-specific/v1.6.0/test-replica-disk-soft-anti-affinity.md
+++ b/docs/content/manual/release-specific/v1.6.0/test-replica-disk-soft-anti-affinity.md
@@ -1,0 +1,63 @@
+---
+title: Test Replica Disk Soft Anti-Affinity
+---
+
+## Related issue
+
+https://github.com/longhorn/longhorn/issues/3823
+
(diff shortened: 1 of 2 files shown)


ejweber:
I have addressed most (hopefully all) of the drone complaints in an amend, but GitHub is currently experiencing PR related problems and my changes aren't reflected here. Will check back later.

innobead:
@khushboo-rancher @yangchiu review this. Thanks.

khushboo-rancher (approved):
Thanks @ejweber, LGTM

nit: We can add one more test case in the manual test cases to verify the existing replicas on same disks can be scheduled properly after the upgrade as below.
```
**Given** An outdated Longhorn cluster with one volume having 3 replicas on same disk

**When** The cluster is upgraded

*And* Replica rebuilding is triggered by crashing the instance manager

*Then* All the replica should get scheduled on different disks.
```

khushboo-rancher:
Merging this as this is good enough, we can take care of the minor [suggestion](https://github.com/longhorn/longhorn-tests/pull/1522#pullrequestreview-1643068857) while testing this

PR #1524, opened 2023-09-07
add test case test_backuptarget_invalid

chriscchien:
Add test case `test_backuptarget_invalid`

ref: [6638](https://github.com/longhorn/longhorn/issues/6638)

Backup in error state did not have correspond `backupvolume` so in test case I cannot utilize below functions. That's the reason I use k8s API to get/delete backup which in error state
 - [delete_backup](https://github.com/longhorn/longhorn-tests/blob/f3324d7535b6f76dd46696c9ec17935ff4123cd0/manager/integration/tests/common.py#L474)
 - [find_backup](https://github.com/longhorn/longhorn-tests/blob/f3324d7535b6f76dd46696c9ec17935ff4123cd0/manager/integration/tests/common.py#L3613)

~longhorn-test pod need to delete backup through k8s API, I modified `ClusterRole` `longhorn-test-role` add permission for pod longhorn-test~ (no need after this [PR](https://github.com/longhorn/longhorn-tests/pull/1486/) 

Pipeline use specific test.yaml, hence I tested on local for reference
image

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -98,6 +98,8 @@ from common import DEFAULT_BACKUP_COMPRESSION_METHOD
 from common import BACKUP_COMPRESSION_METHOD_LZ4
 from common import BACKUP_COMPRESSION_METHOD_GZIP
 from common import BACKUP_COMPRESSION_METHOD_NONE
+from common import create_and_wait_deployment
+from common import get_custom_object_api_client
 
 from backupstore import backupstore_delete_volume_cfg_file
 from backupstore import backupstore_cleanup
@@ -111,6 +113,7 @@ from backupstore import backupstore_get_backup_volume_prefix
 from backupstore import set_backupstore_url, set_backupstore_credential_secret, set_backupstore_poll_interval  # NOQA
 from backupstore import reset_backupstore_setting  # NOQA
 from backupstore import set_backupstore_s3, backupstore_get_secret  # NOQA
+from backupstore import backupstore_invalid # NOQA
 
 from kubernetes import client as k8sclient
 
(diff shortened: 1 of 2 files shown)


chriscchien:
> This pull request is now in conflicts. Could you fix it @chriscchien? 🙏

Fixed

manager/integration/tests/test_basic.py
```diff
@@ -5522,7 +5531,52 @@ def test_backuptarget_invalid(): # NOQA
...
+                                                    "longhorn-system",
+                                                    "backups")
+        for backup in backups['items']:
+            api.delete_namespaced_custom_object("longhorn.io",
+                                                "v1beta2",
+                                                "longhorn-system",
+                                                "backups",
+                                                backup['metadata']['name'])
+
+    request.addfinalizer(delete_all_backups)
```
roger-ryao, line 5548: Perhaps we could consider relocating the `delete_all_backups` function to `common.py` and then executing it within the `cleanup_client()` function. Do you have any thoughts or feedback on this suggestion?

innobead:
@mantissahz you were requested to review. Find some time to help with review. Thanks.

manager/integration/tests/test_basic.py
chriscchien: Updated. I reorganized `delete_all_backups` into  fixure `backupstore_invalid` because only fixture `backupstore_invalid` will create backup in error state. (All other backups have corresponding backup volumes will clean up in `backupstore_cleanup` at `backupstore.py`)

test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4924/)

yangchiu (approved):
LGTM

innobead:
@yangchiu feel free to merge the approved PR.

chriscchien:

PR #1526, opened 2023-09-08
test: negative test case involves 2 volumes and recurring jobs

yangchiu:
test: negative test case involves 2 volumes and recurring jobs

For https://github.com/longhorn/longhorn/issues/6498


yangchiu:
This PR implemented test cases `6. Heavy writing and recurring jobs exists in the Longhorn system` + 
`1. Reboot the node where the volume is attached.` and `2. Reboot a replica node.` in https://confluence.suse.com/pages/viewpage.action?spaceKey=LON&title=Negative+Testing

e2e/keywords/common.resource
```diff
@@ -2,34 +2,21 @@
...
-
-Clean up all Longhorn resources
-    Run keyword And Continue On Failure    cleanup_resources
+Set test environment
+    init_k8s_api_client
+    init_node_exec    ${TEST NAME}
+    @{volume_list} =    Create List
+    Set Test Variable    ${volume_list}
+
+Cleanup resources
```
innobead, line 19: `Clean test resources`

e2e/tests/heavy_writing_and_recurring_jobs.robot
```diff
@@ -0,0 +1,43 @@
...
+Resource    ../keywords/recurring_job.resource
+
+Test Setup    Set test environment
+Test Teardown    Cleanup resources
+
+*** Variables ***
+${LOOP_COUNT}    1
+
+*** Test Cases ***
+Volume Attached Node Reboot While Heavy Writing And Recurring Jobs Exist
```
innobead, line 15: What's the purpose to test this with recurring jobs?

I feel `Reboot volume node while heavy write IO` good enough?

e2e/tests/heavy_writing_and_recurring_jobs.robot
```diff
@@ -0,0 +1,43 @@
...
+    Create volume 0 with size 2 GB and 1 replicas
+    Create volume 1 with size 2 GB and 3 replicas
+    Keep writing data to volume 0
+    Keep Writing data to volume 1
+    Create snapshot and backup recurring job for volume 0
+    Create snapshot and backup recurring job for volume 1
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        Reboot volume 0 attached node
+        Check recurring jobs for volume 0 work
+        Check recurring jobs for volume 1 work
```
innobead, line 25: What things will be verified?

e2e/tests/heavy_writing_and_recurring_jobs.robot
```diff
@@ -0,0 +1,43 @@
...
+        Check volume 0 works
```
innobead, line 26: It seems we want to test the data industry here, so suggest making the keyword concise with clear info.

For example: `Volume {} data should be data integiry`. This is why it's good to have BDD grammar eventually, given (condition), when (action), then (expectation). 

```robot
    Given Create volume 0 with size 2 GB and 1 replicas
    And Create volume 1 with size 2 GB and 3 replicas
    And Keep writing data to volume 0
    And Keep Writing data to volume 1
    And Create snapshot and backup recurring job for volume 0
    And Create snapshot and backup recurring job for volume 1
    
    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
        When Reboot volume 0 attached node
        
        Then Check recurring jobs for volume 0 work
        And Check recurring jobs for volume 1 work
        And Check volume 0 works
        And Check volume 1 works
    END
```

innobead (changes requested):
In genreal, LGTM. However, still have some questions needed to resolve first.

e2e/tests/heavy_writing_and_recurring_jobs.robot
yangchiu: Check if there will be a backup/snapshot created by the corresponding recurring job in expected timing.

e2e/tests/heavy_writing_and_recurring_jobs.robot
yangchiu: Yes, data integrity should be checked here:

https://github.com/longhorn/longhorn-tests/pull/1526/files#diff-6422ed87ee022691c861ebd060f6651dd79d25bef034662a78dc201f40a8838bR54

but since related function isn't implemented yet, refine this later.

e2e/tests/heavy_writing_and_recurring_jobs.robot
innobead: I see. Let's keep the same test plan first.

e2e/tests/heavy_writing_and_recurring_jobs.robot
innobead: Just make sure remaining part on track. Need to ensure we have corresponding tickets for it.

e2e/tests/heavy_writing_and_recurring_jobs.robot
yangchiu: cc @khushboo-rancher It would be great if related context or tickets can be referenced in the confluence page.

e2e/keywords/common.resource
yangchiu: Refined in https://github.com/longhorn/longhorn-tests/pull/1528

PR #1529, opened 2023-09-15
test: add cluster restart test case

yangchiu:
test: add cluster restart test case

### DIFF
--- a/e2e/libs/node/node.py
+++ b/e2e/libs/node/node.py
@@ -3,18 +3,42 @@ import yaml
 import time
 import logging
 from utility.utility import apply_cr_from_yaml, get_cr
+from utility.utility import wait_for_cluster_ready
+import boto3
 
 RETRY_COUNT = 180
 RETRY_INTERVAL = 1
 
 class Node:
 
+    def __init__(self):
+        with open('/tmp/instance_mapping', 'r') as f:
+            self.mapping = yaml.safe_load(f)
+        self.aws_client = boto3.client('ec2')
+        #logging.warn(f"describe_instances = {self.aws_client.describe_instances()}")
+
+    def restart_all_nodes(self):
+        instance_ids = [value for value in self.mapping.values()]
+        print(instance_ids)
+        resp = self.aws_client.stop_instances(InstanceIds=instance_ids)
+        print(resp)
+        waiter = self.aws_client.get_waiter('instance_stopped')
+        waiter.wait(InstanceIds=instance_ids)
+        print(f"all instances stopped")
+        time.sleep(60)
+        resp = self.aws_client.start_instances(InstanceIds=instance_ids)
+        print(resp)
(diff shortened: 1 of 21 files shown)


e2e/libs/node/node.py
```diff
@@ -3,12 +3,36 @@
...
 
 RETRY_COUNT = 180
 RETRY_INTERVAL = 1
 
 class Node:
 
+    def __init__(self):
+        with open('/tmp/instance_mapping', 'r') as f:
+            self.mapping = yaml.safe_load(f)
+        self.aws_client = boto3.client('ec2')
```
innobead, line 17: The implementation should be extendable to support different env in the future. Please create a github issue for this.

e2e/libs/utility/utility.py
```diff
@@ -107,6 +129,246 @@ def wait_delete_ns(name):
         time.sleep(RETRY_INTERVAL)
     assert not found
 
+def create_deployment(filepath):
```
innobead, line 132: if needed, can create other modules (in this case, it could be workload.py) to organize util functions to prevent a long implementation in util.py.

e2e/templates/litmus/reboot-engine.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: litmuschaos.io/v1alpha1
+kind: ChaosEngine
+metadata:
+  name: reboot-engine
```
innobead, line 4: I assume this is for reboot node. If yes, can make it clear like `reboot-node`.

innobead (changes requested):
In general, LGTM. Some comments needs to resolve.

innobead:
Let me know when the code is ready, then we can merge this PR.

e2e/libs/node/node.py
yangchiu: Can all supported environments be listed here? Then the related issue can have a clear description and scope.

innobead: It will depend on the implementation. Surely we can implement here but delegate the call to the corresponding infra implementation. Let u and the team to think about that.

cc @longhorn/qa

e2e/libs/utility/utility.py
yangchiu: moved to `workload.py`.

e2e/templates/litmus/reboot-engine.yaml
yangchiu: renamed.

e2e/libs/node/node.py
yangchiu: https://github.com/longhorn/longhorn/issues/6709 created for this.

yangchiu:
> Let me know when the code is ready, then we can merge this PR.

@innobead This PR is ready for review.

innobead (approved):

PR #1535, opened 2023-09-22
test: add volume node reboot test cases

yangchiu:
test: add volume node reboot test cases

(1) Reboot Volume Node While Workload Heavy Writing
(2) Power Off Volume Node For More Than Pod Eviction Timeout While Workload Heavy Writing


e2e/tests/node_reboot.robot
```diff
@@ -11,22 +12,122 @@ Test Teardown    Cleanup test resources
...
+        Check deployment 0 works
+        Check deployment 1 works
+        Check deployment 2 works
+        Check statefulset 0 works
+        Check statefulset 1 works
+        Check statefulset 2 works
+    END
+
+Reboot Volume Node While Workload Heavy Writing
+    Create statefulset 0 with rwo volume
```
innobead, line 116: We can do the same way you have done before by also adding deployment workload together.

e2e/keywords/common.resource
```diff
@@ -13,7 +13,7 @@ Library             ../libs/keywords/workload_keywords.py
 *** Keywords ***
 Set test environment
```
innobead, line 14: Probably you can make a more gradual setup or teardown for testsuite or test, so some resources are not required to clean up from test teardown but suite teardown. 

This can be figured out over time when having more tests.

innobead (changes requested):
In general, LGTM. Just one comment.

e2e/tests/node_reboot.robot
yangchiu: Created https://github.com/longhorn/longhorn/issues/6753 for this.

e2e/keywords/common.resource
yangchiu: Tackle this later to see if there's any resources can be reused.

innobead (approved):

PR #1300, opened 2023-04-07
feat: negative testing for node not ready

weizhe0422:
[Longhorn 5556](https://github.com/longhorn/longhorn/issues/5556)

Signed-off-by: Ray Chang <ray.chang@suse.com>

Here's a draft PR, still working on the first test case: `RWO volume has a copy on the attached node and closes it,` I'm resolving ssh to AWS nodes to execute commands.

cc @roger-ryao @yangchiu 

## Steps to execution
 1. change directory to `e2e/`
 2. According to your cluster to prepare the environment
     - Cluster@AWS
       1. Setting elastic IP for worker nodes
       2. Adjust `settings.ini`
          ```ini
          CLOUD_PROVIDER = 'aws'
          K8S_DISTRO = 'RKE2'
          LONGHORN_CLIENT_URL = 'http://localhost:8080/'
          ```
       3. Setting Longhorn client URL
           - port forward if the cluster launched in Rancher: `while true; do kubectl port-forward services/longhorn-frontend 8080:http -n longhorn-system; sleep 10; done`
       4. Setting AWS config and credentials in `~/.aws` folder for authorization.
           - config
            ```ini
            [default]
            region = <aws-region, e.g., ap-northeast-1>
            output = json
            ```
           - credentials
            ```ini
            [default]
            aws_access_key_id = <aws-access-key>
            aws_secret_access_key = <aws-secret-access-key>
            ```
        5. It is recommended to set up Elastic IP for each worker node, because the IP address will be reassigned after restarting the shutdown node.
        6. Setting SSH config file (`~/.ssh/config`) for worker nodes
            ```
            Host <external-ip>
                HostName <public-IPv4-DNS>
                User ubuntu
                IdentityFile <private-key-path>
                ForwardAgent yes
            ```
        7. Refer to `e2e/requirements.txt` to install required libraries including AWS Python SDK: [Boto3](https://github.com/boto/boto3)
           - `pip install -r e2e/requirements.txt`
     - Cluster@local
       1. Adjust `settings.ini`
          ```ini
          CLOUD_PROVIDER = 'local'
          LONGHORN_CLIENT_URL = '<longhorn-ui-url>'
          ```
       2. Setting SSH config file (`~/.ssh/config`) for worker nodes
        ```
          Host <external-ip>
              HostName <worker-node-ip-address>
              User ubuntu
              IdentityFile <private-key-path>
              ForwardAgent yes
          ```
  3. Run the test
      ```shell
[...]

PR #1536, opened 2023-09-22
test: add realtime logs to track test execution progress

yangchiu:
test: add realtime logs to track test execution progress

For https://github.com/longhorn/longhorn-tests/pull/1515#discussion_r1311215163

### DIFF
--- a/e2e/libs/keywords/pod_keywords.py
+++ b/e2e/libs/keywords/pod_keywords.py
@@ -1,12 +1,14 @@
-import logging
 import time
-
 from node import Nodes
 
 retry_count = 200
 retry_interval = 5
 
 class pod_keywords:
+
+    #TODO
+    # keywords layer can only call lower implementation layer to complete its work
+    # and should not have any business logic here
 
     def wait_all_pods_evicted(self, node_index):
         node_name = Nodes.get_name_by_index(int(node_index))
--- a/e2e/libs/keywords/recurring_job_keywords.py
+++ b/e2e/libs/keywords/recurring_job_keywords.py
@@ -1,14 +1,13 @@
 from utility.utility import get_test_case_namespace, generate_volume_name
 from utility.utility import get_node, list_nodes
 from utility.utility import get_test_pod_running_node, get_test_pod_not_running_node
+from utility.utility import logging
 from robot.libraries.BuiltIn import BuiltIn
 from recurring_job import RecurringJob
-import logging
 
 class recurring_job_keywords:
 
     def __init__(self):
-        logging.warn("initialize recurring_job_keywords class")
         self.recurring_job = RecurringJob()
 
 
@@ -17,7 +16,7 @@ class recurring_job_keywords:
         self.recurring_job.create(job_name, task="snapshot")
         self.recurring_job.add_to_volume(job_name, volume_name)
         self.recurring_job.get(job_name)
-        logging.warn(f'==> create recurring job {job_name} for volume {volume_name}')
+        logging(f'Created recurring job {job_name} for volume {volume_name}')
 
 
     def create_backup_recurring_job_for_volume(self, volume_name):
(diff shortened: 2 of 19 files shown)


e2e/libs/keywords/pod_keywords.py
```diff
@@ -1,13 +1,15 @@
...
-
 from node import Nodes
 
 retry_count = 200
 retry_interval = 5
 
 class pod_keywords:
 
+    #TODO
+    # keywords layer can only call lower implementation layer to complete its work
```
innobead, line 10: nit: Can add a readme in keywords folder or add this statement to __init__.py in keywords package.

e2e/libs/keywords/recurring_job_keywords.py
```diff
@@ -17,15 +16,15 @@ def create_snapshot_recurring_job_for_volume(self, volume_name):
         self.recurring_job.create(job_name, task="snapshot")
         self.recurring_job.add_to_volume(job_name, volume_name)
         self.recurring_job.get(job_name)
-        logging.warn(f'==> create recurring job {job_name} for volume {volume_name}')
+        logging(f'Created recurring job {job_name} for volume {volume_name}')
```
innobead, line 19: Can do below to have logs around the call functions, especially for exported functions. You can review others to have the same pattern.  

```python
logging("Creating ....")
....
logging("Created ...")
```

Again, this can be automatically implemented via metaclass such as https://stackoverflow.com/questions/16961914/python-logging-all-of-a-class-methods-without-decorating-each-one, so you don't need to manually add a log for each function/method. This is a beauty of dynamic lang or lang supporting metaclass or macro. You can improve this later.

e2e/libs/recurring_job/crd.py
```diff
@@ -10,25 +9,25 @@ def __init__(self):
...
         return self.rest.create(name, task, groups, cron, retain, concurrency, label)
 
     def delete(self, job_name, volume_name):
-        logging.warn("NotImplemented")
+        logging("Delegating the delete call to API because there is no CRD implementation")
         return self.rest.delete(job_name, volume_name)
 
     def get(self, name):
-        logging.warn("NotImplemented")
+        logging("Delegating the get call to API because there is no CRD implementation")
```
innobead, line 20: For getter function, don't suggest having logs to prevent log flooding.

e2e/libs/workload/workload.py
```diff
@@ -31,7 +31,7 @@ def create_storageclass(name):
         try:
             api.create_storage_class(body=manifest_dict)
         except Exception as e:
-            print(f"Exception when create storageclass: {e}")
+            logging(f"Creating storageclass error: {e}")
```
innobead, line 34: Right now you just hard-coded use info log, but for error, you want use warn if it's tolerated.

If it's a real error, just throw out to let the caller to deal with and the call will log with error level then handle it.

innobead (changes requested):
In general, LGTM. Some feedback needs to check first.

e2e/libs/keywords/recurring_job_keywords.py
yangchiu: Created https://github.com/longhorn/longhorn/issues/6758 for this.

e2e/libs/recurring_job/crd.py
yangchiu: Removed.

e2e/libs/workload/workload.py
yangchiu: Removed the try/except blocks.

e2e/libs/keywords/pod_keywords.py
yangchiu: These code is added in https://github.com/longhorn/longhorn-tests/pull/1300. There are still some code in that PR left and need to be refined or removed.

innobead (approved):
Keep improving this over time. Logging is always important for development and testing.

PR #1537, opened 2023-09-26
test: dynamically set wait condition timeout and reduce log size

yangchiu:
test: dynamically set wait condition timeout and reduce log size


innobead (approved):
LGTM, but feel retry count quite general here, because I guess we should have different setting for different cases eventually. However, it's good for now.

PR #1542, opened 2023-10-03
test: inject control plane node network latency

yangchiu:
test: inject control plane node network latency

For https://github.com/longhorn/longhorn/issues/6704 https://github.com/longhorn/longhorn/issues/6818

### DIFF
--- a/e2e/keywords/common.resource
+++ b/e2e/keywords/common.resource
@@ -6,6 +6,7 @@ Library             ../libs/keywords/node_keywords.py
 Library             ../libs/keywords/volume_keywords.py
 Library             ../libs/keywords/recurring_job_keywords.py
 Library             ../libs/keywords/workload_keywords.py
+Library             ../libs/keywords/network_keywords.py
 
 
 *** Variables ***
@@ -22,8 +23,10 @@ Set test environment
     Set Test Variable    ${deployment_list}
     @{statefulset_list} =    Create List
     Set Test Variable    ${statefulset_list}
+    setup_control_plane_network_latency
 
 Cleanup test resources
+    cleanup_control_plane_network_latency
     cleanup_node_exec
     cleanup_stress_helper
     cleanup_recurring_jobs    ${volume_list}
--- a/e2e/libs/network/network.py
+++ b/e2e/libs/network/network.py
@@ -1,0 +1,29 @@
+from robot.libraries.BuiltIn import BuiltIn
+from utility.utility import get_control_plane_nodes
+from node_exec import NodeExec
+
+def get_control_plane_node_network_latency_in_ms():
+    latency_in_ms = int(BuiltIn().get_variable_value("${CONTROL_PLANE_NODE_NETWORK_LATENCY_IN_MS}"))
+    return latency_in_ms
+
+def setup_control_plane_network_latency():
+    latency_in_ms = get_control_plane_node_network_latency_in_ms()
+    if latency_in_ms != 0:
+        nodes = get_control_plane_nodes()
+        for node in nodes:
+            cmd = f"tc qdisc replace dev eth0 root netem delay {latency_in_ms}ms"
+            res = NodeExec.get_instance().issue_cmd(node, cmd)
+            cmd = f"tc qdisc show dev eth0 | grep delay"
+            res = NodeExec.get_instance().issue_cmd(node, cmd)
+            assert res, "setup control plane network latency failed"
+
+def cleanup_control_plane_network_latency():
+    latency_in_ms = get_control_plane_node_network_latency_in_ms()
+    if latency_in_ms != 0:
+        nodes = get_control_plane_nodes()
+        for node in nodes:
+            cmd = "tc qdisc del dev eth0 root"
+            res = NodeExec.get_instance().issue_cmd(node, cmd)
+            cmd = f"tc qdisc show dev eth0 | grep -v delay"
+            res = NodeExec.get_instance().issue_cmd(node, cmd)
+            assert res, "cleanup control plane network failed"
--- a/e2e/libs/node_exec/node_exec.py
+++ b/e2e/libs/node_exec/node_exec.py
@@ -85,12 +85,20 @@ class NodeExec:
             stdout=True,
             tty=False
         )
-        logging(f"Issued command: {cmd} with result {res}")
+        logging(f"Issued command: {cmd} on {node_name} with result {res}")
         return res
 
     def launch_pod(self, node_name):
         if node_name in self.node_exec_pod:
-            return self.node_exec_pod[node_name]
+            for i in range(DEFAULT_POD_TIMEOUT):
+                pod = self.core_api.read_namespaced_pod(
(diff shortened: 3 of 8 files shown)


yangchiu:
Waiting for https://github.com/longhorn/longhorn-tests/pull/1537 and https://github.com/longhorn/longhorn-tests/pull/1539 merged.

e2e/tests/cluster_restart.robot
```diff
@@ -9,6 +9,7 @@ Test Teardown    Cleanup test resources
 
 *** Variables ***
 ${LOOP_COUNT}    1
+${LATENCY_IN_MS}    0
```
innobead, line 12: Make the naming meaningful like `CONTROL_PLANE_NODE_NETWORK_LATENCY_IN_MS`

The same for `${LOOP_COUNT}`.

e2e/libs/utility/utility.py
```diff
@@ -41,6 +41,14 @@ def list_nodes():
             nodes.append(item.metadata.name)
     return sorted(nodes)
 
+def get_control_plane_node():
+    core_api = client.CoreV1Api()
+    obj = core_api.list_node()
+    for item in obj.items:
+        if 'node-role.kubernetes.io/control-plane' in item.metadata.labels or \
+                'node-role.kubernetes.io/master' in item.metadata.labels:
+            return item.metadata.name
```
innobead, line 50: Only get the first node? This seems an implicit assumption because we are running only one master node. I would suggest making them general, like apply latency to all control plane nodes.

e2e/libs/network/network.py
```diff
@@ -0,0 +1,25 @@
+from robot.libraries.BuiltIn import BuiltIn
+from utility.utility import get_control_plane_node
+from node_exec import NodeExec
+
+def get_latency_in_ms():
+    latency_ms_ms = int(BuiltIn().get_variable_value("${LATENCY_IN_MS}"))
+    return latency_ms_ms
+
+def inject_control_plane_network_latency():
```
innobead, line 9: maybe? 

`setup_control_plane_network_latency`
`cleanup_control_plane_network_latency`

innobead (changes requested):
Some feedback.

e2e/tests/cluster_restart.robot
yangchiu: Refined.

e2e/libs/utility/utility.py
yangchiu: Refined.

e2e/libs/network/network.py
yangchiu: Refined.

yangchiu:
@longhorn/qa @innobead This PR is ready for review.

e2e/keywords/common.resource
```diff
@@ -22,8 +23,10 @@ Set test environment
     Set Test Variable    ${deployment_list}
     @{statefulset_list} =    Create List
     Set Test Variable    ${statefulset_list}
+    setup_control_plane_network_latency
```
c3y1huang, line 26: Does this imply that network latency will affect all the tests?

e2e/libs/node_exec/node_exec.py
```diff
@@ -80,11 +81,20 @@ def issue_cmd(self, node_name, cmd):
...
-            return self.node_exec_pod[node_name]
+            for i in range(DEFAULT_POD_TIMEOUT):
+                pod = self.core_api.read_namespaced_pod(
+                        name=node_name,
+                        namespace=self.namespace
+                      )
+                if pod is not None and pod.status.phase == 'Running':
+                    break
+                time.sleep(DEFAULT_POD_INTERVAL)
+            return pod
```
c3y1huang, line 97: Perhaps we should consider verifying this elsewhere. Is there a specific reason for having this here? Shouldn't we be cleaning up after tests?

e2e/libs/network/network.py
```diff
@@ -0,0 +1,29 @@
+from robot.libraries.BuiltIn import BuiltIn
+from utility.utility import get_control_plane_nodes
+from node_exec import NodeExec
+
+def get_latency_in_ms():
+    latency_ms_ms = int(BuiltIn().get_variable_value("${CONTROL_PLANE_NODE_NETWORK_LATENCY_IN_MS}"))
```
c3y1huang, line 6: The `CONTROL_PLANE_NODE_NETWORK_LATENCY_IN_MS` should reflect in the function name.

c3y1huang (changes requested):

e2e/keywords/common.resource
yangchiu: Yes, the default latency is `0`.

e2e/libs/node_exec/node_exec.py
yangchiu: After node reboot or other k8s related negative factors, the pod could not be in `Running` state even though it's created before. Need to wait for it recovering.

e2e/libs/network/network.py
yangchiu: refined.

e2e/libs/node_exec/node_exec.py
c3y1huang: The function's name suggests that it should create a pod. If we can identify the relevant factors, the subsequent step should involve waiting for the pod to recover. This approach enhances transparency in the test steps, reducing ambiguity.

e2e/keywords/common.resource
c3y1huang: Any reason we need this for all tests?

e2e/libs/node_exec/node_exec.py
yangchiu: Contributors don't even have to know there is such an execution pod running.

e2e/keywords/common.resource
yangchiu: A test case isn't restricted to only have one negative factor. For example, in this case we can have `node reboot` + `network latency`.

c3y1huang: Could the test cases run without this line?

e2e/libs/node_exec/node_exec.py
c3y1huang: It's better for the tests to have transparent steps.

c3y1huang: Actually, I think I've misunderstood you. What I meant to ask is why there would be an existing pod running during `launch_pod`.

yangchiu: > why there would be an existing pod running during `launch_pod`

Like a singleton class, when you instantiate it, internally it checks whether there's an instance exists, if so, just returns this existing one, otherwise creates a new one for you.

> why we are reusing the pod

A node execution pod, as a delegate, is meant to be reused throughout the whole test to provide a simpler way to execute commands on a node without ssh hostname/key configuration.

c3y1huang (approved):

PR #1547, opened 2023-10-05
test: update a read-only setting

mantissahz:
Ref: longhorn/longhorn#5989

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -8,7 +8,7 @@ from backupstore import set_random_backupstore  # NOQA
 
 from common import (  # NOQA
     get_longhorn_api_client, get_self_host_id,
-    get_core_api_client, get_apps_api_client,
+    get_core_api_client, get_apps_api_client, get_custom_object_api_client,
     create_and_check_volume, cleanup_volume,
     wait_for_volume_healthy, wait_for_volume_detached,
     write_volume_random_data, check_volume_data,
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_settings.py
```diff
@@ -1241,3 +1242,28 @@ def test_setting_update_with_invalid_value_via_configmap(core_api, request):  #
                        ""])
 
     cleanup_volume_by_name(client, vol_name)
+
+def test_setting_update_readonly():  # NOQA
```
khushboo-rancher, line 1246: Can we use `client` fixture or `reset_settings` fixture, even if the test fails somehow, the settings will be set to the default values. No need to handle the setting to default value.

khushboo-rancher (changes requested):
Test steps look fine, we can consider a small change.

manager/integration/tests/test_settings.py
mantissahz: We still allow modifying the read-only settings by `Longhorn APIs`.
I would change to modify settings by the manifest.
or wait for changing codes if we don't allow modifying the read-only settings by `Longhorn APIs`. cc @innobead

yangchiu:
@mantissahz Could you help to confirm if this test case can pass on master-head? Because I tried to run this test case, but it failed with assertion error: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5308/testReport/junit/tests/test_settings/test_setting_update_readonly/

Thanks!

mantissahz:
> @mantissahz Could you help to confirm if this test case can pass on master-head? Because I tried to run this test case, but it failed with an assertion error: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5308/testReport/junit/tests/test_settings/test_setting_update_readonly/

Sure, and no, it would not pass now because we still allow modifying read-only settings by APIs.
Do we have any method to update the setting by the manifest in our regression test?

yangchiu:
> Sure, and no, it would not pass now because we still allow modifying read-only settings by APIs. Do we have any method to update the setting by the manifest in our regression test?

Yes, `kubectl` is available in the test, Does calling `kubectl patch` to update the setting solve the problem?

mantissahz:
> Does calling kubectl patch to update the setting solve the problem?

@yangchiu, 
Yes, it should solve the problem by the `kubectl patch` cmd.

manager/integration/tests/test_settings.py
mantissahz: Modified to use the patching custom object API.

yangchiu:
Close this in favor of https://github.com/longhorn/longhorn-tests/pull/3302.

PR #1552, opened 2023-10-20
Refactor restoration tests with replica failure and add one

ejweber:
longhorn/longhorn#6905

The behavior requested in longhorn/longhorn#2103 was already added with the inclusion of the allow-volume-creation-with-degraded-availability setting. This PR adds a test with that setting on and confirms volume restoration completes. Because it is so similar to a couple of other tests, it refactors all three tests to use the same base functionality.

- [test_single_replica_unschedulable_restore_failure (new) results](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5057/testReport/tests/test_ha/)
- [test_rebuild_with_restoration results](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5059)
- [test_single_replica_restore_failure results](test_single_replica_restore_failure)
- [test_single_replica_unschedulable_restore_failure (new) backport to 1.5.x results](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5060/)

We should theoretically be able to backport all the way to v1.4.x, since the setting has been around that long. Haven't confirmed yet.

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -95,6 +95,8 @@ from node import taint_non_current_node
 SMALL_RETRY_COUNTS = 30
 BACKUPSTORE = get_backupstores()
 
+REPLICA_FAILURE_MODE_CRASH = "replica_failure_mode_crash"
+REPLICA_FAILURE_MODE_DELETE = "replica_failure_mode_delete"
 
 @pytest.mark.coretest   # NOQA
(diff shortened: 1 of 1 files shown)


ejweber:
@mergify backport v1.5.x

yangchiu (changes requested):
@ejweber The flake8 check failed. Could you help to fix it?

yangchiu (approved):
LGTM

PR #1553, opened 2023-10-23
test: fix flaky test case test_default_storage_class_syncup

yangchiu:
test: fix flaky test case test_default_storage_class_syncup

Since storage_api.read_storage_class is flaky, it should be wrapped in try-catch and for-loop.

For https://github.com/longhorn/longhorn/issues/6141 https://github.com/longhorn/longhorn/issues/6913


chriscchien (approved):
LGTM

khushboo-rancher (approved):

PR #1559, opened 2023-10-24
test: implement kubelet restart

yangchiu:
test: implement kubelet restart

currently only support k3s

For https://github.com/longhorn/longhorn/issues/6707

### DIFF
--- a/e2e/libs/kubelet/kubelet.py
+++ b/e2e/libs/kubelet/kubelet.py
@@ -1,0 +1,22 @@
+from utility.utility import logging
+import time
+
+from workload.pod import new_pod_manifest
+from workload.pod import create_pod
+from workload.pod import wait_for_pod_status
+from workload.pod import delete_pod
+from workload.pod import IMAGE_UBUNTU
+
+def restart_kubelet(node_name, stop_time_in_sec=10):
+    manifest = new_pod_manifest(
+        image=IMAGE_UBUNTU,
+        command=["/bin/bash"],
+        args=["-c", f"sleep 10 && systemctl stop k3s-agent && sleep {stop_time_in_sec} && systemctl start k3s-agent"],
+        node_name=node_name
+    )
+    pod_name = manifest['metadata']['name']
+    create_pod(manifest, is_wait_for_pod_running=True)
+
+    time.sleep(stop_time_in_sec)
+
+    delete_pod(pod_name)
--- a/e2e/libs/workload/pod.py
+++ b/e2e/libs/workload/pod.py
@@ -2,14 +2,14 @@ import time
 
 from kubernetes import client
 
-from robot.libraries.BuiltIn import BuiltIn
-
 from utility.utility import logging
 from utility.utility import generate_name
+from utility.utility import get_retry_count_and_interval
 
 
 IMAGE_BUSYBOX = 'busybox:1.34.0'
 IMAGE_LITMUX = 'litmuschaos/go-runner:latest'
+IMAGE_UBUNTU = 'ubuntu:16.04'
 
 def new_pod_manifest(image="", command=[], args=[],
                      claim_name="", node_name="", labels={}):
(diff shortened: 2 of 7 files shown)


e2e/libs/kubelet/kubelet.py
```diff
@@ -0,0 +1,22 @@
...
+
+def restart_kubelet(node_name, stop_time_in_sec=10):
+    filepath = "./templates/kubelet/kubelet_restart.sh"
+    dst_filepath = "/tmp/kubelet_restart.sh"
+    with open(filepath, 'r') as f:
+        content = f.read()
+        cmd = [
+            "sh",
+            "-c",
+            f"touch {dst_filepath}; chmod +x {dst_filepath}; cat > {dst_filepath} <<-'EOF'\n{content}\nEOF"
```
c3y1huang, line 13: This looks like it will add a file to the host environment.
Why not just have 2 functions `stop_kubelet` and `start_kubelet` with `systemctl stop/start k3s-agent`?

e2e/templates/kubelet/kubelet_restart.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+STOP_TIME_IN_SEC=$1
+
+sleep 3; # avoid remote host connection lost before pod execution returns
```
c3y1huang, line 5: Could you elaborate more?

yangchiu: Otherwise `connect_get_namespaced_pod_exec` will throw exception during command execution since once kubelet stopped, the connection lost.

e2e/libs/kubelet/kubelet.py
yangchiu: This is feasible after `ssh` implemented.

c3y1huang: Is it possible with `nsenter`?

c3y1huang: Never mind, I see your point.

yangchiu: I've tried `nsenter` before, but so far unable to set the options/parameters right to make it work. Do you have any suggestion?

c3y1huang: What about we do something similar to the Litmus helper pod?
https://github.com/litmuschaos/litmus-go/blob/bdddd0d8036cbd57880f64db39817906d69ff823/chaoslib/litmus/kubelet-service-kill/lib/kubelet-service-kill.go#L115

yangchiu: It doesn't seem to have much difference?
The command used in Litmus is:
```
sleep 10 && systemctl stop kubelet && sleep " + strconv.Itoa(experimentsDetails.ChaosDuration) + " && systemctl start kubelet
```

c3y1huang: It doesn't directly create a file on the host. And, we should also consider leverage on `IMAGE_LITMUX = 'litmuschaos/go-runner:latest'`,  which should come equipped with all the necessary packages.

yangchiu: More investigation needed. 

Probably related to what's the approach we'd like to execute commands: `Using a permanent pod to run all commands` or `Using a disposable pod to run every command`.

c3y1huang: We could also enhance later, considering that this PR appears to be fulfilling its purpose. WDYU @yangchiu ?

yangchiu: Node execution pod is a delegate permanently existing on each node to execute any random commands for users.

Helper pod is created to execute pre-defined specific commands in its manifest and destroyed once it finished its job.

Yes, it can be refined later since at this moment there's no way to tell which approach is more suitable for which situation.

yangchiu: I've confirmed this can be done by a helper pod, but the image is still needed to be [`ubuntu:16.04`](https://github.com/litmuschaos/litmus-go/blob/bdddd0d8036cbd57880f64db39817906d69ff823/pkg/generic/kubelet-service-kill/environment/environment.go), and some configuration need to be done. I can refine in this PR.

yangchiu: Refined.

e2e/libs/workload/pod.py
```diff
@@ -31,15 +31,35 @@ def new_pod_manifest(image="", command=[], args=[],
...
+                'name': 'bus',
+                'hostPath': {
+                    'path': '/var/run'
+                }
+            }, {
+                'name': 'rancher',
+                'hostPath': {
+                    'path': '/var/lib/rancher'
+                }
+            }]
```
c3y1huang, line 62: Could we have a new function `new_pod_manifest_for_node_exec` that uses `new_pod_manifest` for the base manifest and then append the volumes fields?

c3y1huang (changes requested):

e2e/libs/workload/pod.py
yangchiu: Can be refined later after sort out what are the possible required configurations for different situations.

c3y1huang (approved):

PR #1566, opened 2023-10-27
fix(negative): duplicated stress pod

c3y1huang:
Avoid duplicating stress pods per node when called inside of a test loop.


e2e/libs/node/constant.py
```diff
@@ -1,6 +1,7 @@
 NODE_STRESS_CPU_LOAD_PERCENTAGE = 100
 NODE_STRESS_MEM_LOAD_PERCENTAGE = 100
 NODE_STRESS_MEM_VM_WORKERS = 1
-NODE_STRESS_TIMEOUT_SECOND = 300
+NODE_STRESS_TIMEOUT_SECOND = 60 * 60  # 1 hour
 
-LABEL_STRESS_HELPER = "longhorn-stress-helper"
+STRESS_HELPER_LABEL = "longhorn-stress-helper"
```
yangchiu, line 6: If a constant is only possible to be used in a single one place, like `STRESS_HELPER_LABEL` will be used only in `stress.py`, Does it still have the benefit of placing in a separate file?

c3y1huang: I think it's easier to locate and manage constants when they are in their own separate file.

yangchiu (approved):
LGTM

PR #1567, opened 2023-10-27
test(negative): stress cpu/mem during volume attach detach

c3y1huang:
https://github.com/longhorn/longhorn/issues/6951

### DIFF
--- a/e2e/keywords/node.resource
+++ b/e2e/keywords/node.resource
@@ -67,8 +67,8 @@ Restart cluster
         wait_for_workload_pod_stable   ${statefulset}
     END
 
-During replica rebuilding, stress volume node cpu
+Stress volume node cpu
     stress_node_cpu_by_volume    ${volume_name}
 
-During replica rebuilding, stress volume node memory
+Stress volume node memory
     stress_node_memory_by_volume    ${volume_name}
--- a/e2e/keywords/volume.resource
+++ b/e2e/keywords/volume.resource
@@ -18,6 +18,12 @@ Create volume ${idx} with ${size} GB and ${replica_count} replicas
     attach_volume    ${volume_name}
     Insert Into List    ${volume_list}    ${idx}    ${volume_name}
 
+Attach volume to node
+    attach_volume    ${volume_name}
+
+Detach volume from node
+    detach_volume    ${volume_name}
+
 Write data to the volume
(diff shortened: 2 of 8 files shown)


yangchiu (changes requested):
According to https://confluence.suse.com/pages/viewpage.action?spaceKey=LON&title=Negative+Testing `Phase 3` - `9. Increase CPU load on the nodes` and `10. High memory consumption`

It first inject the negative factor:
```
1. Deploy multiple apps on the cluster to increase the load of the CPU or on individual node increase load on CPU.
```
```
1. Deploy multiple apps on the cluster to increase memory consumption or on individual node increase the memory consumption.
```
And then run the for loop to do longhorn operations:
```
2. Perform the operations for a time period of 15-30 mins.
```

So probably in these test cases we inject high cpu/memory usage first, just like what we did in `control plane node network latency`, and then enter the for loop. And you don't even have to deal with duplicated stress pod.

c3y1huang:
> According to https://confluence.suse.com/pages/viewpage.action?spaceKey=LON&title=Negative+Testing `Phase 3` - `9. Increase CPU load on the nodes` and `10. High memory consumption`
> 
> It first inject the negative factor:
> 
> ```
> 1. Deploy multiple apps on the cluster to increase the load of the CPU or on individual node increase load on CPU.
> ```
> 
> ```
> 1. Deploy multiple apps on the cluster to increase memory consumption or on individual node increase the memory consumption.
> ```
> 
> And then run the for loop to do longhorn operations:
> 
> ```
> 2. Perform the operations for a time period of 15-30 mins.
> ```
> 
> So probably in these test cases we inject high cpu/memory usage first, just like what we did in `control plane node network latency`, and then enter the for loop. And you don't even have to deal with duplicated stress pod.

This is still a valid test case from my perspective. The user started off with an unstressed environment, and over time, there could be an increase in CPU and memory pressure. The scenario you've highlighted specifically tests the behavior of a pre-stressed environment, which I believe is also covered during the iteration of the test loop.

As for the duplicated stress pod, it's a general fix since we are keeping the LOOP_COUNT as an open parameter, so the timeout of the stress pod should reflect individual test runs and even the test run time. If not, it might terminate before the tests are completed. Do you have any concerns about that fix?

yangchiu (approved):
LGTM

yangchiu (approved):

PR #1568, opened 2023-10-27
test: sync test.yaml file to use cluster admin role

yangchiu:
test: sync test.yaml file with PR https://github.com/longhorn/longhorn-tests/pull/1486 to use cluster admin role to run the test


yangchiu:
Need this otherwise `test_backuptarget_invalid` fails on `v1.5.x`.

innobead (approved):

yangchiu:
@Mergify backport v1.4.x

PR #1570, opened 2023-10-30
refactor(negative): move constants to constant.py

c3y1huang:
Improve readability.


khushboo-rancher (approved):
Nit: Should we use later version of Ubuntu instead of 16.04?

PR #1577, opened 2023-11-01
test(negative): stress cpu/mem during volume expansion

c3y1huang:
https://github.com/longhorn/longhorn/issues/6982


e2e/libs/persistent_volume_claim/rest.py
```diff
@@ -0,0 +1,15 @@
+from persistent_volume_claim.base import Base
+
+from utility.utility import get_longhorn_client
+
+class Rest(Base):
+
+    def __init__(self):
+        self.longhorn_client = get_longhorn_client()
```
yangchiu, line 8: Is pvc part of Longhorn that can be manipulate by Longhorn client? If not, maybe we don't need this Rest implementation.

e2e/libs/persistent_volume_claim/crd.py
```diff
@@ -0,0 +1,40 @@
...
+        self.core_v1_api = client.CoreV1Api()
+        self.retry_count, self.retry_interval = get_retry_count_and_interval()
+
+    def get(self, claim_name, claim_namespace="default"):
+        return self.core_v1_api.read_namespaced_persistent_volume_claim(
+            name=claim_name,
+            namespace=claim_namespace,
+        )
+
+    def expand(self, claim_name, size, namespace="default"):
```
yangchiu, line 21: Besides `expand`, will we support any other operation on pvc needed to be added into this class?

e2e/tests/stress_cpu.robot
```diff
@@ -0,0 +1,61 @@
...
+        Then Wait for statefulset 0 volume size expanded
+        And Check statefulset 0 data is intact
+    END
+
+Stress Volume Node CPU When Volume Is Offline Expanding
+    @{data_checksum_list} =    Create List
+    Set Test Variable    ${data_checksum_list}
+
+    Given Create statefulset 0 with rwo volume
+    And Write 1024 MB data to statefulset 0
```
yangchiu, line 50: We have Write **1024 MB** data to statefulset 0 and **Expand statefulset 0 volume by 100 MiB**

When should we use MB and MiB respectively?

e2e/libs/persistent_volume_claim/crd.py
c3y1huang: This is all we need for the scenarios in this PR. Additional operations will be determined as more cases are added. These can be addressed later.

e2e/keywords/workload.resource
```diff
@@ -21,6 +41,15 @@ Create statefulset ${idx} with ${volume_type} and ${option} volume
     ${statefulset_name} =    create_statefulset    ${volume_type}    ${option}
     Insert Into List    ${statefulset_list}    ${idx}    ${statefulset_name}
 
+Expand statefulset ${idx} volume by ${size} MiB
```
yangchiu, line 44: Since we've been able to get pvc name from workload name, maybe we can hide the details in the lower layer like `keywords folder` or even `workload.py` to keep it cohesive on a best effort and save the variable you need to record in resource layer and reduce resource layer complexity.

e2e/tests/stress_cpu.robot
```diff
@@ -0,0 +1,61 @@
...
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        When Delete replica on volume node to trigger replica rebuilding
+        And Stress the CPU of all volume nodes
+
+        Then Wait until replica on volume node rebuilt
+        And Check data is intact
+    END
+
+Stress Volume Node CPU When Volume Is Online Expanding
+    @{data_checksum_list} =    Create List
```
yangchiu, line 31: Would it be possible if there's a more elegant way to deal with this instead of having lots of variables in resource layer? WDYT?

e2e/libs/node/utility.py
```diff
@@ -7,3 +9,31 @@ def get_node_by_name(node_name):
 def get_node_cpu_cores(node_name):
     node = get_node_by_name(node_name)
     return node.status.capacity['cpu']
+
+def list_node_names_by_role(role="worker"):
```
yangchiu, line 13: There are similar functions in `utility/utility.py`. Is it possible to merge them?

e2e/tests/stress_cpu.robot
c3y1huang: The dd command is set to use a block size of 1M, whereas the requested storage for the StatefulSet was 3Gi. We can enhance later to make them consistent, but I think its trivial.

e2e/libs/node/utility.py
c3y1huang: Sure

e2e/tests/stress_cpu.robot
c3y1huang: I think it's fine for now. We can make improvements later when we have more cases to determine what should be included in the variable section.

e2e/keywords/workload.resource
yangchiu:
```
${expected_size} = expand_pvc_size_by_mib ${persistentvolumeclaim_list}[${idx}] ${size}
```
And
```
${pvc_name} = get_workload_pvc_name ${statefulset_list}[${idx}]
${expected_size} = expand_pvc_size_by_mib ${pvc_name} ${size}
```
Can you help to elaborate more on why the former is more transparent than the latter?

e2e/tests/stress_cpu.robot
yangchiu: Variable section doesn't seems to be helpful. I'm afraid that setting variable (list) straightforwardly without a well-designed mechanism will bloat the code fast and make things hard to control when we deal with more complex test cases. Like the following test case:

```
create deployment
write data to deployment and create snapshot 0
write data to deployment and create snapshot 1
write data to deployment and create snapshot 2
check deployment data
revert to snapshot 1 and check deployment data
revert to snapshot 0 and check deployment data
create statefulset
write data to statefulset and create snapshot 0
write data to statefulset and create snapshot 1
write data to statefulset and create snapshot 2
check statefulset data
revert to snapshot 1 and check statefulset data
revert to snapshot 0 and check statefulset data
```

How can we get the correct checksum from a single checksum list contains checksums from different workloads and different snapshots?

e2e/keywords/workload.resource
c3y1huang: I am confused. Why change to the latter? Is it about the `${persistentvolumeclaim_list}`? We cannot get the workload PVC when StatefulSet are scaled down.

e2e/tests/stress_cpu.robot
c3y1huang: You're probably right. Using a variable section might not be the best approach. We can improve this later when we're dealing with more complex cases.

e2e/keywords/workload.resource
yangchiu: Yes, workload scaled down can be a reason. 

But again, for a test case involves multiple different type of workloads, like the following existing test case, how do we distinguish one from others using a single `${persistentvolumeclaim_list}`?

```
Given Create deployment 0 with rwo volume
And Create deployment 1 with rwx volume
And Create deployment 2 with rwo and strict-local volume
And Create statefulset 0 with rwo volume
And Create statefulset 1 with rwx volume
And Create statefulset 2 with rwo and strict-local volume
```

e2e/libs/node/utility.py
c3y1huang: https://github.com/longhorn/longhorn-tests/pull/1582

e2e/keywords/workload.resource
c3y1huang: 1. What scenario are we testing that would need different types of workloads?
1. Currently, the list of persistent volume claims (and other created lists) is differentiated by index. For your case and the way the current codebase implemented, we use the index to distinguish them:
    ```robot
    Given Create deployment 0 with rwo volume
    And Create deployment 1 with rwx volume
    And Create deployment 2 with rwo and strict-local volume
    And Create statefulset 3 with rwo volume
    And Create statefulset 4 with rwx volume
    And Create statefulset 5 with rwo and strict-local volume
    ```
    But to be honest, I don't think using an index is the best way to go about it. It would probably be more logical and clearer if we distinguished them by the name of the workload instead. And the improvement we are discussing should be done in a separate PR.
1. Introducing of the `${persistentvolumeclaim_list}` is to handle the PVC of the downscaled statefulset.

e2e/libs/persistent_volume_claim/crd.py
c3y1huang: I think it's more organized to have a class for each CR. If it's unnecessary after we add other cases, we can reorganize it later.

e2e/tests/stress_cpu.robot
yangchiu: We probably should figure out this now since we're defining the framework, otherwise when team members or contributors participate in, they'll follow the awkward pattern to write more test cases and make our lives difficult.

e2e/keywords/workload.resource
yangchiu: > I don't think using an index is the best way to go about it. It would probably be more logical and clearer if we distinguished them by the name of the workload instead.

Yes, totally agree.

c3y1huang: Lets handle in https://github.com/longhorn/longhorn/issues/7034.

e2e/tests/stress_cpu.robot
c3y1huang: Let's handle in https://github.com/longhorn/longhorn/issues/7034.

e2e/keywords/workload.resource
yangchiu: > What scenario are we testing that would need different types of workloads?

It's just an example not just for `pvc - workload mapping` but also can be generalize to other situation like the `write data/snapshot/backup - checksum mapping` [below](https://github.com/longhorn/longhorn-tests/pull/1577#discussion_r1379931280). They're actually the same problem.

But for the original question, making the test framework flexible to the maximum helps us to meet any requirements. It makes us be able to randomly combine steps to mimic users' environment/setup and try to reproduce their issue.

e2e/tests/stress_cpu.robot
c3y1huang: See if we can handle in Let's address this in https://github.com/longhorn/longhorn/issues/7034.

e2e/libs/persistent_volume_claim/base.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC, abstractmethod
+
+class Base(ABC):
```
yangchiu, line 3: If pvc has only one possible implementation (which is crd), why do we still need this abstract class to define the interfaces to be implemented? Since there's no other one to extend and have its own implementation.

e2e/keywords/workload.resource
yangchiu: > We cannot get the workload PVC when StatefulSet is scaled down.

How about `get pvc from deployment/statefulset` instead of `get pvc from pod`. If it uses `volumes.persistentVolumeClaim.claimName`, the pvc name has already been defined. If it uses `volumeClaimTemplates`, the pvc name follows the convention of `volumeClaimTemplates.metadata.name` + `StatefulSet name` + `index`. 

Does this solve the problem when StatefulSet is scaled down?

c3y1huang: I prefer not to complicate this further as the `${persistentvolumeclaim_list}` works for this PR. Will handle the refactoring in https://github.com/longhorn/longhorn/issues/7034.

yangchiu (dismissed):
LGTM

yangchiu (approved):

yangchiu (approved):

PR #1578, opened 2023-11-01
Update `Test System Upgrade with New Instance Manager`

roger-ryao:
Ref: longhorn/longhorn#7013


khushboo-rancher (changes requested):
I think, its better we can just modify the lines instead of adding a note. We can just keep the latest test cases as we are not maintaining versions of test cases.

khushboo-rancher (approved):

PR #1581, opened 2023-11-02
refactor(negative): enhance stress cpu/mem keywords

c3y1huang:
- Reorganize to another file for test case expansion
- Improve reliability
- Consolidate duplicated codes

### DIFF
--- a/e2e/tests/stress_cpu.robot
+++ b/e2e/tests/stress_cpu.robot
@@ -20,7 +20,7 @@ Stress Volume Node CPU When Replica Is Rebuilding
 
     FOR    ${i}    IN RANGE    ${LOOP_COUNT}
         When Delete replica on volume node to trigger replica rebuilding
-        And Stress volume node cpu
+        And Stress the CPU of all volume nodes
 
         Then Wait until replica on volume node rebuilt
         And Check data is intact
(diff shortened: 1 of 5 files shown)


e2e/tests/stress_cpu.robot
```diff
@@ -0,0 +1,27 @@
...
+
+*** Test Cases ***
+
+Stress Volume Node CPU When Replica Is Rebuilding
+    Given Create a volume with 5 GB and 3 replicas
+    And Write data to the volume
+
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        When Delete replica on volume node to trigger replica rebuilding
+        And Stress the CPU of all volume nodes
```
yangchiu, line 23: Since there is only **one** volume created, why we say Stress the CPU of **all** volume node**s** here?

c3y1huang: This enhancement involves utilizing 'volume_list' instead of 'single volume_name'. I don't believe we require an additional keyword to manage the plural; it serves the same purpose. Thus, in the scenario of a single volume, it strains the CPU of that specific volume node.

yangchiu (approved):
LGTM

yangchiu (approved):

PR #1582, opened 2023-11-03
refactor(negative): cleanup get/list node functions

c3y1huang:
- Consolidated list node functions.
- Rename and reorganize the get node function.


yangchiu (approved):
LGTM

PR #1585, opened 2023-11-11
test(metrics): PVC namespace included in volume metrics

antoninferrand:
https://github.com/longhorn/longhorn/issues/7077


yangchiu (approved):
LGTM. It's better to include scenarios of different namespace like https://github.com/longhorn/longhorn/issues/7077#issuecomment-1812150822.

But at least it verifies the namespace field included in metrics, it's good enough for now.

PR #1594, opened 2023-11-16
infra: allow ssh into nodes to make debugging easier

yangchiu:
infra: allow ssh into nodes to make debugging easier

For https://github.com/longhorn/longhorn/issues/6601


pipelines/utilities/kubeconfig.sh
```diff
@@ -12,6 +12,8 @@ set_kubeconfig(){
   elif [[ "${TF_VAR_arch}" == "arm64"  ]]; then
     export KUBECONFIG="${PWD}/test_framework/terraform/${LONGHORN_TEST_CLOUDPROVIDER}/${DISTRO}/k3s.yaml"
   fi
+
+  cat "${KUBECONFIG}"
```
innobead, line 16: Could this expose credentials outside in the job build?

innobead (changes requested):
Suggest making exporting worker nodes public optional, so this could make us have more flexibilty for testing env setup.

pipelines/utilities/kubeconfig.sh
yangchiu: From my understanding there are 3 fields in KUBECONFIG: `certificate-authority-data`, `client-certificate-data` and `client-key-data`. `certificate-authority-data` is SSL server certificate. `client-certificate-data` and `client-key-data` are also `openssl` generated certificate and key which to be approved to access the k8s cluster, so they don't involve personal ssh key or credential. Even though one gets this KUBECONFIG file, he can only manipulate the created k8s cluster / ec2 instances, and has nothing to do with our infra machines.

yangchiu:
> Suggest making exporting worker nodes public optional, so this could make us have more flexibilty for testing env setup.

I'll add this option.

yangchiu:
> > Suggest making exporting worker nodes public optional, so this could make us have more flexibilty for testing env setup.
> 
> I'll add this option.

Switching between public ip and private network is more complicated than expected. Simply disabled `associate_public_ip_address` won't work. Need more time to investigate.

yangchiu:
> Switching between public ip and private network is more complicated than expected. Simply disabled `associate_public_ip_address` won't work. Need more time to investigate.

Refined. If it has a public ip, it can directly communicate with Internet. If it doesn't have a public ip, use nat to communicate with Internet.

yangchiu:
> Suggest making exporting worker nodes public optional, so this could make us have more flexibilty for testing env setup.

Added option in https://github.com/longhorn/infra/pull/153.

pipelines/utilities/kubeconfig.sh
yangchiu: We don't directly print out `kubeconfig` now, but users can ssh into control plane node to retrieve it. For example, the default file path for `k3s` is `/etc/rancher/k3s/k3s.yaml`.

innobead (changes requested):
Instead of removing the private network for worker nodes, have you consided bastion way to ssh from local to the internal nodes?

Suggest this way to avoid avoidable instance access exposure (security). 
https://repost.aws/knowledge-center/ec2-troubleshoot-bastion-host-connection

yangchiu:
More investigation needed. Close this for now.

innobead:
Sounds good @yangchiu

PR #1595, opened 2023-11-17
test(manual): support bundle should include kubelet log when on K3s cluster

c3y1huang:
https://github.com/longhorn/longhorn/issues/7121


chriscchien (approved):
LGTM and test result [here](https://github.com/longhorn/longhorn/issues/7121#issuecomment-1818124493) based on the PR test steps.

PR #1607, opened 2023-11-29
Fix test_disk_eviction_with_node_level_soft_anti_affinity_disabled

ejweber:
longhorn/longhorn#7210

The node controller refuses to update a node until its disk monitor has synced. In v1.6.0, the node controller is responsible for triggering replica eviction. This means it may take longer for replica eviction to start than it did in the past. In the long term, we may want to try to tighten up this timing again, but unfortunately, there is no obvious and easy way to do so.

test_disk_eviction_with_node_level_soft_anti_affinity_disabled is currently failing because it does not wait long enough for a replica to be evicted from one disk before checking for it on another. This PR ensures the eviction process has started before proceeding to do the check.

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -2626,6 +2626,14 @@ def test_disk_eviction_with_node_level_soft_anti_affinity_disabled(client, # NOQ
     replica_path = test_disk_path + '/replicas'
     assert os.path.isdir(replica_path)
 
+    # Since https://github.com/longhorn/longhorn-manager/pull/2138, the node
+    # controller is responsible for triggering replica eviction. If the timing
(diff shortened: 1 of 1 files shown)


ejweber:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5323/testReport/tests/test_node/

khushboo-rancher (approved):

ejweber:
I initially requested @roger-ryao's review because he raised the associated issue. I am fine with merging based on @khushboo-rancher's review if that's more convenient, however.

roger-ryao (approved):
LGTM

ejweber:
@mergify backport v1.5.x

PR #1608, opened 2023-12-01
Script to generate test data

chriscchien:
Add generate test data script 

- test_tools/gen_data/clean.sh
   - clean workload
- test_tools/gen_data/config.yaml
   - config
- test_tools/gen_data/deployment.yaml  
   - template
- test_tools/gen_data/README.md
- test_tools/gen_data/run.sh
- test_tools/gen_data/statefulset.yaml    
   - template

`./run.sh`
`./clean.sh`
 
ref: [6665](https://github.com/longhorn/longhorn/issues/6665)

### DIFF
--- a/test_tools/gen_data/README.md
+++ b/test_tools/gen_data/README.md
@@ -1,0 +1,46 @@
+# Generate test data script
+Generate RWO/RWX workloads, write data into `/mnt/data/data` in workloads and record md5 to data.output.
+
+# Usage
+Modify config.yaml
+```yaml
+storage: 1Gi # Each volume size
+storageClass: longhorn-test # Need to prepare your own storage class first
+dataSizeInMb: 500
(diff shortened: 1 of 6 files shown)


yangchiu (changes requested):
Is it necessary to have dependencies on the existing pytest code? Can't we use shell scripts to simulate what we usually do in manual testing? WDYT

chriscchien:
> Is it necessary to have dependencies on the existing pytest code? Can't we use shell scripts to simulate what we usually do in manual testing? WDYT

There are many ways to achieve the goal and looks like you already have shell script to do the same thing, could you share the code for some inspiration? many thanks.

chriscchien:
Updated to shell script

yangchiu (approved):
LGTM

PR #1615, opened 2023-12-06
test(integration): replica-auto-balance when disabled disk scheduling in zone

c3y1huang:
https://github.com/longhorn/longhorn/issues/6508
https://github.com/longhorn/longhorn/issues/7273


chriscchien (approved):
LGTM

PR #1616, opened 2023-12-07
test(integration): replica-auto-balance when no storage available

c3y1huang:
https://github.com/longhorn/longhorn/issues/6671
https://github.com/longhorn/longhorn/issues/7276


yangchiu (approved):
LGTM

PR #1619, opened 2023-12-11
test(manual): upgrade responder should collect SPDK info

c3y1huang:
https://github.com/longhorn/longhorn/issues/6033

Depends on https://github.com/longhorn/longhorn-manager/pull/2346.


chriscchien (approved):
Following step perform testing, did not encounter problem and here is the test [result](https://github.com/longhorn/longhorn/issues/6033#issuecomment-1867473977)

PR #1620, opened 2023-12-12
ci: fix build test env image system-wide python package installation error

yangchiu:
ci: fix build test env image system-wide python package installation error

For https://github.com/longhorn/longhorn/issues/7308


innobead (approved):
LGTM. 

But, why wasn't this issue raised before?

yangchiu:
> LGTM.
> 
> But, why wasn't this issue raised before?

This issue is introduced by the latest version of `alpine` image (3.19.0): 
- Python’s package directory is now [marked](https://gitlab.alpinelinux.org/alpine/infra/aports/-/commit/6c2a300f6787544f3194e7f6cc19058baf9b7419) as [externally managed](https://packaging.python.org/en/latest/specifications/externally-managed-environments/), which means that pip can no longer install to system directory which is managed by apk. Users may use [pipx](https://pipx.pypa.io/stable/) instead.

ref: https://www.alpinelinux.org/posts/Alpine-3.19.0-released.html

PR #1621, opened 2023-12-12
infra: add custom ssh public key for debugging

yangchiu:
infra: add custom ssh public key for debugging

For https://github.com/longhorn/longhorn/issues/7307

### DIFF
--- a/test_framework/Jenkinsfile
+++ b/test_framework/Jenkinsfile
@@ -136,6 +136,7 @@ node {
                                        --env TF_VAR_azure_tenant_id=${AZURE_TENANT_ID} \
                                        --env TF_VAR_azure_subscription_id=${AZURE_SUBSCRIPTION_ID} \
                                        --env TF_VAR_cis_hardening=${CIS_HARDENING} \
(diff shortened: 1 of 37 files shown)


test_framework/Jenkinsfile
```diff
@@ -136,6 +136,7 @@ node {
                                        --env TF_VAR_azure_tenant_id=${AZURE_TENANT_ID} \
                                        --env TF_VAR_azure_subscription_id=${AZURE_SUBSCRIPTION_ID} \
                                        --env TF_VAR_cis_hardening=${CIS_HARDENING} \
+                                       --env TF_VAR_custom_ssh_public_key="${CUSTOM_SSH_PUBLIC_KEY}" \
```
innobead, line 139: Would this be exported in the log of a job build? If yes, we should use this only for authorized users.

innobead (approved):
LGTM.

test_framework/Jenkinsfile
yangchiu: @innobead This PR https://github.com/longhorn/infra/pull/151 isn't merged yet, which only provides `CUSTOM_SSH_PUBLIC_KEY` option in `private/longhorn-tests-regression` job.

PR #1618, opened 2023-12-08
add case test_node_metrics

chriscchien:
Add test case `test_node_metrics`

ref: [4029](https://github.com/longhorn/longhorn/issues/4029)

[doc](https://longhorn.io/docs/1.5.3/monitoring/metrics/#node) and test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5358/testReport/tests/test_metric/)

- `longhorn_node_status`, have 4 condititons
   - **ready**, assert value == 1 (we can not stop node by pytest)
   - **mountpropagation**, assert value == 1
   - **allowScheduling**, assert value ==1 when node allow scheduling
   - **allowScheduling**, assert value ==0 when node not allow scheduling
   - **schedulable**, assert value == 1 when node is uncordoned
   - **schedulable**, assert value == 0  and  condition_reason = "KubernetesNodeCordoned" when node is cordoned
```
longhorn_node_status{condition="allowScheduling",condition_reason="",node="ip-172-31-41-106"} 1
longhorn_node_status{condition="mountpropagation",condition_reason="",node="ip-172-31-41-106"} 1
longhorn_node_status{condition="ready",condition_reason="",node="ip-172-31-41-106"} 1
longhorn_node_status{condition="schedulable",condition_reason="",node="ip-172-31-41-106"} 1
```
- `longhorn_node_count_total`,  aseert value == 3
- `longhorn_node_cpu_capacity_millicpu`, assert value  > 0
- `longhorn_node_cpu_usage_millicpu`, assert value > 0 
- `longhorn_node_memory_capacity_bytes`, assert value > 0
- `longhorn_node_memory_usage_bytes`, assert value > 0
- `longhorn_node_storage_capacity_bytes`, assert value == `disk.storageMaximum`
- `longhorn_node_storage_usage_bytes`, assert value > 0. because `node.storageAvailable` not totally equally to prometheus_client return value
```
root@ip-172-31-41-106:/home/ubuntu# k get node.longhorn.io ip-172-31-46-250 -n longhorn-system -o yaml | grep storageAvailable
      storageAvailable: 75916902400

Sample(name='longhorn_node_storage_usage_bytes', labels={'node': 'ip-172-31-46-250'}, value=7136530432.0, timestamp=None, exemplar=None)
```
- `longhorn_node_storage_reservation_bytes`, assert value == `disk.storageReserved`

PR #1622, opened 2023-12-12
refactor(integration): make `check_volume_metric` a generic function

c3y1huang:
Rename and replace the hardcoded label keys to make the function `check_volume_metric` more generic.


chriscchien (approved):
LGTM.

I will revise https://github.com/longhorn/longhorn-tests/pull/1618 based on this refactor.

PR #1629, opened 2023-12-15
Add skeleton for test_do_not_react_to_brief_kubelet_restart

ejweber:
longhorn/longhorn#7302

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -2806,3 +2806,17 @@ def test_auto_detach_volume_when_node_is_cordoned(client, core_api, volume_name)
 
     volumes = client.list_volume().data
     assert len(volumes) == 0
+
+@pytest.mark.skip(reason="TODO")  # NOQA
+def test_do_not_react_to_brief_kubelet_restart():
+    """
+    Test the node controller ignores Ready == False due to KubeletNotReady for
+    ten seconds before reacting.
+
+    Repeat the following five times:
+    1. Verify status.conditions[type == Ready] == True for the Longhorn node we
+       are running on.
+    2. Kill the kubelet process (e.g. `pkill kubelet`).
+    3. Verify status.conditions[type == Ready] != False for the Longhorn node
+       we are running on at any point for at least ten seconds.
+    """


yangchiu:
Tested on master-head (longhorn-manager 7aa2625) by executing `systemctl restart k3s-agent.service` on a worker node, but I can observe the `Ready` status became `False` during the k3s-agent restart:
```
$ kubectl get nodes.longhorn.io ip-10-0-1-210 -n longhorn-system -ojsonpath={.status.conditions[0]} -w
{"lastProbeTime":"","lastTransitionTime":"2023-12-19T03:54:50Z","message":"Node ip-10-0-1-210 is ready","reason":"","status":"True","type":"Ready"}
{"lastProbeTime":"","lastTransitionTime":"2023-12-19T03:55:25Z","message":"Kubernetes node ip-10-0-1-210 not ready: KubeletNotReady","reason":"KubernetesNodeNotReady","status":"False","type":"Ready"}
{"lastProbeTime":"","lastTransitionTime":"2023-12-19T03:55:26Z","message":"Node ip-10-0-1-210 is ready","reason":"","status":"True","type":"Ready"}
```

@ejweber Could you help to confirm? Thank you!

ejweber:
Thanks for looking into this @yangchiu. So far, I am unable to reproduce it on `1.25.12+rke2r1`. I thought maybe restarting the entire service as you are doing caused kubelet to be down longer than the allowed 10 seconds (my original testing just killed the kubelet process), but my RKE2 testing and @chriscchien's testing in https://github.com/longhorn/longhorn/issues/7302#issuecomment-1859663202 don't agree with that so far.

Can you execute the test with the following simultaneous watches (or similar if you don't use `yq`)? This should allow us to see how long kubelet is actually posting the unready status. Maybe we need to increase the timeout. Also, could you provide a support bundle following test failure so I can examine the logs?

```
kubectl get node -oyaml -w | yq '[{"name":.metadata.name, "conditions":.status.conditions}]'

and

kubectl get node.longhorn.io -n longhorn-system -oyaml -w | yq '[{"name":.metadata.name, "conditions":.status.conditions}]'
```
```

**Comment by ejweber** [comment]
Following up on @yangchiu's test failures.

My original testing was with RKE2 and the specific action I took was to kill the `kubelet` process directly. Generally, `kubelet` only posted a `Ready == False` status for a second or two, so I thought I was being generous with the ten second window.

Testing ten times with `v1.26.11+k3s2`, I see different behavior. There is no independent `kubelet` process to kill, so we must restart the `k3s-agent`. In my environment, this results in `kubelet` posting `Ready == False` for about ten seconds.

```
11 s -> test failed
10 s -> test passed
10 s -> test passed
0 s  -> test passed
10 s -> test passed
10 s -> test failed
10 s -> test passed
11 s -> test passed
11 s -> test passed
0 s  -> test passed
```

I am not inclined to increase the window by very much, especially since Harvester (which is the driver for opening this issue in the first place) is RKE2 based. I'm sure there are lots of factors that determine how long it takes `kubelet` to post ready again when `k3s-agent` restarts, and there is no guarantee we can wait long enough. However, it seems like a good idea to increase it to fifteen seconds in light of these results. I will open a PR to do that.

ejweber:
This can be reevaluated now, @yangchiu.

innobead (approved):

innobead:
https://github.com/longhorn/longhorn/issues/7302 was closed. Please follow up with the later implementation. 

cc @longhorn/qa

PR #1633, opened 2023-12-20
test(manual): upgrade-responder collect average sizes for v1 volumes only

c3y1huang:
https://github.com/longhorn/longhorn/issues/7380


chriscchien (approved):
Following steps to do test, did not encounter problem and here is the test [result](https://github.com/longhorn/longhorn/issues/7380#issuecomment-1866166075)

innobead (approved):

PR #1639, opened 2023-12-26
test: fix flaky test case test_offline_node

yangchiu:
test: fix flaky test case test_offline_node

Adding resource suffix filter to select the correct ec2 instance to be powered off. Originally it only used instance private ip to select the instance, but private ip is not unique, so it's possible to power off the wrong instance which belongs to another cluster instead of the current test case running cluster.

For https://github.com/longhorn/longhorn/issues/7463


innobead (approved):

yangchiu:

PR #1640, opened 2023-12-26
e2e: add test_setting_v1_data_engine

derekbit:
Longhorn/longhorn#7095

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -22,7 +22,7 @@ from common import (  # NOQA
     get_engine_image_status_value,
     create_volume, create_volume_and_backup, cleanup_volume_by_name,
     wait_for_volume_restoration_completed, wait_for_backup_restore_completed,
-    get_engine_host_id,
+    get_engine_host_id, wait_for_instance_manager_count,
     Gi, Mi,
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_settings.py
```diff
@@ -1241,3 +1242,58 @@ def test_setting_update_with_invalid_value_via_configmap(core_api, request):  #
...
     cleanup_volume_by_name(client, vol_name)
+
+
+def test_setting_v1_data_engine(client, request): # NOQA
+    """
+    Test that the v1 data engine setting works correctly.
+    1. Create a volume and attach it.
+    2. Set v1 data engine setting to false. The setting should be rejected.
+    3. Detach the volume.
+    4. Set v1 data engine setting to false again. The setting should be
```
innobead, line 1253: You can add one more step between 4 and 5. After disabling the v1 engine, attach the volume and it should fail.

innobead (changes requested):

manager/integration/tests/test_settings.py
derekbit: Added. Thank you.

innobead (approved):

PR #1641, opened 2023-12-28
Update comment for excluded settings test case.

james-munson:
longhorn/longhorn#2026 

Original issue has since been solved.  This updates the comment for the still-disabled test case to explain why it is so.

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -260,12 +260,12 @@ def test_setting_toleration_extra(core_api, apps_api):  # NOQA
                 },
             ],
         },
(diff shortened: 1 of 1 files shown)


PhanLe1010 (commented):
Are we planing to fix the test in this PR or we are going to update the comment only?

james-munson:
We had mentioned fixing the test in issue 2026, but then it was closed as no longer an issue, so I took that to mean that the work to modify the test case was low-priority.

innobead (commented):
Let's close this comment change, as we are working on https://github.com/longhorn/longhorn/issues/7173.

PR #1643, opened 2023-12-28
 Fix error log "Found unknown instance manager" and bug failed to find a node that is ready and has the default engine image

PhanLe1010:
1. Fix error log "Found unknown instance manager"
When we found an instance manager with the correct type but not in the
running state, don't print out the log "Found unknown instance manager"
because it is misleading
The function wait_for_instance_manager_desire_state() below it already
taking care of waiting for the instance manager to be running.
1. Fix bug failed to find a node that is ready and has the default engine image
Wait for the engine image pods to be running before starting the tests

longhorn/longhorn#7413

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -3538,7 +3538,7 @@ def wait_for_all_instance_manager_running(client):
         node_to_instance_manager_map = {}
         try:
             for im in instance_managers:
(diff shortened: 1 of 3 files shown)


yangchiu (approved):
LGTM

innobead (approved):

PR #1644, opened 2023-12-28
test: fix flaky test case test_ha_salvage

yangchiu:
test: fix flaky test case test_ha_salvage

By separate the `pgrep` command from the `kill` command. The original command format will cause the `pgrep` command also selects the `kill` process itself in addition to the `engine-binary` process, and potentially cause the `kill` process being killed by itself first, so no `engine-binary` process will be killed.

Normally it won't happen because the pid is increasing, a newly created process always has a large pid than an old process, which means the `kill` process always has a large pid than the `engine-binary` process. In this case, the `engine-binary` process will be killed first, and then the `kill` process, the test case won't fail.

But when the pid wraps around at `32768`, the newly created `kill` process gets assigned to a small pid (e.g. `300`), it will kill itself before the `engine-binary` process, and fail the test case.

For https://github.com/longhorn/longhorn/issues/7471

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2312,13 +2312,18 @@ def crash_replica_processes(client, api, volname, replicas=None,
 
     for r in replicas:
         assert r.instanceManagerName != ""
-        kill_command = "kill `pgrep -f " + r['dataPath'] + "`"
+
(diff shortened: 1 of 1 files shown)


innobead (approved):

yangchiu:

PR #1645, opened 2023-12-29
ci: enable live logging in pytest by disabling python buffer

yangchiu:
ci: enable live logging in pytest by disabling python buffer

For https://github.com/longhorn/longhorn/issues/7490

### DIFF
--- a/manager/integration/tests/run.sh
+++ b/manager/integration/tests/run.sh
@@ -1,3 +1,4 @@
 #!/bin/bash
 
+export PYTHONUNBUFFERED=1
 pytest -v "$@"


innobead (approved):

manager/integration/tests/run.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 
+export PYTHONUNBUFFERED=1
```
innobead, line 3: Not sure if there will be memory issue, but I assumed this is a streaming way.

innobead (approved):

PR #1646, opened 2024-01-02
test: add test_pvc_storage_class_name_from_backup_volume

chriscchien:
Add test case `test_pvc_storage_class_name_from_backup_volume`

ref: [7454](https://github.com/longhorn/longhorn/issues/7454)

test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5785/)

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -1849,8 +1849,10 @@ def test_volume_multinode(client, volume_name):  # NOQA
     assert len(volumes) == 0
 
 
-@pytest.mark.skip(reason="TODO")
-def test_pvc_storage_class_name_from_backup_volume(): # NOQA
+def test_pvc_storage_class_name_from_backup_volume(set_random_backupstore, # NOQA
+                                                   core_api, client, volume_name, # NOQA
+                                                   pvc_name, pvc, pod_make, # NOQA
+                                                   storage_class): # NOQA
     """
     Test the storageClasName of the restored volume's PV/PVC
     should be from the backup volume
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_basic.py
```diff
@@ -1891,17 +1892,89 @@ def test_pvc_storage_class_name_from_backup_volume(): # NOQA
...
+    wait_for_backup_completion(client, volume_name, snapshot.name)
+
+    # in nfs backupstore, bv.storageClassName sometimes were empty
+    # due to timing issue
+    for i in range(RETRY_COMMAND_COUNT):
+        bv, b = find_backup(client, volume_name, snapshot.name)
+        if bv.storageClassName != "":
+            break
+        time.sleep(RETRY_INTERVAL)
+    assert bv.storageClassName == storage_class['metadata']['name']
```
khushboo-rancher, line 1942: Just to confirm, according to the test skeleton `status.storageClassName` is asserted. Do we have two fields `StorageClassName` and `status.storageClassName`?

manager/integration/tests/test_basic.py
```diff
@@ -1891,17 +1892,89 @@ def test_pvc_storage_class_name_from_backup_volume(): # NOQA
...
+    create_pv_for_volume(client, core_api, volume, restore_name)
+    create_pvc_for_volume(client, core_api, volume, restore_name)
+
+    claim = core_api.\
+        read_namespaced_persistent_volume_claim(name=restore_name,
+                                                namespace='default')
+
+    assert claim.spec.storage_class_name == storage_class['metadata']['name']
+
+    backup_pod = pod_make(name="backup-pod")
```
khushboo-rancher, line 1965: As we are using same fixture twice, can you please check that after the test completion, all the pods are teared off?

khushboo-rancher (commented):
In general LGTM, couple of doubts

manager/integration/tests/test_basic.py
chriscchien: Checked and there are only `status.storageClassName` in `backupvolumes.longhorn.io`
```bash
> k get backupvolume pvc-d8869f46-c205-40c3-851e-85c8f37c1410 -n longhorn-system -o yaml | grep storageClassName
  storageClassName: longhorn-test
```

manager/integration/tests/test_basic.py
chriscchien: Hi @khushboo-rancher , yes the pods and other resources in this test case all teared off  after test completion (using  longhorn-test maser to examine.)

khushboo-rancher (approved):

chriscchien:
@mergify backport v1.5.x

PR #1656, opened 2024-01-04
fix(integration): test_metric_longhorn_snapshot_actual_size_bytes

c3y1huang:
https://github.com/longhorn/longhorn/issues/7532

### DIFF
--- a/manager/integration/tests/test_metric.py
+++ b/manager/integration/tests/test_metric.py
@@ -1,5 +1,6 @@
 import pytest
 import requests
+import time
 
 from collections import defaultdict
 from prometheus_client.parser import text_string_to_metric_families
@@ -24,6 +25,8 @@ from common import write_volume_random_data
 
 from common import Mi
 from common import LONGHORN_NAMESPACE
+from common import RETRY_COUNTS
+from common import RETRY_INTERVAL
 
 # The dictionaries use float type of value because the value obtained from
 # prometheus_client is in float type.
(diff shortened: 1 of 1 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5825/

Note: Running regression with the current master branch code base will fail due to https://github.com/longhorn/longhorn/issues/6563#issuecomment-1875159217. so testing with at commit 91bc9cb701ac3116d37311fb8c732ac049a560f4.

yangchiu (dismissed):
LGTM

manager/integration/tests/test_metric.py
```diff
@@ -158,6 +161,21 @@ def filter_metric_by_labels(metrics, labels):
         assert total_metrics["value"] >= 0.0
 
 
+def wait_for__metric_count_all_nodes(client, core_api, metric_name, metric_labels, expected_count): # NOQA
```
innobead, line 164: `wait_for_metric_count_all_nodes`

manager/integration/tests/test_metric.py
```diff
@@ -158,6 +161,21 @@ def filter_metric_by_labels(metrics, labels):
...
+        time.sleep(RETRY_INTERVAL)
+
+        try:
+            check_metric_count_all_nodes(client, core_api, metric_name,
+                                         metric_labels, expected_count)
+            break
+        except AssertionError:
+            continue
+
+    check_metric_count_all_nodes(client, core_api, metric_name,
```
innobead, line 175: Why need to run this check again?

innobead (changes requested):

manager/integration/tests/test_metric.py
c3y1huang: We could also `raise AssertionError("Cannot find the metric matching the labels")`, but having another run would reveal the specific assertion error, I think this way gives a bit more clarity.
```
>       assert len(filtered_metrics) == expected_count
E       assert 1 == 4
E         -1
E         +4
```

innobead (approved):

PR #1663, opened 2024-01-09
Fix Longhorn managed components restart at the begining of a test

PhanLe1010:
The test CI deploys Longhorn with the setting value longhorn-critical for the setting priority-class. Don't reset it to empty (which is the default value defined in longhorn-manager code) because this will restart Longhorn managed components and fail the test cases.

longhorn/longhorn#7413

More details are at https://github.com/longhorn/longhorn/issues/7413#issuecomment-1881707958

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -3479,6 +3479,13 @@ def reset_settings(client):
         # resetting this to an empty default value.
         if setting_name == "storage-network":
             continue
+        # The test CI deploys Longhorn with the setting value longhorn-critical
+        # for the setting priority-class. Don't reset it to empty (which is
+        # the default value defined in longhorn-manager code) because this will
+        # restart Longhorn managed components and fail the test cases.
+        # https://github.com/longhorn/longhorn/issues/7413#issuecomment-1881707958
+        if setting.name == SETTING_PRIORITY_CLASS:
+            continue
 
         # The version of the support bundle kit will be specified by a command
         # option when starting the manager. And setting requires a value.


PhanLe1010:
Test result:
* Round 1: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5886/
* Round 2: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5887/
* Round 3: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5888/

yangchiu:
Why priority class is set to `longhorn-critical` by default when installation but the default priority class is `empty` instead?

cc @mantissahz @innobead

innobead:
> Why priority class is set to `longhorn-critical` by default when installation but the default priority class is `empty` instead?
> 
> cc @mantissahz @innobead

The default priority should be `longhorn-critical` even though there is no specific setting. You should see values.yaml has that value. If using the manifest way to install, should see that the `default setting` config map has that default value. However, the code is not fully merged yet, so if you have some components already running, those pods will be restarted. The fix will be https://github.com/longhorn/longhorn-manager/pull/2411. 

@mantissahz Please add if anything is missing here.

PhanLe1010:
> Why priority class is set to longhorn-critical by default when installation but the default priority class is empty instead?

@yangchiu I know this is a bit confusing. Here is my attempt to explain it:
1. The default value of the setting `priority-class` is defined in longhorn-manager code [here](https://github.com/longhorn/longhorn-manager/blob/7a56d7103680ebb96e030a89a6e0a57abbbb5715/types/setting.go#L750-L760) which is an empty value. When Longhorn test queries this setting and looks up the default value [here](https://github.com/longhorn/longhorn-tests/blob/93e1a11721010a9c3926740e9613a7e2ffdba831/manager/integration/tests/common.py#L3474), Longhorn test will always see that the hard-coded default value of the setting definition is empty. This is like a value burned into the code to indicate what was the original value before the user ever changes it.
2. However, the user (or the test CI pipeline) did change it to a non-empty value [here](https://github.com/longhorn/longhorn/blob/b93eebc1341a4da4473b09db4f347b7e8fddc63e/deploy/longhorn.yaml#L67C5-L67C5) during the installation 

Now, I think the best approach would be to make the test pipeline respect the user setting value (the non-empty `longhorn-critical` value)  and don't reset it to empty during the test to avoid component crashes.

PhanLe1010:
> The default priority should be longhorn-critical even though there is no specific setting. You should see values.yaml has that value. If using the manifest way to install, should see that the default setting config map has that default value. However, the code is not fully merged yet, so if you have some components already running, those pods will be restarted. The fix will be https://github.com/longhorn/longhorn-manager/pull/2411.

@innobead I think the PR https://github.com/longhorn/longhorn-manager/pull/2411 alone will not fix this testing issue. That PR delay update for priority-class setting when there are attached volume. However, in this case, there is no volume at the beginning, so even after that PR is merged, the priority-class setting will still be clear and applied at the beginning -> the engine image pods are still crash -> we will still hit the error `'failed to get the parameters: failed to get target node ID: failed to find a node that is ready and has the default engine image longhornio/longhorn-engine:master-head deployed'`

So, I think the best approach would be to make the test pipeline respect the user setting value (the non-empty longhorn-critical value) and don't reset it to empty during the test to avoid component crashes.

innobead:
@PhanLe1010 your explanation is correct. We should not rely on the hard coded settings in general.

innobead:
@mantissahz please help fix any impacted cases later if needed.

innobead (approved):

PR #1669, opened 2024-01-10
Fix flaky test_volume_wait_for_backing_image_condition on ubuntu arm64

roger-ryao:
Resolves issue longhorn/longhorn#7611 by fixing a flaky test case. 
The private image test results for different platforms are as follows:

1. sles amd64 : [link](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5951/)
2. sles arm64  : [link](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5952/)
3. ubuntu arm64  : [link](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5953/)

### DIFF
--- a/manager/integration/tests/test_backing_image.py
+++ b/manager/integration/tests/test_backing_image.py
@@ -45,6 +45,7 @@ from common import BACKING_IMAGE_QCOW2_CHECKSUM
 from common import BACKING_IMAGE_STATE_READY
 from common import BACKING_IMAGE_STATE_FAILED_AND_CLEANUP
 from common import BACKING_IMAGE_STATE_IN_PROGRESS
+from common import RETRY_COUNTS_LONG
 import time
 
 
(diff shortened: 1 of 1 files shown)


yangchiu (changes requested):
I can't find the reason or context why we need to do these changes. Can you help to elaborate more?

Is it a timeout issue? Why both timeout and size modifications are needed? Only extending the timeout won't work?

roger-ryao:
> I can't find the reason or context why we need to do these changes. Can you help to elaborate more?
> 
> Is it a timeout issue? Why both timeout and size modifications are needed? Only extending the timeout won't work?

Hi @yangchiu 
     On `Ubuntu 22.04 arm64` , if we only modify the `TIMEOUT` without changing the SIZE, waiting for the backing image to complete takes around 350 to 390 seconds. However, if we reduce the SIZE by half, the time spent is approximately 200-250 seconds.
     
Screenshot_20240126_112217

yangchiu (approved):
LGTM

roger-ryao:
@mergify backport v1.6.x

yangchiu:
@roger-ryao Do `v1.5.x` and `v1.4.x` also need this fix?

roger-ryao:
> @roger-ryao Do `v1.5.x` and `v1.4.x` also need this fix?

After the merge of https://github.com/longhorn/longhorn-tests/pull/1662, we need to backport https://github.com/longhorn/longhorn-tests/pull/1669 to `v1.5.x`.

roger-ryao:
@mergify backport v1.5.x

PR #1671, opened 2024-01-12
test(manual): cordoned node controller renders the backup target unusable

c3y1huang:
longhorn/longhorn#7619


chriscchien (approved):
LGTM and here is test [result](https://github.com/longhorn/longhorn/issues/7619#issuecomment-1891234297), thank you.

PR #1674, opened 2024-01-15
fix(engineimage): new status `incompatible`.

mantissahz:
Modified the test case `test_engine_image_incompatible` by new introduced status field `incompatible`.

Ref: longhorn/longhorn#7683, longhorn/longhorn#7539


innobead (approved):

innobead:
@mergify backport v1.6.x

PR #1676, opened 2024-01-16
fix(manual): missing example to customize kubelet root dir for workers

c3y1huang:
longhorn/longhorn#7685


innobead:
@roger-ryao should review this.

roger-ryao (approved):
LGTM

PR #1677, opened 2024-01-16
test: new field `Incompatible` of EngineImage

mantissahz:

Issue longhorn/longhorn#5842, longhorn/longhorn#7539

Modify the manual test case for engine version enforcement.

### DIFF
--- a/docs/content/manual/release-specific/v1.6.0/test-engine-version-enforcement.md
+++ b/docs/content/manual/release-specific/v1.6.0/test-engine-version-enforcement.md
@@ -4,6 +4,7 @@ title: Test engine version enforcement
 
 ## Related issue
 https://github.com/longhorn/longhorn/issues/5842
+https://github.com/longhorn/longhorn/issues/7539
 
 ## Test step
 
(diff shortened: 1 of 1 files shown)


docs/content/manual/release-specific/v1.6.0/test-engine-version-enforcement.md
```diff
@@ -24,19 +25,19 @@ And should see incompatible version error in longhorn-manager Pod logs
...
+**When** downgraded Longhorn to v1.5.x
 **Then** Longhorn components should be running
 
-**When** upgraded v1.4.1 volume (volume-1) engine  
+**When** upgraded v1.4.1 volume (volume-1) engine
 And upgraded Longhorn to v1.6.0
-**Then** Longhorn components should be running  
-And v1.4.x EngineImage state should be incompatible
+**Then** Longhorn components should be running
+And v1.4.x EngineImage state should be still deployed but status.incompatible should be true.
```
c3y1huang, line 34:
```suggestion
And v1.4.x EngineImage state should be deployed and incompatible should be true.
```

c3y1huang (changes requested):
The trailing spaces are deliberately placed for line breaks.

It's looking like this without the spaces.
Screenshot from 2024-01-18 12-04-38

docs/content/manual/release-specific/v1.6.0/test-engine-version-enforcement.md
mantissahz: Modified the description and added the spaces back. Thanks.

yangchiu (approved):
LGTM

c3y1huang (commented):
LGTM, however, I noticed that the commit is signed off with a Gmail account instead of the one from SUSE. If this is intended then we could merge this. @mantissahz could you confirm?

mantissahz:
>LGTM, however, I noticed that the commit is signed off with a Gmail account instead of the one from SUSE. If this is intended then we could merge this. @mantissahz could you confirm?

Corrected.
Thank you, @c3y1huang, and it should be the one from SUSE.

c3y1huang (approved):

PR #1681, opened 2024-01-23
fix(test): fix recurring_job_restored_from_backup_test

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7744

1. If we run the tests 100 times, the previous test could still be in the process of being cleaned up while the next one starts running
So we should use random volume name to prevent backup lock issue because of the same name.

2. We should get the volume after Backup is ready, or the Volume.BackupStatus will not be correct

### DIFF
--- a/manager/integration/tests/test_recurring_job.py
+++ b/manager/integration/tests/test_recurring_job.py
@@ -17,6 +17,7 @@ from common import pvc  # NOQA
 from common import random_labels, volume_name  # NOQA
 from common import storage_class, statefulset, pvc  # NOQA
 from common import make_deployment_with_pvc  # NOQA
+from common import generate_volume_name
 
 from common import get_self_host_id
 
@@ -1991,9 +1992,9 @@ def test_recurring_job_restored_from_backup_target(set_random_backupstore, clien
     back1 = BACKUP + "1"
     back2 = BACKUP + "2"
     group1 = "group01"
-    volume_name1 = "record-recurring-job"
-    rvolume_name1 = "restore-record-recurring-job-01"
-    rvolume_name2 = "restore-record-recurring-job-02"
+    volume_name1 = "record-recur" + "-" + generate_volume_name()
+    rvolume_name1 = "restore-01" + "-" + generate_volume_name()
+    rvolume_name2 = "restore-02" + "-" + generate_volume_name()
 
     recurring_jobs = {
         back1: {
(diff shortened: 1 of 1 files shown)


yangchiu:
The cleanup function fixture `set_random_backupstore` has a wait mechanism to wait until the backup volume is gone:
```
# in backupstore.py
@pytest.fixture(params=BACKUPSTORE)
def set_random_backupstore(request, client):
    if request.param == "s3":
        set_backupstore_s3(client)
    elif request.param == "nfs":
        set_backupstore_nfs(client)
        mount_nfs_backupstore(client)

    yield
    cleanup_all_volumes(client)
    backupstore_cleanup(client)
    system_backups_cleanup(client)
    reset_backupstore_setting(client)

    if request.param == "nfs":
        umount_nfs_backupstore(client)
```
```
# in backupstore.py
def backupstore_cleanup(client):
    backup_volumes = client.list_backup_volume()

    # we delete the whole backup volume, which skips block gc
    for backup_volume in backup_volumes:
        delete_backup_volume(client, backup_volume.name)

    backup_volumes = client.list_backup_volume()
    assert backup_volumes.data == []
```

```
# in common.py
def delete_backup_volume(client, volume_name):
    bv = client.by_id_backupVolume(volume_name)
    client.delete(bv)
    wait_for_backup_volume_delete(client, volume_name)
```

Do you have an idea about why it still pollutes the subsequent test case?

ChanYiLin:
Hi @yangchiu from you reply here
https://github.com/longhorn/longhorn/issues/7744#issuecomment-1905501398
After the cleanup, there is still a chance that there are BackupVolumes in the cluster not being deleted, correct?

yangchiu:
> Hi @yangchiu from you reply here [longhorn/longhorn#7744 (comment)](https://github.com/longhorn/longhorn/issues/7744#issuecomment-1905501398) After the cleanup, there is still a chance that there are BackupVolumes in the cluster not being deleted, correct?

It doesn't make sense since it iterated the backup volume list and deleted them one by one, but according to the failed assertion, yes, there are something still in the list..

yangchiu (approved):
LGTM

ChanYiLin:
The correct fix is https://github.com/longhorn/backupstore/pull/158/files
But we still need this PR because of the second reason
"We should get the volume after Backup is ready, or the Volume.BackupStatus will not be correct"

ChanYiLin:
cc @innobead

innobead (approved):

innobead:
@yangchiu Please manage the backport if needed.

cc @longhorn/qa

yangchiu:
@mergify backport v1.6.x v1.5.x v1.4.x

PR #1682, opened 2024-01-23
feat(rwx):  add a test to check mount options override.

james-munson:

Issue  https://github.com/longhorn/longhorn/issues/7639

Add a test to rwx regressions to use a storage class with customized NFS mount options.


yangchiu (approved):
LGTM

PR #1689, opened 2024-01-24
test(recurring-job/fstrim): test RWO and RWX volume

c3y1huang:

Issue:
- longhorn/longhorn#7768
- longhorn/longhorn#7771

Extend `test_recurring_job_filesystem_trim` with pytest.mark.parametrize fixture to test RWO and RWX volumes.

None

None

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -287,6 +287,9 @@ disk_being_syncing = "being syncing and please retry later"
 FS_TYPE_EXT4 = "ext4"
 FS_TYPE_XFS = "xfs"
 
+ACCESS_MODE_RWO = "rwo"
+ACCESS_MODE_RWX = "rwx"
+
 ATTACHER_TYPE_CSI_ATTACHER = "csi-attacher"
 ATTACHER_TYPE_LONGHORN_API = "longhorn-api"
 ATTACHER_TYPE_LONGHORN_UPGRADER = "longhorn-upgrader"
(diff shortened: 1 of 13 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6077/

manager/integration/tests/common.py
```diff
@@ -490,7 +493,7 @@ def delete_backup_volume(client, volume_name):
     wait_for_backup_volume_delete(client, volume_name)
 
 
-def create_and_check_volume(client, volume_name,
+def create_and_check_volume(client, volume_name, access_mode=ACCESS_MODE_RWO,
                             num_of_replicas=3, size=SIZE, backing_image="",
```
james-munson, line 497: I'm confused.  Wouldn't this require an added parameter in all those places that call with positional arguments (or maybe a change to use named arguments)?  Such as this one:  https://github.com/longhorn/longhorn-tests/blob/078c17d8ce66d649ce0d5ed68307cfa28549ef17/manager/integration/tests/test_basic.py#L1000

c3y1huang: Good catch, you are correct, this indeed messes up the arguments when optional arguments are invoked as positional arguments.

derekbit (approved):
LGTM

yangchiu:
There are many test cases modified, but only test case `test_recurring_job_filesystem_trim` is run. Maybe another test is required to make sure affected test cases still work.

james-munson (approved):
Looks good.

c3y1huang:
> There are many test cases modified, but only test case `test_recurring_job_filesystem_trim` is run. Maybe another test is required to make sure affected test cases still work.

The refactor involves assigning names to the optional argument when they are called. There isn't any behavioral change.

yangchiu:
> 
> Could you share the reason you think this change would make the related cases not work as expected?

Like this one https://github.com/longhorn/longhorn-tests/pull/1689#discussion_r1465338260?

c3y1huang:
> Like this one [#1689 (comment)](https://github.com/longhorn/longhorn-tests/pull/1689#discussion_r1465338260)?

It has been addressed. Any other concerns?

yangchiu (approved):
LGTM

Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6157/

roger-ryao:

PR #1697, opened 2024-01-29
Fix flaky test_volume_metrics

chriscchien:
ref: [7626](https://github.com/longhorn/longhorn/issues/7626)

Issue #7626

```
 crash_replica_processes(client, core_api, volume_name)
>       volume = wait_for_volume_faulted(client, volume_name)

test_metric.py:291: 
```
Volume faulted when there are no replica running, I tried 80 times looks  use `delete_replica_processes` replace `crash_replica_processes` to combine with `wait_for_volume_faulted` is more stable.

80 times test [result](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6168/) - all passed

For the more detail about why `wait_for_volume_faulted` is not stable when combine with `crash_replica_processes` I will do more observation later.


khushboo-rancher (approved):
I'm ok with the change as the objective of the test is to check the metrics when volume's status is faulted not how we made the volume faulted.

PR #1787, opened 2024-02-28
Fix HA tests for scheduling change

ejweber:

longhorn/longhorn#8043

https://github.com/longhorn/longhorn-manager/pull/2650 disallows scheduling a replica to a node, zone, or disk if it already contains a failed replica (unless the appropriate soft anti-affinity setting is enabled). This behavior requires minor changes to a couple of tests.

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -80,6 +80,7 @@ from common import wait_for_backup_volume_backing_image_synced
 from common import RETRY_COMMAND_COUNT
 from common import wait_for_snapshot_count
 from common import DEFAULT_BACKUP_COMPRESSION_METHOD
+from common import SETTING_REPLICA_DISK_SOFT_ANTI_AFFINITY
 
 from backupstore import set_random_backupstore # NOQA
 from backupstore import backupstore_cleanup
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_ha.py
```diff
@@ -2455,22 +2462,25 @@ def test_replica_failure_during_attaching(settings_reset, client, core_api, volu
...
-    9. Attach volume2.
+    10. Attach volume2.
        --> Verify volume will be attached with state Degraded.
-    10. Wait for the replenishment interval.
+    11. Wait for the replenishment interval.
         --> Verify a new replica will be created but it cannot be scheduled.
-    11. Enable the default disk for the host node.
-    12. Wait for volume2 becoming Healthy.
-    13. Verify data content and r/w capability for volume2.
+    12. Enable the default disk for the host node.
```
ejweber, line 2481: At this point, the purpose of the test is complete. We successfully recognized a failed replica during attaching and (after `replica-replenishment-wait-interval`) scheduled a new replica to replace it.

Previously, the test relied on `replica-node-soft-anti-affinity == false` to ensure the new replica could not schedule anywhere until the default disk on the test node was enabled. Once the default disk was enabled, the failed replica did not prevent scheduling to the node.

Now, the failed replica does prevent scheduling to the node. However, we can rely on `replica-disk-soft-anti-affinity == false` to continue to ensure it does not get scheduled until we enabled the default disk.

ejweber:
With the changes to https://github.com/longhorn/longhorn-manager/pull/2650, we likely do not need this PR. I will keep it open as a draft until that one merges.

PR #1805, opened 2024-03-04
ci(negative): add option to running negative testing in container

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8047

add option to running negative testing in container


khushboo-rancher (approved):

pipelines/e2e/scripts/longhorn-setup.sh
```diff
@@ -53,11 +53,7 @@ main(){
   setup_longhorn_ui_nodeport
   export_longhorn_ui_url
 
-  if [[ -n "${LONGHORN_TESTS_CUSTOM_IMAGE}" ]]; then
-    run_longhorn_e2e_test
-  else
-    run_longhorn_e2e_test_out_of_cluster
-  fi
+  run_longhorn_e2e_test_out_of_cluster
```
khushboo-rancher, line 56: nit: Do we want to change the name more specific, something like run_longhorn_e2e_test_dockerized ?

yangchiu: I'll refine this in the following PR. Thank you!

PR #1808, opened 2024-03-05
ci: Update livenessProbe.repository value in Rancher chart

roger-ryao:
longhorn/longhorn#8095

When the `LONGHORN_REPO` is set to `longhornio`, we need to update the `livenessProbe.repository` value to `longhornio/livenessprobe` in the Rancher chart.

https://github.com/longhorn/longhorn/issues/8095#issuecomment-1977976920


yangchiu:
@roger-ryao Do you think the file in `terraform_upgrade` should be updated as well?

roger-ryao:
> @roger-ryao Do you think the file in `terraform_upgrade` should be updated as well?

Update.

yangchiu (approved):
LGTM

PR #1814, opened 2024-03-08
Revert "ci: use more cost-effective ec2 instance type"

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/7957

Revert "ci: use more cost-effective ec2 instance type"


chriscchien (approved):
LGTM

PR #1818, opened 2024-03-13
fix: check for timeout in wait_deployment_replica_ready and fail if so.

james-munson:

Issue https://github.com/longhorn/longhorn/issues/8183

Adds a check for timeout when waiting for deployment to create successfully.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -4783,6 +4783,7 @@ def make_deployment_cpu_request(request):
 
 def wait_deployment_replica_ready(apps_api, deployment_name,
                                   desired_replica_count, namespace='default'):  # NOQA
+    ok = False
     for i in range(DEFAULT_DEPLOYMENT_TIMEOUT):
         deployment = apps_api.read_namespaced_deployment(
             name=deployment_name,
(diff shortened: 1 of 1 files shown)


james-munson:
Still testing, so created PR as draft.

james-munson:
RWX tests failed as hoped.  The "broken" longhorn-share-manager image I am using is `jmunson4suse/longhorn-share-manager:8183-test1`
Removing the "draft" status.  This is ready to go.

manager/integration/tests/common.py
```diff
@@ -4790,9 +4791,11 @@ def wait_deployment_replica_ready(apps_api, deployment_name,
         # deployment is none if deployment is not yet created
         if deployment is not None and \
            deployment.status.ready_replicas == desired_replica_count:
+            ok = True
             break
 
         time.sleep(DEFAULT_DEPLOYMENT_INTERVAL)
+    assert ok
```
c3y1huang, line 4798: Nit: When the `assert ok` statement fails, it doesn't output what was `not ok`, suggest adding a brief message behind `assert ok, f"something"`. Also, you can use the else statement in conjunction with a for loop to execute a block of code if the loop completes normally, so we don't need the extra `ok` variable.
```
    for i in range(DEFAULT_DEPLOYMENT_TIMEOUT):
        # do something and break
    else:
        raise TimeoutError(f"Deployment '{deployment_name}' not ready\n"
                           f"Expected replica count: {desired_replica_count}\n"
                           f"Got replica count: {deployment.status.ready_replicas}")
```

james-munson: Will do.  I was just copying the idiom that is used elsewhere in the tests.  This is clearer on failure.

c3y1huang (approved):

PR #1820, opened 2024-03-18
Add test_provisioner_generic_ephemeral skeleton

ejweber:

longhorn/longhorn#8198

longhorn/longhorn#8198 uncovered the fact that a change had unknowingly broken Longhorn generic ephemeral volumes. We need an automated test to ensure this can't happen in the future.

I wasn't really sure if this skeleton fit best in `test_provisioner.py` (since we want to test the ability to provision generic ephemeral volumes) or `test_kubernetes.py`, since the actual change I made in https://github.com/longhorn/longhorn-manager/pull/2701 ensures `volume.status.kubernetesStatus.workloadStatus` was correctly populated by the Longhorn Kuberntes PV controller. I went with my gut, but I'm fine with moving it if QA thinks it makes better sense elsewhere.

### DIFF
--- a/manager/integration/tests/test_provisioner.py
+++ b/manager/integration/tests/test_provisioner.py
@@ -63,6 +63,24 @@ def test_provisioner_mount(client, core_api, storage_class, pvc, pod):  # NOQA
     assert volumes.data[0].state == "attached"
 
 
+@pytest.mark.skip(reason="TODO")
+def test_provisioner_generic_ephemeral():
+    """
+    Test that a Longhorn generic ephemeral volume can be created, mounted,
+    unmounted, and deleted properly on the Kubernetes cluster.
+
+    1. Create a StorageClass and Pod with a generic ephemeral volume spec:
+       https://kubernetes.io/docs/concepts/storage/ephemeral-volumes
+    2. Verify:
+       - The Pod is running.
+       - The volume parameters match the StorageClass parameters.
+       - The volume.status.kubernetesStatus.workloadStatus reflects the
+         running Pod.
+    3. Write data to the volume using the Pod, read it back, and verify it.
+    """
+    pass
+
+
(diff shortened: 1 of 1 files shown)


chriscchien (approved):
LGTM

ejweber:
Fixed the linting issues.

james-munson (dismissed):
Looks fine as a placeholder.  Is there an associated issue to fill it in?

ejweber:
> Looks fine as a placeholder. Is there an associated issue to fill it in?

It is linked in the issue: https://github.com/longhorn/longhorn/issues/8203.

manager/integration/tests/test_provisioner.py
```diff
@@ -63,6 +63,23 @@ def test_provisioner_mount(client, core_api, storage_class, pvc, pod):  # NOQA
...
+    unmounted, and deleted properly on the Kubernetes cluster.
+
+    1. Create a StorageClass and Pod with a generic ephemeral volume spec:
+       https://kubernetes.io/docs/concepts/storage/ephemeral-volumes
+    2. Verify:
+       - The Pod is running.
+       - The volume parameters match the StorageClass parameters.
+       - The volume.status.kubernetesStatus.workloadStatus reflects the
+         running Pod.
+    """
```
khushboo-rancher, line 79: Should we add a test step attempting to write data in ephemeral volume?

ejweber: It definitely can't hurt, so I added it in an amend. For context, I wrote this skeleton to mirror `test_provisioner_mount` (directly above in the file), which does not. Importantly, we cannot follow the pattern of some other tests and read the data back from a different pod, because the life cycle of a generic ephemeral volume mirrors the life cycle of the pod it is created in.

PR #1821, opened 2024-03-19
fix(negative): wait_for_workload_pods_stable

c3y1huang:

Issue longhorn/longhorn#8193

Check whether the retry count exceeds the `WAIT_FOR_POD_STABLE_MAX_RETRY` instead of verifying the exact match to avoid the loop.

`None`

https://github.com/longhorn/longhorn/issues/8193#issuecomment-2006061894


yangchiu (approved):
LGTM

innobead (approved):

PR #1822, opened 2024-03-20
fix(negative): write_random_data

c3y1huang:

Issue longhorn/longhorn#8208

Add retries for the `dd` command to capture failures before invoking the `md5sum` calculation.  This also allows delay for failure investigation with the `RETRY_COUNT` variable.

### DIFF
--- a/e2e/libs/volume/crd.py
+++ b/e2e/libs/volume/crd.py
@@ -263,10 +263,25 @@ class CRD(Base):
     def write_random_data(self, volume_name, size):
         node_name = self.get(volume_name)["spec"]["nodeID"]
         endpoint = self.get_endpoint(volume_name)
-        checksum = self.node_exec.issue_cmd(
-            node_name,
-            f"dd if=/dev/urandom of={endpoint} bs=1M count={size} status=none; md5sum {endpoint} | awk \'{{print $1}}\'")
-        return checksum
+
+        dd_command = f"dd if=/dev/urandom of={endpoint} bs=1M count={size} status=none"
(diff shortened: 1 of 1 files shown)


yangchiu:
If we keeps retrying the `dd` command until it succeeds, does it mean we could any ignore potential issue? For example, if the `dd` command tried 100 times until it succeeded, the test case is still passed, but the previous 99 failed `dd` commands are still problematic, right? Since the test case is eventually marked passed, nobody will check the previous 99 failed `dd` commands and they will be ignored.

Instead of retrying, could we just sleep and halt the test if the `dd` command failed or the checksum doesn't match?

On the other hand, in the original issue https://github.com/longhorn/longhorn/issues/8208, is it possible that the pod was not stable (restarted) yet when you did the IO operation? The `Input/output error` looks like the symptom of this scenario.

c3y1huang:
> If we keeps retrying the `dd` command until it succeeds, does it mean we could any ignore potential issue? For example, if the `dd` command tried 100 times until it succeeded, the test case is still passed, but the previous 99 failed `dd` commands are still problematic, right? Since the test case is eventually marked passed, nobody will check the previous 99 failed `dd` commands and they will be ignored.
> 
> Instead of retrying, could we just sleep and halt the test if the `dd` command failed or the checksum doesn't match?
> 
> On the other hand, in the original issue [longhorn/longhorn#8208](https://github.com/longhorn/longhorn/issues/8208), is it possible that the pod was not stable (restarted) yet when you did the IO operation? The `Input/output error` looks like the symptom of this scenario.

As discussed in the team meeting, we will pause the test when the dd command fails.

yangchiu (approved):

PR #654, opened 2021-07-13
Backup refactoring

jenting:
#### Issue

https://github.com/longhorn/longhorn/issues/1761

PR #1837, opened 2024-03-28
fix(backup lock): fix test case to follow the test purpose

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8269

Refactor the code to test the true purpose of this test

When creating Backup during previous Backup deletion, the Backup should be failed because of the lock

So adjust the code to
1. Create the first Snapshot1 and Backup1
2. Create the second Snapshot2
3. Delete the first Backup1
4. Create the second Backup2
5. Backup2 should be failed
6. Wait for Backup1 to be deleted
7. Create third Snapshot3 and Backup3
8. Backup3 should be completed.

### DIFF
--- a/manager/integration/tests/test_basic.py
+++ b/manager/integration/tests/test_basic.py
@@ -36,7 +36,7 @@ from common import generate_random_data, write_volume_data
 from common import VOLUME_RWTEST_SIZE
 from common import write_pod_volume_data
 from common import find_backup, find_replica_for_backup
-from common import wait_for_backup_completion
+from common import wait_for_backup_completion, wait_for_backup_failed
 from common import create_storage_class
 from common import wait_for_backup_restore_completed
 from common import wait_for_volume_restoration_completed
@@ -3337,14 +3337,15 @@ def test_backup_lock_creation_during_deletion(set_random_backupstore, client, co
     steps:
     1. Create a volume, then create the corresponding PV, PVC and Pod.
     2. Wait for the pod running and the volume healthy.
-    3. Write data (DATA_SIZE_IN_MB_2) to the pod volume and get the md5sum.
+    3. Write data (DATA_SIZE_IN_MB_4) to the pod volume and get the md5sum.
     4. Take a backup.
     5. Wait for the backup to be completed.
     6. Delete the backup.
     7. Create another backup of the same volume.
-    8. Wait for the delete backup to be completed.
-    8. Wait for the backup to be completed.
-    9. Assert there is 1 backup in the backup store.
+    8. The newly created backup should failed because there is a deletion lock.
+    9. Wait for the first backup to be Deleted
+    10. Create another backup of the same volume.
+    11. Wait for the backup to be completed.
     """
     backupstore_cleanup(client)
     std_volume_name = volume_name + "-std"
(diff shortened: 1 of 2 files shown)


ChanYiLin:
## Local test: PASSED

ChanYiLin:
Run test x 20 times: 
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6690/console

ChanYiLin:
## Run test x 20 times - PASSED

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6690/console

manager/integration/tests/test_basic.py
```diff
@@ -3337,46 +3337,51 @@ def test_backup_lock_creation_during_deletion(set_random_backupstore, client, co
...
+
+    # create first snapshot and backup
     snap1 = create_snapshot(client, std_volume_name)
     std_volume.snapshotBackup(name=snap1.name)
     wait_for_backup_completion(client, std_volume_name, snap1.name)
     _, b1 = common.find_backup(client, std_volume_name, snap1.name)
 
-    write_pod_volume_random_data(core_api, std_pod_name,
-                                 "/data/test2", DATA_SIZE_IN_MB_1)
+    # create second snapshot
```
ejweber, line 3365: It looks like both the original test and the modified version from https://github.com/longhorn/longhorn-tests/pull/654/commits/79d3f75b3e02c8d78890851cbace731b48260f9d wrote data before creating the second snapshot. I guess we are not doing that here to ensure the backup starts quickly (while the delete lock is active)?

Should write data before creating the third snapshot? Or does it not matter?

innobead (approved):

innobead:

ejweber (commented):
The explanation provided at https://github.com/longhorn/longhorn/issues/8269#issuecomment-2025020671 for why the test logic should change is compelling, and these changes look like they achieve the goal.

There is one thing I am not understanding. If the changes to the test in https://github.com/longhorn/longhorn-tests/pull/654/commits/79d3f75b3e02c8d78890851cbace731b48260f9d caused the issue, why did we not see regular test failures until now?

khushboo-rancher:
>There is one thing I am not understanding. If the changes to the test in https://github.com/longhorn/longhorn-tests/commit/79d3f75b3e02c8d78890851cbace731b48260f9d caused the issue, why did we not see regular test failures until now?

I suspect that the data in the pod was 1 MB, which was getting deleted quickly and the 2nd backup trigger didn't hit the backup lock. That's why the test was not failing all the time.

manager/integration/tests/test_basic.py
ChanYiLin: Yes, as you said, we are testing the lock to ensure that the backup creation will fail during backup deletion
so the content of the snapshot/backup doesn't matter, we don't even check the content in this test.

ChanYiLin:
> The explanation provided at [longhorn/longhorn#8269 (comment)](https://github.com/longhorn/longhorn/issues/8269#issuecomment-2025020671) for why the test logic should change is compelling, and these changes look like they achieve the goal.
> 
> There is one thing I am not understanding. If the changes to the test in [79d3f75](https://github.com/longhorn/longhorn-tests/commit/79d3f75b3e02c8d78890851cbace731b48260f9d) caused the issue, why did we not see regular test failures until now?

Hi @ejweber actually, it happened in our master head and 1.5.x as well (information provided by QA).
It is a flaky issue. If the deletion completes too fast, then the creation will success
In addition, the `create_snapshot()` in our test code is a blocking call. It request the API to create the snapshot and wait until the snapshot is ready.
So that's why the test has higher success rate

PR #1841, opened 2024-03-29
Verify message of invalid backup target

ejweber:

This test enhancement should allow us to catch longhorn/longhorn#8210 in a future regression.

We don't currently have test coverage that can detect the issue.

The test fails in `v1.6.0` as expected.

```
        # If the condition/message was not updated, it is likely still this one.
>       assert backup_targets[0]["message"] != BACKUP_TARGET_MESSAGE_EMPTY_URL
E       AssertionError: assert 'backup target URL is empty' != 'backup target URL is empty'

test_basic.py:5711: AssertionError
```

It passes on `master-head`.


khushboo-rancher (approved):

khushboo-rancher:

PR #1852, opened 2024-04-10
test(negative): add fs stress operations

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/6379

add fs stress operations


e2e/tests/stress_filesystem.robot
```diff
@@ -0,0 +1,73 @@
...
+    And Stress filesystem of all worker nodes
+
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        And Scale down statefulset 0 to detach volume
+
+        When Expand statefulset 0 volume by 100 MiB
+        Then Wait for statefulset 0 volume size expanded
+
+        And Scale up statefulset 0 to attach volume
+        And Check statefulset 0 data in file data.txt is intact
```
chriscchien, line 72: Could we also add a check for volume health in `Stress Volume Node FS When Volume Is Offline Expanding` and `Stress Volume Node Filesystem When Volume Is Detaching and Attaching` ? My thougt was that read data through pod or endpoint success not guaranteen all replicas are in running state. Thank you.

yangchiu: Added. Thank you!

chriscchien (approved):
LGTM

PR #1855, opened 2024-04-11
test(robot): add test case tag and description

yangchiu:

add test case tag and description


chriscchien (approved):
LGTM

PR #1863, opened 2024-04-18
ci: setup test environment for v2 engine

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/7202

setup test environment for v2 engine


khushboo-rancher:
@yangchiu LGTM,  As discussed, could you please add the parameter `run v2 test cases` (Extra block device) configurable, so that we if don't want the robot test to run v2 test, it won't create the extra volumes on the cloud.
Robot framework - By default `run v2 test cases` will be yes
Pytest framework - By default `run v2 test cases` will be No (Non editable)as it is now.

yangchiu:
> @yangchiu LGTM, As discussed, could you please add the parameter `run v2 test cases` (Extra block device) configurable, so that we if don't want the robot test to run v2 test, it won't create the extra volumes on the cloud. Robot framework - By default `run v2 test cases` will be yes Pytest framework - By default `run v2 test cases` will be No (Non editable)as it is now.

Created https://github.com/longhorn/longhorn-tests/pull/1881 and https://github.com/longhorn/infra/pull/169 for it. Thank you!

khushboo-rancher (approved):

PR #1864, opened 2024-04-18
feat(system-backup): add system backup backingimage test

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/5085

Create a Volume with the BackingImage
Create the system backup
Delete the Volume and the BackingImage
Restore the system backup
Both Volume and BackingImage will be restored.


manager/integration/tests/test_system_backup_restore.py
```diff
@@ -112,6 +116,66 @@ def test_system_backup_and_restore_volume_with_data(client, volume_name, set_ran
...
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    assert backing_image.sourceType == BACKING_IMAGE_SOURCE_TYPE_RESTORE
+
+    restored_volume = client.by_id_volume(volume_name)
+    wait_for_volume_restoration_completed(client, volume_name)
+    wait_for_volume_detached(client, volume_name)
+
+    restored_volume.attach(hostId=host_id)
+    restored_volume = wait_for_volume_healthy(client, volume_name)
+
```
c3y1huang, line 178: Should also check the data?

c3y1huang (changes requested):

manager/integration/tests/test_system_backup_restore.py
ChanYiLin: Thanks for the advice
I think no, we can focus on the volume backup/restore with BackingImage
And for the data consistency
1. Volume already has its own test above
2. BackingImage use checksum to verified the data during backup/restore. If it is not consistency, BackingImage will be failed and so is the test.

So I think we can remove the data test in this test case to make it simple.

manager/integration/tests/test_system_backup_restore.py
```diff
@@ -112,6 +117,66 @@ def test_system_backup_and_restore_volume_with_data(client, volume_name, set_ran
...
+    Given a backingimage
+    And a volume created with the backingimage
+    When system backup created
+    Then system backup in state Ready
+
+    When volume deleted
+    And backingimage deleted
+    And restore system backup
+
+    When restore system backup
```
c3y1huang, line 136: `When` should be followed by `And` or `Then`.
```suggestion
    And restore system backup
```

manager/integration/tests/test_system_backup_restore.py
c3y1huang: I see. Can you add a comment to ensure @longhorn/qa is informed that the data integrity testing is covered somewhere else by looking at the case?

c3y1huang (changes requested):

manager/integration/tests/test_system_backup_restore.py
ChanYiLin: Sure!
Added.

manager/integration/tests/test_system_backup_restore.py
ChanYiLin: Fixed

manager/integration/tests/test_system_backup_restore.py
```diff
@@ -112,6 +117,72 @@ def test_system_backup_and_restore_volume_with_data(client, volume_name, set_ran
...
+    Given a backingimage
+    And a volume created with the backingimage
+    When system backup created
+    Then system backup in state Ready
+
+    When volume deleted
+    And backingimage deleted
+    And restore system backup
+
+    And restore system backup
```
c3y1huang, line 142: Sorry, bad eye, I just noticed these are duplicated steps.

c3y1huang (changes requested):

manager/integration/tests/test_system_backup_restore.py
ChanYiLin: me too, fixed.

c3y1huang (approved):

yangchiu (approved):
LGTM

chriscchien:
@mergify backport v1.6.x

PR #1868, opened 2024-04-24
ci: fix go tool cover error: go.mod requires go >= 1.22.2

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/7166

fix go tool cover error: go.mod requires go >= 1.22.2

### DIFF
--- a/test_framework/Dockerfile.setup
+++ b/test_framework/Dockerfile.setup
@@ -1,4 +1,4 @@
-FROM golang:1.21-alpine3.19
+FROM golang:1.22-alpine3.19
 
 ARG KUBECTL_VERSION=v1.20.2
 


test_framework/Dockerfile.setup
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.21-alpine3.19
+FROM golang:1.22.2-alpine3.19
```
FrankYang0529, line 1: May we use [1.22-alpine3.19](https://hub.docker.com/layers/library/golang/1.22-alpine3.19/images/sha256-963da5f97ab931c0df6906e8c0ebc7db28c88d013735ae020f9558c3e6cf0580?context=explore)? So we can always get latest patch version. Thanks.

FrankYang0529 (approved):
LGTM. Thanks.

c3y1huang (approved):

PR #1869, opened 2024-04-25
test(robot): add rwx workload restart kubelet test

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8405

add rwx workload restart kubelet test


chriscchien (approved):
LGTM, went through the code and did not have any concern.

PR #1874, opened 2024-04-29
ci: clean up instances without an owner

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8412

Clean up instances without an owner if they have been running for more than 1 day.
Send a slack notification about instances with an owner if they have been running for more than 3 days.
Clean up instances with an owner if they have been running for more than 1 week.


cleanup/scripts/cleanup.sh
```diff
@@ -140,4 +148,70 @@ for SUFFIX in ${SUFFIX_ARR[@]}; do
...
+done
+for INSTANCE_ID in ${INSTANCE_IDS}
+do
+  aws ec2 terminate-instances --instance-ids "${INSTANCE_ID}"
+  echo "   instance ${INSTANCE_ID} shutting-down"
+done
+while [[ -n $(aws ec2 describe-instances --instance-ids "${INSTANCE_ID}" | jq '.Reservations[].Instances[].State.Name' | grep -v "terminated") ]]; do
+  echo "Wait for instances terminated ..."
+  sleep 5s
+done
```
chriscchien, line 173: Just curious if the while loop only wait the last `${INSTANCE_ID}` in above for loop to terminated? (and the same question at the similar part of next section) thank you.

yangchiu: Refined. Thank you!

chriscchien (approved):
LGTM

PR #1875, opened 2024-04-29
test(robot): remove unused test pod functions

yangchiu:

remove unused test pod functions


khushboo-rancher:
@yangchiu Could you please explain in brief why we had test pod function before and why are deleting it now?

yangchiu:
@khushboo-rancher Test pod running node is the node where the test is running. Unlike the original pytest, we don't run the robot test in a pod anymore. We run it directly from command line or in a docker container instead, so we don't have the restrictions that something must be run on the test pod running node or something must not be run on the test pod running node. That makes us be able to manipulate any arbitrary node. That's why we are able to run cluster restart test case to reboot ALL nodes.

khushboo-rancher (approved):

PR #1877, opened 2024-04-29
feat(backingimage): add backing image ha, eviction and selector testing

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/2856

Add e2e tests for backing image enhancement
- HA
- selector
- eviction

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -221,6 +221,8 @@ SETTING_ALLOW_EMPTY_NODE_SELECTOR_VOLUME = \
 SETTING_REPLICA_DISK_SOFT_ANTI_AFFINITY = "replica-disk-soft-anti-affinity"
 SETTING_ALLOW_EMPTY_DISK_SELECTOR_VOLUME = "allow-empty-disk-selector-volume"
 SETTING_NODE_DRAIN_POLICY = "node-drain-policy"
+SETTING_MIN_NUMBER_OF_BACKING_IMAGE_COPIES = \
+    "default-min-number-of-backing-image-copies"
 
 DEFAULT_BACKUP_COMPRESSION_METHOD = "lz4"
 BACKUP_COMPRESSION_METHOD_LZ4 = "lz4"
@@ -313,6 +315,10 @@ BACKUP_TARGET_MESSAGE_EMPTY_URL = "backup target URL is empty"
 BACKUP_TARGET_MESSAGES_INVALID = ["failed to init backup target clients",
                                   "failed to list backup volumes in",
                                   "error listing backup volume names"]
+
+FAILED_DELETING_REASONE = "FailedDeleting"
+BACKINGIMAGE_FAILED_EVICT_MSG = \
+    "since there is no other healthy backing image copy"
 
 # customize the timeout for HDD
 disktype = os.environ.get('LONGHORN_DISK_TYPE')
@@ -5315,7 +5321,9 @@ def assert_backup_state(b_actual, b_expected):
     assert b_expected.messages == b_actual.messages is None
 
 
-def create_backing_image_with_matching_url(client, name, url):
+def create_backing_image_with_matching_url(client, name, url,
+                                           minNumberOfCopies=1,
+                                           nodeSelector=[], diskSelector=[]):
     backing_images = client.list_backing_image()
     found = False
     for bi in backing_images:
@@ -5340,13 +5348,16 @@ def create_backing_image_with_matching_url(client, name, url):
             expected_checksum = BACKING_IMAGE_QCOW2_CHECKSUM
         bi = client.create_backing_image(
             name=name, sourceType=BACKING_IMAGE_SOURCE_TYPE_DOWNLOAD,
-            parameters={"url": url}, expectedChecksum=expected_checksum)
+            parameters={"url": url}, expectedChecksum=expected_checksum,
+            minNumberOfCopies=minNumberOfCopies,
+            nodeSelector=nodeSelector, diskSelector=diskSelector)
     assert bi
 
     is_ready = False
     for i in range(RETRY_COUNTS):
         bi = client.by_id_backing_image(name)
-        if len(bi.diskFileStatusMap) == 1 and bi.currentChecksum != "":
+        if (len(bi.diskFileStatusMap) == minNumberOfCopies and
+                bi.currentChecksum != ""):
             for disk, status in iter(bi.diskFileStatusMap.items()):
                 if status.state == "ready":
                     is_ready = True
@@ -6322,3 +6333,104 @@ def wait_for_volume_replica_rebuilt_on_same_node_different_disk(client, node_nam
 
     assert new_disk_name != old_disk_name, \
         "Failed to rebuild replica disk to another disk"
+
+
+def set_tags_for_node_and_its_disks(client, node, node_tags, disks_tags_map): # NOQA
+    for disk_name in node.disks.keys():
+        if disk_name in disks_tags_map:
+            node.disks[disk_name].tags = disks_tags_map[disk_name]
+
+    node = update_node_disks(client, node.name, disks=node.disks)
+    for disk_name in node.disks.keys():
+        if disk_name in disks_tags_map:
+            assert node.disks[disk_name].tags == disks_tags_map[disk_name]
+
+    if len(node_tags) == 0:
+        expected_tags = []
+    else:
+        expected_tags = list(node_tags)
+
+    node = set_node_tags(client, node, node_tags)
+    assert node.tags == expected_tags
+
+
+def set_tags_for_node_and_its_disks(client, node, tags): # NOQA
+    if len(tags) == 0:
+        expected_tags = []
+    else:
+        expected_tags = list(tags)
+
+    for disk_name in node.disks.keys():
+        node.disks[disk_name].tags = tags
+    node = update_node_disks(client, node.name, disks=node.disks)
+    for disk_name in node.disks.keys():
+        assert node.disks[disk_name].tags == expected_tags
+
+    node = set_node_tags(client, node, tags)
+    assert node.tags == expected_tags
+
+    return node
+
+
+def get_node_by_disk_id(client, disk_id): # NOQA
+    nodes = client.list_node()
+
+    for node in nodes:
+        disks = node.disks
+        for name, disk in iter(disks.items()):
+            if disk.diskUUID == disk_id:
+                return node
+    # should handle empty result in caller
+    return ""
+
+
+def check_backing_image_single_copy_disk_eviction(client, bi_name, old_disk_id): # NOQA
+    for i in range(RETRY_COUNTS):
+        backing_image = client.by_id_backing_image(bi_name)
+        current_disk_id = next(iter(backing_image.diskFileStatusMap))
+        if current_disk_id != old_disk_id:
+            break
+
+        time.sleep(RETRY_INTERVAL)
+
+    assert current_disk_id != old_disk_id
+
+
+def check_backing_image_single_copy_node_eviction(client, bi_name, old_node): # NOQA
+    for i in range(RETRY_COUNTS):
+        backing_image = client.by_id_backing_image(bi_name)
+        current_disk_id = next(iter(backing_image.diskFileStatusMap))
+        current_node = get_node_by_disk_id(client, current_disk_id)
+        if current_node.name != old_node.name:
+            break
+
+        time.sleep(RETRY_INTERVAL)
+
+    assert current_node.name != old_node.name
+
+
+def check_backing_image_eviction_failed(name): # NOQA
+    core_client = get_core_api_client()
+    selector = "involvedObject.kind=BackingImage,involvedObject.name=" + name
+    check = False
+
+    for i in range(RETRY_COUNTS_LONG):
+        events = core_client.list_namespaced_event(
+            namespace=LONGHORN_NAMESPACE,
+            field_selector=selector,
+        ).items
+        if len(events) == 0:
+            continue
+
+        for j in range(len(events)):
+            if (events[j].reason == FAILED_DELETING_REASONE and
+                    BACKINGIMAGE_FAILED_EVICT_MSG in events[j].message):
+                check = True
+                break
+
+        if check:
+            break
+
+        time.sleep(RETRY_INTERVAL)
+
+    assert check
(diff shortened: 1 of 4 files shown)


ChanYiLin:
This PR depends on the
longhorn-manager PR: https://github.com/longhorn/longhorn-manager/pull/2742

ChanYiLin:
> This pull request is now in conflict. Could you fix it @ChanYiLin? 🙏

Fixed.

ChanYiLin:
Hi @khushboo-rancher @shuo-wu @derekbit 
the feature has been merged, we can review and merge the tests
thanks

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +701,190 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+def test_backing_image_eviction(client): # NOQA
+    """
+    Test Node Eviction
+    - Create BackingImage with one copy
+    - Evict the node where the copy is on
+    - The BackingImage will be evicted to another node
+
+    Test Disk Eviction
+    - Disable the eviction on the previous node
+    - Evict the disk where the copy is currently on
```
shuo-wu, line 813: It seems that this test:
1. Does not prepare the 2nd disk for the test node
2. always evicts node rather than disks

ChanYiLin: You are right
I change the test to
1. evict the node in the first phase
2. evict the disk in the second phase
    - since each node has only one disk, the backing image will be evicted to other node, and we also check the diskID is different

the reason why we don't prepare another disk is because
=> we create the host disk by getting the host ID of the testing pod and mount the volume to the same node

But when preparing the backing image, we don't know where the backing image is
It may not on the host of the testing pod, thus, we can't prepare the host disk on the different node then the testing pod.
```
def create_host_disk(client, vol_name, size, node_id):
    # create a single replica volume and attach it to node
    volume = create_volume(client, vol_name, size, node_id, 1)

    # prepare the disk in the host filesystem
    disk_path = prepare_host_disk(get_volume_endpoint(volume), volume.name)
    return disk_path

```

manager/integration/tests/common.py
```diff
@@ -6270,3 +6274,58 @@ def wait_delete_dm_device(api, name):
             break
         time.sleep(RETRY_INTERVAL)
     assert not found
+
+
+def set_tags_for_node_and_its_disks(client, node, tags): # NOQA
```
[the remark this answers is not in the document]
ChanYiLin: This is not writing by me
It is used in test_ha long time ago

My test cases and `test_ha` test cases don't have such needs to set flexible tags.

But I agree with you
We can have another util function for this flexibility.
I can add another one, but lets keep the original one, because I don't want to change the `test_ha` testing code in this PR.

chriscchien (changes requested):
Test cases looks not contain the scenario of setting `nodeSelector` and `diskSelector` backing image csi VolumeSnapshot. But referenced in ticket [#6526](https://github.com/longhorn/longhorn/issues/6526#issuecomment-2089737138), does it need another e2e ticket for implement/track? Beside this LGTM, thank you

ChanYiLin:
Hi @chriscchien 
I have added the test of passing the `nodeSelector` and `diskSelector` through csi volumesnapshot
And since we already test the scheduling in backing image e2e tests.
We can simply check if the `spec.nodeSelector` and `spec.diskSelector` have the values then pass the csi test.

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +703,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+        diskSelector=backing_image_disk_selector)
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 1, "ready")
+
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    previous_disk_id = next(iter(backing_image.diskFileStatusMap))
+
+    node_1 = set_node_scheduling_eviction(
+        client, node_1, allowScheduling=False, evictionRequested=True)
+
+    time.sleep(60)
```
shuo-wu, line 895: Is there a better way to verify the BI eviction failure? e.g., Checking the related events.

This kind of verification is not deterministic and a waste of test resources.

manager/integration/tests/test_csi_snapshotter.py
```diff
@@ -1207,6 +1218,123 @@ def finalizer():
...
+        metadata:
+            name: test-snapshot-backing
+        spec:
+            volumeSnapshotClassName: longhorn-snapshot-vsc
+            source:
+                persistentVolumeClaimName: test-vol
+        ```
+
+    Then
+    - A BackingImage is created with the correct selectors
```
shuo-wu, line 1272: It's better to verify the BI data as well.

ChanYiLin: The content verification will be perform in the following test
`test_csi_volumesnapshot_backing_image_basic`
to save the testing time, I prefer focus on the selectors parameters passing through csi in this this.

manager/integration/tests/test_backing_image.py
ChanYiLin: Fixed.

ChanYiLin:
Waiting for testing
- `-vvs test_backing_image.py`: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7188/console
- `-vvs test_csi_snapshotter.py`: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7189/console

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +704,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 2, "ready")
+
+    update_setting(client, "backing-image-cleanup-wait-interval", "1")
+
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    backing_image = backing_image.updateMinNumberOfCopies(minNumberOfCopies=1)
+
+    time.sleep(60)
+
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 1, "ready")
```
derekbit, line 750: Can it be replaced a polling for loop such as
```
max_retries = 60
for (i = 0; i < max_retries; i++)
  check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 1, "ready")
  if i == max_retries - 1
    raise error
```

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +704,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+    When
+    - Update the minNumberOfCopies to 1
+    - Wait for 1 minute
+
+    Then
+    - There will be one ready BackingImage file left.
+    """
+    bi_url = BACKING_IMAGE_RAW_URL
+    create_backing_image_with_matching_url(
+        client, BACKING_IMAGE_NAME, bi_url)
```
derekbit, line 736: Is the default mimNumberOfCopies 3? Can we check the value after the creation as well?

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +704,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    backing_image = backing_image.updateMinNumberOfCopies(minNumberOfCopies=2)
+
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 2, "ready")
```
derekbit, line 741: No need to wait?

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +704,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+def test_backing_image_selector_setting(client, volume_name): # NOQA
+    """
+    - Set node1 with nodeTag = [node1], diskTag = [node1]
+    - Set node2 with nodeTag = [node2], diskTag = [node2]
+    - Create a BackingImage with
+        - minNumberOfCopies = 2
+        - nodeSelector = [node1]
+        - diskSelector = [node1]
+    - After creation, the first BackingImage copy will be on node1, disk1
+    - Wait for a while, the second one will never show up
```
derekbit, line 764: What will happen for the case
```
- Set node1 with nodeTag = [node1], disk1 diskTag = [ssd], disk2 diskTag = [ssd] 
- Set node2 with nodeTag = [node2], diskTag = [ssd]

    - Create a BackingImage with
        - minNumberOfCopies = 2
        - nodeSelector = [node1]
        - diskSelector = [ssd]
```
Can we allow the two copies on the same node's different disks?

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +704,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+    assert vol.diskSelector == volume_disk_selector
+    assert vol.nodeSelector == volume_node_selector
+
+    assert vol.conditions.Scheduled.status == "False"
+    cleanup_all_volumes(client)
+    cleanup_all_backing_images(client)
+
+
+@pytest.mark.backing_image  # NOQA
+def test_backing_image_eviction(client): # NOQA
```
derekbit, line 807: nit: suggest separating it into two functions `test_backing_image_node_eviction` and `test_backing_image_disk_eviction`.

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +704,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+    backing_image = backing_image.updateMinNumberOfCopies(minNumberOfCopies=2)
+
+    check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 2, "ready")
+
+    update_setting(client, "backing-image-cleanup-wait-interval", "1")
+
+    backing_image = client.by_id_backing_image(BACKING_IMAGE_NAME)
+    backing_image = backing_image.updateMinNumberOfCopies(minNumberOfCopies=1)
+
+    time.sleep(60)
```
derekbit, line 748: We already check the number of copies and status by `check_backing_image_disk_map_status`. Do you need the 60-second sleep?

manager/integration/tests/test_backing_image.py
```diff
@@ -694,3 +704,199 @@ def test_volume_wait_for_backing_image_condition(client): # NOQA
...
+def test_backing_image_unable_eviction(client): # NOQA
+    """
+    - Set node1 nodeTag = [node1], diskTag = [node1]
+    - Create a BackingImage with following settings to place the copy on node1
+        - minNumberOfCopies = 1
+        - nodeSelector = [node1]
+        - diskSelector = [node1]
+    - Evict node1
+    - The copy would not be deleted because it is the only copy
+    - The copy can't be copied to other nodes because of the selector settings.
```
derekbit, line 873: What can a user do for the case? Will it introduce a different behavior than before?

derekbit (approved):
In general, LGTM.
There are some behaviors or limitations we need to mention in the official doc.

manager/integration/tests/test_backing_image.py
ChanYiLin: `check_backing_image_disk_map_status()` itself is a for loop polling already
```
def check_backing_image_disk_map_status(client, bi_name, expect_cnt, expect_disk_state): # NOQA
    # Number of expect_cnt should equal to number of disk map
    # that have expect_disk_state

    for i in range(RETRY_COUNTS):
        backing_image = client.by_id_backing_image(bi_name)

        count = 0
        for disk_id, status in iter(backing_image.diskFileStatusMap.items()):
            if status.state == expect_disk_state:
                count = count + 1

        if expect_cnt == count:
            break
        time.sleep(RETRY_INTERVAL)

    assert expect_cnt == count
```

manager/integration/tests/test_backing_image.py
ChanYiLin: the default of `create_backing_image_with_matching_url()` function is 1
this test first create a minNumberOfCopies=1 backingimage
then update it to 2
then check if there are 2 files ready.

```
def create_backing_image_with_matching_url(client, name, url,
                                           minNumberOfCopies=1,
                                           nodeSelector=[], diskSelector=[]):
```

derekbit: I see. Can we check number of copies 1 after `create_backing_image_with_matching_url()`?

manager/integration/tests/test_backing_image.py
ChanYiLin: yes we can
backing image manager is disk level
and just like replicas, it allows to create file on the different disks of the same node

ChanYiLin: after all, if two replicas on disk1 and disk2 of the same node node1
we still need to prepare two backing image file on disk1 and disk2

manager/integration/tests/test_backing_image.py
ChanYiLin: you are right
check_backing_image_disk_map_status will poll the status for 150seconds
the cleanup time interval is set to 1min in this test case
we can rely on the check_backing_image_disk_map_status() to wait for the 1min

manager/integration/tests/test_backing_image.py
ChanYiLin: no, we don't have backing image eviction feature before
so when user evict a node, it won't even try to create the copy to other nodes in the previous version

This behavior is to prevent users from draining the node without noticing that this is the last backing image file
There is a event for this, and users should notice that the backing image is not evicted.

Moreover, the behavior is exactly the same as replicas
https://github.com/longhorn/longhorn-manager/blob/dcd5164185d7130732f72bb2e978dfdc0338acd1/controller/volume_controller.go#L1127-L1146

manager/integration/tests/test_backing_image.py
ChanYiLin: is already verified
```
    is_ready = False
    for i in range(RETRY_COUNTS):
        bi = client.by_id_backing_image(name)
        if len(bi.diskFileStatusMap) == minNumberOfCopies and bi.currentChecksum != "":
            for disk, status in iter(bi.diskFileStatusMap.items()):
                if status.state == "ready":
                    is_ready = True
                    break
            if is_ready:
                break
        time.sleep(RETRY_INTERVAL)

```

manager/integration/tests/test_backing_image.py
ChanYiLin: fixed.

derekbit:
Need to fix codespell and flake8 errors

derekbit (approved):

PR #1882, opened 2024-05-03
test(robot): add single replica node down cases

chriscchien:

Issue longhorn/longhorn#8350

Make manual test case [Single replica node down](https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-down/single-replica-node-down/) an automtaion test case

1. Create workload with `3` replicas then set the PVC replica count to `1`. After that,  delete specific replicas to make volume and the only one replica located on the same node or on diffent node. In the end power off the volume node to observe workload become ready or not by setting `node-down-pod-deletion-policy` and the only one replica locality.

2. And when volume faluted, the nodeID field is empty so it need `power_on_all_nodes` to boot up off node.

[test-report.zip](https://github.com/longhorn/longhorn-tests/files/15198264/test-report.zip)

### DIFF
--- a/e2e/keywords/common.resource
+++ b/e2e/keywords/common.resource
@@ -37,3 +37,7 @@ Cleanup test resources
     cleanup_backupstore
     cleanup_disks
     cleanup_backing_images
+
+Cleanup test resources include off nodes
+    Power on off node
+    Cleanup test resources
--- a/e2e/libs/keywords/workload_keywords.py
+++ b/e2e/libs/keywords/workload_keywords.py
@@ -12,10 +12,14 @@ from workload.workload import keep_writing_pod_data
 from workload.workload import write_pod_random_data
 from workload.workload import wait_for_workload_pods_running
 from workload.workload import wait_for_workload_pods_stable
+from workload.workload import wait_for_workload_pod_kept_in_state
+from workload.workload import get_pod_node
 
 from utility.constant import ANNOT_CHECKSUM
 from utility.constant import ANNOT_EXPANDED_SIZE
 from utility.utility import logging
+from node.utility import check_replica_locality
+from node.node import Node
 
 from volume import Volume
 from volume.constant import MEBIBYTE
@@ -121,3 +125,10 @@ class workload_keywords:
         logging(f'Waiting for {workload_name} volume {volume_name} to expand to {expanded_size}')
         self.volume.wait_for_volume_expand_to_size(volume_name, expanded_size)
         self.volume.wait_for_volume_detached(volume_name)
+
+    def wait_for_pod_kept_in_state(self, workload_name, expect_state, namespace="default"):
+        assert expect_state in ["Terminating", "ContainerCreating", "Running"], f"Unknown expected pod state: {expect_state}: "
+        return wait_for_workload_pod_kept_in_state(workload_name, expect_state, namespace=namespace)
+
+    def get_pod_node(self, pod):
+        return get_pod_node(pod)
--- a/e2e/libs/volume/crd.py
+++ b/e2e/libs/volume/crd.py
@@ -224,6 +224,16 @@ class CRD(Base):
             time.sleep(self.retry_interval)
         assert volume["status"]["state"] == desired_state
 
+    def wait_for_volume_keep_in_state(self, volume_name, desired_state):
+        self.wait_for_volume_state(volume_name, desired_state)
+
+        keep_state_desire_time = 20
+        for i in range(keep_state_desire_time):
+            volume = self.get(volume_name)
+            logging(f"Checking volume {volume} kept in status {desired_state}")
+            assert volume["status"]["state"] == desired_state
+            time.sleep(self.retry_interval)
+
     def wait_for_volume_robustness(self, volume_name, desired_state):
         volume = None
         for i in range(self.retry_count):
@@ -374,3 +384,29 @@ class CRD(Base):
         for replica in replica_list:
             node_set.add(replica['status']['ownerID'])
         assert len(replica_list) == len(node_set), f"unexpected replicas on the same node: {replica_list}"
+
+    def update_volume_spec(self, volume_name, key, value):
+        # retry conflict error
+        for i in range(self.retry_count):
+            try:
+                volume = self.get(volume_name)
+                spec = volume['spec']
+                if key == "numberOfReplicas":
+                    spec[key] = int(value)
+                else:
+                    spec[key] = value
+                self.obj_api.replace_namespaced_custom_object(
+                    group="longhorn.io",
+                    version="v1beta2",
+                    namespace="longhorn-system",
+                    plural="volumes",
+                    name=volume_name,
+                    body=volume
+                )
+                break
+            except Exception as e:
+                if e.status == 409:
+                    logging(f"Conflict error: {e.body}, retry ({i}) ...")
+                else:
+                    raise e
+            time.sleep(self.retry_interval)
--- a/e2e/libs/workload/workload.py
+++ b/e2e/libs/workload/workload.py
@@ -8,6 +8,7 @@ from utility.utility import logging
 
 from workload.pod import is_pod_terminated_by_kubelet
 from workload.constant import WAIT_FOR_POD_STABLE_MAX_RETRY
+from workload.constant import WAIT_FOR_POD_KEPT_IN_STATE_TIME
 
 
 def get_workload_pod_names(workload_name):
@@ -193,3 +194,35 @@ async def wait_for_workload_pods_stable(workload_name, namespace="default"):
         await asyncio.sleep(retry_interval)
 
     assert False, f"Timeout waiting for {workload_name} pods {wait_for_stable_pod} stable)"
+
+def wait_for_workload_pod_kept_in_state(workload_name, expect_state, namespace="default"):
+    def count_pod_in_specifc_state_duration(count_pod_in_state_duration, pods, expect_state):
+        for pod in pods:
+            pod_name = pod.metadata.name
+            if pod_name not in count_pod_in_state_duration:
+                count_pod_in_state_duration[pod_name] = 0
+            elif (expect_state == "ContainerCreating" and pod.status.phase == "Pending") or \
+                ((expect_state == "Terminating" and hasattr(pod.metadata, "deletion_timestamp") and pod.status.phase == "Running")) or \
+                (expect_state == "Running" and pod.status.phase == "Running"):
+                count_pod_in_state_duration[pod_name] += 1
+            else:
+                count_pod_in_state_duration[pod_name] = 0
+
+    retry_count, retry_interval = get_retry_count_and_interval()
+    count_pod_in_state_duration = {}
+
+    for i in range(retry_count):
+        pods = get_workload_pods(workload_name, namespace=namespace)
+        if len(pods) > 0:
+            count_pod_in_specifc_state_duration(count_pod_in_state_duration, pods, expect_state)
+            for pod in pods:
+                logging(f'Waiting for workload {workload_name} pod kept in {expect_state}')
+                if count_pod_in_state_duration[pod.metadata.name] > WAIT_FOR_POD_KEPT_IN_STATE_TIME:
+                    logging(f'Found pod {pod.metadata.name} kept in {expect_state}')
+                    return pod
+        time.sleep(retry_interval)
+
+    assert False, f"Timeout waiting for {workload_name} pod in state: {expect_state})"
+
+def get_pod_node(pod):
+    return pod.spec.node_name
--- a/e2e/tests/test_cases/single_replica_node_down.robot
+++ b/e2e/tests/test_cases/single_replica_node_down.robot
@@ -1,0 +1,135 @@
+*** Settings ***
+Documentation    Single replica node down
+...              https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-down/single-replica-node-down/
+
+Test Tags    manual_test_case
+
+Resource    ../keywords/common.resource
+Resource    ../keywords/deployment.resource
+Resource    ../keywords/persistentvolumeclaim.resource
+Resource    ../keywords/recurringjob.resource
+Resource    ../keywords/statefulset.resource
+Resource    ../keywords/volume.resource
+Resource    ../keywords/setting.resource
+Resource    ../keywords/setting.resource
+Resource    ../keywords/workload.resource
+Resource    ../keywords/host.resource
+
+Test Setup    Set test environment
+Test Teardown    Cleanup test resources include off nodes
+
+*** Variables ***
+${LOOP_COUNT}    1
+${RETRY_COUNT}    300
+${RETRY_INTERVAL}    1
+
+*** Test Cases ***
+Single Replica Node Down Deletion Policy do-nothing With RWO Volume Replica Locate On Replica Node
+    Given Set setting node-down-pod-deletion-policy to do-nothing
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+
+    # Delete replicas to have the volume with its only replica located on different nodes.
+    And Update volume of deployment 0 replica count to 1
+    And Delete replica of deployment 0 volume on replica node
+    And Delete replica of deployment 0 volume on volume node
+    And Power off volume node of deployment 0
+    Then Wait for volume of deployment 0 stuck in state attaching
+    And Wait for deployment 0 pod stuck in Terminating on the original node
+
+    When Power on off node
+    And Wait for deployment 0 pods stable
+    And Check deployment 0 pod is Running on another node
+    Then Check deployment 0 data in file data is intact
+
+Single Replica Node Down Deletion Policy do-nothing With RWO Volume Replica Locate On Volume Node
+    Given Set setting node-down-pod-deletion-policy to do-nothing
+
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+
+    # Delete replicas to have the volume with its only replica located on the same node.
+    And Update volume of deployment 0 replica count to 1
+    And Delete replica of deployment 0 volume on all replica node
+    And Power off volume node of deployment 0
+    Then Wait for volume of deployment 0 faulted
+    And Wait for deployment 0 pod stuck in Terminating on the original node
+
+    When Power on off node
+    And Wait for deployment 0 pods stable
+    And Check deployment 0 pod is Running on the original node
+    Then Check deployment 0 data in file data is intact
+
+Single Replica Node Down Deletion Policy delete-deployment-pod With RWO Volume Replica Locate On Replica Node
+    Given Set setting node-down-pod-deletion-policy to delete-deployment-pod
+
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+
+    # Delete replicas to have the volume with its only replica located on different nodes.
+    And Update volume of deployment 0 replica count to 1
+    And Delete replica of deployment 0 volume on replica node
+    And Delete replica of deployment 0 volume on volume node
+    And Power off volume node of deployment 0
+    Then Wait for volume of deployment 0 attaching
+
+    And Wait for deployment 0 pods stable
+    Then Check deployment 0 data in file data is intact
+    And Power on off node
+
+Single Replica Node Down Deletion Policy delete-deployment-pod With RWO Volume Replica Locate On Volume Node
+    Given Set setting node-down-pod-deletion-policy to delete-deployment-pod
+
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+
+    # Delete replicas to have the volume with its only replica located on the same node
+    And Update volume of deployment 0 replica count to 1
+    And Delete replica of deployment 0 volume on all replica node
+    And Power off volume node of deployment 0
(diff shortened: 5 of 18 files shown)


e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,118 @@
...
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+    And Update persistentvolumeclaim 0 volume replica count to ${1}
+    And Delete replica of deployment 0 volume on all replica node
+    And Power off volume node of deployment 0
+    And Sleep   ${SHUTDOWN_WAIT_DURATION}
+    Then Wait for volume of deployment 0 faulted
+
+    When Power on off nodes
```
yangchiu, line 54: How about keeping the powered off node node as a variable like: https://github.com/longhorn/longhorn-tests/blob/master/e2e/keywords/k8s.resource#L31

It makes test cases more consistent.

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,118 @@
...
+${RETRY_COUNT}    300
+${RETRY_INTERVAL}    1
+${SHUTDOWN_WAIT_DURATION}    180
+
+*** Test Cases ***
+Single Replica Node Down Deletion Policy do-nothing With RWO Volume Data Locality disabled
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+    And Update persistentvolumeclaim 0 volume replica count to ${1}
```
yangchiu, line 33: Is it possible to make comments here to help us understand what it would like to achieve?
```
Create workload with 3 replicas then set the PVC replica count to 1. 
After that, delete specific replicas to make volume and the only one replica located on the same node or on different node. 
```

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,118 @@
...
+Test Teardown    Cleanup test resources
+
+*** Variables ***
+${LOOP_COUNT}    1
+${RETRY_COUNT}    300
+${RETRY_INTERVAL}    1
+${SHUTDOWN_WAIT_DURATION}    180
+
+*** Test Cases ***
+Single Replica Node Down Deletion Policy do-nothing With RWO Volume Data Locality disabled
```
yangchiu, line 29: Why some test cases are named as `With RWO Volume Data Locality disabled` but some are just named as `With RWO Volume Data Locality`.

Don't they all have data locality `disabled` by default?

e2e/tests/test_cases/single_replica_node_down.robot
yangchiu, line 33: What will happen if the `${1}` is replaced with `1`?

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,118 @@
...
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+    And Update persistentvolumeclaim 0 volume replica count to ${1}
+    And Delete replica of deployment 0 volume on replica node
+    And Delete replica of deployment 0 volume on volume node
+    And Power off volume node of deployment 0
+    And Sleep   ${SHUTDOWN_WAIT_DURATION}
+    Then Wait for volume of deployment 0 attaching
```
yangchiu, line 38: There is a step `The workload pod will get stuck in the terminating state.` in the original manual test case. Does it need to be verified as well?

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,118 @@
...
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+    And Update persistentvolumeclaim 0 volume replica count to ${1}
+    And Delete replica of deployment 0 volume on all replica node
+    And Power off volume node of deployment 0
+    And Sleep   ${SHUTDOWN_WAIT_DURATION}
+    Then Wait for volume of deployment 0 faulted
+
+    When Power on off nodes
+    And Wait for deployment 0 pods stable
```
yangchiu, line 55: Will the pod be created on another node or the same node? Will the volume be attached to another node or the same node? 

Does it need to be verified as well?

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,118 @@
...
+*** Test Cases ***
+Single Replica Node Down Deletion Policy do-nothing With RWO Volume Data Locality disabled
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+    And Update persistentvolumeclaim 0 volume replica count to ${1}
+    And Delete replica of deployment 0 volume on replica node
+    And Delete replica of deployment 0 volume on volume node
+    And Power off volume node of deployment 0
+    And Sleep   ${SHUTDOWN_WAIT_DURATION}
```
yangchiu, line 37: Just wondering is the sleep required to pass the test case? Since the original manual test case doesn't mention the sleep time.

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,118 @@
...
+    Given Set setting node-down-pod-deletion-policy to delete-deployment-pod
+
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+    And Update persistentvolumeclaim 0 volume replica count to ${1}
+    And Delete replica of deployment 0 volume on all replica node
+    And Power off volume node of deployment 0
+    And Sleep   ${SHUTDOWN_WAIT_DURATION}
+    Then Wait for volume of deployment 0 faulted
```
yangchiu, line 84: According to the original test case, seems we have to check the following as well?
```
Wait for pod deletion & recreation on another node. The pod recreation will not happen immediately.
The replacement workload pod will get stuck in the ContainerCreating state.
```

e2e/libs/volume/crd.py
```diff
@@ -367,3 +367,27 @@ def validate_volume_replicas_anti_affinity(self, volume_name):
...
         assert len(replica_list) == len(node_set), f"unexpected replicas on the same node: {replica_list}"
+
+    def update_volume_spec(self, volume_name, key, value):
+        # retry conflict error
+        for i in range(self.retry_count):
+            try:
+                volume = self.get(volume_name)
+                spec = volume['spec']
+                spec[key] = value
+                volume['sepc'] = spec
```
yangchiu, line 378: There is a typo `sepc` here and seems this line is not required.

e2e/tests/test_cases/single_replica_node_down.robot
yangchiu, line 33: How about `Update volume of deployment 0 replica count to`?

Since we already have many keywords like `volume of ${workload_kind} ${workload_id}`.

WDYT? Does it make sense to you?

yangchiu (changes requested):

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated, using `Power on off node` in host.resources

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated test case name to 

- Single Replica Node Down Deletion Policy do-nothing With RWO Volume Replica Locate On Replica Node
- Single Replica Node Down Deletion Policy do-nothing With RWO Volume Replica Locate On Volume Node
- Single Replica Node Down Deletion Policy delete-deployment-pod With RWO Volume Replica Locate On Replica Node
- Single Replica Node Down Deletion Policy delete-deployment-pod With RWO Volume Replica Locate On Volume Node
- Single Replica Node Down Deletion Policy delete-both-statefulset-and-deployment-pod With RWO Volume Replica Locate On Replica Node
- Single Replica Node Down Deletion Policy delete-both-statefulset-and-deployment-pod With RWO Volume Replica Locate On Volume Node

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: 500 error, but I updated pytthon code to accept str for test case consistent.

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated
```
    And Wait for deployment 0 pods on volume node kept in state Terminating
    And Wait for deployment 0 pods on replica node kept in state ContainerCreating
```

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: IIRC, we do not have restrict rule for this and the seems the pod lcation decided by scheduler,  therefore check volume/pod OK and data intact maybe enough for now.

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: In my test env I did not encounter wait time not enough problem, because the previous step to wait node really stopped also have wait time.

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated

e2e/libs/volume/crd.py
chriscchien: Updated

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated

chriscchien:
Updated and tested on local environment.

For `wait_for_workload_pods_kept_in_state` in `workload_keywords.py`, because a faulted volume did not have nodeID hence we can not utilize `get_node_ids_by_replica_locality`, so this function using extra code to get replica nodes and volume node, thank you.

e2e/tests/test_cases/single_replica_node_down.robot
yangchiu: So why the sleep is required here?

e2e/tests/test_cases/single_replica_node_down.robot
yangchiu: But the original manual test case has steps to check pod created on another node, right?
```
Power up node or delete the workload pod so that kubernetes will recreate pod on another node.
````

```
Wait for pod deletion & recreation on another node.
```

File: e2e/tests/test_cases/single_replica_node_down.robot (hunk 4 above)
Comment (yangchiu) [reply e2e/tests/test_cases/single_replica_node_down.robot:40]: Could you help elaborate more on why do we need 2 separate steps here?

File: e2e/keywords/common.resource [hunk 9]
```diff
@@ -26,6 +27,7 @@ Set test environment
     set_backupstore
 
 Cleanup test resources
+    power_on_all_nodes
```
yangchiu, line 30: Is this still required? It introduces extra dependencies to the cleanup function. It would be great if we can figure out a better cleanup mechanism and avoid adding everything to this cleanup function. Do you have any idea?

e2e/tests/test_cases/single_replica_node_down.robot
yangchiu: Why node info is required when waiting for a pod state?

yangchiu (changes requested):

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated.

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated keyword name below to more close to the test plan. (change `kept` to `stuck`)
```robotframework
Wait for ${workload_kind} ${workload_id} pods on ${replica_locality} stuck in state ${expect_state}
```

Because in some scenarios there will not be only 1 pod in the test workload(1 stuck on volume node, 1 stuck in creating etc.).  and then it need node info to check pod state on specific node for the keyword

Back to the review comment
Step 
`The workload pod will get stuck in the terminating state.` 
should match 
`And Wait for deployment 0 pods on volume node stuck in state Terminating`

Step 
`The replacement workload pod will get stuck in the ContainerCreating state.` 
should match 
`And Wait for deployment 0 pods on replica node kept in state ContainerCreating`

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: > Will the pod be created on another node or the same node? Will the volume be attached to another node or the same node?
> 
> Does it need to be verified as well?

Added `Wait for ${workload_kind} ${workload_id} pod ${expect_state} on ${replica_locality}` for check

e2e/keywords/common.resource
chriscchien: Updated in common.resouce as below

```
Cleanup test resources include off nodes
    Power on off node
```

chriscchien:
Updated and [report.zip](https://github.com/longhorn/longhorn-tests/files/15273450/report.zip) here

e2e/tests/test_cases/single_replica_node_down.robot
yangchiu: In the original test case, it seems we don't need to know the pod running on which node? It only knows the original pod and replaced pod. Can we achieve this by retrieving pod's `creationTimestamp` or `startTime`?

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,136 @@
...
+    When Create persistentvolumeclaim 0 using RWO volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Write 100 MB data to file data in deployment 0
+
+    # Delete replicas to have the volume with its only replica located on different nodes.
+    And Update volume of deployment 0 replica count to 1
+    And Delete replica of deployment 0 volume on replica node
+    And Delete replica of deployment 0 volume on volume node
+    And Power off volume node of deployment 0
+    Then Wait for volume of deployment 0 attaching
```
yangchiu, line 39: The original step is `Volume will fail to attach since volume is not ready (i.e remains faulted, since single replica is on downed node).`

But `Then Wait for volume of deployment 0 attaching` doesn't seem to indicate the volume is stuck in a state. WDYT?

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,136 @@
...
+    And Wait for deployment 0 pods on volume node stuck in state Terminating
+
+    When Power on off node
+    And Wait for deployment 0 pods stable
+    And Wait for deployment 0 pod Running on replica node
```
yangchiu, line 44: The original test step is `kubernetes will recreate pod on another node`, so it looks like we should check the pod is created on a node which is not the previously powered off one instead?

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,136 @@
...
+    # Delete replicas to have the volume with its only replica located on the same node.
+    And Update volume of deployment 0 replica count to 1
+    And Delete replica of deployment 0 volume on all replica node
+    And Power off volume node of deployment 0
+    Then Wait for volume of deployment 0 faulted
+    And Wait for deployment 0 pods on volume node stuck in state Terminating
+
+    When Power on off node
+    And Wait for deployment 0 pods stable
+    And Wait for deployment 0 pod Running on volume node
```
yangchiu, line 63: In this case, will the pod run on the previously powered off node? Or another node?

Volume node and replica node are dynamically changing depend on the volume is attached to which node and where are replicas of this volume located. 

Just try to clarify things here. Thank you!

yangchiu (changes requested):

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Yes, pod will on the origin(powered off) node. 
Updated keyword to `Check deployment 0 pod is Running on the original node`

[report.zip](https://github.com/longhorn/longhorn-tests/files/15318396/report.zip)

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated keyword to `Check deployment 0 pod is Running on another node`

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated `Then Wait for volume of deployment 0 stuck in state attaching`

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: IIRC, the origin pod will be volume/engine node, after the volume node powed off, the replaced pod should create on the other scheduleable node. 

The step `The replacement workload pod will get stuck in the ContainerCreating state.` be performed by `And Wait for deployment 0 pod stuck in ContainerCreating on another node`, thank you. Or if it's necessary to use `startTime`, I will update it, thank you.

e2e/libs/keywords/workload_keywords.py
```diff
@@ -121,3 +124,21 @@ def wait_for_workload_claim_size_expanded(self, workload_name, claim_index=0):
         logging(f'Waiting for {workload_name} volume {volume_name} to expand to {expanded_size}')
         self.volume.wait_for_volume_expand_to_size(volume_name, expanded_size)
         self.volume.wait_for_volume_detached(volume_name)
+
+    def wait_for_workload_pods_kept_in_state_on_origin_node(self, workload_name, last_volume_node, expect_state, namespace="default"):
```
yangchiu, line 128: Can we just wait and se if there is any workload pods stuck in the expected state, and return the pod information, then compare the pod node name with `last_volume_node`?

Like:
```
pod = wait_for_pod_kept_in_state(workload_name, expected_state)
Should Be Equal    pod.node    last_volume_node
or
Should Not Be Equal    pod.node    last_volume_node
```

Do you think it's feasible?

yangchiu (changes requested):

e2e/libs/workload/workload.py
```diff
@@ -182,3 +183,45 @@ async def wait_for_workload_pods_stable(workload_name, namespace="default"):
         await asyncio.sleep(retry_interval)
 
     assert False, f"Timeout waiting for {workload_name} pods {wait_for_stable_pod} stable)"
+
+
+def wait_for_workload_pods_kept_in_state(workload_name, nodes_to_check, expect_state, namespace="default"):
```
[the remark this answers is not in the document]
chriscchien: Updated to `wait_for_workload_pod_kept_in_state` and change the function to wait any pod of workload kept in expect state for a duration then return pod, thank you.

e2e/libs/keywords/workload_keywords.py
chriscchien: Updated and test [report here](https://github.com/longhorn/longhorn-tests/files/15376596/report.zip)

e2e/libs/volume/crd.py
```diff
@@ -224,6 +224,16 @@ def wait_for_volume_state(self, volume_name, desired_state):
             time.sleep(self.retry_interval)
         assert volume["status"]["state"] == desired_state
 
+    def wait_for_volume_keep_at_state(self, volume_name, desired_state):
```
yangchiu, line 227: `at_state` is used here, but `in_state` is used elsewhere. Could we make it consistent?

yangchiu (changes requested):

e2e/tests/test_cases/single_replica_node_down.robot
```diff
@@ -0,0 +1,135 @@
...
+Test Setup    Set test environment
+Test Teardown    Cleanup test resources include off nodes
+
+*** Variables ***
+${LOOP_COUNT}    1
+${RETRY_COUNT}    300
+${RETRY_INTERVAL}    1
+
+*** Test Cases ***
+Single Replica Node Down Deletion Policy do-nothing With RWO Volume Replica Locate On Replica Node
```
yangchiu, line 27: These `Volume Replica Locate On Replica Node` type test cases are not mentioned in [the original manual test case](https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-down/single-replica-node-down/). Do you consider adding them and refining the manual one?

yangchiu (changes requested):

e2e/tests/test_cases/single_replica_node_down.robot
chriscchien: Updated, manual test case refine PR [here](https://github.com/longhorn/longhorn-tests/pull/1913), verified agagin on local, thank you.

yangchiu (approved):
LGTM

khushboo-rancher:
@chriscchien Could we move the test cases to the node_reboot.robot file instead of creating in a separate folder?

cc: @yangchiu

chriscchien:
> @chriscchien Could we move the test cases to the node_reboot.robot file instead of creating in a separate folder?
> 
> cc: @yangchiu

I would like to move the fiile, just wondering if all those manaul test [cases](https://github.com/longhorn/longhorn/issues/7211) need all add into negative test case file? thank you.

cc @yangchiu

yangchiu:
I think we should distinguish manual test cases from negative test cases instead of mixing them together.

Negative test cases are meant to be run repeatedly to test the resilience, but manual test cases only check the functionality.

Also the size of `node_reboot.robot` is bloating and I'm thinking moving some test cases out of it now.

PR #1898, opened 2024-05-12
ci: update kubectl version in Dockerfile to fix node drain test case failed

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/6485

update kubectl version in Dockerfile to fix node drain test case failed


yangchiu:
cc @longhorn/qa

chriscchien (approved):
LGTM

PR #1901, opened 2024-05-14
test: specify the namespace when uninstall Longhorn

roger-ryao:

Issue longhorn/longhorn#8568

After [89270bf](https://github.com/longhorn/longhorn/commit/89270bf0fabe40dcffe4eb8b345ca21326ae6b0c) was merged, we also need to update [Uninstal Longhorn step](https://github.com/longhorn/longhorn-tests/blob/master/benchmark_test/scripts/benchmark-test-setup.sh#L231). 

https://github.com/longhorn/longhorn/commit/89270bf0fabe40dcffe4eb8b345ca21326ae6b0c
https://ci.longhorn.io/job/private/job/longhorn-benchmark-test/


yangchiu (approved):
LGTM

PR #1690, opened 2024-01-24
feat(rwx): add rwx nfsOptions to robot test scaffolding.

james-munson:
#### Which issue(s) this PR fixes:

Issue # https://github.com/longhorn/longhorn/issues/7639

#### What this PR does / why we need it:

This is a draft for adding a new storage class and rwx volume with overridden mount options to an `e2e` test.

#### Special notes for your reviewer:

Untested - hope to test locally or when the pipeline can be triggered with a private build of longhorn-tests.  Just checking to see that this is the way it would be done.

PR #1904, opened 2024-05-16
test(robot): add rwx nfsOptions to robot test scaffolding

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/7639
PR https://github.com/longhorn/longhorn-tests/pull/1690

adding a new storage class and rwx volume with overridden mount options to an e2e test.


yangchiu:
cc @longhorn/qa

e2e/libs/workload/statefulset.py
```diff
@@ -10,7 +10,7 @@
 from utility.utility import logging
 
 
-def create_statefulset(statefulset_name, volume_type, option):
+def create_statefulset(statefulset_name, volume_type, sc_name):
     filepath = "./templates/workload/statefulset.yaml"
```
roger-ryao, line 14: For this change, do we need to add `sc_name` the default value, similar to what was done in https://github.com/longhorn/longhorn-tests/pull/1904/files#diff-cb5abaea0420a63451de0d96179c6a7859721fccec1cbdcd80940c8920eba665R34?

yangchiu: Not necessary. `create_statefulset` in `statefulset.py` is called by `create_statefulset` in `statefulset_keywords.py` which already specifies the default value. You can reference `create_volume` in `volume_keywords.py` which uses the same strategy.

roger-ryao (approved):
LGTM

PR #1905, opened 2024-05-16
test(manual) update Test Backup Creation With Old Engine Image

chriscchien:

Issue longhorn/longhorn#8580

Update manual test case Test Backup Creation With Old Engine Image


yangchiu:
I believe this test case has a specific objective to test the behavior of backup creation with incompatible engine images at that time (`v1.2.0`). Not sure if we should overwrite it or create a new test case for `Test backup creation with previous feature/patch release engine image`. Since once we overwrite it, it's not relevant to incompatible engine images and https://github.com/longhorn/longhorn/issues/2897 anymore. WDYT @longhorn/qa

roger-ryao (approved):
LGTM

PR #1909, opened 2024-05-21
test: add retry mechanism for update setting

yangchiu:
to fix test case test_setting_toleration and test_setting_priority_class

Issue https://github.com/longhorn/longhorn/issues/8582

add retry mechanism for update setting


chriscchien (approved):
LGTM

PR #1910, opened 2024-05-22
test(robot): physical node reboot

chriscchien:

Issue longhorn/longhorn#8567

Automate manual test case `physical node reboot
`

Verified on local


yangchiu (changes requested):
The original test case is:
Reboot _the node_ that the controller (StatefulSet/Deployment) attached to. After reboot complete, the volume should be reattached to _the node_.

But in this implementation, whether the volume is _reattached_ to the original node is not verified.

e2e/tests/test_cases/physical_node_reboot.robot
```diff
@@ -0,0 +1,52 @@
...
+    And Create deployment 1 with persistentvolumeclaim 1
+    And Write 100 MB data to file data in deployment 0
+    And Write 100 MB data to file data in deployment 1
+    And Reboot volume node of deployment 0
+    And Reboot volume node of deployment 1
+
+    And Wait for deployment 0 pods stable
+    Then Check deployment 0 data in file data is intact
+
+    And Wait for deployment 1 pods stable
```
yangchiu, line 37: `Wait for workloads pods stable` can be used to save wait time.

yangchiu (changes requested):

e2e/tests/test_cases/physical_node_reboot.robot
chriscchien: Updated!

e2e/tests/test_cases/physical_node_reboot.robot
```diff
@@ -0,0 +1,52 @@
...
+${RETRY_COUNT}    300
+${RETRY_INTERVAL}    1
+
+*** Test Cases ***
+Physical Node Reboot With Attached Deployment
+    Given Create persistentvolumeclaim 0 using RWO volume
+    And Create persistentvolumeclaim 1 using RWX volume
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Create deployment 1 with persistentvolumeclaim 1
+    And Record workloads attached node    deployment 0    deployment 1
```
yangchiu, line 29: It's complicated because you want to test everything in a single test case.

You can make `RWX` and `RWO` be a variable too. Like:
```
*** Variables ***
${VOLUME_TYPE}    RWO

*** Test Cases ***
Physical Node Reboot With Attached Deployment
Given Create persistentvolumeclaim 0 using ${VOLUME_TYPE} volume
```

And only dealing with a single workload in a test case, so the test case can be simplified.

No need to explicitly call a keyword to record the volume node:
```
And Record workloads attached node deployment 0 deployment 1
```
Instead, like what you implemented in the previous PR. you can embed it in a keyword:
```
Power off volume node of ${workload_kind} ${workload_id}
${last_volume_node} = get_volume_node ${volume_name}
Set Test Variable ${last_volume_node}
```

Move this logic to a more general keyword like:
```
Create deployment ${deployment_id} with persistentvolumeclaim ${claim_id}
```

Then the volume node recording will be automatic.

yangchiu (changes requested):

chriscchien:
Updated, also updated manual test case in [1924](https://github.com/longhorn/longhorn-tests/pull/1924)

yangchiu (approved):

PR #1914, opened 2024-05-24
test(robot): RWX volume data integrity after CSI plugin pod restart

c3y1huang:

Issue longhorn/longhorn#8184

In storage-network support of RWX volumes (longhorn/longhorn#8184), the CSI plugin pod no longer uses the custom mounter, `hostPID`, and host `proc` container volume mount. Instead, it relies on `hostNetwork: true`.

This proposed test PR verifies data accessibility and integrity for RWX volumes after a CSI plugin pod restart.

The custom mounter, `hostPID`, and host `proc` container volume mount have been removed and replaced with `hostNetwork: true` in the CSI plugin pod as part of the implementation for [[FEATURE] Support storage network for RWX volumes](https://github.com/longhorn/longhorn/issues/8184), hence this proposed test ensures the data accessibility and integrity for the RWX volumes after the CSI plugin pod restarts.

`None`

https://github.com/longhorn/longhorn/pull/8616


e2e/libs/keywords/workload_keywords.py
```diff
@@ -28,13 +32,23 @@
...
+        self.node = Node()
         self.persistentvolumeclaim = PersistentVolumeClaim()
         self.volume = Volume()
 
     def check_pod_data_checksum(self, expected_checksum, pod_name, file_name):
         logging(f'Checking checksum for file {file_name} in pod {pod_name}')
         check_pod_data_checksum(expected_checksum, pod_name, file_name)
 
+    def delete_workload_pod_on_node(self, workload_name, node_index, namespace="default"):
+        node_name = self.node.get_node_by_index(node_index)
```
yangchiu, line 44: It would be more consistent if we can get `node_name` in `*.resource` layer.

e2e/libs/workload/deployment.py
```diff
@@ -9,7 +9,7 @@
 from utility.utility import get_retry_count_and_interval
 
 
-def create_deployment(name, claim_name):
+def create_deployment(name, claim_name, replicaset=1):
```
yangchiu, line 12: Is it possible to make the function argument name identical to the field name? So we can have:
```
manifest_dict['spec']['replicas'] = replicas
```
I think it could make users lives easier.

e2e/keywords/deployment.resource
```diff
@@ -4,13 +4,21 @@ Documentation    Deployment Keywords
...
 Library    ../libs/keywords/deployment_keywords.py
+Library    ../libs/keywords/node_keywords.py
 
 *** Keywords ***
 Create deployment ${deployment_id} with persistentvolumeclaim ${claim_id}
     ${deployment_name} =    generate_name_with_suffix    deployment    ${deployment_id}
     ${claim_name} =    generate_name_with_suffix    claim    ${claim_id}
     create_deployment    ${deployment_name}    ${claim_name}
 
+Create deployment ${deployment_id} with persistentvolumeclaim ${claim_id} with max replicaset
```
yangchiu, line 15: We can simplify this by making an assumption that we have 3 worker nodes like:
```
def scale_statefulset_up(self, statefulset_name, replicaset_count=3):
```

yangchiu (changes requested):

e2e/libs/workload/deployment.py
c3y1huang: I prefer to use `replicaset` in this context because it aligns with the actual behavior which corresponds with the CRD kind ([ReplicaSet](https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/)) that is used by the controller. Additionally, it also simplifies when searching in the code base considering that Longhorn also uses the term `replica` for volume replicas. So this makes the distinction between the two.

e2e/keywords/deployment.resource
c3y1huang: IMO, we should avoid making assumptions about the infrastructure, as the robot isn't confined to running solely within the pipeline; it may also execute on our preferred cluster, which could possess a different setup in terms of node count.

yangchiu (approved):

PR #1922, opened 2024-05-30
ci: config vpn, add harvester terraform script, and implement harvester vm reboot operation

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8628

config vpn, add harvester terraform script, and implement harvester vm reboot operation


yangchiu:
cc @longhorn/qa

e2e/libs/host/harvester.py
```diff
@@ -0,0 +1,92 @@
...
+
+        stopped = False
+        for i in range(self.retry_count):
+            logging(f"Waiting for vm {vm_id} stopped ... ({i})")
+            resp = requests.get(url, cookies=self.cookies, verify=False)
+            if "Stopped" in resp.json()['metadata']['fields']:
+                stopped = True
+                break
+            time.sleep(self.retry_interval)
+        assert stopped, f"Expect vm {vm_id} is stopped, but it's not"
```
khushboo-rancher, line 73: nit: `Expected vm {vm_id} to be stopped but it's not` and similar logging messages at other places can be corrected. :)

test_framework/terraform/harvester/ubuntu/main.tf
```diff
@@ -0,0 +1,146 @@
+terraform {
+  required_providers {
+    rancher2 = {
+      source  = "rancher/rancher2"
+      version = "3.0.0-rc2"
```
khushboo-rancher, line 5: Why are we using `3.0.0-rc2`, the latest is [v4.1.0](https://github.com/rancher/terraform-provider-rancher2/releases/tag/v4.1.0)?

khushboo-rancher (changes requested):
In general LGTM, just a doubt regarding the terraform version.

e2e/libs/host/harvester.py
yangchiu: Refined. Thank you!

test_framework/terraform/harvester/ubuntu/main.tf
yangchiu: Refined. Thank you!

test_framework/terraform/harvester/ubuntu/main.tf
```diff
@@ -0,0 +1,146 @@
...
+
+    vm_namespace = "longhorn-qa"
+
+    cpu_count = "4"
+    memory_size = "8"
+
+    disk_info = <<EOF
+    {
+        "disks": [{
+            "imageName": "longhorn-qa/image-nbv7f",
```
khushboo-rancher, line 51: We might need a better image name like `ubuntu-22-04`. WDYT?

test_framework/terraform/harvester/ubuntu/main.tf
```diff
@@ -0,0 +1,146 @@
...
+      worker_role = true
+      quantity = 3
+      machine_config {
+        kind = rancher2_machine_config_v2.e2e-machine-config.kind
+        name = rancher2_machine_config_v2.e2e-machine-config.name
+      }
+    }
+    machine_selector_config {
+      config = <<EOF
+        cloud-provider-name: ""
```
khushboo-rancher, line 119: Do we need `harvester` as cloud provider name?

yangchiu: According to https://harvester.github.io/tests/manual/harvester-rancher2-terraform-integration/terraform_rancher2_provider_testing/, it's not required, and things work without problem when it's empty.

test_framework/terraform/harvester/ubuntu/main.tf
yangchiu: It looks like it expects an image id instead of an image name?

```
docker-machine-driver-harvester: ELF 64-bit LSB executable, x86-64, version
    1 (SYSV), statically linked, stripped

    Trying to access option  which does not exist

    THIS ***WILL*** CAUSE UNEXPECTED BEHAVIOR

    Type assertion did not go smoothly to string for key 

    Running pre-create checks...

    Error with pre-create check: "virtualmachineimages.harvesterhci.io
    \"ubuntu-2204\" not found"

    The default lines below are for a sh/bash shell, you can specify the shell
    you're using, with the --shell flag.
  failureReason: CreateError
```

khushboo-rancher (approved):

khushboo-rancher:
@yangchiu One last question regarding the code factor issue?

resp = requests.post(f"{url}?action=start", cookies=self.cookies, verify=False)

What's the reason for verify=False?

yangchiu:
> @yangchiu One last question regarding the code factor issue?
> 
> resp = requests.post(f"{url}?action=start", cookies=self.cookies, verify=False)
> 
> What's the reason for verify=False?

It avoid requests failed due to failed SSL certificate verification:
ssl-invalid

PR #1924, opened 2024-05-31
test(manual): update Physical node reboot

chriscchien:
longhorn/longhorn-8567

Issue https://github.com/longhorn/longhorn/issues/8567

In v1.6.2 and master, reboot workload volume node, the workload not 100% attached to the origin node after the rebooted node ready.

Update test case for only check volume healthy and data correct.

### DIFF
--- a/docs/content/manual/pre-release/node-not-ready/node-restart/physical-node-reboot.md
+++ b/docs/content/manual/pre-release/node-not-ready/node-restart/physical-node-reboot.md
@@ -1,4 +1,5 @@
 ---
 title: Physical node reboot
 ---
-1. Reboot the node that the controller (StatefulSet/Deployment) attached to. After reboot complete, the volume should be reattached to the node.
+1. Create StatefulSet/Deployment with 3 a replicas volume.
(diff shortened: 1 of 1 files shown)


yangchiu (changes requested):
Does the behavior conflict with the [`single replica node down` test case](https://github.com/longhorn/longhorn-tests/blob/master/e2e/tests/test_cases/single_replica_node_down.robot) you just implemented? I believe it checks if the volume is attached to a specific node.

chriscchien:
> Does the behavior conflict with the [`single replica node down` test case](https://github.com/longhorn/longhorn-tests/blob/master/e2e/tests/test_cases/single_replica_node_down.robot) you just implemented? I believe it checks if the volume is attached to a specific node.

In that test case, there is only one replica exist, it is different from this test case.

yangchiu:
I'm not familiar with the behavior. Can other team members confirm it? Many thanks! cc @longhorn/qa @longhorn/dev @derekbit

roger-ryao (approved):
LGTM

yangchiu (approved):

PR #1913, opened 2024-05-23
test(manual) update Single replica node down

chriscchien:
#### Which issue(s) this PR fixes:

Issue  longhorn/longhorn#8350

#### What this PR does / why we need it:

Add test scenario for volume and the only replica not on the same node.

#### Special notes for your reviewer:

N/A

#### Additional documentation or context

N/A

PR #1925, opened 2024-05-31
test(robot): update single_replica_node_down.robot

chriscchien:

Issue https://github.com/longhorn/longhorn/issues/8350

Remove the unavailable manul test case link (since https://github.com/longhorn/longhorn-tests/pull/1913)


chriscchien:
It would be appreciate if developer can help review this ticket. cc @derekbit @longhorn/dev

yangchiu (changes requested):
Since the original reference of these test cases has been deleted, is it possible to add documentation section for each test case to have a summary/description about what it's testing?

yangchiu (approved):
```

PR #1929, opened 2024-06-05
test(robot): migrate test_basic.py::test_restore_inc

yangchiu:

migrate test_basic.py::test_restore_inc

### DIFF
--- a/e2e/libs/keywords/volume_keywords.py
+++ b/e2e/libs/keywords/volume_keywords.py
@@ -24,9 +24,9 @@ class volume_keywords:
         for volume in volumes['items']:
             self.delete_volume(volume['metadata']['name'])
 
-    def create_volume(self, volume_name, size=2, replica_count=3, frontend="blockdev", migratable=False, access_mode="RWO", data_engine="v1", backing_image=""):
+    def create_volume(self, volume_name, size=2, numberOfReplicas=3, frontend="blockdev", migratable=False, accessMode="RWO", dataEngine="v1", backingImage="", Standby=False, fromBackup=""):
         logging(f'Creating volume {volume_name}')
-        self.volume.create(volume_name, size, replica_count, frontend, migratable, access_mode, data_engine, backing_image)
+        self.volume.create(volume_name, size, numberOfReplicas, frontend, migratable, accessMode, dataEngine, backingImage, Standby, fromBackup)
 
     def delete_volume(self, volume_name):
         logging(f'Deleting volume {volume_name}')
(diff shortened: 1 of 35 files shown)


e2e/libs/keywords/volume_keywords.py
```diff
@@ -24,9 +24,9 @@ def cleanup_volumes(self):
         for volume in volumes['items']:
             self.delete_volume(volume['metadata']['name'])
 
-    def create_volume(self, volume_name, size=2, replica_count=3, frontend="blockdev", migratable=False, access_mode="RWO", data_engine="v1", backing_image=""):
+    def create_volume(self, volume_name, size=2, numberOfReplicas=3, frontend="blockdev", migratable=False, accessMode="RWO", dataEngine="v1", backingImage="", Standby=False, fromBackup=""):
```
roger-ryao, line 27: I have a question. Why don't we continue to use snake_case for parameters? 
Are there any concerns?

yangchiu: I try to make the naming exactly identical to the longhorn volume crd, so users can set up a test case just like they would configure a Longhorn volume, without having to worry about the differences between Longhorn and Longhorn tests.

yangchiu:
cc @longhorn/qa

e2e/tests/regression/test_backup.robot
```diff
@@ -44,3 +49,58 @@ Test Backup Volume List
...
+    And Wait for volume 1 restoration from backup 0 completed
+    And Create DR volume 2 from backup 0
+    And Wait for volume 2 restoration from backup 0 completed
+    And Create DR volume 3 from backup 0
+    And Wait for volume 3 restoration from backup 0 completed
+
+    # Verify snapshot creation and PV/PVC creation are not allowed on a DR volume
+    Then Create snapshot 0 of volume 1 will fail
+    And Create persistentvolume for volume 1 will fail
+    And Create persistentvolumeclaim for volume 1 will fail
```
roger-ryao, line 71: One keyword that might help us better understand the story in the test report.
```
*** Keywords ***
Snapshot PV PVC could not be created on DR volume 1
    Create snapshot 0 of volume 1 will fail
    Create persistentvolume for volume 1 will fail
    Create persistentvolumeclaim for volume 1 will fail

Backup target could not be changed when DR volume exist
    Set setting backup-target to random.backup.target will fail


*** Test Cases ***
...
    And Wait for volume 3 restoration from backup 0 completed
    
    Then Snapshot PV PVC could not be created on DR volume 1
    And Backup target could not be changed when DR volume exist

    When Activate DR volume 1
...
```

Screenshot_20240607_161752

yangchiu: Refined. Thank you!

roger-ryao (approved):
LGTM

PR #1931, opened 2024-06-06
test(robot): storage-network-for-rwx-volume-enabled setting

c3y1huang:

Issue longhorn/longhorn#8184, longhorn/longhorn#8185

Propose a new robot test case for the `storage-network-for-rwx-volume-enabled` setting.

Depending on https://github.com/longhorn/longhorn-manager/pull/2815.

### DIFF
--- a/e2e/libs/workload/workload.py
+++ b/e2e/libs/workload/workload.py
@@ -1,14 +1,16 @@
 import time
 import asyncio
+
 from kubernetes import client
 from kubernetes.stream import stream
 
 from utility.utility import get_retry_count_and_interval
 from utility.utility import logging
 from utility.utility import list_namespaced_pod
-from workload.pod import is_pod_terminated_by_kubelet
+
 from workload.constant import WAIT_FOR_POD_STABLE_MAX_RETRY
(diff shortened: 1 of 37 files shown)


e2e/libs/workload/workload.py
```diff
@@ -225,3 +229,10 @@ def count_pod_in_specifc_state_duration(count_pod_in_state_duration, pods, expec
 
 def get_pod_node(pod):
     return pod.spec.node_name
+
+def is_workload_pods_has_annotations(workload_name, annotation_key, namespace="default", label_selector=""):
+    pods = get_workload_pods(workload_name, namespace=namespace, label_selector=label_selector)
+    for pod in pods:
+        if pod.metadata.annotations and annotation_key in pod.metadata.annotations:
+            return True
```
chriscchien, line 237: Looks like the statement will be satisfied if one pod has the expect annotation but the CSI plugin has many pods, should we check all pods instead of stopping after finding just one?, beside this, no other question, thank you.

c3y1huang: Good catch. Fixed, thank you.

chriscchien (approved):
LGTM, thank you.

PR #1932, opened 2024-06-06
ci: fix docker build error

FrankYang0529:
Original CI error: https://github.com/longhorn/longhorn-tests/actions/runs/9315512758/job/25641815189
This PR CI result: https://github.com/longhorn/longhorn-tests/actions/runs/9396482830/job/25877751459?pr=1932

### DIFF
--- a/manager/integration/Dockerfile
+++ b/manager/integration/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1.7.0
+
 FROM registry.suse.com/bci/python:3.11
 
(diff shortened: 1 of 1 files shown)


roger-ryao:
After upgrading Docker from `24.07` to `26.1.4`, I was able to build the private image without this PR. I'm not sure if we need to submit that.

FrankYang0529:
Thanks @roger-ryao for validation. We need this PR to fix CI error. If you are okay with it, may you approve it? Thanks.

PR #1933, opened 2024-06-06
test(manual) update Node disconnection test

chriscchien:

Issue #8617

Update manual test case [Node disconnection test](https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-disconnection/node-disconnection/). Because the last content updated was 2 years ago.

Tested on v1.4.4 and master-head, after volume node network back, the volume does not detach, but if data is written during the volume node network disconnect, replica rebuilding is observed..


roger-ryao (approved):
LGTM

PR #1943, opened 2024-06-18
test: refine test_drain_with_block_for_eviction_if_contains_last_replica_success

chriscchien:

longhorn/longhorn#8753

Do not use stict-local volume for disk to avoid misunderstanding.

Verified on local

### DIFF
--- a/manager/integration/tests/test_node.py
+++ b/manager/integration/tests/test_node.py
@@ -2927,8 +2927,7 @@ def test_drain_with_block_for_eviction_if_contains_last_replica_success(client, 
     disk_volume_name = 'vol-disk'
     disk_volume = client.create_volume(name=disk_volume_name,
                                        size=str(2 * Gi),
-                                       numberOfReplicas=1,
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_node.py
```diff
@@ -2927,8 +2927,7 @@ def test_drain_with_block_for_eviction_if_contains_last_replica_success(client,
     disk_volume_name = 'vol-disk'
     disk_volume = client.create_volume(name=disk_volume_name,
```
yangchiu, line 2928: It seems the `disk_volume` is used to create an extra disk on the host node and is not involved in the subsequent node eviction process. So is the modification necessary?

yangchiu (changes requested):

manager/integration/tests/test_node.py
chriscchien: It's from https://github.com/longhorn/longhorn/issues/8753#issuecomment-2172325125, I am fine with either ways.

PR #1945, opened 2024-06-18
ci: install cryptsetup for volume encryption test

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8688

install cryptsetup for volume encryption test


yangchiu:
cc @longhorn/qa

innobead (approved):

PR #1946, opened 2024-06-20
Add a test skeleton for filesystem freeze functionality

ejweber:

longhorn/longhorn#2187

### DIFF
--- a/manager/integration/tests/test_snapshot.py
+++ b/manager/integration/tests/test_snapshot.py
@@ -726,3 +726,22 @@ def test_snapshot_cr(client, volume_name, settings_reset):  # NOQA
         assert len(snapshots) == 1
         assert snapshots[0].created != created_time
         created_time = snapshots[0].created
+
+@pytest.mark.skip(reason="TODO") # NOQA
+def test_freeze_file_system_for_snapshot(): # NOQA
+    """
+    1. Record the test start time.
+    2. Set freeze-filesystem-for-snapshot to true.
+    3. Create a Longhorn volume 6 GiB with a replica on every Longhorn node.
+    4. Create a workload pod that mounts the volume and runs
+       "dd if=/dev/random of=/path/to/file/under/mount/point bs=1M count=5000".
+       (Add "status=progress" if running manually to see the progress freeze.)
+    5. After the workload is running, take a snapshot.
+    6. Verify the snapshot succeeded.
+    7. Verify the pod completed successfully.
+    8. Verify the following logs appeared in the instance-manager running the
+       volume engine after the test began. (Checking logs this way isn't ideal,
+       but there isn't really a better way to confirm the freeze.)
+       - "Freezing filesystem mounted at"
+       - "Unfreezing filesystem mounted at"
+    """


manager/integration/tests/test_snapshot.py
```diff
@@ -726,3 +726,22 @@ def test_snapshot_cr(client, volume_name, settings_reset):  # NOQA
...
+
+@pytest.mark.skip(reason="TODO") # NOQA
+def test_freeze_file_system_for_snapshot(): # NOQA
+    """
+    1. Record the test start time.
+    2. Set freeze-filesystem-for-snapshot to true.
+    3. Create a Longhorn volume 6 GiB with a replica on every Longhorn node.
+    4. Create a workload pod that runs
+       "dd if=/dev/random of=/path/to/file/on/volume bs=1M count=5000".
+       (Add "status=progress" if running manually to see the progress freeze.)
```
chriscchien, line 738: Hello @ejweber , does step 4 means to use a workload to mount volume created in step 3 and then write data into mount point?

ejweber: It does. Sorry for any confusion. I changed the wording a bit. Please let me know if it is still unclear.

chriscchien (approved):
~LGTM~

Hi @ejweber 
 According to step 8, the logs should be appeared in all instance-manager pods but I can only see related log appeared in the instance-manager which running on the volume attached node, could you help confirm if this is correct behavior? thank you.
```
8. Verify the following logs appeared in all instance-managers after the
       test began. (Checking logs this way isn't ideal, but there isn't really
       a better way to confirm the freeze.)
       - "Freezing filesystem mounted at"
       - "Unfreezing filesystem mounted at"
```

My steps
1. A 3 nodes cluster, from UI create volume `vol1` with 6 Gi size and 3 replicas
2. Create PVC `vol1` from UI
3. Apply below manifest
```
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  containers:
  - name: test-container
    image: ubuntu:latest
    command: ["/bin/sh", "-c"]
    args: ["dd if=/dev/random of=/mnt/data/1 bs=1M count=5000 && sleep infinity"]
    volumeMounts:
    - name: test-vol
      mountPath: /mnt/data
  volumes:
  - name: test-vol
    persistentVolumeClaim:
      claimName: vol1
```
4. After pod running, create snapshot from UI.
5. Wait data writing complete, check logs in all instance-manager pods
6. Only find related log appears in the instance-manager pod which run on the volume attaced node.


[supportbundle_94204353-8e83-473e-a788-b94f4699eb73_2024-07-01T02-36-02Z.zip](https://github.com/user-attachments/files/16047372/supportbundle_94204353-8e83-473e-a788-b94f4699eb73_2024-07-01T02-36-02Z.zip)

ejweber:
> Hi @ejweber
> According to step 8, the logs should be appeared in all instance-manager pods but I can only see related log appeared in the instance-manager which running on the volume attached node, could you help confirm if this is correct behavior? thank you.

Thanks for the careful reading of the skeleton!

There is no reason to expect the logs anywhere except the instance-manager running the volume engine. TBH, I'm not sure why I wrote it that way. Changing it in an amend.

chriscchien (approved):
LGTM

PR #1948, opened 2024-06-24
Rework HA volume migration test cases for v1.5.0+

ejweber:

longhorn/longhorn#8735
longhorn/longhorn#3401

Many of the manual HA volume migration test cases no longer pass with the introduction of the volume attachment controller in `v1.5.0`. I reworked the behavior in longhorn/longhorn#8735, but the manual tests cases also need to be updated. In this iteration, I removed a couple of cases that weren't particularly interesting, added a couple of additional cases, and added detail regarding the expected behavior in `v1.5.0+`.

### DIFF
--- a/docs/content/manual/pre-release/ha/ha-volume-migration.md
+++ b/docs/content/manual/pre-release/ha/ha-volume-migration.md
@@ -2,30 +2,400 @@
 title: HA Volume Migration
 ---
 
-1. Create a migratable volume:
-    1. Deploy a migratable StorageClass. e.g., https://github.com/longhorn/longhorn/blob/master/examples/rwx/storageclass-migratable.yaml
-    2. Create a PVC with access mode `ReadWriteMany` via this StorageClass.
-2. Attach a volume to a node and wait for volume running. Then write some data into the volume. Here I would recommend directly restoring a volume (set `fromBackup` in the StorageClass) and attach it instead.
-3. Start the migration by request attaching to another node for the volume.
-4. Trigger the following scenarios then confirm or rollback the migration:
+## Basic instructions
 
-   --------------------------
-   | **#** | **Scenario**          | **Migration Confirmation** | **Migration Rollback** |
-   | ----- | --------------------- | -------------------------- | ---------------------- |
-   | 1     | New engine crash      | If the request is successfully sent before the migration engine & replicas being removed, the volume should be auto reattached after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
-   | 2     | All new replica crash | If the request is successfully sent before the migration engine & replicas being removed, the volume should keep running (then may start rebuilding) after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
-   | 3     | Old engine crash      | If the request is successfully sent before the volume becoming Faulted, the volume should keep running after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
-   | 4     | All old replica crash | If the request is successfully sent before the volume becoming Faulted, the volume should keep running (then may start rebuilding) after migration. <br> Otherwise, the request will be rejected until the volume is reattached and restarts the migration components | Should succeed |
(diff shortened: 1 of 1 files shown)


yangchiu (commented):
LGTM. We can merge this after https://github.com/longhorn/longhorn-manager/pull/2894.

docs/content/manual/pre-release/ha/ha-volume-migration.md
```diff
@@ -2,30 +2,400 @@
...
+kl delete --wait=false pod instance-manager-699da83c0e9d22726e667344227e096b && sleep 0.5 && k delete -f va-1.yaml
+# OR
+kl delete --wait=false pod instance-manager-699da83c0e9d22726e667344227e096b && sleep 1 && k delete -f va-1.yaml
+```
+
+- "...selected to detach from <old>"
+- "Cancelling migration for detached volume..."
+- MigrationFailed event
+- "...selected to attach to <new>"
+- Same engine and replicas
```
yangchiu, line 206: Since the volume has been attached to the new node, what does "same engine and replicas" mean here?

yangchiu (changes requested):

docs/content/manual/pre-release/ha/ha-volume-migration.md
```diff
@@ -2,30 +2,400 @@
...
+k delete --wait=false -f va-1.yaml && sleep 1 && kl delete pod instance-manager-699da83c0e9d22726e667344227e096b
+```
+
+- "Confirming migration..."
+- No detachment
+- New engine and replicas
+
+#### 2.4 Rollback immediately after crash
+
+The volume automatically detaches from the old node. Then, it reattaches to the new node.
```
yangchiu, line 224: Since it's selected to attach to "old", should it be "reattaches to the old node"?

yangchiu (changes requested):

docs/content/manual/pre-release/ha/ha-volume-migration.md
```diff
@@ -2,30 +2,400 @@
...
+
+#### 5.1 Confirmation
+
+The volume is allowed to detach from the old node (special logic in code). It attempts to attach to cleanly attach to
+the migration node, but is stuck until it comes back.
+
+- "Waiting to confirm migration until migration engine is ready"
+- "Detaching volume for attempted migration to down node"
+- "...selected to attach to <new>"
+- Same engine and replicas
```
yangchiu, line 351: Is it possible to provide more information on how to check "Same engine and replicas"? Sorry but I'm a bit confused..

yangchiu (changes requested):

docs/content/manual/pre-release/ha/ha-volume-migration.md
ejweber: You are right. Copy/paste error. Sorry!

docs/content/manual/pre-release/ha/ha-volume-migration.md
ejweber: Since the migration actually failed, the new engine and replicas created for the migration should be cleaned up. Then, the old engine and replicas are used to attach to the new node. It's a migration failure -> detach -> attach instead of a successful migration.

In practice, it means if you check the engine and replica names before the scenario and then after, they are the same. If the migration succeeded, they would be different.

docs/content/manual/pre-release/ha/ha-volume-migration.md
ejweber: I added https://github.com/longhorn/longhorn-tests/pull/1948/files#diff-b892c032654f9b2b7dcd1d87cdd7c8333ab3eb449f5a71213990f3bb8ebdc6ffR86-R91 for clarity.

In practice, it means if you check the engine and replica names before the scenario and then after, they may be the same (e.g. for a migration failure or a successful migration rollback) or different (e.g. for a successful migration confirmation). In the middle of the migration, we always have two engines with different names and two sets of three replicas with different names.

Please let me know if you think I should add additional context!

yangchiu (approved):
LGTM

PR #1949, opened 2024-06-26
test(robot): power off replica node should not rebuild new replica on same node

c3y1huang:

Issue longhorn/longhorn#8825 , longhorn/longhorn#1992

Add a new case to ensure that no new replica is created and rebuilt on the same node if the node is powered off for a duration longer than the replica-replenishment-wait-interval. When the node is powered on, the existing replica should be reused.

`None`

`None`


yangchiu (approved):
LGTM

PR #1950, opened 2024-06-26
test: print message when `system_restore_wait_for_state` assert failed

roger-ryao:
longhorn/longhorn#8826

It appears to fail only during a full regression. We are unable to reproduce the issue when running the test case individually.

https://github.com/longhorn/longhorn/issues/8826
https://github.com/longhorn/longhorn/issues/8515#issuecomment-2177537181


yangchiu (approved):
LGTM

PR #1951, opened 2024-07-02
fix(integration): test_extra_replica_cleanup

c3y1huang:

Issue longhorn/longhorn#8871

Adjust the test case following the changes in https://github.com/longhorn/longhorn-manager/pull/2918. We implemented a precheck before creating a new Replica CR, so no new Replica will be created is no suitable node and disk are available.

`None`

`None`

### DIFF
--- a/manager/integration/tests/test_ha.py
+++ b/manager/integration/tests/test_ha.py
@@ -1932,21 +1932,20 @@ def test_rebuild_after_replica_file_crash(client, volume_name): # NOQA
     check_volume_data(volume, data)
 
 
-def test_extra_replica_cleanup(client, volume_name, settings_reset): # NOQA
+def test_replica_should_not_be_created_when_no_suitable_node_found(client, volume_name, settings_reset): # NOQA
     """
-    Test extra failed to scheduled replica cleanup
-    when no eviction requested
+    Test replica should not be created when no suitable node is found.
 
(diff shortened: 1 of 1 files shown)


c3y1huang:
To PASS the test case is dependent on the fix in https://github.com/longhorn/longhorn-manager/pull/2939 for the volume condition. However, this test case is ready to review.

manager/integration/tests/test_ha.py
```diff
@@ -1932,21 +1932,20 @@ def test_rebuild_after_replica_file_crash(client, volume_name): # NOQA
...
     3. Attach the volume to a node and write some data to it and
     save the checksum.
     4. Increase the volume replica number to 4.
-    5. Volume should show failed to schedule and an extra stop replica.
-    6. Decrease the volume replica number to 3.
-    7. Volume should show healthy and the extra failed to scheduled replica
-    should be removed.
-    8. Check the data in the volume and make sure it's same as the checksum.
+    5. No Replica should be created.
+    7. Volume should show failed to schedule.
```
roger-ryao, line 1945: LGTM, but there is a small error with the numbering of the cases.

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7106/

roger-ryao (approved):
LGTM

PR #1956, opened 2024-07-03
test: specify which container to execute the command in pod exec

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8878

specify which container to execute the command in pod exec


chriscchien (approved):
LGTM

PR #1970, opened 2024-07-08
test(robot): wait for disk updated when adding disks

yangchiu:

Issue N/A

wait for disk updated when adding disks


chriscchien (approved):
LGTM

PR #1971, opened 2024-07-08
test(robot): make cluster restart test case be able to run out of cluster

yangchiu:

Issue N/A

make cluster restart test case be able to run out of cluster


chriscchien (approved):
LGTM

PR #1972, opened 2024-07-08
test(robot): fix flaky test case single replica node down and refine logging

yangchiu:

Issue N/A

fix flaky test case single replica node down and refine logging


chriscchien (approved):
LGTM

PR #1974, opened 2024-07-12
test(robot): extend regression test cases to v2 volume

yangchiu:

Issue N/A

extend regression test cases to v2 volume


roger-ryao (approved):
LGTM

PR #1988, opened 2024-07-19
fix(backupbackingimage): cleanup backup backing image in the backupstore_cleanup function

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9041

The issue is caused by the backup backing image left over from the previous test.
We should clean up the backup backing image in the backupstore_cleanup function

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -517,6 +517,12 @@ def delete_backup_volume(client, volume_name):
     wait_for_backup_volume_delete(client, volume_name)
 
 
+def delete_backup_backing_image(client, backing_image_name):
+    bbi = client.by_id_backupBackingImage(backing_image_name)
+    client.delete(bbi)
+    wait_for_backup_backing_image_delete(client, backing_image_name)
+
+
 def create_and_check_volume(client, volume_name,
                             num_of_replicas=3, size=SIZE, backing_image="",
                             frontend=VOLUME_FRONTEND_BLOCKDEV,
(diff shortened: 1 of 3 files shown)


manager/integration/tests/common.py
```diff
@@ -2054,6 +2060,20 @@ def wait_for_backup_volume_delete(client, name):
...
+    for _ in range(RETRY_BACKUP_COUNTS):
+        bbis = client.list_backupBackingImage()
+        found = False
+        for bbi in bbis:
+            if bbi.name == name:
+                found = True
+                break
+        if not found:
+            break
+        time.sleep(RETRY_BACKUP_INTERVAL)
```
derekbit, line 2073:
```suggestion
def wait_for_backup_backing_image_delete(client, name):
    for _ in range(RETRY_COUNTS):
        bbis = client.list_backupBackingImage()
        found = False
        for bbi in bbis:
            if bbi.name == name:
                found = True
                break
        if not found:
            break
        time.sleep(RETRY_INTERVAL)
```

ChanYiLin:
e2e test is running
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/

manager/integration/tests/common.py
ChanYiLin: fixed.

derekbit:
> e2e test is running
> 
> * https://ci.longhorn.io/job/private/job/longhorn-tests-regression/

We can merge it after the regression completed.

yangchiu (approved):

ChanYiLin:
Somehow the previous e2e test got ERROR for all the tests
Here is the new one: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7279/console

roger-ryao:
@mergify backport v1.7.x

roger-ryao:
@mergify backport v1.6.x

PR #1993, opened 2024-07-22
fix(engineimage): check if engine image is deployed correctly

mantissahz:

Issue # longhorn/longhorn#9038

After we tainted a node as `NoSchedule` and delete the EngineImage DaemonSet, `nodeDeploymentMap` will be unstable for a period and it is not enough only to check if the `nodeDeploymentMap` has the expected count for deloyed node.


derekbit (approved):
LGTM

mantissahz:
@Mergifyio backport v1.7.x v1.6.x

PR #1995, opened 2024-07-22
ci: use larger azure instance type

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9024

use larger azure instance type


chriscchien (approved):
LGTM

PR #1996, opened 2024-07-22
ci: avoid generating go-cover in sle-micro read-only file system

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9001

avoid generating go-cover in sle-micro read-only file system


yangchiu:
cc @longhorn/qa

khushboo-rancher (approved):

PR #2008, opened 2024-07-24
test: refine test_rwx_delete_share_manager_pod for share manager HA

yangchiu:
#### Which issue(s) this PR fixes:

Issue https://github.com/longhorn/longhorn/issues/6205

#### What this PR does / why we need it:

refine test_rwx_delete_share_manager_pod for share manager HA

PR #2011, opened 2024-07-26
test: put back test_rwx_delete_share_manager_pod expects workload res…

james-munson:
…tart.

Issue # https://github.com/longhorn/longhorn/issues/9081

This PR reverts the change made in https://github.com/longhorn/longhorn-tests/pull/2008.  With Phan's changes, the sync flow for share-manager is closer to the original, and the test can expect the RWX workload pods to be restarted, at least for the moment.


yangchiu (approved):
LGTM

yangchiu:
@Mergify backport v1.7.x

PR #2013, opened 2024-07-26
fix(backup): check if dummy backup resource is deleted when the config is deleted

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9037

Longhorn automatically creates backup resource in the cluster when there is a backup in the backupstore.
We need to check if the backup resource is deleted as well before creating the next backup.
Otherwise, the following backup creation might fail due to the deletion lock introduced by previous backup CR deletion.


ChanYiLin:
## Wait for e2e test

yangchiu (approved):
LGTM

yangchiu:

PR #2014, opened 2024-07-27
test(robot): extend negative test cases to test rwx volume fast failover

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/6205

extend negative test cases to test rwx volume fast failover


roger-ryao (approved):
LGTM

PR #2017, opened 2024-07-27
ci: workaround for sle-micro not rebooting via ssh

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9001

workaround for sle-micro not rebooting via ssh


roger-ryao (approved):
LGTM

PR #2023, opened 2024-07-29
ci: use longhornctl to handle prerequisite installations and perform environment checks

yangchiu:

Issue N/A

use longhornctl to handle prerequisite installations and perform environment checks

### DIFF
--- a/test_framework/scripts/longhorn-setup.sh
+++ b/test_framework/scripts/longhorn-setup.sh
@@ -351,6 +351,17 @@ create_aws_secret(){
 }
 
 
+longhornctl_check(){
+  curl -L https://github.com/longhorn/cli/releases/download/v1.7.0-rc2/longhornctl-linux-amd64 -o longhornctl
+  chmod +x longhornctl
+  ./longhornctl install preflight
+  ./longhornctl check preflight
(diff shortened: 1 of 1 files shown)


test_framework/scripts/longhorn-setup.sh
```diff
@@ -351,6 +351,17 @@ create_aws_secret(){
 }
 
 
+longhornctl_check(){
+  curl -L https://github.com/longhorn/cli/releases/download/v1.7.0-rc2/longhornctl-linux-amd64 -o longhornctl
```
roger-ryao, line 355: For `arm64`, we should download https://github.com/longhorn/cli/releases/download/v1.7.0-rc2/longhornctl-linux-arm64 instead of `longhornctl-linux-amd64`, right?

roger-ryao (changes requested):

test_framework/scripts/longhorn-setup.sh
yangchiu: No. it’s run in a Jenkins agent docker container instead of the nodes you created.

roger-ryao (approved):
LGTM

derekbit (approved):
LGTM

yangchiu:
@Mergify backport v1.7.x

PR #2031, opened 2024-07-31
ci: install cryptsetup on sle-micro worker nodes

yangchiu:

Issue N/A

install cryptsetup on sle-micro worker nodes


chriscchien (approved):
LGTM

PR #2032, opened 2024-07-31
test: skip test case test_backup_lock_creation_during_deletion

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8269

skip test case test_backup_lock_creation_during_deletion since in the current Longhorn implementation, it's almost impossible to catch the moment when the lock is held and the test case always fails.


chriscchien (approved):
LGTM

yangchiu:
@Mergify backport v1.7.x

yangchiu:
@mergify backport v1.6.x

PR #2052, opened 2024-08-09
ci: using larger instances in gke to prevent test pod crashes

yangchiu:

Issue N/A

using larger instances in gke to prevent test pod crashes


chriscchien (approved):
LGTM

PR #2056, opened 2024-08-12
ci: enable both uio_pci_generic and vfio_pci

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8813

enable both uio_pci_generic and vfio_pci


roger-ryao (approved):
LGTM

PR #2057, opened 2024-08-12
ci: add talos provisioning

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/6596

add talos provisioning


pipelines/e2e/scripts/longhorn-setup.sh
```diff
@@ -25,6 +25,16 @@ create_instance_mapping_configmap(){
   kubectl create configmap instance-mapping --from-file=/tmp/instance_mapping
 }
 
+longhornctl_check(){
+  curl -L https://github.com/longhorn/cli/releases/download/v1.7.0-rc2/longhornctl-linux-amd64 -o longhornctl
```
roger-ryao, line 29: We could use `v1.7.1-rc2` from https://github.com/longhorn/cli/releases/download/v1.7.1-rc2/longhornctl-linux-amd64 for now.

yangchiu: Updated. Thank you!

khushboo-rancher (approved):
We can update the cli version to the released version 1.7.1 with this PR or later.

PR #2058, opened 2024-08-12
test(robot): wait for block disks ready and schedulable

yangchiu:

Issue N/A

wait for block disks ready and schedulable

### DIFF
--- a/e2e/libs/node/node.py
+++ b/e2e/libs/node/node.py
@@ -35,7 +35,10 @@ class Node:
                 all_updated = True
                 disks = node.disks
                 for d in disks:
-                    if disks[d]["diskUUID"] == "":
+                    if disks[d]["diskUUID"] == "" or
+                        not disks[d]["conditions"] or
+                        disks[d]["conditions"]["Ready"]["status"] != "True" or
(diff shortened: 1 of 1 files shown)


roger-ryao (approved):
LGTM

e2e/libs/node/node.py
```diff
@@ -35,7 +35,10 @@ def wait_for_disk_update(self, node_name, disk_num):
                 all_updated = True
                 disks = node.disks
                 for d in disks:
-                    if disks[d]["diskUUID"] == "":
+                    if disks[d]["diskUUID"] == "" or
+                        not disks[d]["conditions"] or
+                        disks[d]["conditions"]["Ready"]["status"] != "True" or
+                        disks[d]["conditions"]["Schedulable"]["status"] != "True":
```
derekbit, line 41: If a block disk in `node.spec.disks`  is not added to the `node.status.diskStatus` yet, the check will succeed. Then. a volume will be created before the disk becomes ready.

yangchiu: It assert len(node.disks) == disk_num, and disk_num equals 2 here, which means default disk + newly added block disk.

PR #2063, opened 2024-08-20
Fix flaky test_exporting_backing_image_from_volume on sle-micro arm64

roger-ryao:
longhorn/longhorn#9254

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7486/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7487/

After executing the test case 20 times, I found that `volume3 = wait_for_volume_healthy` takes 350 to 450 seconds. 

Screenshot_20240820_124311


yangchiu (approved):
LGTM

PR #2067, opened 2024-08-21
Test resilience handling for the last replica timeout

ejweber:

longhorn/longhorn#8711

Adds a robot test for longhorn/longhorn#8711.

Please feel free to take a close look at this one. I'm not super familiar with writing tests for the new framework, so I may have broken a convention.


khushboo-rancher (approved):
This looks good to me.
Do we need to skip this test for OS other than SLES?

cc: @yangchiu for help in review

ejweber:
> Do we need to skip this test for OS other than SLES?

No. The iptables command is executed inside an instance-manager pod, which is definitely running a suse-like OS. I intended the comment there as a way to communicate that you can't use the iptables functionality inside other arbitrary pods.

ejweber:
@mergify backport v1.7.x

PR #2068, opened 2024-08-22
ci: refactor robot test folder structure and generate junit.xml for qase sync

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/7412 https://github.com/longhorn/longhorn/issues/8853

refactor robot test folder structure and generate junit.xml for qase sync


khushboo-rancher (commented):
I think, node_disconnection, physical_node_reboot, single_replica_node_down can be moved in the negative folder. All the negative tests are supposed to run before any release.
Further negative folder can be categorized as below
- i. node_failure
- ii. stress
- iii. network_outage
- iv. cluster_instability
- etc

Do we need pre-release test? We will be running all the automated test cases before the release.
Do we want to rename the `regression` as `functional`, will it make more sense? WDYT?
In future, we might add performance, load testing etc.

yangchiu:
> I think, node_disconnection, physical_node_reboot, single_replica_node_down can be moved in the negative folder. 

Negative test cases are designed to be run repeatedly over an extended period to test resiliency. However `node_disconnection`, `physical_node_reboot`, `single_replica_node_down` only need to be run and verified once per pre-release version. They are fundamentally different.

> Do we need pre-release test? 

Could you suggest a suitable name? Or would you like to put them in `regression` folder?

> Do we want to rename the `regression` as `functional`, will it make more sense?

Traditionally, we refer to our daily tests as `regression`, like:

https://github.com/longhorn/longhorn/issues/8156
https://github.com/longhorn/longhorn/issues/6659
https://github.com/longhorn/longhorn/issues/6596
https://github.com/longhorn/longhorn/issues/6347
https://github.com/longhorn/longhorn/issues/5638
https://github.com/longhorn/longhorn/issues/5460
https://github.com/longhorn/longhorn/issues/5459

However, we don’t typically refer to them as `functional` tests.

So I'm uncertain about changing the naming. Can you help make a decision? Then all team members can follow the new naming and avoid confusion.

@khushboo-rancher

khushboo-rancher:
I see your point. `Regression` folder name seems good then.

Regarding `pre-release` tests, in automation we would like to run all the automated test cases during release. So, there is no point in keeping specific test cases in pre-release section. `node_disconnection`, `physical_node_reboot`, `single_replica_node_down` are also negative in nature, so we should ideally run them multiple times to test resiliency like other negative test cases.

Let's discuss this in QA meeting in detail.

yangchiu:
> Regarding `pre-release` tests, in automation we would like to run all the automated test cases during release. So, there is no point in keeping specific test cases in pre-release section. `node_disconnection`, `physical_node_reboot`, `single_replica_node_down` are also negative in nature, so we should ideally run them multiple times to test resiliency like other negative test cases.

I've removed `pre-release` folder and moved test cases to `negative`. Thank you!

khushboo-rancher (approved):

PR #2070, opened 2024-08-26
ci: sync automated test cases between github and qase and post test run for daily regression

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8853 https://github.com/longhorn/longhorn/issues/7412

sync automated test cases between github and qase and post test run for daily regression

### DIFF
--- a/pipelines/utilities/junit_to_qase.py
+++ b/pipelines/utilities/junit_to_qase.py
@@ -1,0 +1,274 @@
+#!/usr/bin/python
+import os
+import sys
+from urllib.parse import urlencode
+from datetime import date
+import requests
+from bs4 import BeautifulSoup
+
+
+def collect_test_results(test_cases):
+
+    test_results = []
+
+    for test_case in test_cases:
+
+        test_result = {
+            "case": "",
+            "case_id": "",
+            "suite": "",
+            "suite_id": "",
+            "skipped": "",
+            "failure": ""
+        }
+
+        test_suite = test_case.get("classname")
+        test_result['suite'] = test_suite.split(".")[-1]
+
+        test_result['case'] = test_case.get("name").split("[")[0]
+
+        skipped = test_case.findChildren("skipped", recursive=False)
+        if len(skipped) > 0:
+            test_result['skipped'] = skipped[0].get("message")
+
+        failure = test_case.findChildren("failure", recursive=False)
+        if len(failure) > 0:
+            test_result['failure'] = failure[0].get("message")
+            test_result['failure'] += '\n\n' + failure[0].get_text()
+
+        test_results.append(test_result)
+
+    return test_results
+
+
+def get_suite_id(suite_name):
+
+    print(f"getting suite {suite_name}")
+
+    query_string = urlencode({ "search": suite_name })
+    headers = {
+        "accept": "application/json",
+        "Token": qase_token
+    }
+    url = f"https://api.qase.io/v1/suite/{qase_project}?{query_string}"
+
+    resp = requests.get(url, headers=headers)
+    res = resp.json()
+    if res["result"]["count"] > 0:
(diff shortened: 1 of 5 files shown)


pipelines/utilities/junit_to_qase.py
```diff
@@ -0,0 +1,274 @@
...
+
+    print(f"adding missing case {case_name} under parent suite {parent_suite_id}")
+
+    if len(case_name) > 128:
+        case_name = case_name[:127]
+
+    payload = {
+        "title": case_name,
+        "suite_id": parent_suite_id,
+        "automation": 2
```
khushboo-rancher, line 126: There won't be any description of the test case which is getting added here?

pipelines/utilities/junit_to_qase.py
```diff
@@ -0,0 +1,274 @@
...
+    if res["result"]["count"] > 0:
+        id = res["result"]["entities"][0]["id"]
+        print(f"got suite {suite_name} with id {id}")
+        return id
+    else:
+        print(f"missed suite {suite_name}")
+        return None
+
+
+def add_missing_suite(suite_name):
```
khushboo-rancher, line 66: This will add the test suit based on the test execution, right? Have we considered that Pytest cases & few robot test cases are already added in Qase which is under certain test suit. `automation->e2e-Pytest` & `automation->e2e-robot`

khushboo-rancher (dismissed):

pipelines/utilities/junit_to_qase.py
yangchiu: No. There is no description section for a test case in a junit report.

khushboo-rancher: Ok, in that case, should we rely on syncing the test case from github to Qase like manual test case syncing?

pipelines/utilities/junit_to_qase.py
yangchiu: Yes, it takes into account the test cases that have been added to Qase. That's why it only adds "missing" items.

pipelines/utilities/junit_to_qase.py
yangchiu: We don't have a better approach for this right now, but we can figure it out in the future.

khushboo-rancher:
@yangchiu Looks like the test cases with suffix created by specific Jenkins parameter are getting added. E.g. test case running with NFS get the suffix [nfs] and test case running with s3 get suffix [S3], they are both getting added in the test repository in Qase. We have duplicated test cases now.

Screenshot 2024-08-30 at 1 02 00 PM

yangchiu:
> @yangchiu Looks like the test cases with suffix created by specific Jenkins parameter are getting added. E.g. test case running with NFS get the suffix [nfs] and test case running with s3 get suffix [S3], they are both getting added in the test repository in Qase. We have duplicated test cases now.
> 
This is expected since you tried to copy daily regression test results from a junit test report to a qase test run. Every test case must have a one to one mapping between junit and qase.

@khushboo-rancher

pipelines/utilities/junit_to_qase.py
```diff
@@ -0,0 +1,274 @@
...
+            "case": "",
+            "case_id": "",
+            "suite": "",
+            "suite_id": "",
+            "skipped": "",
+            "failure": ""
+        }
+
+        test_suite = test_case.get("classname")
+        test_result['suite'] = test_suite.split(".")[-1]
```
khushboo-rancher, line 26: Let's trim the test case name and eliminate anything after `[`. Will this solve the problem of adding duplicated test cases?

yangchiu: Updated!

khushboo-rancher (approved):

yangchiu:

PR #2077, opened 2024-08-30
test(robot): add replica rebuilding after engine image upgrade test

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9343

add replica rebuilding after engine image upgrade test


e2e/tests/regression/test_engine_image.robot
```diff
@@ -0,0 +1,32 @@
+*** Settings ***
+Documentation    Engine Image Test Cases
+
+Test Tags    regression    engine_image
```
khushboo-rancher, line 4: qq: Are we maintaining `core` test tag in robot also?

khushboo-rancher (approved):

khushboo-rancher:
@roger-ryao Would you like to take a look?

e2e/tests/regression/test_engine_image.robot
yangchiu: Yes, the `pytest` test cases tagged with `coretest` will retain the `coretest` tag after being migrated to `robot`.

PR #2079, opened 2024-09-02
ci: fix syntax error in e2e Jenkinsfile

yangchiu:

Issue N/A

fix syntax error in e2e Jenkinsfile


khushboo-rancher (commented):
Could you elaborate why we don't need to handle the upgrade test result separately now?

yangchiu:
> Could you please elaborate why we don't need to handle the upgrade test result separately now?

Running 2 stage upgrade test separately won't catch bugs like https://github.com/longhorn/longhorn/issues/9267#issuecomment-2303607612

cc @longhorn/qa

khushboo-rancher (approved):

PR #2080, opened 2024-09-03
test(robot): add test strict local volume disabled revision conuter by default

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9318

add test strict local volume disabled revision conuter by default

### DIFF
--- a/e2e/libs/replica/crd.py
+++ b/e2e/libs/replica/crd.py
@@ -10,11 +10,11 @@ class CRD(Base):
     def __init__(self):
         self.obj_api = client.CustomObjectsApi()
 
-    def get(self, volume_name, node_name):
+    def get(self, volume_name, node_name=None):
         label_selector = []
         if volume_name != "":
             label_selector.append(f"longhornvolume={volume_name}")
-        if node_name != "":
+        if node_name:
             label_selector.append(f"longhornnode={node_name}")
 
         replicas = self.obj_api.list_namespaced_custom_object(
@@ -24,7 +24,7 @@ class CRD(Base):
             plural="replicas",
             label_selector=",".join(label_selector)
         )
-        return replicas
+        return replicas["items"]
 
     def delete(self, volume_name, node_name):
         if volume_name == "" or node_name == "":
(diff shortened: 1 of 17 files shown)


khushboo-rancher (commented):
LGTM, few minor comments.

Also, do we want to add the test skeleton also for https://github.com/longhorn/longhorn/issues/9279#issuecomment-2307833776?

e2e/libs/replica/crd.py
```diff
@@ -33,10 +33,7 @@ def delete(self, volume_name, node_name):
             logging(
                 f"Deleting volume {volume_name} on node {node_name} replicas")
 
-        resp = self.get(volume_name, node_name)
-        assert resp != "", f"failed to get replicas"
```
roger-ryao, line 37: Regarding the removal of `assert resp != "", f"failed to get replicas"`, could you explain the main reasoning behind removing this step? I don't quite understand it.

e2e/libs/volume/crd.py
```diff
@@ -495,3 +497,7 @@ def upgrade_engine_image(self, volume_name, engine_image_name):
 
     def wait_for_engine_image_upgrade_completed(self, volume_name, engine_image_name):
         return Rest(self.node_exec).wait_for_engine_image_upgrade_completed(volume_name, engine_image_name)
+    def validate_volume_setting(self, volume_name, setting_name, value):
```
roger-ryao, line 500: Add an extra line of separation in the function.

e2e/libs/replica/crd.py
roger-ryao, line 37: Regarding the removal of `assert resp != "", f"failed to get replicas"`, could you explain the main reasoning behind removing this step? I don't quite understand it.

khushboo-rancher (approved):
LGTM from my side.

@roger-ryao has few questions.

khushboo-rancher:
@roger-ryao Could you take a look whenever get a chance?

roger-ryao (approved):
LGTM, but I noticed that @khushboo-rancher and I have similar questions.  If possible, could you also explain the reason for removing these two sections, so we can understand more?

https://github.com/longhorn/longhorn-tests/pull/2080#discussion_r1768129952
https://github.com/longhorn/longhorn-tests/pull/2080#discussion_r1761984542

PR #2081, opened 2024-09-03
test: skeleton for don't orphan processes when node not ready

ejweber:

longhorn/longhorn#6552

Skeleton for an automated test.


e2e/tests/regression/test_ha.robot
```diff
@@ -30,3 +30,13 @@ Disrupt Data Plane Traffic For Less Than Long Engine Replica Timeout
     And Wait for volume of statefulset 0 attached and degraded
     And Wait for volume of statefulset 0 healthy
     And Check statefulset 0 data in file data is intact
+
+Don't Orphan Processes When Node Not Ready
+    [Documentation]    Don't orphan processes when a node becomes not ready.
+    ...
+    ...                1. Create a volume of any size and a number of replicas equal to the number of nodes.
+    ...                2. Attach the volume to a node.
+    ...                3. Stop kubelet on the attached node long enough the node to become node ready and its instance
```
yangchiu, line 39: Should it be `become not ready`?

e2e/tests/regression/test_ha.robot
```diff
@@ -30,3 +30,13 @@ Disrupt Data Plane Traffic For Less Than Long Engine Replica Timeout
...
+    ...                   manager to have state unknown.
+    ...                4. Detach the volume.
+    ...                5. Verify that no related engine or replica process exists in any instance-manager.
```
yangchiu, line 42: Is it possible to provide more details on the command used to check the process and the expected output?

yangchiu (changes requested):

e2e/tests/regression/test_ha.robot
ejweber: Does the change I made in an amend make it more clear?

yangchiu: @ejweber I tried to follow the test steps, please help confirm if the following steps work as expected:

1. Create a volume with 3 replicas through UI
2. Create PV/PVC for this volume
3. Create a pod to attach and use this volume
4. The instance manager CR has status `instanceEngines` and `instanceReplicas`:
```
$ kubectl get -n longhorn-system lhim   instance-manager-4fd47ccee643a72f27d794b1bfb24b20 -oyaml
  status:
    apiMinVersion: 1
    apiVersion: 6
    currentState: running
    dataEngineStatus:
      v2:
        cpuMask: ""
    instanceEngines:
      test-1-e-0:
        spec:
          backendStoreDriver: ""
          dataEngine: v1
          name: test-1-e-0
        status:
          conditions:
            FilesystemReadOnly: false
          endpoint: ""
          errorMsg: ""
          listen: ""
          portEnd: 10000
          portStart: 10000
          resourceVersion: 0
          state: running
          targetPortEnd: 10000
          targetPortStart: 10000
          type: engine
    instanceReplicas:
      test-1-r-37128e19:
        spec:
          backendStoreDriver: ""
          dataEngine: v1
          name: test-1-r-37128e19
        status:
          conditions: null
          endpoint: ""
          errorMsg: ""
          listen: ""
          portEnd: 10010
          portStart: 10001
          resourceVersion: 0
          state: running
          targetPortEnd: 0
          targetPortStart: 0
          type: replica
```
5. Stop kubelet on the volume attached node by:
```
systemctl stop k3s-agent.service
```
6. Wait for the node not ready:
```
$ kubectl get nodes
NAME            STATUS     ROLES                  AGE   VERSION
ip-10-0-1-252   NotReady   <none>                 57m   v1.30.0+k3s1
ip-10-0-1-189   Ready      <none>                 56m   v1.30.0+k3s1
ip-10-0-1-53    Ready      <none>                 57m   v1.30.0+k3s1
ip-10-0-1-77    Ready      control-plane,master   57m   v1.30.0+k3s1
```
7. The instance manager CR becomes `error` instead of `unknown`:
```
$ kubectl get -n longhorn-system lhim -w
NAME                                                DATA ENGINE   STATE     TYPE   NODE            AGE
instance-manager-4fd47ccee643a72f27d794b1bfb24b20   v1            error     aio    ip-10-0-1-252   48m
instance-manager-b0ad49fa6ba72f4e19a51503d82d6330   v1            running   aio    ip-10-0-1-189   47m
instance-manager-a59fb876633905a381bf548807fb1b64   v1            running   aio    ip-10-0-1-53    48m
```
8. The instance manager pod gets stuck in `Terminating`:
```
 kubectl get pods -n longhorn-system -owide | grep instance-manager
instance-manager-a59fb876633905a381bf548807fb1b64   1/1     Running       0          55m   10.42.1.11   ip-10-0-1-53    <none>           <none>
instance-manager-b0ad49fa6ba72f4e19a51503d82d6330   1/1     Running       0          54m   10.42.3.16   ip-10-0-1-189   <none>           <none>
instance-manager-4fd47ccee643a72f27d794b1bfb24b20   1/1     Terminating   0          24m   10.42.0.15   ip-10-0-1-252   <none>           <none>
```
9. The volume gets stuck in `attaching`:
```
$ kubectl get volumes -n longhorn-system
NAME     DATA ENGINE   STATE       ROBUSTNESS   SCHEDULED   SIZE          NODE   AGE
test-1   v1            attaching   unknown                  21474836480          54m
```
10. Force delete the pod to detach the volume
11. The instance manager CR doesn't have status `instanceEngines` and `instanceReplicas`:
```
$ kubectl get -n longhorn-system -o yaml lhim | grep 'name:.*-e-' | sort | uniq -c
# empty
$ kubectl get -n longhorn-system -o yaml lhim | grep 'name:.*-r-' | sort | uniq -c
# empty
```

ejweber: Sorry @yangchiu, I added this additional wording to the skeleton in an amend for clarity:

```
    ...                3. Stop kubelet on the attached node long enough the node to become not ready and its instance
    ...                   manager to have state unknown.
    ...                   NOTE: The timing is important here. The instance manager reaches state unknown after
    ...                   approximately 30 seconds. Approximately 6 minutes later, Kubernetes evicts most pods on the
    ...                   node, including the instance manager pod, so the instance manager transitions to state error.
    ...                   The remaining steps must be executed before this happens.
```

It looks like in your example you executed the final steps long after the kubelet went down, but there is only a six minute window in which they make sense. For a bit of background, the problem, ultimately, is our unknown state. The instance-manager may come back from it. If it does, none of the processes running in it are killed as expected. Once the instance-manager reaches the error state, we will definitely restart its pod, so the issue doesn't exist.

ejweber: Also, to be clear, the fix hasn't been merged yet, so the steps in this skeleton are currently expected to fail according to https://github.com/longhorn/longhorn/issues/6552#issuecomment-2319177183. (The expected failure is different than the one you described above, but you should get the same failure by executing the post-kubelet failure steps in the "unknown window".

khushboo-rancher (approved):

khushboo-rancher:
Thanks @ejweber for adding the test case.
Thanks @yangchiu for running the test steps. I think, we can merge this and once the issue is ready for testing, we can once again verify the test steps.

yangchiu (approved):

yangchiu:
Verification passed on master-head (longhorn-manager ).

1. Create a volume of any size and a number of replicas equal to the number of nodes.
2. Attach the volume to a node.
3. Check there are `instanceEngines` and `instanceReplicas` in instance-manager CR:
```
$ kubectl get -n longhorn-system -o yaml lhim | grep 'name:.*-e-' | sort | uniq -c
      1           name: test-1-e-0
$ kubectl get -n longhorn-system -o yaml lhim | grep 'name:.*-r-' | sort | uniq -c
      1           name: test-1-r-2ad34fce
      1           name: test-1-r-3e0831c9
      1           name: test-1-r-461a6e67
```
4. Stop kubelet on the attached node, and wait for 30 seconds to make kubernetes notice that and change node status to `NotReady` or `Down`
5. Detach the volume.
6. Verify there is no `instanceEngines` or instanceReplicas in instance-manager CR:
```
$ kubectl get -n longhorn-system -o yaml lhim | grep 'name:.*-e-' | sort | uniq -c
$ kubectl get -n longhorn-system -o yaml lhim | grep 'name:.*-r-' | sort | uniq -c
```

PR #2086, opened 2024-09-06
ci: fix missing longhorn-system namespace during iscsi installation

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/8855

fix missing longhorn-system namespace during iscsi installation


roger-ryao (approved):
LGTM

PR #2091, opened 2024-09-11
test: fix unable to set default sc to a non-existing sc

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9391

fix unable to set default sc to a non-existing sc


roger-ryao (approved):
LTGM

yangchiu:
@mergify backport v1.7.x v1.6.x

PR #2108, opened 2024-09-19
feat(metric): add e2e test for volume filesystem readonly metrics

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8508

Export the metrics in the following form
- metric name: `longhorn_volume_file_system_read_only`
- type: gauge
- if the volume becomes read-only, it has metric value 1 with the corresponding labels
    - So you know which volume is in read-only mode now
    - You can sum up the metrics to see how many volumes are in read-only mode now 
```
# HELP longhorn_volume_file_system_read_only Volume whose mount point is in read-only mode
# TYPE longhorn_volume_file_system_read_only gauge
longhorn_volume_file_system_read_only{node="kworker1",pvc="nginx",pvc_namespace="default",volume="pvc-3a7a208b-c8c9-46a9-b2ce-4c5f740c8673"} 1
```

Note:
If Longhorn remounts the mount point to read-write successfully, the metrics will be cleaned up.
The metric is either 1 or no record for each volume

### DIFF
--- a/manager/integration/tests/test_metric.py
+++ b/manager/integration/tests/test_metric.py
@@ -3,15 +3,18 @@ import requests
 import time
 
 from collections import defaultdict
+from kubernetes.stream import stream
 from prometheus_client.parser import text_string_to_metric_families
 
-from common import client, core_api, volume_name  # NOQA
+from common import client, core_api, pod, volume_name  # NOQA
 
+from common import crash_engine_process_with_sigkill
 from common import delete_replica_processes
 from common import create_pv_for_volume
 from common import create_pvc_for_volume
 from common import create_snapshot
 from common import create_and_check_volume
+from common import create_and_wait_pod
 from common import generate_random_data
 from common import get_self_host_id
 from common import wait_for_volume_degraded
@@ -20,6 +23,7 @@ from common import wait_for_volume_detached_unknown
 from common import wait_for_volume_expansion
 from common import wait_for_volume_faulted
 from common import wait_for_volume_healthy
+from common import write_pod_volume_data
 from common import write_volume_data
 from common import set_node_scheduling
 from common import set_node_cordon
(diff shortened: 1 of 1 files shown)


manager/integration/tests/test_metric.py
```diff
@@ -54,9 +58,9 @@
 def get_metrics(core_api, metric_node_id): # NOQA
     pods = core_api.list_namespaced_pod(namespace=LONGHORN_NAMESPACE,
                                         label_selector="app=longhorn-manager")
-    for pod in pods.items:
-        if pod.spec.node_name == metric_node_id:
-            manager_ip = pod.status.pod_ip
+    for po in pods.items:
```
c3y1huang, line 61: Is there a specific reason for renaming this?

ChanYiLin: because I import the `pod()` function from common
```
from common import pod
```
so we can't use the same name here in the loop

ChanYiLin: I feel like the function name should be a verb to avoid this kind of conflict.

manager/integration/tests/test_metric.py
```diff
@@ -359,6 +377,58 @@ def test_volume_metrics(client, core_api, volume_name, pvc_namespace): # NOQA
...
+    When create a volume and attach to a pod
+    Then verify the volume is healthy
+
+    When crash the volumes engine process
+    Then verify the volume is healthy
+
+    When write the data to the pod
+    And issue sync command
+    Then has 1 metrics longhorn_volume_file_system_read_only with
+        volume label is the volume name
```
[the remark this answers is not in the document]
ChanYiLin: fixed. Thanks

ChanYiLin:
Hi @c3y1huang 
The PR has been updated. 
Thanks

c3y1huang (approved):

PR #2109, opened 2024-09-19
fix: attach volumes to eliminate duplicated replicas

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8573

https://github.com/longhorn/longhorn/issues/8573#issuecomment-2348405896
When the engine image is not stable, there is a chance that the volume changes the owner and creates duplicated replicas.
This might fail this e2e test because it checks the number of replicas before the volume is attached.

We should attach the volume and let Longhorn deletes duplicated replicas before starting the test.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -6494,3 +6494,14 @@ def check_backing_image_eviction_failed(name): # NOQA
         time.sleep(RETRY_INTERVAL)
 
     assert check
+
+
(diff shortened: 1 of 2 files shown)


ChanYiLin:
## E2E test PASSED

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7622/testReport/tests/test_ha/

yangchiu (approved):
LGTM

yangchiu:

PR #2116, opened 2024-09-24
test(robot): add drain cases from manual test cases

chriscchien:

Issue #9292

Implement test case in https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-down/node-drain-deletion/ to automation test cases

Verified on local
image

N/A


e2e/tests/negative/node_drain.robot
```diff
@@ -89,3 +93,142 @@ Force Drain Replica Node While Replica Rebuilding
...
+    And Set setting node-drain-policy to block-if-contains-last-replica
+    And Given Create volume 0 with    size=5Gi    numberOfReplicas=2
+    And Attach volume 0 to node 1
+    And Wait for volume 0 healthy
+    And Write data to volume 0
+    And Detach volume 0 from attached node
+    And Wait for volume 0 detached
+    And Power off node 1
+
+    When Force drain node 2 and wait for 90 second    dataEngine=${DATA_ENGINE}
```
yangchiu, line 193: Test steps of a robot test case should be self-explanatory, but the following steps are unclear in terms of what they are checking:
```
When Force drain node 2 and wait for 90 second dataEngine=${DATA_ENGINE}
And The drain process not completed
And Check instance-manager pod is running on node 2    dataEngine=${DATA_ENGINE}
```
Could these steps be made more readable to clarify what exactly is being validated?

e2e/libs/utility/utility.py
```diff
@@ -285,3 +285,25 @@ def get_name_suffix(*args):
         if arg:
             suffix += f"-{arg}"
     return suffix
+
+
+def check_popen_process_not_completed(process):
```
yangchiu, line 290: Is it necessary to introduce `popen`, `process.poll()`, `process.terminate()` and `process.communicate()`?

Can `subporcess.check_output()` with the `timeout` parameter accomplish the same goal?

ref: https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

e2e/keywords/k8s.resource
```diff
@@ -51,5 +52,63 @@ Force drain volume of ${workload_kind} ${workload_id} replica node
...
+        ${node_name} =    get_node_by_index    ${node_id}
+        force_drain_node    ${node_name}
+        wait_for_all_pods_evicted    ${node_name}
+    END
+
+Check node ${node_id} cordoned
+    ${node_name} =    get_node_by_index    ${node_id}
+    check_node_cordoned    ${node_name}
+
+Force drain node ${node_id} and wait for ${duration} second
```
yangchiu, line 90: Is it possible to use `subprocess.check_output()` with the `timeout` parameter to eliminate the need for the explicit `wait for ${duration} second`?

yangchiu (changes requested):

e2e/tests/negative/node_drain.robot
chriscchien: Updated,  removed `dataEngine=${DATA_ENGINE}` in `When Force drain node 2 and wait for 90 second    dataEngine=${DATA_ENGINE}`and use `data_engine = BuiltIn().get_variable_value("${DATA_ENGINE}")` to get data engine setting
```python
def get_instance_manager_on_node(node_name):
    data_engine = BuiltIn().get_variable_value("${DATA_ENGINE}")
    pods = get_all_pods_on_node(node_name)
    for pod in pods:
        labels = pod.metadata.labels
        if labels.get("longhorn.io/data-engine") == data_engine and \
           labels.get("longhorn.io/component") == "instance-manager":
            return pod.metadata.name
    return None
```


[Test result](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/1491/)

e2e/libs/utility/utility.py
chriscchien: Yes, in test case `Setting Allow Node Drain with the Last Healthy Replica protects the last healthy replica with Pod Disruption Budget (PDB)`,  fter change `node-drain-policy` from `block-if-contains-last-replica` to `always-allow`, it 
 need to validate if the drain process completed or not, currently I can not find a way to let `subporcess.check_output()` execute in background. that why I choosed `popen`

e2e/keywords/k8s.resource
chriscchien: In some scenario, the drain process need to running in background, using timeout seems force terminated the process.

e2e/libs/utility/utility.py
yangchiu: Is the following approach feasible?

```
def force_drain_node():
    subprocess.check_output(timeout=90)
```

```
Force drain node ${node_id} and expect failure
    Run Keyword And Expect Failure    force drain node
```

```
Force drain node ${node_id} and expect success
    force drain node
```

```
And Set setting node-drain-policy to block-if-contains-last-replica
...
...
When Force drain node 2 and expect fail
And Check instance-manager pod is running on node 2

When Set setting node-drain-policy to always-allow
And Force drain node 2 and expect success
And Check PDB not exist    
```

e2e/keywords/k8s.resource
yangchiu: Isn't `wait for ${duration} second` also an explicit timeout?

e2e/libs/utility/utility.py
chriscchien: Hi @yangchiu, updated and test [result](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/1492/)

chriscchien:
@longhorn/qa , this ticket updated and ready for revirew, thank you.

yangchiu (approved):
LGTM

PR #2135, opened 2024-10-11
fix(robot): Single Replica Node Down Deletion Policy do-nothing With RWO Volume Replica Locate On Replica Node

c3y1huang:

Issue longhorn/longhorn#9498

When the `node-down-pod-deletion-policy` is set to `do-nothing`, Longhorn doesn't actively delete the workload pod. In this case, reattaching may not happen, as this behavior will depend on Kubernetes' response. We should remove the check of the volume's attaching state and just ensure that the workload pod is in a terminating state, which is expected.

`None`

`None`

### DIFF
--- a/e2e/libs/workload/pod.py
+++ b/e2e/libs/workload/pod.py
@@ -122,7 +122,7 @@ def create_pod(manifest, is_wait_for_pod_running=False):
 def delete_pod(name, namespace='default'):
     core_api = client.CoreV1Api()
     try:
-        core_api.delete_namespaced_pod(name=name, namespace=namespace, delete_namespaced_pod=0)
+        core_api.delete_namespaced_pod(name=name, namespace=namespace, grace_period_seconds=0)
         wait_delete_pod(name)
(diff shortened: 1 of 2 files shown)


e2e/libs/workload/pod.py
```diff
@@ -122,7 +122,7 @@ def create_pod(manifest, is_wait_for_pod_running=False):
 def delete_pod(name, namespace='default'):
     core_api = client.CoreV1Api()
     try:
-        core_api.delete_namespaced_pod(name=name, namespace=namespace, delete_namespaced_pod=0)
+        core_api.delete_namespaced_pod(name=name, namespace=namespace, grace_period_seconds=0)
```
c3y1huang, line 125: ref: https://github.com/longhorn/longhorn-tests/pull/2134

derekbit (approved):
LGTM. The fix makes send to me.

derekbit:
@mergify backport v1.7.x

PR #2139, opened 2024-10-14
fix(robot): Single Replica Node Down Deletion Policy do-nothing With RWO Volume Replica Locate On Volume Node

c3y1huang:

Issue longhorn/longhorn#9622

Remove redundant intermediate check as when node down deletion policy is set to `do-nothing`, the workload pod deletion isn't handled by Longhorn.

`None`

`Name`


chriscchien (approved):
LGTM

c3y1huang:
@mergify backport v1.7.x

PR #2141, opened 2024-10-14
fix(robot): fix flaky test case Disrupt Data Plane Traffic For Less Than Long Engine Replica Timeout

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9276

fix flaky test case Disrupt Data Plane Traffic For Less Than Long Engine Replica Timeout


roger-ryao (approved):
LGTM

PR #2151, opened 2024-10-24
test(robot): pull backup from another longhorn

chriscchien:

Issue #9646

Implement manual test case [pull backup created by another Longhorn system](https://longhorn.github.io/longhorn-tests/manual/pre-release/backup-and-restore/pull-backup-created-by-another-longhorn/)

If running on local, below enviromnet need to be set up at first

example:
```
export LONGHORN_INSTALL_METHOD=manifest
export LONGHORN_REPO_BRANCH=v1.7.x
export CUSTOM_LONGHORN_MANAGER_IMAGE=longhornio/longhorn-manager:v1.7.2
export CUSTOM_LONGHORN_ENGINE_IMAGE=longhornio/longhorn-engine:v1.7.2
export CUSTOM_LONGHORN_INSTANCE_MANAGER_IMAGE=longhornio/longhorn-instance-manager:v1.7.2
export CUSTOM_LONGHORN_SHARE_MANAGER_IMAGE=longhornio/longhorn-share-manager:v1.7.2
export CUSTOM_LONGHORN_BACKING_IMAGE_MANAGER_IMAGE=longhornio/backing-image-manager:v1.7.2
export LONGHORN_STABLE_VERSION=v1.6.3
```

Tested on my local and Jenkins
image

Manual test case (in case manual case removed in the future)
```markdown
1. Prepare 2 k8s clusters: cluster A and cluster B.
2. Install previous version of Longhorn which doesn’t include this fix e.g v1.3.1, v1.2.5 on cluster A.
3. Install the release version of Longhorn on cluster B.
4. Set the same backup target on both cluster A and cluster B.
5. Create volume, write some data, and take backup on cluster A.
6. Wait for backup target polling update on cluster B.
7. Make sure the backup created by cluster A can be pulled on cluster B.
8. Restore the pulled backup and verify the data on cluster B.
9. Repeat the test with both clusters installed the release version of Longhorn.
```

Hi @mantissahz , this PR is related to backup, could you take a look on this? thank you.

## Release Notes

- **New Features**
  - Introduced keywords for managing Longhorn installations: 
    - Install and Uninstall Longhorn stable versions.
  - Added functionality for writing data to volumes and creating volumes from backups in different clusters.

- **Bug Fixes**
  - Enhanced error handling for backup metadata access to prevent unhandled exceptions.

- **Tests**
  - Added a new test case to validate backup restoration across different Longhorn installations.

- **Chores**
  - Updated installation scripts to support stable version installations and improved logging during the process.
  - Modified test scripts to include the new environment variable for stable version testing.

### DIFF
--- a/e2e/utilities/longhorn-install.sh
+++ b/e2e/utilities/longhorn-install.sh
@@ -13,6 +13,8 @@ source ../pipelines/utilities/longhorn_manifest.sh
 # create and clean tmpdir
 TMPDIR="/tmp/longhorn"
 LONGHORN_NAMESPACE="longhorn-system"
+LONGHORN_REPO_DIR="${TMPDIR}/longhorn"
+LONGHORN_REPO_URI=${LONGHORN_REPO_URI:-"https://github.com/longhorn/longhorn.git"}
 mkdir -p ${TMPDIR}
 rm -rf "${TMPDIR}/"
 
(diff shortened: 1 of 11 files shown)


e2e/utilities/longhorn-install.sh
```diff
@@ -23,19 +25,46 @@ install_longhorn_by_chart(){
...
+  if [[ "${LONGHORN_INSTALL_METHOD}" == "helm" ]]; then
+    install_longhorn_stable_by_chart
+  elif [[ "${LONGHORN_INSTALL_METHOD}" == "manifest" ]]; then
+    install_longhorn_stable_by_manifest
+  fi
+  setup_longhorn_ui_nodeport
 }
 
-install_longhorn
+if [[ $# -gt 0 ]]; then
```
yangchiu, line 66: Could you elaborate on why this is needed?

yangchiu (changes requested):

e2e/utilities/longhorn-install.sh
chriscchien: @yangchiu, From file libs/longhorn_deploy/base.py of this PR, a new statement has been added to handle the following scenarios:
- longhorn-install.sh is run with no arguments, the `install_longhorn` function is called which is the default design as before.
- longhorn-install.sh is run with an argument, the function named by the argument is executed, in this test case, `install_longhorn_stable_version` will be executed to perform steps for keyword `Then Install Longhorn stable version`
```python
def install_longhorn(self, is_stable_version=False):
        current_path=os.getcwd()
        full_path = os.path.join(current_path, LONGHORN_INSTALL_SCRIPT_PATH)

        if is_stable_version is True:
            cmd = ['bash', full_path, LONGHORN_INSTALL_STABLE_SHELL_FUNCTION]
        else:
            cmd = ['bash', full_path]
```

yangchiu: @chriscchien Why do we need to pass the function name from robot test code? What would be the drawback of letting the shell script handle it, like:
```
if [[ -n "${LONGHORN_STABLE_VERSION}" ]]; then
 install_longhorn_stable_version
else
  install_longhorn
fi
```

chriscchien: @yangchiu, test case `Uninstallation Checks` also uses `longhorn-install.sh` to install Longhorn test version. If the script use environment variable to determin which version to install, this will result in `Uninstallation Checks` and `Pull backup created by another Longhorn system` can not running on the same environment.

yangchiu: What about the following?
```
IS_INSTALL_STABLE_VERSION="${1:-false}"
if [[ "${IS_INSTALL_STABLE_VERSION}" == "true" ]]; then
  # do something
fi
```
Do you think it's more readable?

chriscchien: Current code, run function name as first argument if the argument exist
```bash
./longhorn-install.sh install_longhorn_stable_version
```
suggestion (if I understand correctly)
```bash
./longhorn-install.sh true
```

Just think if enhance comment for more clear realize the scope as below is acceptable? if not, I will use the suggestion, thank you.
```bash
# Check if an argument is passed
if [[ $# -gt 0 ]]; then
    # Run the function specified in the first argument, 
    # ex: ./longhorn-install.sh install_longhorn_stable_version
    $1 
else
    # Run the default install_longhorn function if no argument is provided
    install_longhorn
fi
```

yangchiu: How about the following:
```
IS_INSTALL_STABLE_VERSION=true ./longhorn-install.sh 
```
Can the shell script get the variable?

chriscchien: Updated and tested.

yangchiu (approved):
LGTM

PR #2195, opened 2024-12-16
test(robot): fix migration confirmation after migration node down test case for v2 volumes

yangchiu:

Issue N/A

Fix migration confirmation after migration node down test case for v2 volumes.
Since v2 volumes delete stopped replicas on down nodes, replica name checks will fail for v2 volumes.

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated the migration confirmation test case to streamline the process by removing unnecessary steps.


chriscchien (approved):
LGTM

PR #2208, opened 2024-12-18
test(robot): add node down during migration test cases

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9905 https://github.com/longhorn/longhorn/issues/8817 https://github.com/longhorn/longhorn/issues/8906

add node down during migration test cases

## Summary by CodeRabbit

- **New Features**
	- Introduced four new test cases for negative scenarios in volume migration, enhancing the testing framework's coverage.
		- Migration Rollback After Migration Node Down
		- Migration Confirmation After Original Node Down
		- Migration Rollback After Original Node Down
		- Migration Confirmation After Migration Node Down

### DIFF
--- a/e2e/tests/negative/live_migration.robot
+++ b/e2e/tests/negative/live_migration.robot
@@ -37,3 +37,70 @@ Migration Confirmation After Migration Node Down
     Then Wait for volume 0 to migrate to node 1
     And Wait for volume 0 healthy
     And Check volume 0 data is intact
+
+Migration Rollback After Migration Node Down
+    Given Create volume 0 with    migratable=True    accessMode=RWX    dataEngine=${DATA_ENGINE}
+    And Attach volume 0 to node 0
+    And Wait for volume 0 healthy
+    And Write data to volume 0
+
+    And Attach volume 0 to node 1
+    And Wait for volume 0 migration to be ready
+
+    # power off migration node
+    When Power off node 1
+    # migration rollback by detaching from the migration node
+    And Detach volume 0 from node 1
+
+    # migration rollback succeed
+    Then Wait for volume 0 to stay on node 0
+    And Wait for volume 0 degraded
+    And Check volume 0 data is intact
+
+Migration Confirmation After Original Node Down
+    Given Create volume 0 with    migratable=True    accessMode=RWX    dataEngine=${DATA_ENGINE}
+    And Attach volume 0 to node 0
+    And Wait for volume 0 healthy
+    And Write data to volume 0
+
+    And Attach volume 0 to node 1
+    And Wait for volume 0 migration to be ready
+
+    # power off original node
+    When Power off node 0
+    # migration confirmation by detaching from the original node
+    And Detach volume 0 from node 0
+
+    # migration is stuck until the Kubernetes pod eviction controller decides to
+    # terminate the instance-manager pod that was running on the original node.
+    # then Longhorn detaches the volume and cleanly reattaches it to the migration node.
+    Then Wait for volume 0 to migrate to node 1
+    And Wait for volume 0 degraded
+    And Check volume 0 data is intact
+
+Migration Rollback After Original Node Down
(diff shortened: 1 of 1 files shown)


e2e/tests/negative/live_migration.robot
```diff
@@ -37,3 +37,70 @@ Migration Confirmation After Migration Node Down
...
+    And Wait for volume 0 healthy
+    And Write data to volume 0
+
+    And Attach volume 0 to node 1
+    And Wait for volume 0 migration to be ready
+
+    # power off original node
+    When Power off node 0
+    # migration confirmation by detaching from the original node
+    And Detach volume 0 from node 0
```
khushboo-rancher, line 72: Do we need to detach explicitly from node 0? Should we just wait for the instance-manager to terminate?

e2e/tests/negative/live_migration.robot
```diff
@@ -37,3 +37,70 @@ Migration Confirmation After Migration Node Down
...
+    And Detach volume 0 from node 1
+
+    # migration is stuck until the Kubernetes pod eviction controller decides to
+    # terminate the instance-manager pod that was running on the original node.
+    # then Longhorn detaches the volume and attempts to cleanly reattach it to the original node,
+    # but it is stuck in attaching until the node comes back.
+    Then Check volume 0 kept in attaching
+
+    # power on original node
+    When Power on off nodes
```
khushboo-rancher, line 102: Was the original node restarted at this point? Why are powering off the node 1 too?

e2e/tests/negative/live_migration.robot
yangchiu: The test case is `Migration Confirmation After Original Node Down`, so we should detach the original node to `Confirm the Migration`.

If no action was taken after the original node went down, it seems to be another test case, called `Original Node Down After Migration Ready`. 
And I just tested this scenario: after the migration was ready, powered off the original node, and then did nothing. Eventually the volume was clearly detached:
```
$ kubectl get volumes -n longhorn-system pvc-8bf2ae3a-be4d-4110-b9e0-a9a124db8095 -oyaml -w | grep -i nodeid
  migrationNodeID: ""
  nodeID: ""
  currentMigrationNodeID: ""
  currentNodeID: ""
  pendingNodeID: ""
```
Even after powering on the original node again, the volume remains in `detached` state permanently.

[supportbundle_ef8a6972-0c68-48f9-bd45-1e575e519d95_2024-12-19T00-29-41Z.zip](https://github.com/user-attachments/files/18190700/supportbundle_ef8a6972-0c68-48f9-bd45-1e575e519d95_2024-12-19T00-29-41Z.zip)

If we need this test case, it needs @derekbit and @PhanLe1010 to confirm whether this is the expected behavior first.

e2e/tests/negative/live_migration.robot
yangchiu: > Was the original node restarted at this point? 

Yes, powered off node is restarted at this point. In this case, it's the original node that was powered off in the previous step.

> Why are powering off the node 1 too?

No, node 1 won't be powered off or on in this test case.

e2e/tests/negative/live_migration.robot
PhanLe1010: Yes, this is expected behavior. The volume will remain in detached until CSI flow or user detach the volume from the original node. This is the new design that when volume crash, migration will be stop, volume remain in detach state waiting for the user/csi to make decision about what is the only node Longhorn attach to. The reason behind this is that volume already crash so we don't need live migration anymore. This reduces the risk of unnecessary migration and chaotic. 

Ref https://github.com/longhorn/longhorn/issues/8735#issuecomment-2177309258
Manual test case is updated at https://github.com/longhorn/longhorn-tests/pull/1948

khushboo-rancher (approved):

e2e/tests/negative/live_migration.robot
yangchiu: I'll add this test case in the next PR.

yangchiu:
@mergify backport v1.8.x

PR #2211, opened 2024-12-19
fix: fix volume size smaller than bi size issue

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10017

we added a new webhook rule which rejects creating a volume which size is smaller than the backing image
update the volume size with larger size.


ChanYiLin:
## Test

- `-vs test_backing_image.py`: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7848/console
- whole: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7847/console

derekbit (approved):

roger-ryao:
@mergify backport v1.8.x

PR #2225, opened 2024-12-24
test: fix test_system_backup_and_restore_volume_with_backingimage

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10052

fix test_system_backup_and_restore_volume_with_backingimage

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved timing and sequence handling in the system backup and restore test case.

- **New Features**
	- Introduced a new method to wait for volume creation during system restore testing.

### DIFF
--- a/manager/integration/tests/test_system_backup_restore.py
+++ b/manager/integration/tests/test_system_backup_restore.py
@@ -8,6 +8,7 @@ from common import core_api # NOQA
 from common import check_volume_data
 from common import cleanup_volume
 from common import create_and_check_volume
+from common import wait_for_volume_creation
 from common import create_backup
 from common import find_backup_volume
 from common import get_self_host_id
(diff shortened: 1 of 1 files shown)


yangchiu:
Test results: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7964/

manager/integration/tests/test_system_backup_restore.py
```diff
@@ -199,7 +200,7 @@ def test_system_backup_and_restore_volume_with_backingimage(client, core_api, vo
     check_backing_image_disk_map_status(client, BACKING_IMAGE_NAME, 3, "ready")
 
     restored_volume = client.by_id_volume(volume_name)
-    wait_for_volume_restoration_completed(client, volume_name)
+    wait_for_volume_creation(client, volume_name)
```
mantissahz, line 203: Does it mean we don't know whether the volume is restored?
Cause we don't check if the data is correct after restoring the system backup.

yangchiu: Yes, you're right. The test case is missing a data integrity check. I'll create a separate ticket for that. Thank you!

yangchiu: Using https://github.com/longhorn/longhorn/issues/10052 to track.

PR #2289, opened 2025-01-23
ci: expose os and k8s version info in gke pipeline

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10285

expose os and k8s version info in gke pipeline


pipelines/gke/scripts/longhorn-setup.sh
```diff
@@ -26,6 +26,32 @@ set_kubeconfig_envvar(){
...
+      mountPath: /mnt/host
+  volumes:
+  - name: host
+    hostPath:
+      path: /
+      type: Directory
+EOF
+  kubectl wait --for=condition=Ready pod/print-os-release --timeout=60s
+  kubectl exec -it print-os-release -- cat /mnt/host/etc/os-release
+  kubectl delete pod/print-os-release
```
khushboo-rancher, line 52: Can we get the OS info from any other pod which is already running instead of creating a pod?

khushboo-rancher (changes requested):
LGTM in general, just one concern about creating a pod just to get the OS info

pipelines/gke/scripts/longhorn-setup.sh
yangchiu: Probably not, because the other existing pods don’t mount the host’s `/etc` folder.

khushboo-rancher (approved):

PR #2308, opened 2025-02-06
test(integration): support v2 volume test on test_provisioner.py

chriscchien:

Issue #9760

support v2 volume test on test_provisioner.py

Test result
- [v1 volume](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8422/testReport/)
- [v2 volume](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8423/)

N/A

## Summary by CodeRabbit

- **Tests**
  - Updated provisioning validations to dynamically adjust disk configuration checks based on the current engine version.
  - Enhanced storage provisioning tests by adding an engine-specific parameter and categorizing them for clearer test coverage.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -1684,7 +1684,13 @@ def node_default_tags():
 
     tag_mappings = {}
     for tags, node in zip(DEFAULT_TAGS, nodes):
-        assert len(node.disks) == 1
+        if DATA_ENGINE == "v2":
+            # if the v2 data engine is enabled, both a file system disk
+            # and a block disk will coexist. This is because a v2 backing image
+            # requires a file system disk to function.
+            assert len(node.disks) == 2
+        else:
+            assert len(node.disks) == 1
 
         update_disks = get_update_disks(node.disks)
         update_disks[list(update_disks)[0]].tags = tags["disk"]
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
```diff
@@ -1684,7 +1684,10 @@ def node_default_tags():
 
     tag_mappings = {}
     for tags, node in zip(DEFAULT_TAGS, nodes):
-        assert len(node.disks) == 1
+        if DATA_ENGINE == "v2":
+            assert len(node.disks) == 2
+        else:
+            assert len(node.disks) == 1
```
derekbit, line 1690: Could you leave a comment explaining why the len(node.disks) is 2 while enabling v2 data engine?

chriscchien: Updated, thank you.

manager/integration/tests/common.py
```diff
@@ -1684,7 +1684,12 @@ def node_default_tags():
 
     tag_mappings = {}
     for tags, node in zip(DEFAULT_TAGS, nodes):
-        assert len(node.disks) == 1
+        if DATA_ENGINE == "v2":
+            # if v2 data engine is enabled, both a file system disk
```
derekbit, line 1688: Can you elaborate on the purpose? For v2 backing image, right?

chriscchien: Yes, I have updated the comment to below, because a v2 volume use block disk but a v2 backing image can not function without file system disk so the two types of disks need to coexist if we want to test v2 volume with v2 backing image.
```
            # if the v2 data engine is enabled, both a file system disk
            # and a block disk will coexist. This is because a v2 backing image
            # requires a file system disk to function.
```

derekbit (approved):
LGTM

roger-ryao:
@mergify backport v1.8.x

PR #2313, opened 2025-02-17
test(robot): add Crash Instance Manager While Workload Pod Is Starting test case

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10431

add Crash Instance Manager While Workload Pod Is Starting test case

## Summary by CodeRabbit

- **New Features**
  - Enhanced secret management with new capabilities for crypto secret creation and cleanup.
  - Expanded storage configuration to support encryption by integrating secret associations.
  - Improved workload management with options for dynamic pod deletion based on workload identifiers.
  - Introduced a new method to wait for workload pods to be in the container creating state.

- **Tests**
  - Added a negative test suite that simulates instance manager crashes during workload initialization to validate resilience and data integrity.


chriscchien (approved):
LGTM

yangchiu:
@mergify backport v1.8.x

PR #2325, opened 2025-02-24
fix(setting): cleanup the toleration setting

mantissahz:

Issue # longhorn/longhorn#10475

The setting `toleration` might be re-applied with volumes detached after an hour (reconciliation interval) if the toleration is not cleaned up, and it will cause the API `list_backupVolume` to fail to find a valid node with the default engine image.

Regression test:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8237/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8238/

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the process for resetting default settings by consolidating multiple configuration flows into one efficient approach.
- **Tests**
  - Updated settings validation tests with enhanced parameter consistency.
  - Improved cleanup process in tests to ensure configuration environment resets reliably after execution.

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -1119,27 +1119,16 @@ def update_settings_via_configmap(core_api, client, setting_names, setting_value
                                          body=configmap_body)
 
     def reset_default_settings():
-        if configmap_name == DEFAULT_SETTING_CONFIGMAP_NAME:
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

PR #2366, opened 2025-03-14
test(robot): add test case Test Snapshot Purge Rejection While Migration

derekbit:

Issue 
Longhorn/longhorn#10495
Longhorn/longhorn#10496

Reject snapshot purge when volume is being live migration. If snapshot purge is executed while live migrating a volume, the disk files will be changed. The new engine and replicas are unable to be aware of the change and cannot list the new disk files, resulting in ERR replicas.

## Summary by CodeRabbit

- **Tests**
  - Added a negative test to confirm that volume snapshot purge operations fail gracefully when error conditions occur.
  - Introduced a new scenario to verify that purge requests are properly rejected during active migrations, ensuring robust volume management safeguards.
  - Added a new keyword for testing failure cases in snapshot purges.

### DIFF
--- a/e2e/keywords/snapshot.resource
+++ b/e2e/keywords/snapshot.resource
@@ -37,6 +37,10 @@ Purge volume ${volume_id} snapshot
     ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
     purge_snapshot    ${volume_name}
 
+Purge volume ${volume_id} snapshot should fail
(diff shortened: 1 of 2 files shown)


derekbit:
@c3y1huang @yangchiu Could you help review the robot test case and see if there are bad smells? Thank you.

yangchiu (approved):
LGTM

yangchiu:
@mergify backport v1.8.x

PR #2371, opened 2025-03-15
test(robot): Add case `The node the restore volume attached to is down`

roger-ryao:
longhorn/longhorn#9865
Implement manual test case [The node the restore volume attached to is down](https://longhorn.github.io/longhorn-tests/manual/pre-release/node-not-ready/node-down/restore-volume-node-down/) in robot

Implement manual test case `The node the restore volume attached to is down` in robot

https://github.com/longhorn/longhorn/issues/7211
- PR test log : 
  - 
[9865-log.tar.gz](https://github.com/user-attachments/files/19261850/9865-log.tar.gz)

- Here's how to execute test case: ` ./run.sh -i 'longhorn-9865'`

Screenshot_20250315_204951

## Summary by CodeRabbit

- **Documentation**
  - Expanded the user manual with a new precondition section linking to CoreDNS best practices.

- **New Features**
  - Introduced enhanced node and volume management capabilities, including improved controls for powering off nodes, checking their health, creating volumes from backups, and attaching volumes with optional encryption and attachment waiting parameters.

- **Tests**
  - Added a restoration test scenario to validate volume recovery when a node is down for both encrypted and non-encrypted volumes.

### DIFF
--- a/e2e/libs/k8s/k8s.py
+++ b/e2e/libs/k8s/k8s.py
@@ -80,6 +80,15 @@ def wait_all_pods_evicted(node_name):
 
     assert evicted, 'failed to evict pods'
 
+def is_node_ready(node_name):
+    api = client.CoreV1Api()
+    node = api.read_node(node_name)
+    conditions = node.status.conditions
+    for condition in conditions:
+        if condition.type == "Ready" and condition.status == "True":
+            return True
+    return False
+
 def check_node_cordoned(node_name):
     api = client.CoreV1Api()
     node = api.read_node(node_name)
--- a/e2e/libs/volume/crd.py
+++ b/e2e/libs/volume/crd.py
@@ -66,11 +66,12 @@ class CRD(Base):
                 plural="volumes",
                 body=body
             )
-            if not Standby:
-                self.wait_for_volume_state(volume_name, "detached")
-            else:
+            if fromBackup or Standby:
                 self.wait_for_volume_state(volume_name, "attached")
-            self.wait_for_restore_required_status(volume_name, Standby)
+                self.wait_for_restore_required_status(volume_name, True)
+            else:
+                self.wait_for_volume_state(volume_name, "detached")
+                self.wait_for_restore_required_status(volume_name, False)
             volume = self.get(volume_name)
             assert volume['metadata']['name'] == volume_name, f"expect volume name is {volume_name}, but it's {volume['metadata']['name']}"
             assert volume['spec']['size'] == size, f"expect volume size is {size}, but it's {volume['spec']['size']}"
--- a/e2e/libs/volume/crd.py
+++ b/e2e/libs/volume/crd.py
@@ -26,7 +26,7 @@ class CRD(Base):
         self.retry_count, self.retry_interval = get_retry_count_and_interval()
         self.engine = Engine()
 
-    def create(self, volume_name, size, numberOfReplicas, frontend, migratable, dataLocality, accessMode, dataEngine, backingImage, Standby, fromBackup):
+    def create(self, volume_name, size, numberOfReplicas, frontend, migratable, dataLocality, accessMode, dataEngine, backingImage, Standby, fromBackup, encrypted):
         size_suffix = size[-2:]
         size_number = size[:-2]
         size_unit = MEBIBYTE if size_suffix == "Mi" else GIBIBYTE
@@ -42,6 +42,7 @@ class CRD(Base):
                 }
             },
             "spec": {
+                "encrypted": encrypted,
                 "frontend": frontend,
                 "replicaAutoBalance": "ignored",
                 "size": size,
@@ -81,6 +82,7 @@ class CRD(Base):
             assert volume['spec']['backingImage'] == backingImage, f"expect volume backingImage is {backingImage}, but it's {volume['spec']['backingImage']}"
             assert volume['spec']['Standby'] == Standby, f"expect volume Standby is {Standby}, but it's {volume['spec']['Standby']}"
             assert volume['spec']['fromBackup'] == fromBackup, f"expect volume fromBackup is {fromBackup}, but it's {volume['spec']['fromBackup']}"
+            assert volume['spec']['encrypted'] == encrypted, f"expect volume encrypted is {encrypted}, but it's {volume['spec']['encrypted']}"
         except ApiException as e:
             logging(e)
 
--- a/e2e/tests/negative/test_restore_volume_node_reboot.robot
+++ b/e2e/tests/negative/test_restore_volume_node_reboot.robot
@@ -1,0 +1,68 @@
+*** Settings ***
+Documentation    The node the restore volume attached to is down
+...              - Issue: https://github.com/longhorn/longhorn/issues/1355
+...              - Issue: https://github.com/longhorn/longhorn/issues/9865
+...              - Test the restoration process of a Longhorn volume when the attached node goes down.
+...              - Includes verification for both encrypted and non-encrypted volumes.
+
+Test Tags    manual    negative    longhorn-9865
+
+Resource    ../keywords/variables.resource
+Resource    ../keywords/common.resource
+Resource    ../keywords/longhorn.resource
+Resource    ../keywords/host.resource
+Resource    ../keywords/volume.resource
+Resource    ../keywords/snapshot.resource
+Resource    ../keywords/backup.resource
+Resource    ../keywords/k8s.resource
+
+
+Test Setup    Set test environment
+Test Teardown    Cleanup test resources
+Test Template    Restore volume attached node is down
+
+*** Keywords ***
+Restore volume attached node is down
+    [Arguments]    ${description}    ${encrypted}
+    [Documentation]    Test the behavior of restoring a Longhorn volume when the attached node goes down.
+    ...                Arguments:
+    ...                - ${description}: Description of the test case.
+    ...                - ${encrypted}: Boolean to specify if the volume is encrypted (true/false).  
+    Given Create volume 0 with    dataEngine=${DATA_ENGINE}    encrypted=${encrypted}
+    And Attach volume 0
+    And Wait for volume 0 healthy
+    And Write data 0 to volume 0
+    And Create backup 0 for volume 0
+    And Verify backup list contains backup 0 of volume 0
+
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        When Create volume 1 from backup 0 of volume 0  dataEngine=${DATA_ENGINE}  encrypted=${encrypted}
+        And Wait for volume 1 restoration from backup 0 of volume 0 start
+        And Power off volume 1 volume node without waiting
+        Then Wait for volume 1 restoration to complete
+
+        When Attach volume 1 to healthy node
+        Then Wait for volume 1 degraded
+        And Check volume 1 data is backup 0 of volume 0
+        And Detach volume 1 from attached node
+        And Delete volume 1
+        And Power on off nodes
+    END
+
+    And Detach volume 0 from attached node
+    And Delete volume 0
+
+*** Test Cases ***    DESCRIPTION    ENCRYPTED
+The Restore Volume Attached Node is Down
+    [Documentation]    - Manual Test Plan
+    ...                - Given Create a volume 0 with 3 replicas.
+    ...                - And Attach the volume 0 to a node.
+    ...                - And Write data 0 to the volume 0.
+    ...                - Then Create a backup of the volume 0.
+    ...                - When Restore the backup to volume 1.
+    ...                - And During the restoration, power off the node where the volume 1 is attached.
+    ...                - And Wait for the restoration to complete.
+    ...                - Then Attach the restored volume 1 to a healthy node.
+    ...                - Then Verify the restored data matches the backup.
+    Test Restore Non-encrypted Volume Node Down    false
+    Test Restore Encrypted Volume Node Down    true
--- a/docs/content/manual/pre-release/node-not-ready/node-down/restore-volume-node-down.md
+++ b/docs/content/manual/pre-release/node-not-ready/node-down/restore-volume-node-down.md
@@ -2,6 +2,9 @@
 title: "[#1355](https://github.com/longhorn/longhorn/issues/1355) The node the restore volume attached to is down"
 ---
 
+## Precondition
+- https://longhorn.io/docs/1.8.1/best-practices/#coredns-setup
+
 ### Case 1:
 1. Create a backup.
 2. Restore the above backup.
--- a/e2e/keywords/host.resource
+++ b/e2e/keywords/host.resource
@@ -25,6 +25,34 @@ Reboot node ${idx}
 Restart all worker nodes
     reboot_all_worker_nodes
 
+Power off volume ${volume_id} volume node
+    ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
+    ${node_name} =    get_volume_node    ${volume_name}
+    ${powered_off_node} =    Set Variable    ${node_name}
+    Append to list    ${powered_off_nodes}    ${powered_off_node}
+    power_off_node_by_name    ${node_name}
+
+Power off volume ${volume_id} replica node
+    ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
+    ${node_name} =    get_replica_node    ${volume_name}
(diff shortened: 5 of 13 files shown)


e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,51 @@
...
+Restore volume attached node is down
+    [Arguments]    ${description}    ${encrypted}
+    Given Create volume 0 with    dataEngine=${DATA_ENGINE}    encrypted=${encrypted}
+    And Attach volume 0
+    And Wait for volume 0 healthy
+    And Write data 0 to volume 0
+    Then Volume 0 backup 0 should be able to create
+
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
+        When Create volume 1 from backup 0 of volume 0  dataEngine=${DATA_ENGINE}  wait_attached=True  encrypted=${encrypted}
```
yangchiu, line 34: Could I know why we need to explicitly wait it to be attached during the volume creation? Isn't it also attached in the next step `And Wait for volume 1 restoration from backup 0 of volume 0 start`?

roger-ryao: Due to https://github.com/longhorn/longhorn-tests/blob/master/e2e/libs/volume/crd.py#L68-L71, our original  logic was that, except for cases where the DR volume's `Standby = true` and `self.wait_for_volume_state(volume_name, "attached")`, all other volumes should be in the state of `self.wait_for_volume_state(volume_name, "detached")`.

Under this design, the next step `And Wait for volume 1 restoration from backup 0 of volume 0 start` cannot obtain the `attached` state for `volume 1`, which causes this step to fail.

yangchiu: Could it be this?
```
if Standby or fromBackup:
    self.wait_for_volume_state(volume_name, "attached")
```

e2e/libs/volume/crd.py
```diff
@@ -65,11 +66,20 @@ def create(self, volume_name, size, numberOfReplicas, frontend, migratable, data
...
-            self.wait_for_restore_required_status(volume_name, Standby)
+            else:
+                if fromBackup:
+                    if wait_attached:
+                        self.wait_for_volume_state(volume_name, "attached")
+                    else:
+                        self.wait_for_volume_state(volume_name, "detached")
+                else:
+                    self.wait_for_volume_state(volume_name, "detached")
+            if wait_attached:
```
yangchiu, line 79: Why does whether to `wait_for_restore_required_status` depends on `wait_attached`? Should it depend on if it's restored from a backup?

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,51 @@
...
+
+        Then Detach volume 1 from attached node
+        And Delete volume 1
+        And Power on off nodes
+    END
+
+*** Test Cases ***
+The Restore Volume attached node is down
+    Non-Encrypted Volume    false
+    Encrypted Volume    true
```
yangchiu, line 51: The whole test case steps are wrapped in a keyword, and the test case itself doesn't contain much readable content. Could people have no idea what the test case is doing just by reading it? Does that really make sense?

e2e/libs/volume/crd.py
```diff
@@ -65,11 +66,20 @@ def create(self, volume_name, size, numberOfReplicas, frontend, migratable, data
...
-            else:
+            if Standby:
                 self.wait_for_volume_state(volume_name, "attached")
-            self.wait_for_restore_required_status(volume_name, Standby)
+            else:
+                if fromBackup:
+                    if wait_attached:
+                        self.wait_for_volume_state(volume_name, "attached")
+                    else:
+                        self.wait_for_volume_state(volume_name, "detached")
```
yangchiu, line 76: If `fromBackup` is not empty and `wait_attached` is `False`,
from https://github.com/longhorn/longhorn-tests/pull/2371/files#r1998210595, does it mean it will implicitly wait for volume restoration complete and the following `And Wait for volume 1 restoration from backup 0 of volume 0 start` will fail?

roger-ryao: Yes, because when executing `create volume from the backup`, the operation will wait until the volume is created and its state becomes `detached`. Therefore, during the step `Wait for volume 1 restoration from backup 0 of volume 0 start`, it will failed to obtain the restoration state of `volume 1`

e2e/libs/volume/crd.py
roger-ryao: When we create a volume from a backup, `restoreRequired` will be `true`. Once the restore is complete and the volume is `detached`, `restoreRequired` will become `false`. If we use `fromBackup` as the condition, it may cause failures in other existing cases that involve `creating volume from backup`.

e2e/libs/volume/crd.py
yangchiu: So why do we need the logic for `fromBackup is not empty and wait_attached is False`?

e2e/libs/volume/crd.py
yangchiu: Could I know which existing cases will fail?

roger-ryao: > Why does whether to `wait_for_restore_required_status` depends on `wait_attached`? Should it depend on if it's restored from a backup?

Hi @yangchiu 
    I understand your point, and I will update it.

e2e/libs/volume/crd.py
roger-ryao: After discussing with @yangchiu we concluded that when creating a volume from  a backup, subsequent keywords must include a check for the `volume wait for detach` action.  For example:https://github.com/longhorn/longhorn-tests/blob/master/e2e/tests/negative/pull_backup_from_another_longhorn.robot#L65-L66

e2e/libs/k8s/k8s.py
```diff
@@ -80,6 +80,31 @@ def wait_all_pods_evicted(node_name):
...
+def get_node_status(node_name):
+    api = client.CoreV1Api()
+    node = api.read_node(node_name)
+    conditions = node.status.conditions
+    for condition in conditions:
+        if condition.type == "Ready" and condition.status == "True":
+            return True
+    return False
+
+def get_node_status_debug(node_name):
```
yangchiu, line 92: Is this function required? If so, could we try to merge the debug messages into the `get_node_status` function?

yangchiu (changes requested):

e2e/libs/k8s/k8s.py
roger-ryao: > Is this function required? If so, could we try to merge the debug messages into the `get_node_status` function?

This function was used during the development process, but I forgot to remove it. I will remove it.

e2e/libs/volume/crd.py
```diff
@@ -26,7 +26,7 @@ def __init__(self):
         self.retry_count, self.retry_interval = get_retry_count_and_interval()
         self.engine = Engine()
 
-    def create(self, volume_name, size, numberOfReplicas, frontend, migratable, dataLocality, accessMode, dataEngine, backingImage, Standby, fromBackup):
+    def create(self, volume_name, size, numberOfReplicas, frontend, migratable, dataLocality, accessMode, dataEngine, backingImage, Standby, fromBackup, encrypted, wait_attached):
```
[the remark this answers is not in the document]
roger-ryao: I will remove it.

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
...
+        And Wait for volume 1 degraded
+        Then Check volume 1 data is backup 0 of volume 0
+
+        Then Detach volume 1 from attached node
+        And Delete volume 1
+        And Power on off nodes
+    END
+
+    Then Detach volume 0 from attached node
+    And Delete volume 0
```
roger-ryao, line 53: The reason for adding these [two lines](https://github.com/longhorn/longhorn-tests/pull/2371/files#diff-97eb5b1444f916ec1e626003844002ca0c7c96bdc4422e41a7e987131d819e58R52-R53) is that when using the [Test Template: Restore volume attached node is down](https://github.com/longhorn/longhorn-tests/pull/2371/files#diff-97eb5b1444f916ec1e626003844002ca0c7c96bdc4422e41a7e987131d819e58R22), the initially created volume, which is a `non-encrypted volume`, continues to be used. 

As a result, when executing the test case for [Test Restore Encrypted Volume Node Down](https://github.com/longhorn/longhorn-tests/pull/2371/files#diff-97eb5b1444f916ec1e626003844002ca0c7c96bdc4422e41a7e987131d819e58R68), the non-encrypted volume will be backed up and restored as `an encrypted volume`.

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
...
+    [Arguments]    ${description}    ${encrypted}
+    [Documentation]    Test the behavior of restoring a Longhorn volume when the attached node goes down.
+    ...                Arguments:
+    ...                - ${description}: Description of the test case.
+    ...                - ${encrypted}: Boolean to specify if the volume is encrypted (true/false).  
+    Given Create volume 0 with    dataEngine=${DATA_ENGINE}    encrypted=${encrypted}
+    And Attach volume 0
+    And Wait for volume 0 healthy
+    And Write data 0 to volume 0
+    Then Volume 0 backup 0 should be able to create
```
c3y1huang, line 35: This step feels a bit awkward. I understand that the keyword includes some validations, but it doesn't quite fit naturally.

Suggest to unfold the keywords for clarity.
```suggestion
    And Create backup 0 for volume 0
    And Verify backup list contains backup 0 of volume 0
```

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
...
+        And Power off volume 1 volume node without waiting
+        And Wait for volume 1 restoration to complete
+
+        Then Attach volume 1 to healthy node
+        And Wait for volume 1 degraded
+        Then Check volume 1 data is backup 0 of volume 0
+
+        Then Detach volume 1 from attached node
+        And Delete volume 1
+        And Power on off nodes
```
c3y1huang, line 49:
```suggestion
        When Create volume 1 from backup 0 of volume 0  dataEngine=${DATA_ENGINE}  encrypted=${encrypted}
        And Wait for volume 1 restoration from backup 0 of volume 0 start
        And Power off volume 1 volume node without waiting
        Then Wait for volume 1 restoration to complete
        
        When Attach volume 1 to healthy node
        Then Wait for volume 1 degraded
        And Check volume 1 data is backup 0 of volume 0
        And Detach volume 1 from attached node
        And Delete volume 1
        And Power on off nodes
```
If a new action requires validation, it should be introduced with `When`.
There's no need for multiple `Then` steps if the following steps are all test expectations (`Then`) of the preceding action (`When`).

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
...
+    END
+
+    Then Detach volume 0 from attached node
```
c3y1huang, line 52:
```suggestion
    And Detach volume 0 from attached node
```

e2e/libs/k8s/k8s.py
```diff
@@ -80,6 +80,15 @@ def wait_all_pods_evicted(node_name):
 
     assert evicted, 'failed to evict pods'
 
+def get_node_status(node_name):
```
c3y1huang, line 83:
```suggestion
def is_node_ready(node_name):
```
The function name should reflect its purpose, which here checks for the node `Ready` condition. A typical naming convention for boolean-returning functions is to prefix with `is_`, `has_`, etc.

roger-ryao: Update

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
...
+        And Check volume 1 data is backup 0 of volume 0
+        And Detach volume 1 from attached node
+        And Delete volume 1
+        And Power on off nodes
+    END
+
+    And Detach volume 0 from attached node
+    And Delete volume 0
+
+*** Test Cases ***
```
c3y1huang, line 55:
```suggestion
*** Test Cases ***	DESCRIPTION	ENCRYPTED
```

c3y1huang, line 55:
```suggestion
*** Test Cases ***	DESCRIPTION	ENCRYPTED
```

c3y1huang (commented):
DCO failed.

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
...
+    ...                - And Attach the volume 0 to a node.
+    ...                - And Write data 0 to the volume 0.
+    ...                - Then Create a backup of the volume 0.
+    ...                - When Restore the backup to volume 1.
+    ...                - And During the restoration, power off the node where the volume 1 is attached.
+    ...                - And Wait for the restoration to complete.
+    ...                - Then Attach the restored volume 1 to a healthy node.
+    ...                - Then Verify the restored data matches the backup.
+    Test Restore Non-encrypted Volume Node Down    encrypted=${false}
+    Test Restore Encrypted Volume Node Down    encrypted=${true}
```
c3y1huang, line 68:
```suggestion
    Test Restore Non-encrypted Volume Node Down    false
    Test Restore Encrypted Volume Node Down    true
```

roger-ryao:
> DCO failed.

Update

c3y1huang (dismissed):

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
...
+Restore volume attached node is down
+    [Arguments]    ${description}    ${encrypted}
+    [Documentation]    Test the behavior of restoring a Longhorn volume when the attached node goes down.
+    ...                Arguments:
+    ...                - ${description}: Description of the test case.
+    ...                - ${encrypted}: Boolean to specify if the volume is encrypted (true/false).  
+    Given Create volume 0 with    dataEngine=${DATA_ENGINE}    encrypted=${encrypted}
+    And Attach volume 0
+    And Wait for volume 0 healthy
+    And Write data 0 to volume 0
```
derekbit, line 34: I'm a bit confused. What does `Write data 0` mean?

derekbit (commented):
In general, LGTM. Only two questions

e2e/tests/negative/test_restore_volume_node_reboot.robot
yangchiu: In scenarios where random data is written to a volume multiple times and the data checksum is validated after a snapshot reversion or a backup restoration, it can be helpful to assign an index to each data write to keep its checksum. You can find test case `Test V1 Snapshot` demonstrates this scenario.

e2e/tests/negative/test_restore_volume_node_reboot.robot
```diff
@@ -0,0 +1,68 @@
+*** Settings ***
+Documentation    The node the restore volume attached to is down
+...              - Issue: https://github.com/longhorn/longhorn/issues/1355
+...              - Issue: https://github.com/longhorn/longhorn/issues/9865
+...              - Test the restoration process of a Longhorn volume when the attached node goes down.
+...              - Includes verification for both encrypted and non-encrypted volumes.
+
+Test Tags    manual  longhorn-9865
```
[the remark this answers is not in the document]
roger-ryao: Because in the past, when I implemented manual test cases, I had the habit of adding tags such as `manual` and the related issue number., I have added `negative` tag.

e2e/tests/negative/test_restore_volume_node_reboot.robot
derekbit: Thanks for the explanation.
nit: we can rename to a better keyword to reflect what its meaning later.

derekbit (dismissed):
LGTM. Good implementation and review. Also learn a lot from the comments.

yangchiu (approved):

yangchiu:
@mergify backport v1.8.x

PR #2393, opened 2025-03-25
test(robot): trim leading and trailing double quotes only when both are present

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10591

trim leading and trailing double quotes only when both are present


roger-ryao (approved):
LGTM

PR #2405, opened 2025-03-31
test(pytest): refine test_settings.py by count instance manager based on data engine

chriscchien:

Refine `test_settings.py` by counting instance manager based on the data engine

Test result:  https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8389/testReport/tests/test_settings/

### DIFF
--- a/manager/integration/tests/test_settings.py
+++ b/manager/integration/tests/test_settings.py
@@ -80,7 +80,9 @@ def check_workload_update(core_api, apps_api, count):  # NOQA
 
     im_pod_list = core_api.list_namespaced_pod(
         LONGHORN_NAMESPACE,
-        label_selector="longhorn.io/component=instance-manager").items
+        label_selector="longhorn.io/component=instance-manager, \
+                        longhorn.io/data-engine={}".format(DATA_ENGINE)).items
+
     if len(im_pod_list) != count:
         return False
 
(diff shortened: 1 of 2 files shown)


manager/integration/tests/test_settings.py
```diff
@@ -484,11 +486,12 @@ def guaranteed_instance_manager_cpu_setting_check(  # NOQA
         for im in instance_managers:
             pod = core_api.read_namespaced_pod(name=im.name,
                                                namespace=LONGHORN_NAMESPACE)
-            if cpu_val:
-                assert (pod.spec.containers[0].resources.requests['cpu'] ==
-                        cpu_val)
-            else:
-                assert (not pod.spec.containers[0].resources.requests)
+            if pod.metadata.labels["longhorn.io/data-engine"] == "v1":
```
yangchiu, line 489: Does it mean test case `test_instance_manager_cpu_reservation` doesn't support `v2`? Since it doesn't have the `@pytest.mark.v2_volume_test` annotation, will the test case still be executed when `RUN_V2_TEST` is enabled? When will it fail without this modification?

chriscchien: Hi @yangchiu,
- The test case not support v2 volume test yet: [ref](https://github.com/longhorn/longhorn/issues/9760#issuecomment-2582142211)
- Enabling `RUN_V2_TEST` in pytest use v2 data engine for testing. A test case marked with mark `v2_volume_test` indicates support for v2 data engine testing.
- Failure of test_instance_manager_cpu_reservation https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/1192/testReport/junit/tests/test_settings/test_instance_manager_cpu_reservation/

yangchiu: If the test case doesn't support v2, why does it fail because of v2 instance manager?

If `RUN_V2_TEST` is disabled so v2 data engine is not used for testing, why does the test case failed because of `pod.metadata.labels["longhorn.io/data-engine"] == "v2"`?

yangchiu (approved):

chriscchien:
@mergify backport v1.8.x

PR #2420, opened 2025-04-10
feat(v2 snapshot): add skeleton for data integrity

DamiaSan:
Issue https://github.com/longhorn/longhorn/issues/9023

Added a skeleton for a new test to verify data integrity between replicas when executing a volume snapshot during I/O over the volume.


e2e/tests/regression/test_v2.robot
```diff
@@ -116,3 +116,23 @@ V2 Volume Should Cleanup Resources When Instance Manager Is Deleted
...
+    ...                4. Make a snapshot of the volume.
+    ...                5. Stop data writing.
+    ...                6. For every replica of the volume, execute the following operations:
+    ...                 6.1 Enter into the pod where the spdk_tgt containing the replica data is running.
+    ...                 6.2 Using go-spdk-helper, find the snapshot created above with lvol command.
+    ...                 6.3 Using go-spdk-helper, start exposing the snapshot bdev with expose command.
+    ...                 6.4 Using go-spdk-helper, connect to the exposed snapshot bdev with nvmecli command.
+    ...                 6.5 Compute a checksum of the block device created with the command above.
+    ...                 6.6 Using go-spdk-helper, stop exposing the snapshot bdev with expose command.
+    ...                 6.1 Exit from the pod.
```
derekbit, line 136: This is an e2e test case. Shouldn't use go-spdk-helper that is transparent to users.
Please use Longhorn CR resources instead.

DamiaSan: In the pre-ready-for-testing checklis of the issue https://github.com/longhorn/longhorn/issues/2187 I found a link to the PR https://github.com/longhorn/longhorn-tests/pull/1946, but the skeleton inside of it doesn't contain any indication for what I need to do. Moreover, in the old manual test case for that issue, https://longhorn.github.io/longhorn-tests/manual/pre-release/basic-operations/snapshot-while-writing-data/, there is simply a  "_Create another snapshot/backup and verify the data_."
How can I compute the checksum of one replica's data?

derekbit: > How can I compute the checksum of one replica's data?

I think you should check the data's checksum after reverting a snapshot.

DamiaSan: > I think you should check the data's checksum after reverting a snapshot.

I want to compare data of different replicas.
If I revert a snapshot, the data are read from all the available replicas, because SPDK RAID1 balance the reading load between all available replicas.


Looking at the test you mentioned, for example in the file `test_snapshot.py` there is the test `check_snapshot_checksums_and_change_timestamps` , which open the snapshot file and read the checksum from metadata. This operation is analogue to the one I used in the skeleton, i.e export the snapshot via nvmf, connect to it via nvme cli and calculate the checksum.

derekbit: > If I revert a snapshot, the data are read from all the available replicas, because SPDK RAID1 balance the reading load between all available replicas.
> Looking at the tests you mentioned, for example in the file test_snapshot.py there is the test check_snapshot_checksums_and_change_timestamps , which open the snapshot file and read the checksum from metadata. This operation is analogue to the one I used in the skeleton, i.e export the snapshot via nvmf, connect to it via nvme cli and calculate the checksum.

If so, why need the complicated steps? why not just check snapshot.status.checksum? 

One more question, could you validate the backup and restore as well?
If you write a big file like 2G file, create a backup and restore it back. If the snapshot is not synced, the restored volume will have wrong data. This is the purpose of the sync. The test can validate it as well.

DamiaSan: > If so, why need the complicated steps? why not just check snapshot.status.checksum?
> 

First of all, the engine API `SnapshotHashStatus` has been implementd on `longhorn-spdk-engine` only 2 days ago, so when I wrote down the test skeleton didn't know about its existence. Until then, the field `SnapshotChecksum` remained only internal in `longhorn-spdk-engine`.
Now that we have this API implemented, looking at the way `snapshot.status.checksum` is calculated in `determineChecksumFromHashStatus` in `longhorn-manager`, there is a vote mechanism to determine the checksum, so we can rely on this variable also if we cannot access directly to the checksum of each replica, which would have been safer for me.
However I will update the skeleton, thanks.

> 
> Another question, after cmputing a checksum of the block device, how to make sure data is really sync?
> 

SPDK doesn't have any cache, so if we use a NVMe disk (and then SPDK NVMe driver), when computing the checksum data are read from the disk. I will put this in the skeleton too, thanks.

> --
> 
> One more question, could you validate the backup and restore as well? If you write a big file like 2G file, create a backup and restore it back. If the snapshot is not synced, the restored volume will have wrong data. This is the purpose of the sync. The test can validate it as well, or create another test case for it.

The same as above, if we use a NVMe disk and the checksum is correctly set, which means that the checksums of different replicas are equals, we can be sure that data are synced on disk.
BTW, with my development I have simply changed the sync mechanism, a sync command was present in the code also before, so we should already have this kind of test.

derekbit: > First of all, the engine API SnapshotHashStatus has been implementd on longhorn-spdk-engine only 2 days ago, so when I wrote down the test skeleton didn't know about its existence.

Hmm, I think snapshot.status.checksum is already existing for a long while, so you can use it and make the test simpler instead. Because this is a skeleton, you don't need to worry about the implementation. 

> SPDK doesn't have any cache, so if we use a NVMe disk (and then SPDK NVMe driver), when computing the checksum data are read from the disk. I will put this in the skeleton too, thanks.

I mean I don't understand the purpose of the checksum calculation. The steps seem not clear and not complete. Calculate the checksum and then?

> BTW, with my development I have simply changed the sync mechanism, a sync command was present in the code also before, so we should already have this kind of test.

We don't have this kind of test in e2e at the moment. It's only in the regression tests. Since we're migrating to e2e, this is a good opportunity to add an e2e test case in the PR.

Again, this is not a high-priority task. You can address the high-priority tasks first.

DamiaSan: > > First of all, the engine API SnapshotHashStatus has been implementd on longhorn-spdk-engine only 2 days ago, so when I wrote down the test skeleton didn't know about its existence.
> 
> Hmm, I think snapshot.status.checksum is already existing for a long while, so you can use it and make the test simpler instead. Because this is a skeleton, you don't need to worry about the implementation.
> 

I have already made the test simpler, please look at my last changes.
`snapshot.status.checksum` was present also before, but for v2 engine it wasn't set.

> > SPDK doesn't have any cache, so if we use a NVMe disk (and then SPDK NVMe driver), when computing the checksum data are read from the disk. I will put this in the skeleton too, thanks.
> 
> I mean I don't understand the purpose of the checksum calculation. The steps seem not clear and not complete. Calculate the checksum and then?
> 

If checksum of different replica's data are equal, it means that these data are the same and so that the suspend operation has correctly been executed. Because with my development I have changed a `sync` command with a `dmsetup suspend` and then a `dmsetup resume`.

> > BTW, with my development I have simply changed the sync mechanism, a sync command was present in the code also before, so we should already have this kind of test.
> 
> We don't have this kind of test in e2e at the moment. It's only in the regression tests. Since we're migrating to e2e, this is a good opportunity to add an e2e test case in the PR.
> 
> Again, this is not a high-priority task. You can address the high-priority tasks first.

The issue this test is related to, and whose PR is blocked by this test definition, is marked as `priority/0`, which means high priority. What should I do?

derekbit: test case is managed by https://github.com/longhorn/longhorn/issues/9707 rather than https://github.com/longhorn/longhorn/issues/9023.  https://github.com/longhorn/longhorn/issues/9707 is not milestoned.

e2e/tests/regression/test_v2.robot
```diff
@@ -116,3 +116,17 @@ V2 Volume Should Cleanup Resources When Instance Manager Is Deleted
...
+    [Tags]    robot:skip
+    [Documentation]    V2 Volume Snapshot Data Integrity Check
+    ...
+    ...                1. Install Longhorn with v2 engine enabled on nodes with an available nvme disk, i.e. not the system disk.    
+    ...                2. Create a volume of any size and more than one replica.
+    ...                3. Attach the volume to a node.
+    ...                4. Start intensive data writing over the volume.
+    ...                5. Make a snapshot of the volume while data writing is ongoing.
+    ...                6. Stop data writing.
+    ...                6. Wait snapshot checksum to be set.
```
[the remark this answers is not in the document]
DamiaSan: Hi @shuo-wu , could you help us trying to understand the reason why the snapshot checksum remains empty? I want to use it to ensure data integrity between different replicas. Any idea? Thanks.
cc @derekbit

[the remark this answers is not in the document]
derekbit: @DamiaSan No rush. You can check the codes and figure this out later. The skeleton is not a priority for v1.9.0.

cc @yangchiu

PR #2424, opened 2025-04-15
test(robot): add a workload test case for offline rebuilding

mantissahz:

Issue # longhorn/longhorn#8443

### DIFF
--- a/e2e/tests/negative/replica_rebuilding.robot
+++ b/e2e/tests/negative/replica_rebuilding.robot
@@ -66,6 +66,26 @@ Reboot Replica Node While Replica Rebuilding
         And Check volume 0 data is intact
     END
 
+Reboot Replica Node While Offline Replica Rebuilding
+    [Documentation]    Reboot a replica node during offline replica rebuilding is in progress.
+    ...
+    ...                Issue: https://github.com/longhorn/longhorn/issues/8443
+    Given Create volume 0 with    size=5Gi    numberOfReplicas=3    dataEngine=${DATA_ENGINE}
+    And Attach volume 0
+    And Wait for volume 0 healthy
+    And Write data to volume 0
+    And Write 4 GB data to volume 0
+    And Detach volume 0
+    And Wait for volume 0 detached
+
+    When Delete volume 0 replica on node 0
+    Then Enable volume 0 offline replica rebuilding
+    And Wait until volume 0 replica rebuilding started on node 0
+
+    When Reboot volume 0 replica node
+    And Wait for volume 0 detached
+    And Volume 0 should have 3 replicas when detached
+
 Delete Replicas One By One After The Volume Is Healthy
     Given Create storageclass longhorn-test with    dataEngine=${DATA_ENGINE}
     And Create persistentvolumeclaim 0 using RWO volume with longhorn-test storageclass
--- a/e2e/tests/regression/test_replica.robot
+++ b/e2e/tests/regression/test_replica.robot
@@ -70,3 +70,36 @@ Test Offline Replica Rebuilding
     And Wait for volume 0 detached
     And Volume 0 should have 3 replicas when detached
     And Set setting offline-replica-rebuilding to false
+
+Test Preempt Offline Replica Rebuilding By A Workload
+    [Tags]    coretest    offline-rebuilding
+    [Documentation]    Test preempt offline replica rebuilding by a workload.
+    ...
+    ...                Issue: https://github.com/longhorn/longhorn/issues/8443
+    Given Create volume 0 with    size=6Gi    numberOfReplicas=3    dataEngine=${DATA_ENGINE}
+    And Attach volume 0
+    And Wait for volume 0 healthy
+    And Detach volume 0
+    And Wait for volume 0 detached
+    And Create deployment 0 with volume 0
+    And Wait for volume 0 healthy
+    And Write 5120 MB data to file data.txt in deployment 0
+    And Scale down deployment 0 to detach volume
+    And Wait for volume 0 detached
+
+    # enable offline replica rebuilding.
+    When Delete volume 0 replica on node 0
+    And Enable volume 0 offline replica rebuilding
+    # wait for offline rebuilding to start.
+    Then Wait until volume 0 replica rebuilding started on node 0
+    And Volume 0 should have 3 replicas running
+
+    # scale up the workload and it will preempt the offline rebuilding.
+    When Scale up deployment 0 to attach volume
+    # the volume might be detched or not so we only wait for volume becomes healthy.
+    # we should try to check if offline rebuilding will be canceled and online rebuilding takes over.
+    # then the volume becomes healthy without offline rebuilding.
+    And Wait for volume 0 healthy
+    # write some data with workload pod to check if the workload works.
+    And Write 64 MB data to file data.txt in deployment 0
+    Then Check deployment 0 data in file data.txt is intact
--- a/e2e/keywords/deployment.resource
+++ b/e2e/keywords/deployment.resource
@@ -21,6 +21,14 @@ Create deployment ${deployment_id} with persistentvolumeclaim ${claim_id} with m
     ${node_count} =    Get length  ${node_names}
     create_deployment    ${deployment_name}    ${claim_name}   ${node_count}
 
+Create deployment ${deployment_id} with volume ${volume_id}
+    ${volume_id}=  Convert To String  ${volume_id}
+    Create persistentvolume for volume ${volume_id}
+    Create persistentvolumeclaim for volume ${volume_id}
+    ${deployment_name}=    generate_name_with_suffix    deployment    ${deployment_id}
+    ${pvc_name}=    generate_name_with_suffix    volume    ${volume_id}
+    create_deployment    ${deployment_name}   ${pvc_name}
+
 Delete deployment ${deployment_id}
     ${deployment_name} =    generate_name_with_suffix    deployment    ${deployment_id}
     delete_deployment    ${deployment_name}
--- a/e2e/libs/volume/rest.py
+++ b/e2e/libs/volume/rest.py
@@ -476,4 +476,4 @@ class Rest(Base):
 
     def update_offline_replica_rebuild(self, volume_name, rebuild_type):
         volume = self.get(volume_name)
-        volume.offlineRebuilding(OfflineRebuild=rebuild_type)
+        volume.offlineReplicaRebuilding(OfflineRebuilding=rebuild_type)
--- a/e2e/tests/negative/test_backup_listing.robot
+++ b/e2e/tests/negative/test_backup_listing.robot
@@ -47,14 +47,6 @@ Create volume ${volume_id} from ${workload_kind} ${workload_id} volume random ba
     create_volume   ${volume_name}    size=3Gi    numberOfReplicas=3    fromBackup=${backup_url}    dataEngine=${DATA_ENGINE}
     Set Test Variable    ${random_backup_id}    ${backup_id}
 
-Create deployment ${deployment_id} with volume ${volume_id}
-    ${volume_id}=  Convert To String  ${volume_id}
-    Create persistentvolume for volume ${volume_id}
-    Create persistentvolumeclaim for volume ${volume_id}
-    ${deployment_name}=    generate_name_with_suffix    deployment    ${deployment_id}
-    ${pvc_name}=    generate_name_with_suffix    volume    ${volume_id}
-    create_deployment    ${deployment_name}   ${pvc_name}
-
 Check volume ${volume_id} data is ${workload_kind} ${workload_id} volume backup ${backup_id}
     ${workload_name}=   generate_name_with_suffix    ${workload_kind}    ${workload_id}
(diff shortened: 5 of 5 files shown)


mantissahz:
Hi @yangchiu, @roger-ryao, 
Could you review this test case when you are available?

e2e/tests/regression/test_replica.robot
```diff
@@ -69,4 +78,28 @@ Test Offline Replica Rebuilding
...
+    And Wait for volume 0 detached
+    Then Create deployment 0 with volume 0
+    And Wait for volume 0 healthy
+    Then Write 5120 MB data to file data.txt in deployment 0
+    And Scale down deployment 0 to detach volume
+    And Wait for volume 0 detached
+
+    When Delete volume 0 replica on node 0
+    Then Enable volume 0 offline replica rebuilding
+    And Wait until volume 0 replica rebuilding started on node 0
```
yangchiu, line 101: Could we check the volume is detached and has 3 replicas like the above test case to ensure the replica is rebuilt?

e2e/tests/regression/test_replica.robot
```diff
@@ -69,4 +78,28 @@ Test Offline Replica Rebuilding
...
+    And Scale up deployment 0 to attach volume
+    And Wait for volume 0 healthy
+    And Write 512 MB data to file data.txt in deployment 0
```
yangchiu, line 104: Could we also check the data integrity here?

e2e/tests/regression/test_replica.robot
```diff
@@ -69,4 +78,28 @@ Test Offline Replica Rebuilding
...
+    And Delete deployment 0
```
yangchiu, line 105: Is this manual resource cleanup required?

yangchiu (changes requested):

e2e/tests/regression/test_replica.robot
mantissahz: Here we test if the workload could preempt the offline rebuilding process, so we need to ensure the replica is still rebuilding.

yangchiu: What will happen here? Will the ongoing rebuilding be affected? Will it stop and be automatically retriggered?

mantissahz: It's like the test plan in the [offline volume rebuilding LEP](https://github.com/longhorn/longhorn/blob/0a83092479c522f5c71dcca382f60f6b7759b8ca/enhancements/20250407-volume-offline-rebuilding.md#test-plan)
```txt
- The CSI attaching request should preempt the offline rebuilding process:
  1. Create a workload with Longhorn volume having 3 replicas in a 3 worker nodes cluster
  2. Write some data to the volume.
  3. Scale down the workload to detach the volume.
  4. Delete a replica of the volume.
  5. Enable the offline rebuilding by the API `volume.offlineReplicaRebuilding`.
  6. When the rebuilding is in progress, scale up the workload.
  7. Check if the offline rebuilding process is canceled.
  8. Check if the workload is working well.
```

What will happen here is:
1. Offline rebuilding is in progress.
2. The workload starts, and offline rebuilding will be canceled (will not be automatically retriggered).
3. The workload works, and online rebuilding takes over

e2e/tests/regression/test_replica.robot
mantissahz: Yes, added a check for data integrity. Thanks.

e2e/tests/regression/test_replica.robot
mantissahz: Removed. Thanks.

e2e/tests/regression/test_replica.robot
yangchiu: If so, is it possible to add steps that:
```
And Wait until volume 0 replica rebuilding stopped on node 0
And Wait until volume 0 replica rebuilding started on node 0
```
to clarify the behavior in the test case? Would these steps be able to catch the timing reliably without causing the test to fail?"

mantissahz:
```robot
And Wait until volume 0 replica rebuilding stopped on node 0
And Wait until volume 0 replica rebuilding started on node 0
```
No, we can not add these steps because the volume might not be detached when the workload preempts the offline rebuilding, and detaching the volume and then attaching the volume might be faster than the detection logic.

yangchiu: Can we add some comments between the test steps to clarify the behavior? Since a robot test case should serve as documentation or a specification that helps we understand the software’s functionality.

If some test steps are not feasible to implement right now, at least we can add some comments or todos and revisit them for improvement in the future.

e2e/tests/regression/test_replica.robot
```diff
@@ -14,6 +14,15 @@ Resource    ../keywords/workload.resource
...
 
+*** Keywords ***
+Create deployment ${deployment_id} with volume ${volume_id}
+    ${volume_id}=  Convert To String  ${volume_id}
+    Create persistentvolume for volume ${volume_id}
+    Create persistentvolumeclaim for volume ${volume_id}
+    ${deployment_name}=    generate_name_with_suffix    deployment    ${deployment_id}
+    ${pvc_name}=    generate_name_with_suffix    volume    ${volume_id}
+    create_deployment    ${deployment_name}   ${pvc_name}
+
```
roger-ryao, line 25: The keyword `Create deployment ${deployment_id} with volume ${volume_id}` appears in two files:
1. [test_replica.robot](https://github.com/longhorn/longhorn-tests/blob/b431dafb10250af0f595061fc63b35e270fcfe1c/e2e/tests/regression/test_replica.robot#L18-L24)
2. [test_backup_listing.robot](https://github.com/longhorn/longhorn-tests/blob/b431dafb10250af0f595061fc63b35e270fcfe1c/e2e/tests/negative/test_backup_listing.robot#L50-L56)

Since these keywords are identical, they should be moved to `../keywords/deployment.resource` to avoid code duplication.

mantissahz:
```robot
*** Keywords ***
Create deployment ${deployment_id} with volume ${volume_id}
    ${volume_id}=  Convert To String  ${volume_id}
    Create persistentvolume for volume ${volume_id}
    Create persistentvolumeclaim for volume ${volume_id}
    ${deployment_name}=    generate_name_with_suffix    deployment    ${deployment_id}
    ${pvc_name}=    generate_name_with_suffix    volume    ${volume_id}
    create_deployment    ${deployment_name}   ${pvc_name}
```
I did not move this keyword to the resource file because the `Create deployment ${deployment_id} ...` contains the keywords from `volume.resource` and `deployment.resource`.
Does it allow for including another resource in a resource?

roger-ryao: I have tested it with your PR, and it looks good without any errors.

test report :  
[log.tar.gz](https://github.com/user-attachments/files/19860801/log.tar.gz)

Screenshot_20250423_123223

e2e/tests/regression/test_replica.robot
```diff
@@ -69,4 +69,44 @@ Test Offline Replica Rebuilding
...
+    And Wait until volume 0 replica rebuilding started on node 0
+    And Volume 0 should have 3 replicas running
+    # scale up the workload and it preempts the offline rebuilding.
+    And Scale up deployment 0 to attach volume
+    # the volume might be detched or not so we only wait for volume becomes healthy.
+    # we should try to check if the volume becomes healthy without offline rebuilding.
+    And Wait for volume 0 healthy
+    # write some data with workload pod to check if the workload works.
+    Then Write 64 MB data to file data.txt in deployment 0
+    And Write 64 MB data to file data.txt in deployment 0
```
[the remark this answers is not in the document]
mantissahz: Yes, removed. It was automatically added by my IDE. Thanks.

e2e/tests/regression/test_replica.robot
```diff
@@ -69,4 +69,43 @@ Test Offline Replica Rebuilding
...
+    # enable offline replica rebuilding.
+    When Delete volume 0 replica on node 0
+    Then Enable volume 0 offline replica rebuilding
+    # wait for offline rebuilding to start.
+    And Wait until volume 0 replica rebuilding started on node 0
+    And Volume 0 should have 3 replicas running
+    # scale up the workload and it preempts the offline rebuilding.
+    And Scale up deployment 0 to attach volume
+    # the volume might be detched or not so we only wait for volume becomes healthy.
+    # we should try to check if the volume becomes healthy without offline rebuilding.
```
yangchiu, line 107: I think we should indicate that `online rebuilding takes over`.

yangchiu (changes requested):

e2e/tests/regression/test_replica.robot
```diff
@@ -69,4 +69,43 @@ Test Offline Replica Rebuilding
...
+    And Wait for volume 0 healthy
+    # write some data with workload pod to check if the workload works.
+    Then Write 64 MB data to file data.txt in deployment 0
+    And Check deployment 0 data in file data.txt is intact
```
c3y1huang, line 111:
```suggestion
    Given Create volume 0 with    size=6Gi    numberOfReplicas=3    dataEngine=${DATA_ENGINE}
    And Attach volume 0
    And Wait for volume 0 healthy
    And Detach volume 0
    And Wait for volume 0 detached
    And Create deployment 0 with volume 0
    And Wait for volume 0 healthy
    And Write 5120 MB data to file data.txt in deployment 0
    And Scale down deployment 0 to detach volume
    And Wait for volume 0 detached

    # enable offline replica rebuilding.
    When Delete volume 0 replica on node 0
    And Enable volume 0 offline replica rebuilding
    # wait for offline rebuilding to start.
    Then Wait until volume 0 replica rebuilding started on node 0
    And Volume 0 should have 3 replicas running
  
    # scale up the workload and it preempts the offline rebuilding.
    When Scale up deployment 0 to attach volume
    # the volume might be detched or not so we only wait for volume becomes healthy.
    # we should try to check if the volume becomes healthy without offline rebuilding.
    And Wait for volume 0 healthy
    # write some data with workload pod to check if the workload works.
    And Write 64 MB data to file data.txt in deployment 0
    Then Check deployment 0 data in file data.txt is intact
```

e2e/tests/regression/test_replica.robot
mantissahz: Added. Thanks.

e2e/tests/regression/test_replica.robot
mantissahz: Modified. Thanks.

e2e/tests/negative/replica_rebuilding.robot
```diff
@@ -66,6 +66,20 @@ Reboot Replica Node While Replica Rebuilding
...
+    ...                Issue: https://github.com/longhorn/longhorn/issues/8443
+    ...    1. Create a volume and write some data to the volume.
+    ...    2. Detach the volume.
+    ...    3. Delete a replica of the volume.
+    ...    4. Enable offline rebuilding of the volume by the API `volume.offlineReplicaRebuilding`.
+    ...    5. Check if offline rebuilding starts.
+    ...    6. Reboot a replica node.
+    ...    7. Check if the volume offline rebuilding is in progress.
+    ...    8. The volume is detached after healthy replicas count of the volume equals to `Volume.Spec.NumberOfReplicas`.
+    Given Create volume 0 with    size=5Gi    numberOfReplicas=3    dataEngine=${DATA_ENGINE}
```
c3y1huang, line 81: It looks like the test case only includes a single Given step. Is this finished? What is this test intended to verify?

e2e/tests/regression/test_replica.robot
```diff
@@ -69,4 +69,45 @@ Test Offline Replica Rebuilding
...
+    ...
+    ...                Issue: https://github.com/longhorn/longhorn/issues/8443
+    ...    1. Create a workload with Longhorn volume having 3 replicas in a 3 worker nodes cluster
+    ...    2. Write some data to the volume.
+    ...    3. Scale down the workload to detach the volume.
+    ...    4. Delete a replica of the volume.
+    ...    5. Enable the offline rebuilding by the API `volume.offlineReplicaRebuilding`.
+    ...    6. When the rebuilding is in progress, scale up the workload.
+    ...    7. Check if the offline rebuilding process is canceled.
+    ...    8. Check if the workload is working well.
```
c3y1huang, line 86: The Given, When, Then should be self-explanatory. Consider removing the steps from the documentation section.

e2e/tests/negative/replica_rebuilding.robot
roger-ryao: > It looks like the test case only includes a single Given step. Is this finished? What is this test intended to verify?

I think @mantissahz  plans to add the `robot:skip` tag and then provide the  skeletons for negative test scenarios, which is not yet completed.

mantissahz: Added the test steps for the negative test case.

e2e/tests/regression/test_replica.robot
mantissahz: Removed. Thanks.

yangchiu (approved):

yangchiu:
@mergify backport v1.9.x

PR #2425, opened 2025-04-16
test(robot): Add case Checksum enabled large volume with multiple reb…

roger-ryao:
…uilding

#### Which issue(s) this PR fixes:
longhorn/longhorn#10711
Implement manual test case [Checksum enabled large volume with multiple rebuilding](https://longhorn.github.io/longhorn-tests/manual/pre-release/stability/checksum-enabled-large-volume/) in robot

#### What this PR does / why we need it:
Implement manual test case `Checksum enabled large volume with multiple rebuilding` in robot

#### Additional documentation or context
https://github.com/longhorn/longhorn/issues/7211

- PR test log : https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2613/robot/report/log.html
- Here's how to execute test case: ` ./run.sh -t 'Checksum Enabled Large Volume With Multiple Rebuilding' -v RETRY_COUNT:600
`

PR #2429, opened 2025-04-18
test(manual): Update Checksum enabled large volume with multiple rebiilding

roger-ryao:
longhorn/longhorn#10711

After discussing with @derekbit , we found that the original case design has issues:
1. To verify the deletion of the replica and ensure that Longhorn does not take more time than the first iteration, this is related to Snapshot Data Integrity. To trigger this functionality, it must be enabled through shutting down the node or disconnecting the node network, not by deleting the replica.
2. Create a recurring backup job every 15 minutes. I asked @khushboo-rancher  about this step, and she mentioned that the purpose was to add some parallel operations while rebuilding is happening. It's just an extra step to check if Longhorn can handle multiple operations together. Since Longhorn can use the same replica for backup, which might have been deleted in the test or is being rebuilt. However, @derekbit  suggested that this part should specifically target large volume deletions with multiple rebuilds and ensure data consistency.

@derekbit  @khushboo-rancher 

https://github.com/longhorn/longhorn-tests/pull/2425#discussion_r2050126804


yangchiu (approved):
LGTM

PR #2436, opened 2025-04-25
test(robot): add selective v2 data engine activation test cases

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/7016

add selective v2 data engine activation test cases

### DIFF
--- a/e2e/tests/regression/test_v2.robot
+++ b/e2e/tests/regression/test_v2.robot
@@ -31,7 +31,7 @@ Test V2 Volume Basic
     [Tags]  coretest
     [Documentation]    Test basic v2 volume operations
     When Create volume 0 with    dataEngine=v2
-    And Attach volume 0
+    And Attach volume 0 to node 0
     And Wait for volume 0 healthy
     And Write data to volume 0
     Then Check volume 0 data is intact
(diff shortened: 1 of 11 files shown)


e2e/tests/regression/test_v2.robot
```diff
@@ -116,3 +116,88 @@ V2 Volume Should Cleanup Resources When Instance Manager Is Deleted
...
+    And Check v1 instance manager is running on node 2
+
+    When Label node 1 with node.longhorn.io/disable-v2-data-engine=true
+    Then Check v2 instance manager is running on node 1
+    And Check v1 instance manager is running on node 1
+
+    When Detach volume 0
+    And Wait for volume 0 detached
+    Then Check v2 instance manager is not running on node 2
+    And Check v1 instance manager is running on node 2
```
chriscchien, line 137: According to https://github.com/longhorn/longhorn/issues/7015#issuecomment-1808553645 step 6, should test script also check v2 instance manager not running node 1 as well?

e2e/tests/regression/test_v2.robot
```diff
@@ -116,3 +116,90 @@ V2 Volume Should Cleanup Resources When Instance Manager Is Deleted
...
+    When Label node 2 with node.longhorn.io/disable-v2-data-engine-
+    Then Check v2 instance manager is running on node 2
+    And Wait for volume 0 attached
+    And Wait for volume 0 healthy
+    And Check volume 0 data is intact
+
+Test V2 Data Engine Selective Activation During Replica Rebuilding
+    Given Create volume 0 with    dataEngine=v2
+    And Attach volume 0 to node 0
+    And Wait for volume 0 healthy
```
chriscchien, line 197: Could we align the format of `Attach volume 0` and `Attach volume 0 to node 0` in this test suite?
Since attaching a volume without specifying node attaching the volume to node 0, both keywords effectively perform the same action, besides this, no other questions, thank you.
https://github.com/longhorn/longhorn-tests/blob/299d02fdabca3960414d1e5517ea52896d3938c4/e2e/libs/keywords/volume_keywords.py#L42-L45

chriscchien (approved):

e2e/tests/regression/test_v2.robot
yangchiu: Yes, you're right! Thanks for pointing this out!

e2e/tests/regression/test_v2.robot
yangchiu: Updated!

chriscchien (dismissed):
LGTM

roger-ryao (approved):
LGTM

PR #2442, opened 2025-04-28
test: fix test case drain node failure

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10804

fix test case drain node failure


chriscchien (approved):

yangchiu:
@mergify backport v1.9.x

PR #2454, opened 2025-05-02
ci: update gke to 1.32.2-gke.1182003

chriscchien:

Update gke to 1.32.2-gke.1182003 which is the current default version


yangchiu (approved):
LGTM

PR #2463, opened 2025-05-05
ci: prevent cloud credential from being deleted before harvester cluster is deleted on hal

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10849

prevent cloud credential from being deleted before harvester cluster is deleted on hal by a 2-stage `terraform destroy`, the 1st stage deletes the cluster only, and the 2nd stage deletes the remaining resources, including the credential.


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.9.x

PR #2481, opened 2025-05-12
ci: support accessing hal vms using custom ssh key

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10812

support accessing hal vms using custom ssh key


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.9.x

PR #2493, opened 2025-05-14
test(robot): refine vagrant node reboot function logging

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/6708 https://github.com/longhorn/longhorn/issues/6741

refine vagrant node reboot function logging

### DIFF
--- a/e2e/libs/host/vagrant.py
+++ b/e2e/libs/host/vagrant.py
@@ -37,37 +37,35 @@ class Vagrant(Base):
         return {node: node for node in nodes}
 
     def reboot_all_nodes(self, shut_down_time_in_sec=NODE_REBOOT_DOWN_TIME_SECOND):
+        logging(f'Rebooting all vms')
         self._vagrant_cmd(self._CMD_RELOAD)
-
-        logging(f'Wait for {shut_down_time_in_sec} seconds before starting all vms')
-        time.sleep(shut_down_time_in_sec)
         wait_for_cluster_ready()
-        logging(f'Started all instances')
+        logging(f'Rebooted all vms')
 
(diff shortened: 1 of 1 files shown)


e2e/libs/host/vagrant.py
```diff
@@ -37,37 +37,35 @@ def _get_node_mapping(cls, cmd_bin):
         return {node: node for node in nodes}
 
     def reboot_all_nodes(self, shut_down_time_in_sec=NODE_REBOOT_DOWN_TIME_SECOND):
+        logging(f'Rebooting all vms')
         self._vagrant_cmd(self._CMD_RELOAD)
-
-        logging(f'Wait for {shut_down_time_in_sec} seconds before starting all vms')
-        time.sleep(shut_down_time_in_sec)
```
roger-ryao, line 43: If we remove `time.sleep(shut_down_time_in_sec)`, should we also remove `shut_down_time_in_sec=NODE_REBOOT_DOWN_TIME_SECOND`  on https://github.com/longhorn/longhorn-tests/pull/2493/files#diff-187df4685021721945c2cab74539a50af88bf4617550e5183529919604872938L39 ?

Or should we keep `time.sleep(shut_down_time_in_sec)` and just improve the logging for the vagrant node reboot function?

yangchiu: Unlike other host providers, the `vagrant` implementation of `reboot_node` simply reboots the VM instead of shutting down, waiting for `shut_down_time_in_sec`, and then powering on. As a result, the `shut_down_time_in_sec` parameter in the `vagrant` implementation is basically non-functional.

yangchiu:
@mergify backport v1.9.x

PR #2519, opened 2025-05-22
test(robot): reduce disk space usage in robot test to avoid running out of disk space on HAL

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9149

reduce disk space usage in robot test to avoid running out of disk space on HAL


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.9.x

PR #2525, opened 2025-05-23
test(robot): fix flaky test case Test Longhorn Dynamic Provisioned RWO/RWX Volume Recovery

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10973

fix flaky test case Test Longhorn Dynamic Provisioned RWO/RWX Volume Recovery

### DIFF
--- a/e2e/tests/negative/component_resilience.robot
+++ b/e2e/tests/negative/component_resilience.robot
@@ -155,7 +155,7 @@ Test Longhorn Dynamic Provisioned RWX Volume Recovery
         When Create storageclass longhorn-test with    dataEngine=${DATA_ENGINE}
         And Create persistentvolumeclaim 0 using RWX volume with longhorn-test storageclass
         And Create deployment 0 with persistentvolumeclaim 0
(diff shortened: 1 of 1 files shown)


roger-ryao (commented):
The original test case aims to `delete the IM-e while it is rebuilding`. I think we can improve the step. Change the size of the `data.txt` file written in deployment 0 from `500MB` to `2000MB`. This should make it easier to capture the condition where `Wait until the volume of deployment 0 replica rebuilding started on the replica node`.

What do you think about this adjustment?

yangchiu:
> The original test case aims to `delete the IM-e while it is rebuilding`. I think we can improve the step. Change the size of the `data.txt` file written in deployment 0 from `500MB` to `2000MB`. This should make it easier to capture the condition where `Wait until the volume of deployment 0 replica rebuilding started on the replica node`.
> 
> What do you think about this adjustment?

Yes, you're right. Thanks for pointing out this!

roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.9.x

PR #2526, opened 2025-05-23
fix(tests): add `dataEngine` parameter to storage class in `test_provisioner_tags`

roger-ryao:
https://github.com/longhorn/longhorn/issues/10974

- Added the `dataEngine` parameter to the storage class configuration in the `test_provisioner_tags` test case.
- Ensures that the test aligns with the updated storage class requirements.

@chriscchien


yangchiu (approved):
LGTM

PR #2555, opened 2025-06-03
test(robot): implement vm snapshot on hal

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10918

implement vm snapshot on hal


yangchiu:
@mergify backport v1.9.x

PR #2560, opened 2025-06-05
test(robot): use a non-root user to test a workload and add file ownership and permission test case

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10991 https://github.com/longhorn/longhorn/issues/10624

use a non-root user to test a workload and add file ownership and permission test case

### DIFF
--- a/e2e/keywords/workload.resource
+++ b/e2e/keywords/workload.resource
@@ -47,6 +47,11 @@ Write ${size} GB large data to file ${file_name} in pod ${pod_id}
     ${pod_name} =    generate_name_with_suffix    pod    ${pod_id}
     write_workload_pod_large_data    ${pod_name}    ${size}    ${file_name}
 
+Run commands in ${workload_kind} ${workload_id}
+    [Arguments]    &{config}
+    ${workload_name} =   generate_name_with_suffix    ${workload_kind}    ${workload_id}
+    run_commands_workload_pod    ${workload_name}    &{config}
+
 Check pod ${pod_id} data in file ${file_name} is intact
     ${pod_name} =    generate_name_with_suffix    pod    ${pod_id}
     check_workload_pod_data_checksum    ${pod_name}    ${file_name}
--- a/e2e/keywords/common.resource
+++ b/e2e/keywords/common.resource
@@ -27,6 +27,9 @@ Library             ../libs/keywords/k8s_keywords.py
 Library             ../libs/keywords/secret_keywords.py
 
 *** Keywords ***
+Test should pass
+    No Operation
+
 Set disk path based on host provider and architecture
     ${HOST_PROVIDER}=    Get Environment Variable    HOST_PROVIDER    vagrant
     ${ARCH}=    Get Environment Variable    ARCH    amd64
--- a/e2e/libs/keywords/workload_keywords.py
+++ b/e2e/libs/keywords/workload_keywords.py
@@ -28,6 +28,7 @@ from workload.workload import wait_for_workload_pods_running
 from workload.workload import wait_for_workload_pods_stable
 from workload.workload import wait_for_workload_pod_kept_in_state
 from workload.workload import get_pod_node
+from workload.workload import run_commands_in_pod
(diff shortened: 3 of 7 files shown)


e2e/keywords/workload.resource
```diff
@@ -47,6 +47,11 @@ Write ${size} GB large data to file ${file_name} in pod ${pod_id}
     ${pod_name} =    generate_name_with_suffix    pod    ${pod_id}
     write_workload_pod_large_data    ${pod_name}    ${size}    ${file_name}
 
+Run commands in ${workload_kind} ${workload_id}
+    [Arguments]    &{config}
+    ${workload_name} =   generate_name_with_suffix    ${workload_kind}    ${workload_id}
+    run_commands_workload_pod    ${workload_name}    &{config}
+
```
roger-ryao, line 54: Regarding the function `def run_commands_in_pod(pod_name, commands)`, I think replacing it with Robot Framework's `Run And Return RC And Output` keyword might be a better approach. This would allow us to directly retrieve the return code and output, reducing the complexity of custom functions.  

I have already run the tests on Jenkins, and the results show that the test passed successfully on the master-head branch. You can refer to the report here:
- Test Passed: https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3354/robot/report/log.html
- Issue Reproduced: On version `v1.8.1`: https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3355/robot/report/log.html
The test case is `Test File Ownership And Permission By Executing Git Clone2`.

What do you think about this change?

```suggestion
Get Workload Pod Names
    [Arguments]    ${workload_name}
    ${rc}    ${output}=    Run And Return RC And Output    kubectl get pods -l app=${workload_name} -o jsonpath='{.items[*].metadata.name}'
    Run Keyword If    ${rc} != 0    Fail    Failed to get pod names for workload ${workload_name}: ${output}
    @{pod_names}=    Split String    ${output}
    [Return]    @{pod_names}

Run Commands Workload Pod
    [Arguments]    ${workload_name}    ${command}
    @{pod_names}=    Get Workload Pod Names    ${workload_name}
    ${pod_name}=    Set Variable    ${pod_names}[0]
    Log    Running command "${command}" in pod ${pod_name}
    Run Command In Pod    ${pod_name}    ${command}

Run Command In Pod
    [Arguments]    ${pod_name}    ${command}
    ${full_command}=    Set Variable    cd /data && ${command}
    ${rc}    ${output}=    Run And Return RC And Output    kubectl exec ${pod_name} -- /bin/sh -c "${full_command}"
    Log    Command output: ${output}
    Run Keyword If    ${rc} != 0    Fail    Command failed with return code ${rc}: ${output}

Run commands into ${workload_kind} ${workload_id}
    [Arguments]    &{command}
    ${workload_name}=    Generate Name With Suffix    ${workload_kind}    ${workload_id}
    Run Commands Workload Pod    ${workload_name}    &{command}
```

yangchiu: I try my best to avoid implementing logic directly in robot framework layer to maximize the portability and minimize coupling between the test code and the test framework. By keeping the logic in pure python, the test functions remain reusable regardless of the framework.

Imagine one day you'd like to switch to a different test framework, or even just call a function for a quick test. If the logic is written in pure python, it still works. But if it's written in robot, it becomes tightly bound to the framework. WDYT?

roger-ryao (approved):
LGTM

e2e/keywords/workload.resource
roger-ryao: LGTM

> I try my best to avoid implementing logic directly in robot framework layer to maximize the portability and minimize coupling between the test code and the test framework. By keeping the logic in pure python, the test functions remain reusable regardless of the framework.
> 
> Imagine one day you'd like to switch to a different test framework, or even just call a function for a quick test. If the logic is written in pure python, it still works. But if it's written in robot, it becomes tightly bound to the framework. WDYT?

LGTM

PR #2570, opened 2025-06-17
test(robot): Optimize prerelease checks to prevent space issue

roger-ryao:
longhorn/longhorn#11128

Enhance the Prerelease Checks test case to address the issue of `no space left on device` encountered during the validation of the `longhorn-rancher-chart-test`.
The detailed report : https://ci.longhorn.io/job/private/job/longhorn-rancher-chart-test/197/

The PR test result:
- `106.1.1+up1.7.3` > `106.2.1+up1.8.2`
  -  https://ci.longhorn.io/job/private/job/longhorn-rancher-chart-test/233/
- `106.2.1+up1.8.2`
  -  https://ci.longhorn.io/job/private/job/longhorn-rancher-chart-test/234/

### DIFF
--- a/e2e/tests/prerelease_checks.robot
+++ b/e2e/tests/prerelease_checks.robot
@@ -65,17 +65,17 @@ Pre-release Checks
 
         # (1) create a volume with revision counter enabled
         When Set setting disable-revision-counter to false
-        And Create volume vol-revision-enabled with    dataEngine=v1
+        And Create volume vol-revision-enabled with    size=1Gi    dataEngine=v1
         And Attach volume vol-revision-enabled
         And Wait for volume vol-revision-enabled healthy
-        And Write data data-vol-revision-enabled to volume vol-revision-enabled
+        And Write data data-vol-revision-enabled 256 MB to volume vol-revision-enabled
 
         # (2) create a volume with revision counter disabled
         When Set setting disable-revision-counter to true
-        And Create volume vol-revision-disabled with    dataEngine=v1
+        And Create volume vol-revision-disabled with    size=1Gi    dataEngine=v1
         And Attach volume vol-revision-disabled
         And Wait for volume vol-revision-disabled healthy
-        And Write data data-vol-revision-disabled to volume vol-revision-disabled
+        And Write data data-vol-revision-disabled 256 MB to volume vol-revision-disabled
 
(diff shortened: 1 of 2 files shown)


yangchiu (changes requested):
https://github.com/longhorn/longhorn/issues/11128#issuecomment-2978627558

e2e/tests/prerelease_checks.robot
```diff
@@ -128,31 +128,31 @@ Pre-release Checks
...
         And Create support bundle
         # system backup
         And Create system backup 0
         # orphaned replica
-        And Create volume v1 with    dataEngine=v1
+        And Create volume v1 with    size=1Gi    dataEngine=v1
         And Attach volume v1
         And Wait for volume v1 healthy
-        And Write data 0 to volume v1
+        And Write data 0 300 MB to volume v1
```
yangchiu, line 149: Could the write size be a power of 2?

yangchiu (changes requested):

yangchiu:
> Hi @yangchiu : After executing the test case, the original test case required `19GB` of disk space for backup. However, after adjusting the volume configurations and reducing the write data size, it now only requires `9GB` of disk space.

The refined test case still requires `9GB` of disk space for backups. So it will still fail on the `nfs` backup store if `NFS_DISK_IMAGE_SIZE_MB` is not adjusted, is that correct?

roger-ryao:
> > Hi @yangchiu : After executing the test case, the original test case required `19GB` of disk space for backup. However, after adjusting the volume configurations and reducing the write data size, it now only requires `9GB` of disk space.
> 
> The refined test case still requires `9GB` of disk space for backups. So it will still fail on the `nfs` backup store if `NFS_DISK_IMAGE_SIZE_MB` is not adjusted, is that correct?

Yes, that is correct. If the `NFS_DISK_IMAGE_SIZE_MB` is not adjusted, the refined test case will still fail on the `nfs` backup store due to the `9GB` disk space requirement for backups.

e2e/tests/prerelease_checks.robot
roger-ryao: Update

roger-ryao:
@mergify backport v1.9.x

PR #2573, opened 2025-06-19
test(robot): Automate manual test case Test System Upgrade with New Instance Manager

chriscchien:

longhorn/longhorn#11122

Automate manual test case [Test System Upgrade with New Instance Manager](https://longhorn.github.io/longhorn-tests/manual/pre-release/upgrade/upgrade-with-new-instance-manager/)

### DIFF
--- a/e2e/keywords/longhorn.resource
+++ b/e2e/keywords/longhorn.resource
@@ -61,6 +61,9 @@ Check ${engine_type} instance manager is not running on node ${node_id}
 Check v1 instance manager pods did not restart
     check_all_instance_managers_not_restart
 
+Check v1 instance manager pods recreated
+    wait_all_instance_managers_recreated
+
 Uninstall Longhorn
     ${backups_before_uninstall} =     list_all_backups
     uninstall_longhorn_system
@@ -99,10 +102,25 @@ Uninstall Longhorn stable version
     Set Test Variable    ${backups_before_uninstall}
 
 Upgrade Longhorn to transient version
-    upgrade_longhorn    upgrade_to_transient_version=True
+    ${upgraded} =    upgrade_longhorn    upgrade_to_transient_version=True
+    IF    "${upgraded}" == ${False}
+        Log To Console    Upgrade failed
+        Fail    Upgrading Longhorn to transient failed
+    END
 
 Upgrade Longhorn to custom version
-    upgrade_longhorn
+    ${upgraded} =    upgrade_longhorn
+    IF    "${upgraded}" == ${False}
+        Log To Console    Upgrade failed
+        Fail    Upgrading Longhorn failed
+    END
+
+Upgrade Longhorn to custom version should fail
+    ${upgraded} =     upgrade_longhorn    wait_when_fail=False
+    IF    "${upgraded}" == ${True}
+        Log To Console    Upgrade succeed
+        Fail    Upgrading Longhorn succeed
+    END
 
 Check Longhorn node ${type} state on power off node is ${expect_state}
     ${status} =    get_longhorn_node_condition_status    ${last_volume_node}    ${type}
--- a/e2e/libs/longhorn_deploy/longhorn_argocd.py
+++ b/e2e/libs/longhorn_deploy/longhorn_argocd.py
@@ -59,7 +59,7 @@ class LonghornArgocd(Base):
         process.wait()
         return True if process.returncode == 0 else False
 
-    def upgrade(self, upgrade_to_transient_version):
+    def upgrade(self, upgrade_to_transient_version, timeout):
         if upgrade_to_transient_version:
             upgrade_function = "upgrade_longhorn_transient"
         else:
@@ -67,5 +67,11 @@ class LonghornArgocd(Base):
         command = "./pipelines/utilities/argocd.sh"
         process = subprocess.Popen([command, upgrade_function],
                                    shell=False)
-        process.wait()
-        return True if process.returncode == 0 else False
+        try:
+            process.wait(timeout=timeout)
+            return True if process.returncode == 0 else False
+        except subprocess.TimeoutExpired:
+            logging(f"Upgrade timeout after {timeout}s. Killing process...")
+            process.kill()
+            process.wait()
+            return False
--- a/e2e/libs/longhorn_deploy/longhorn_deploy.py
+++ b/e2e/libs/longhorn_deploy/longhorn_deploy.py
@@ -15,6 +15,7 @@ class LonghornDeploy(Base):
 
     def __init__(self):
 
+        super().__init__()
         if self._method == "manifest":
             self.longhorn = LonghornKubectl()
         elif self._method == "helm":
@@ -52,14 +53,16 @@ class LonghornDeploy(Base):
         self.longhorn.setup_longhorn_ui_nodeport()
         logging(f"Installed Longhorn")
 
-    def upgrade(self, upgrade_to_transient_version):
+    def upgrade(self, upgrade_to_transient_version, timeout, wait_when_fail):
         logging(f"Upgrading Longhorn to {'transient' if upgrade_to_transient_version else 'the latest'} version")
-        upgraded = self.longhorn.upgrade(upgrade_to_transient_version)
+        upgraded = self.longhorn.upgrade(upgrade_to_transient_version, timeout)
         if not upgraded:
             logging(f"Upgrading Longhorn failed")
-            time.sleep(self.retry_count)
-            assert False, "Upgrading Longhorn failed"
+            if wait_when_fail is True:
+                time.sleep(self.retry_count)
+            return False
         else:
             # add some delay between 2 upgrades
             time.sleep(60)
         logging(f"Upgraded Longhorn")
+        return upgraded
--- a/e2e/tests/negative/upgrade_with_new_instance_manager.robot
+++ b/e2e/tests/negative/upgrade_with_new_instance_manager.robot
@@ -1,0 +1,54 @@
+*** Settings ***
+Documentation    Negative Test Cases
+
+Test Tags    negative
+
+Resource    ../keywords/variables.resource
+Resource    ../keywords/common.resource
+Resource    ../keywords/longhorn.resource
+Resource    ../keywords/host.resource
+Resource    ../keywords/volume.resource
+Resource    ../keywords/setting.resource
+
+Test Setup    Set up test environment
+Test Teardown    Cleanup test resources
+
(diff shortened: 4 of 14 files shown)


e2e/tests/negative/upgrade_with_new_instance_manager.robot
```diff
@@ -0,0 +1,54 @@
...
+    And Create volume vol1 with    dataEngine=v1
+    And Attach volume vol1
+    And Wait for volume vol1 healthy
+
+    When Upgrade Longhorn to transient version
+    And Create volume vol2 with    dataEngine=v1
+    And Attach volume vol2
+    And Wait for volume vol2 healthy
+
+    # Instance Manager flapping between ContainerCreating, OutOfcpu, Terminating,
```
yangchiu, line 46: Since it's clear that the instance manager will get stuck in `OutOfcpu` state. Could we verify that in the test step?

Like `Wait for instance manager to be in OutOfcpu state`.

e2e/tests/negative/upgrade_with_new_instance_manager.robot
```diff
@@ -0,0 +1,54 @@
...
+    And Check Longhorn CRD removed
+
+    When Install Longhorn stable version
+    And Set setting guaranteed-instance-manager-cpu to 40
+    And Check v1 instance manager pods recreated
+    And Create volume vol1 with    dataEngine=v1
+    And Attach volume vol1
+    And Wait for volume vol1 healthy
+
+    When Upgrade Longhorn to transient version
```
yangchiu, line 41: Sorry I'm not familiar with this test case. Do you know why the 1st upgrade will succeed but the 2nd upgrade will fail?

yangchiu (changes requested):

e2e/tests/negative/upgrade_with_new_instance_manager.robot
chriscchien: it is flapping like below, I can try cope this if this is necessary
```
instance-manager-e9dbd0ed3f82c9987005ee858aff55aa   0/1     OutOfcpu                 0               0s
instance-manager-6a689d5c0bce9ab0334359b124234fca   0/1     Pending                  0               0s
instance-manager-e9dbd0ed3f82c9987005ee858aff55aa   0/1     OutOfcpu                 0               0s
instance-manager-e9dbd0ed3f82c9987005ee858aff55aa   0/1     Pending                  0               0s
instance-manager-6a689d5c0bce9ab0334359b124234fca   0/1     OutOfcpu                 0               0s
instance-manager-d2a07cdbe635f201b2dca0331e0a40d4   0/1     OutOfcpu                 0               0s
instance-manager-d2a07cdbe635f201b2dca0331e0a40d4   0/1     OutOfcpu                 0               0s
instance-manager-d2a07cdbe635f201b2dca0331e0a40d4   0/1     Pending                  0               0s
instance-manager-e9dbd0ed3f82c9987005ee858aff55aa   0/1     OutOfcpu                 0               0s
instance-manager-d2a07cdbe635f201b2dca0331e0a40d4   0/1     OutOfcpu                 0               0s
instance-manager-6a689d5c0bce9ab0334359b124234fca   0/1     OutOfcpu                 0               0s
instance-manager-6a689d5c0bce9ab0334359b124234fca   0/1     OutOfcpu                 0               0s
instance-manager-6a689d5c0bce9ab0334359b124234fca   0/1     Pending                  0               0s
instance-manager-6a689d5c0bce9ab0334359b124234fca   0/1     OutOfcpu                 0               0s
instance-manager-e9dbd0ed3f82c9987005ee858aff55aa   0/1     OutOfcpu                 0               0s
instance-manager-e9dbd0ed3f82c9987005ee858aff55aa   0/1     OutOfcpu                 0               0s
instance-manager-e9dbd0ed3f82c9987005ee858aff55aa   0/1     Pending                  0    
```

e2e/libs/longhorn_deploy/longhorn_argocd.py
```diff
@@ -68,4 +68,24 @@ def upgrade(self, upgrade_to_transient_version):
...
+        process = subprocess.Popen([command, upgrade_function],
+                                   shell=False)
+        try:
+            process.communicate(timeout=timeout)
+            return process.returncode == 0
+        except subprocess.TimeoutExpired:
+            print(f"Timeout after {timeout}s. Killing upgrade process...")
+            process.kill()
+            process.communicate()            
+            return True
```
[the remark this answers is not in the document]
chriscchien: Updated to below because `upgrade_longhorn` have 50% pass 50% fail due to `wait_longhorn_status_running()` not everytime catch instance manager flapping `OutOfcpu`
```robotframework
Upgrade Longhorn to custom version and ignore fail
    Run Keyword And Ignore Error    upgrade_longhorn
```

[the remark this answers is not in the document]
yangchiu: Does that mean the `wait_longhorn_status_running` function doesn't correctly reflect the Longhorn components status? When it passed, were all Longhorn components in `Running` state including the instance manager pods? Or did it pass because, at that moment, the failed instance manager pods had already been terminated, and the new ones hadn’t been created yet?

If so, could we make it more robust by ensuring that instance manager pods exist before checking for the `Running` state? Like:

```
while [[ -z `kubectl get pods -n ${LONGHORN_NAMESPACE} --no-headers 2>&1 | awk '{print $1}' | grep csi-` ]] || \
    [[ -z `kubectl get pods -n ${LONGHORN_NAMESPACE} --no-headers 2>&1 | awk '{print $1}' | grep engine-image-` ]] || \
    [[ -z `kubectl get pods -n ${LONGHORN_NAMESPACE} --no-headers 2>&1 | awk '{print $1}' | grep instance-manager` ]] 
    [[ -n `kubectl get pods -n ${LONGHORN_NAMESPACE} --no-headers 2>&1 | awk '{print $3}' | grep -v "Running\|Completed"` ]]; do
```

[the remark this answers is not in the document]
chriscchien: yes it pass because at that moment, the failed instance manager pods had already been terminated, and the new ones hadn’t been created yet? I’ll run test using the code you shared to confirm. thank you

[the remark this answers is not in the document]
chriscchien: Update: It appears to be a timing issue. The new instance manager was not started immediately, but we checked longhorn compoment status right after applying the manifest. Hence I have update `wait_longhorn_status_running` for sleep 2 minutes before cheeck pod status

e2e/tests/negative/upgrade_with_new_instance_manager.robot
chriscchien: This PR is ready for review and here is test [result](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3420/)

e2e/libs/longhorn_deploy/longhorn_deploy.py
```diff
@@ -47,19 +48,20 @@ def install(self, install_stable_version):
         installed = self.longhorn.install(install_stable_version)
         if not installed:
             logging(f"Installing Longhorn failed")
-            time.sleep(self.retry_count)
+            time.sleep(self.retry_interval)
```
yangchiu, line 51: If it only sleeps for `self.retry_interval` second, it only sleeps for `1` second before failing the test case and tearing down the test environment, leaving us no opportunity to investigate the issue.

e2e/libs/longhorn_deploy/longhorn_deploy.py
```diff
@@ -47,19 +48,20 @@ def install(self, install_stable_version):
...
+    def upgrade(self, upgrade_to_transient_version, timeout):
         logging(f"Upgrading Longhorn to {'transient' if upgrade_to_transient_version else 'the latest'} version")
-        upgraded = self.longhorn.upgrade(upgrade_to_transient_version)
+        upgraded = self.longhorn.upgrade(upgrade_to_transient_version, timeout)
         if not upgraded:
             logging(f"Upgrading Longhorn failed")
-            time.sleep(self.retry_count)
-            assert False, "Upgrading Longhorn failed"
+            time.sleep(self.retry_interval)
+            return False
```
yangchiu, line 62: It is modified to simply return the result to the upper layer (in this case, `longhorn.resource`), so the responsibility of evaluating the result now belongs to `longhorn.resource`.

For example,
```
logging(f"Upgrading Longhorn failed")
time.sleep(self.retry_count)
assert False, "Upgrading Longhorn failed"
```
should be moved to `longhorn.resource`

e2e/keywords/longhorn.resource
```diff
@@ -104,6 +107,10 @@ Upgrade Longhorn to transient version
 Upgrade Longhorn to custom version
     upgrade_longhorn
```
yangchiu, line 108: The result of `upgrade_longhorn` is not evaluated.

For example,
```
${upgraded} =     upgrade_longhorn
IF    "${upgraded}" == ${False}
    Log To Console
    Sleep
    Fail    Upgrading Longhorn failed
END
```

pipelines/utilities/longhorn_status.sh
```diff
@@ -1,10 +1,13 @@
 wait_longhorn_status_running(){
-  local RETRY_COUNTS=60 # in minutes
+  local RETRY_COUNTS=10 # in minutes
```
yangchiu, line 2: Will this `RETRY_COUNTS` affect the test case? I assume the timeout set in https://github.com/longhorn/longhorn-tests/pull/2573/files#diff-360108ce1e2de83204c6e79a21f402be4d6a9abd6f4b128e855c749bf2e68f19R18 is sufficient for running the test case?

yangchiu (changes requested):

pipelines/utilities/longhorn_status.sh
chriscchien: Reverted the change. The idea is simply to align the timeout between the shell script and the Python function.

e2e/libs/longhorn_deploy/longhorn_deploy.py
chriscchien: Updated and [result](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3428/)

pipelines/utilities/longhorn_status.sh
chriscchien: Hi @yangchiu, I have updated and ready for review, thank you.

PR #2585, opened 2025-06-23
test(robot): Automate test case Test system upgrade with a new storageclass being default

chriscchien:

Automate test case Test system upgrade with a new storage class being default

longhorn/longhorn-11147


e2e/libs/storageclass/storageclass.py
```diff
@@ -50,3 +51,28 @@ def cleanup(self):
...
+    def set_storageclass_default_state(self, name, make_default=True):
+        if make_default:
+            logging(f"Set default annotation for storageclass {name}")
+        else:
+            logging(f"Remove default annotation for storageclass {name}")
+
+        value = '"true"' if make_default else 'null'
+        patch_cmd = [
+            "kubectl", "patch", "storageclass", name,
+            "-p", f'{{"metadata": {{"annotations": {{"storageclass.kubernetes.io/is-default-class": {value}}}}}}}'
```
yangchiu, line 64: Is it possible to use python k8s api [patch_storage_class](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/StorageV1Api.md#patch_storage_class)? It allows us to get rid of tedious `}}}}}}}`.

chriscchien: Updated!

yangchiu (approved):
LGTM

chriscchien:
@mergify backport v1.9.x

PR #2589, opened 2025-06-24
test(robot): update e2e README.md

chriscchien:

If run e2e on local and perform upgrade test case, it is necessary to manually copy folder `pipelines` into folder `e2e` now, or the test case will fail due to not find installation script. Add the step into `README.md`


chriscchien:
> Should it be placed under the `1. To run upgrade/uninstallation related test cases` section?

Updated!

PR #2591, opened 2025-06-24
test(robot): add rescheduling workload pod test cases

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11144

add rescheduling workload pod test cases


yangchiu:
No response from https://github.com/longhorn/longhorn/issues/1985. Close this for now.

PR #2595, opened 2025-06-24
fix: robot test node_reboot_with_different_anti_affinity should enable fast rebuilding for v2

shuo-wu:

Issue longhorn/longhorn#10828

### DIFF
--- a/e2e/tests/negative/node_reboot_with_different_anti_affinity.robot
+++ b/e2e/tests/negative/node_reboot_with_different_anti_affinity.robot
@@ -41,7 +41,11 @@ Power Off Node With Anti-Affinity Settings
     And Set setting replica-disk-soft-anti-affinity to ${disk_affinity}
     And Set setting replica-soft-anti-affinity to ${node_affinity}
     And Set setting replica-zone-soft-anti-affinity to ${zone_affinity}
-    And Create volume 0 with    dataEngine=${DATA_ENGINE}
+    IF    "${DATA_ENGINE}" == "v2"
+        And Set setting v2-data-engine-fast-replica-rebuilding to true
+    END
+
+    Given Create volume 0 with    dataEngine=${DATA_ENGINE}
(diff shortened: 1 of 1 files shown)


shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3449/

shuo-wu:
The e2e test https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3459/ got passed

e2e/keywords/common.resource
```diff
@@ -48,6 +48,10 @@ Set disk path based on host provider and architecture
 
 Enable v2 data engine and add block disks
     ${V2_DATA_ENGINE_ENABLED}=    get_setting    v2-data-engine
+    ${V1_DATA_ENGINE_ENABLED}=    get_setting    v1-data-engine
```
yangchiu, line 51: Could you help elaborate on this?

Does this mean: 

If `v1-data-engine` is `false`, set `v1-data-engine` to `false`?

yangchiu (changes requested):

e2e/keywords/common.resource
shuo-wu: Oh, typo here. Fixed

yangchiu: Could I know why it's necessary to disable `v1-data-engine` when testing a v2 volume?

yangchiu: This means it's impossible to run an integration test that tests both v1 and v2 volumes at the same time.

Also, if I remember correctly, v2 backing images still depend on v1 disks.

yangchiu (changes requested):

e2e/keywords/common.resource
shuo-wu: Removed

yangchiu (approved):
LGTM

PR #2600, opened 2025-06-26
test(robot): migrate test_csi_volumesnapshot_basic

yangchiu:
#### Which issue(s) this PR fixes:

Issue https://github.com/longhorn/longhorn/issues/10939

#### What this PR does / why we need it:

migrate test_csi_volumesnapshot_basic

PR #2598, opened 2025-06-25
test(robot): restart CSI plugin should not impact migratable RWX volume workload pods

c3y1huang:
#### Which issue(s) this PR fixes:

Issues:
- longhorn/longhorn#11158
- longhorn/longhorn#11173

#### What this PR does / why we need it:

New test to cover longhorn/longhorn#11158.

#### Special notes for your reviewer:

`None`

#### Additional documentation or context

`None`

PR #2604, opened 2025-06-28
fix: e2e test resource cleanup should not fail the test itself

shuo-wu:

https://github.com/longhorn/longhorn/issues/11202
https://github.com/longhorn/longhorn/issues/10828
https://github.com/longhorn/longhorn/issues/11188

### DIFF
--- a/e2e/libs/csi_volume_snapshot/csi_volume_snapshot.py
+++ b/e2e/libs/csi_volume_snapshot/csi_volume_snapshot.py
@@ -128,14 +128,17 @@ class CSIVolumeSnapshot:
             for item in classes.get("items", []):
                 name = item["metadata"]["name"]
                 logging(f"Deleting csi volume snapshot class {name}")
-                self.api.delete_cluster_custom_object(
-                    group=self.group,
-                    version=self.version,
-                    plural="volumesnapshotclasses",
-                    name=name
-                )
-        except ApiException as e:
-            assert e.status == 404
+                try:
+                    self.api.delete_cluster_custom_object(
+                        group=self.group,
+                        version=self.version,
+                        plural="volumesnapshotclasses",
+                        name=name
+                    )
+                except ApiException as delete_exception:
+                    assert delete_exception.status == 404
+        except ApiException as list_exception:
+            assert list_exception.status == 404
 
     def delete_csi_volume_snapshot(self, csi_volume_snapshot_name):
         for i in range(self.retry_count):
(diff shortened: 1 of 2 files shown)


shuo-wu:
The e2e test https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3459/ got passed

roger-ryao:
> The e2e test https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3459/ got passed
Hi @shuo-wu 
https://github.com/longhorn/longhorn-tests/pull/2600 should include fixing this. But we have not reviewed it yet. We will review it first, then look at your PR.

shuo-wu:
Hi @roger-ryao. I just quickly glanced at the PR you mentioned and didn't think it resolved the issue:
The 404 Not Found error is caused by `self.api.list_cluster_custom_object` rather than the subsequent `self.api.delete_namespaced_custom_object`

e2e/keywords/common.resource
```diff
@@ -70,32 +70,37 @@ Cleanup test resources
...
-    cleanup_disks
-    cleanup_engine_images
-    reset_backupstore
-    reset_settings
-    reset_node_tags
-    reset_node_disks_tags
-    cleanup_k8s_node_zone
-    cleanup_csi_volume_snapshots
-    cleanup_csi_volume_snapshot_classes
+    TRY
```
yangchiu, line 73: I think failures in resource cleanup should fail the test case, otherwise the leftover resources will silently pollute the environment and unexpectedly affect the subsequent test cases. If the subsequent test cases fail as a result, we will need more efforts to identify the issue.

e2e/libs/csi_volume_snapshot/csi_volume_snapshot.py
```diff
@@ -128,14 +128,17 @@ def cleanup_csi_volume_snapshot_classes(self):
...
                 logging(f"Deleting csi volume snapshot class {name}")
-                self.api.delete_cluster_custom_object(
-                    group=self.group,
-                    version=self.version,
-                    plural="volumesnapshotclasses",
-                    name=name
-                )
-        except ApiException as e:
-            assert e.status == 404
+                try:
```
yangchiu, line 131: There are nested `try/catch` blocks in this function. PR https://github.com/longhorn/longhorn-tests/pull/2598 has introduced a `try/catch`.

yangchiu (changes requested):

e2e/libs/csi_volume_snapshot/csi_volume_snapshot.py
shuo-wu: The nested try/catch blocks may help the following case:
1. The func lists all CRs before cleanup. One entry in the list is already in deletion stage
2. The entry disappears, then the func sends a deletion request for it

Without the nested try/catch, the execution will be directly trapped in the exception block then return. The subsequent resources in the list won't be removed.

e2e/keywords/common.resource
shuo-wu: OK make sense. Just re-considered this part, returning an error in the teardown stage is better than silently polluting the next test.

e2e/libs/csi_volume_snapshot/csi_volume_snapshot.py
yangchiu: This PR is pending https://github.com/longhorn/longhorn-tests/pull/2604#discussion_r2181468764. Will the teardown function return an error in your current design?

shuo-wu: Now I am fine with returning an error during teardowm stage. That commit will be dropped anyway. 
If this nested try/catch is not required either, there is no commit in this PR. We can directly close it.

yangchiu: I can still find the `TRY/EXCEPT` block in the teardown function of this PR. Could you help update it?

shuo-wu: Updated. PTAL

yangchiu:
@mergify backport v1.9.x

PR #2615, opened 2025-07-09
Fix pod deleteion func and add debug logs for robot tests

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/10800

### DIFF
--- a/e2e/libs/workload/pod.py
+++ b/e2e/libs/workload/pod.py
@@ -134,7 +134,8 @@ def delete_pod(name, namespace='default', wait=True):
         try:
             core_api.delete_namespaced_pod(name=name, namespace=namespace, grace_period_seconds=0)
             if wait:
-                wait_delete_pod(name)
+                wait_delete_pod(name, namespace)
+            break
         except rest.ApiException as e:
             if e.status == 404:
                 logging(f"Deleted pod {name} in namespace {namespace}")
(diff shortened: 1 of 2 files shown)


yangchiu (changes requested):

e2e/libs/workload/pod.py
```diff
@@ -134,7 +134,8 @@ def delete_pod(name, namespace='default', wait=True):
         try:
             core_api.delete_namespaced_pod(name=name, namespace=namespace, grace_period_seconds=0)
             if wait:
-                wait_delete_pod(name)
+                wait_delete_pod(name, namespace)
+            break
```
yangchiu, line 138: Do we need the `break` here? If the pod has been deleted, `delete_namespaced_pod` function will throw `404 Not found` exception, and be caught by:
```
except rest.ApiException as e:
    if e.status == 404:
        logging(f"Deleted pod {name} in namespace {namespace}")
        return
``` 
then this `delete_pod` function can successfully return.

yangchiu (changes requested):

yangchiu (approved):

e2e/libs/workload/pod.py
yangchiu: It's because longhorn-manager pod will be automatically recreated immediately.

PR #2616, opened 2025-07-10
ci(argocd): make crds sync by ignoring preserveUnknownFields

yangchiu:

Issue https://github.com/longhorn/longhorn/discussions/10786

make crds sync by ignoring preserveUnknownFields


derekbit (approved):
LGTM

PR #2621, opened 2025-07-15
test: fix backup target default is not available error

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10087

fix backup target default is not available error


yangchiu:
@mergify backport v1.9.x

PR #2624, opened 2025-07-16
test: fix getting longhorn api client could hang

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11316

fix getting longhorn api client could hang

(1) set timeout for socket connect function
(2) avoid trying to get longhorn api client from a terminating longhorn manager pod


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.9.x

PR #2627, opened 2025-07-16
test(robot): fix backup target unavailable error when trying to create an error backup

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10087

fix backup target unavailable error when trying to create an error backup

### DIFF
--- a/e2e/tests/regression/test_backup.robot
+++ b/e2e/tests/regression/test_backup.robot
@@ -116,12 +116,9 @@ Test Uninstallation With Backups
     And Wait for volume 0 healthy
     And Write data 0 to volume 0
 
-    # create failed backups by resetting backupstore and create backups without available backup targets
-    And Reset Backupstore
-    And Create backup 0 for volume 0    wait=False
+    And Create backup 0 for volume 0
+    And Create error backup for volume 0
     And Verify backup list contains errors for volume 0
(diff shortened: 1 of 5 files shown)


e2e/tests/regression/test_backup.robot
```diff
@@ -116,12 +116,9 @@ Test Uninstallation With Backups
     And Wait for volume 0 healthy
     And Write data 0 to volume 0
 
-    # create failed backups by resetting backupstore and create backups without available backup targets
-    And Reset Backupstore
-    And Create backup 0 for volume 0    wait=False
+    And Create backup 0 for volume 0
+    And Create error backup for volume 0
     And Verify backup list contains errors for volume 0
-    And Set Backupstore
```
roger-ryao, line 123: The original test case intends to test `Test Uninstallation With Backups`. Therefore, the following two steps should be retained:
```
And Set Backupstore
And Create backup 1 for volume 0
```

yangchiu: If you didn't `Reset Backupstore`, you don't need to `Set Backupstore` again.

`Create backup 0 for volume 0` is still there.

yangchiu:
@mergify backport v1.9.x

PR #2634, opened 2025-07-18
test(robot): fix Test system upgrade with a new storage class being default on v2 volume

yangchiu:

Issue N/A

fix Test system upgrade with a new storage class being default on v2 volume


yangchiu:
@mergify backport v1.9.x

PR #2644, opened 2025-07-22
test(robot): add Test backupstore with existing backups test case

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11349

add Test backupstore with existing backups test case


yangchiu:
@longhorn/qa This PR is ready for review.

e2e/libs/longhorn_deploy/base.py
```diff
@@ -63,6 +63,16 @@ def create_registry_secret(self):
             time.sleep(self.retry_count)
             assert False, "Creating registry secret failed"
 
+    def create_aws_secret(self):
+        command = "./pipelines/utilities/create_aws_secret.sh"
+        process = subprocess.Popen([command, "create_aws_secret"],
+                                   shell=False)
+        process.wait()
+        if process.returncode != 0:
```
chriscchien, line 71: Does this mean `subprocess_exec_cmd` is not suitable for this scenario?
If yes, would it be okay to share the experience situation encountered here for reference? Thank you!

e2e/tests/regression/test_backup.robot
```diff
@@ -11,14 +11,20 @@ Resource    ../keywords/persistentvolume.resource
...
 
 Test Setup    Set up test environment
-Test Teardown    Cleanup test resources
+Test Teardown    Cleanup test resources without corrupting remote backupstore
 
 *** Keywords ***
+Cleanup test resources without corrupting remote backupstore
+    Set backupstore
+    Cleanup test resources
+
```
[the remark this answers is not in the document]
yangchiu: Updated!

e2e/libs/longhorn_deploy/base.py
yangchiu: The main difference is that you can get outputs of `subprocess.Popen` in real time, just like when you run the commands yourself, so you can clearly track the current status of the running command on the console. On the other hand, for `subprocess.check_output`, the function simply blocks before it returns. It doesn't display anything during its execution, and you have to explicitly print its outputs yourself all at once after it returns.

chriscchien (approved):

PR #2645, opened 2025-07-22
chore: update external-snapshotter to v8.3.0

derekbit:

Issue Longhorn/longhorn#11238

Ref: https://github.com/longhorn/longhorn/blob/v1.9.1-rc1/deploy/longhorn-images.txt#L11


yangchiu:
Should parsing the version from longhorn-images.txt on demand: https://github.com/longhorn/longhorn-tests/blob/c4bff28b119724087f9049c820d6f4f107f1db04/test_framework/scripts/longhorn-setup.sh#L66

The logic was missing during the refactoring. I’ll add it back.

PR #2646, opened 2025-07-23
ci: install corresponding csi snapshotter based on longhorn branch

yangchiu:

Issue N/A

install corresponding csi snapshotter based on longhorn branch


roger-ryao (approved):
LGTM

PR #2651, opened 2025-07-24
ci(regression): add v2 mark when RUN_V2_TEST enabled

chriscchien:

Issue longhorn/longhorn#11376

Add v2 mark when RUN_V2_TEST enabled to avoid manially input "-m v2_volume_test"


derekbit (approved):
LGTM

PR #2654, opened 2025-07-25
feat(regression): update test cases because of the consolidation of settings

derekbit:

Issue Longhorn/longhorn#10926


yangchiu:
@derekbit flake8 check is failing.

chriscchien (approved):

PR #2658, opened 2025-07-29
test(robot): fix uninstallation test case failure caused by inspecting logs from a duplicated failed longhorn uninstall pod

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10483

fix uninstallation test case failure caused by inspecting logs from a duplicated failed longhorn uninstall pod


chriscchien (approved):
LGTM

roger-ryao (approved):
LGTM

PR #2668, opened 2025-08-05
test(robot): fix flaky System upgrade with the same backing image manager image

chriscchien:

Issue longhorn/longhorn#11424

fix flaky `System upgrade with the same backing image manager image` 

Remove test step 
```markdown
Check at latest one disk file status of the backing image becomes `unknown` then `ready` during the longhorn manager 
pods termination and restart. (May need to refresh the UI page after restart.)
```
This step was removed because the backing image disk file status does not **always** transition as described, leading to test flakiness.

Upgraded from `v1.9.1` to `master` while keeping the backing image manager image unchanged. As shown in the video below, the disk status of the backing image `bi` remains `ready` across all nodes.

https://github.com/user-attachments/assets/a2a54393-d97a-4a3c-b9fe-7cc36652df80

https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3713/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/3712/


yangchiu (approved):
LGTM

yangchiu:
@mergify backport v1.9.x

PR #2669, opened 2025-08-06
ci: setup talos longhorn environment with UserVolumeConfig and refine regression test to support non-default data path

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11417 https://github.com/longhorn/longhorn/issues/11426

setup talos longhorn environment with UserVolumeConfig and refine regression test to support non-default data path


roger-ryao (approved):
LGTM

PR #2678, opened 2025-08-07
refactor: introduce "Assert setting" Keyword

derekbit:

Issue Longhorn/longhorn#10926


yangchiu (changes requested):

e2e/tests/regression/test_settings.robot
```diff
@@ -17,9 +17,35 @@ Test Setup    Set up test environment
 Test Teardown    Cleanup test resources
 
 *** Test Cases ***
+Test Setting Update With Correct Value
+    [Documentation]    Test multiple settings update and verify the result.
+
+    Update Setting And Assert Value    default-replica-count    1    {"v1":"1","v2":"1"}
```
yangchiu, line 23: Since we’re validating setting update functionality here, I think separate them into 2 steps:

```
When update setting key to value
Then setting key should be value 
```
will be more clear?

e2e/tests/regression/test_settings.robot
```diff
@@ -17,9 +17,35 @@ Test Setup    Set up test environment
...
+    ${value1}=    Get setting    snapshot-data-integrity-immediate-check-after-snapshot-creation
+    ${value2}=    Get setting    disable-revision-counter
+    ${value3}=    Get setting    default-replica-count
+
+    ${original_values}=    Create Dictionary
+    ...    snapshot-data-integrity-immediate-check-after-snapshot-creation=${value1}
+    ...    disable-revision-counter=${value2}
+    ...    default-replica-count=${value3}
+
+    Run Invalid Setting Update And Verify    snapshot-data-integrity-immediate-check-after-snapshot-creation    {"v1":"true","v2":"true"}    ${original_values['snapshot-data-integrity-immediate-check-after-snapshot-creation']}
```
yangchiu, line 40: How about:
```
When update setting key to invalid-value should fail
Then setting key should remain at default value
```

yangchiu (changes requested):

e2e/keywords/setting.resource
```diff
@@ -5,7 +5,41 @@ Library    ../libs/keywords/setting_keywords.py
 
 *** Keywords ***
 Set setting ${setting_key} to ${setting_value}
-    update_setting    ${setting_key}    ${setting_value}
+    [Arguments]    ${retry}=True
+    Run Keyword If    '${retry}' == 'true'
+    ...    Wait Until Keyword Succeeds    5 times    3 seconds
```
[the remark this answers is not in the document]
derekbit: IMO, this should be fine, because the logic is for the `retry` argument. WDYT?

e2e/tests/regression/test_settings.robot
derekbit: Sounds good

e2e/tests/negative/checksum_enabled_large_volume_with_multiple_rebuilding.robot
```diff
@@ -25,8 +25,10 @@ Verify Large Volume Data Integrity During Replica Rebuilding with Recurring Jobs
     ...                - Create a recurring job of snapshot & backup.
     ...                - Delete a replica and wait for the replica rebuilding.
     ...                - Check volume data is intact
-    Given Set setting snapshot-data-integrity to enabled
-    And Set setting snapshot-data-integrity-immediate-check-after-snapshot-creation to true
+    Given Set setting snapshot-data-integrity to enabled    retry=true
+    And Assert setting snapshot-data-integrity is enabled
```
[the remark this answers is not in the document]
c3y1huang: I think any operation should include a validation step.
If we don’t validate the operation, a silent failure (e.g., the setting update not taking effect) could lead to false positives in our tests.

The redundancy here is not about whether validation happens, but about **where** it happens.

We can address this by introducing a new keyword, for example:
`Given setting snapshot-data-integrity is enabled`

This keyword would:
1. Update the setting.
1. Validate that the setting is applied before returning.

e2e/keywords/setting.resource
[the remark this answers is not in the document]
c3y1huang: Why not use the retry in [update_setting](https://github.com/longhorn/longhorn-tests/blob/663a9708e21f7ce931ceca73318ab94a12ed1c47/e2e/libs/setting/setting.py#L29)?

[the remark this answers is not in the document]
derekbit: Since the retry is removed in this PR.

e2e/tests/negative/checksum_enabled_large_volume_with_multiple_rebuilding.robot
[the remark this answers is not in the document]
c3y1huang: > Although I think the non-atomic is not really necessary, I'm fine with `Given Ensure setting..is set to xxx`.

Suggest leaving out the `Ensure`since it represented as a precondition rather than an action. If we prefer not to show the validation as part of the test case, we can use `Given Setting... is set to xxx` instead.

e2e/keywords/setting.resource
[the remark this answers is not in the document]
c3y1huang: How about moving it to [libs/keywords/setting_keywords.py](https://github.com/longhorn/longhorn-tests/blob/663a9708e21f7ce931ceca73318ab94a12ed1c47/e2e/libs/keywords/setting_keywords.py)?

derekbit:
@yangchiu @c3y1huang All comments are addressed. Please help review it again. Thank you.

e2e/keywords/setting.resource
```diff
@@ -4,8 +4,29 @@ Documentation    Setting Keywords
...
-    Run Keyword And Expect Error    *    update_setting    ${setting_key}    ${setting_value}    retry=False
\ No newline at end of file
+    [Arguments]    ${retry}=False
+    Run Keyword And Expect Error    *    update_setting    ${setting_key}    ${setting_value}    retry=${retry}
+
+Setting ${setting_key} is set to ${setting_value}
+    [Arguments]    ${retry}=True
+    update_setting    ${setting_key}    ${setting_value}    retry=${retry}
+    ${value} =    get_setting    ${setting_key}
+    Should Be Equal    ${value}    ${expected_value}
```
roger-ryao, line 15: How should `${expected_value}` be defined?  But I don't see any related action.

e2e/keywords/setting.resource
```diff
@@ -4,8 +4,29 @@ Documentation    Setting Keywords
...
 
 *** Keywords ***
-Set setting ${setting_key} to ${setting_value}
-    update_setting    ${setting_key}    ${setting_value}
-
 Set setting ${setting_key} to ${setting_value} will fail
-    Run Keyword And Expect Error    *    update_setting    ${setting_key}    ${setting_value}    retry=False
\ No newline at end of file
+    [Arguments]    ${retry}=False
+    Run Keyword And Expect Error    *    update_setting    ${setting_key}    ${setting_value}    retry=${retry}
```
yangchiu, line 9: No need to use `[Arguments]` block here because this keyword doesn't accept arbitrary arguments. `update_setting retry=False` should be good enough.

e2e/keywords/setting.resource
```diff
@@ -4,8 +4,29 @@ Documentation    Setting Keywords
...
+
+Setting ${setting_key} is set to ${setting_value}
+    [Arguments]    ${retry}=True
+    update_setting    ${setting_key}    ${setting_value}    retry=${retry}
+    ${value} =    get_setting    ${setting_key}
+    Should Be Equal    ${value}    ${expected_value}
+
+Get setting ${setting_key}
```
yangchiu, line 17: Is this keyword being used?

e2e/libs/setting/setting.py
```diff
@@ -26,26 +26,46 @@ def __init__(self):
         self.retry_count, self.retry_interval = get_retry_count_and_interval()
         self.backupstore = BackupStore()
 
+    def _update_setting(self, key, value):
+        setting = get_longhorn_client().by_id_setting(key)
+        get_longhorn_client().update(setting, value=value)
+
     def update_setting(self, key, value, retry=True):
```
yangchiu, line 33: Could we know what the improvement were made to the `update_setting` function? It seems we don't need to change anything in this function as it already fulfills the requirements of the test cases you're trying to add.

e2e/libs/setting/setting.py
```diff
@@ -26,26 +26,46 @@ def __init__(self):
...
-            get_longhorn_client().update(setting, value=value)
-            return
-
-        assert False, f"Failed to update setting {key} to {value} ... {setting}"
+            try:
+                self._update_setting(key, value)
+                logging(f"Successfully updated setting {key} to {value}.")
+            except Exception as e:
+                logging(f"Failed to update setting {key} to {value}: {e}")
+                raise
```
yangchiu, line 68: It seems the `try/except/raise` is unnecessary, as re-raising the caught exception is the same as not having a `try/except` block at all.

e2e/libs/setting/setting.py
```diff
@@ -70,10 +90,26 @@ def set_backupstore(self):
                         self.SETTING_BACKUP_TARGET_CREDENTIAL_SECRET,
                         backupsettings[1],
                         retry=False)
+                    value = self.get_setting(
```
yangchiu, line 93: I believe we no longer need to maintain `set_backupstore` and `reset_backupstore` in `setting.py`, since backup target has been moved from `setting` to `backup target CR`. these 2 functions have become unreachable.

yangchiu (changes requested):

e2e/libs/setting/setting.py
derekbit: > Could we know what the improvement were made to the update_setting function?

`update_setting` shouldn't include any validation logic.

e2e/keywords/setting.resource
derekbit: We can remove it.

e2e/libs/setting/setting.py
derekbit: > Could we know what the improvement were made to the update_setting function?

`update_setting` shouldn't include any validation logic.
Based on its literal meaning, it should only perform the update without any validation, to avoid misleading the upper-level callers.

e2e/tests/negative/upgrade_with_new_instance_manager.robot
```diff
@@ -27,12 +27,12 @@ Test System Upgrade with New Instance Manager
         Fail    Environment variable LONGHORN_TRANSIENT_VERSION is not set
     END
 
-    Given Set setting deleting-confirmation-flag to true
+    Given Setting deleting-confirmation-flag is set to {"v1":"true","v2":"true"}
```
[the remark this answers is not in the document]
derekbit: Good catch. It should be `true`

e2e/keywords/setting.resource
```diff
@@ -4,8 +4,24 @@ Documentation    Setting Keywords
...
 *** Keywords ***
-Set setting ${setting_key} to ${setting_value}
-    update_setting    ${setting_key}    ${setting_value}
-
 Set setting ${setting_key} to ${setting_value} will fail
-    Run Keyword And Expect Error    *    update_setting    ${setting_key}    ${setting_value}    retry=False
\ No newline at end of file
+    Run Keyword And Expect Error    *    update_setting    ${setting_key}    ${setting_value}    retry=False
+
+Setting ${setting_key} is set to ${setting_value}
```
shuo-wu, line 10: NIT: Do we need this renaming? All other functions use a Simple Present Tense verb as the 1st word.

PR #2679, opened 2025-08-08
ci: appco pipeline support download support bundle

chriscchien:

appco pipeline support download support bundle

Using a temp pod having curl to create support bundle


yangchiu (approved):
LGTM

PR #2685, opened 2025-08-11
test: mark test_filesystem_trim as a flaky test case

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10956#issuecomment-3173044494

mark test_filesystem_trim as a flaky test case


roger-ryao (approved):
LGTM

PR #2689, opened 2025-08-11
test: test skeleton for backing image download timeout

COLDTURNIP:

Issue longhorn/longhorn#11309

This issue is observed from the log and currently no clear way to reproduce it. A possible root cause is the network or disk I/O latency.


e2e/tests/regression/test_backing_image.robot
```diff
@@ -54,3 +54,20 @@ Test Backup Backing Image
...
+    ...    2. And create a backing image CR to download the given image
+    ...    3. And wait for backing image data source pod get launched
+    ...    4  When temporary block the download connection for 5 second
+    ...    5. And resume the download connection
+    ...    6. Then the backing image data source should be able to download the image without timeout error
+    ...    7. When create a backing image CR to download the given image
+    ...    8. And wait for backing image data source pod get launched
+    ...    9  When temporary block the download connection for 35 second
+    ...    10. And resume the download connection
+    ...    11. Then the backing image data source should fail to download the image
```
yangchiu, line 72: Will the backing image get stuck in failed state and never retry the download to recover itself?

roger-ryao: > retry the download to recover itself

While verifying https://github.com/longhorn/longhorn/issues/11309, I noticed that `backing-image-ds` retries failed downloads and recovers itself, but I’m not sure whether there is a retry limit.

yangchiu:
> This pull request is now in conflict. Could you fix it @COLDTURNIP? 🙏

@COLDTURNIP Please help resolve the conflicts. Thank you!

PR #2703, opened 2025-08-19
test(robot): fix default backup target is not available error

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10087#issuecomment-3195768071

fix default backup target is not available error


roger-ryao (approved):
LGTM

PR #2705, opened 2025-08-20
test(robot): skip test case when required environment variable is not set

chriscchien:

Issue #

Skip test case if required environment variable is missing instead of failing


yangchiu (commented):
I'm not sure if it's the right thing to do. Since if a test case is skipped, people usually ignore it and don't realize there is a test case should be run but wasn't.

PR #2710, opened 2025-08-20
ci: Force `Ubuntu` use `IPv4` for install dependency

chriscchien:

Issue longhorn/longhorn#11529

Force `Ubuntu` use `IPv4` for install dependency to avoid dependency install fail

Before the fix, v2 regression test case all failed
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9226/

Use this fix, v2 test case worked properly with Ubuntu
- https://10.115.5.5/job/private/job/longhorn-tests-regression-experiment/16/

### DIFF
--- a/test_framework/terraform/aws/ubuntu/user-data-scripts/provision_k3s_agent.sh.tpl
+++ b/test_framework/terraform/aws/ubuntu/user-data-scripts/provision_k3s_agent.sh.tpl
@@ -1,7 +1,8 @@
 #!/bin/bash
 
-apt-get update
-apt-get install -y nfs-common cryptsetup dmsetup samba linux-modules-extra-`uname -r`
+apt-get -o Acquire::ForceIPv4=true update
+apt-get -o Acquire::ForceIPv4=true install -y "linux-modules-extra-$(uname -r)"
+apt-get -o Acquire::ForceIPv4=true install -y nfs-common cryptsetup dmsetup samba
 
 systemctl stop multipathd.socket
 systemctl disable multipathd.socket


yangchiu (commented):
I was wondering why it works, because in https://github.com/longhorn/longhorn/issues/11529#issue-3337316903, the first error eventually falls back to ipv4, but it still times out:
```
Could not connect to security.ubuntu.com:80 (91.189.91.81), connection timed out
```
And in the second error, it actually tries to access to the resources by ipv4:
```
Not Found [IP: 54.87.19.168 80]
```

chriscchien:
> I was wondering why it works, because in [longhorn/longhorn#11529 (comment)](https://github.com/longhorn/longhorn/issues/11529#issue-3337316903), the first error eventually falls back to ipv4, but it still times out:
> 
> ```
> Could not connect to security.ubuntu.com:80 (91.189.91.81), connection timed out
> ```
> 
> And in the second error, it actually tries to access to the resources by ipv4:
> 
> ```
> Not Found [IP: 54.87.19.168 80]
> ```

This is Ubuntu [issue](https://askubuntu.com/questions/759524/problem-with-ipv6-sudo-apt-get-update-upgrade) and from the test result https://10.115.5.5/job/private/job/longhorn-tests-regression-experiment/16/, this fix worked

yangchiu:
I know it works, but I'm trying to figure out the reason behind it.

chriscchien:
> I know it works, but I'm trying to figure out the reason behind it.

I think it may because apt update failed at IPV6, package indexes weren’t refreshed, since the command is one line, after `samba`(need update success for install) fail, the next arg `linux-modules-extra-"uname -r"` not performed

PR #2711, opened 2025-08-21
ci: longhornctl install/check preflight enable spdk

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11530

longhornctl install/check preflight enable spdk


roger-ryao (commented):
The `--enable-spdk` option should be added to the current workflow. 
This ensures that the original functionality remains intact while adding support for SPDK.

roger-ryao (approved):
Hi @yangchiu 
      I noticed that `./longhornctl install preflight --enable-spdk` and `./longhornctl check preflight --enable-spdk` can cover the functionality of the commands without `--enable-spdk`, so this PR can be merged.

PR #2712, opened 2025-08-21
test(robot): fix unreachable sleep in waiting for system backup to be created

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11533

Fix unreachable sleep in waiting for system backup to be created.
And attaching the volume before creates backups since allow-recurring-job-while-volume-detached is disabled by default

### DIFF
--- a/e2e/tests/regression/test_recurringjob.robot
+++ b/e2e/tests/regression/test_recurringjob.robot
@@ -16,6 +16,8 @@ Test Teardown    Cleanup test resources
 Test System Backup Recurring Job When volume-backup-policy is disabled
     [Tags]    recurring-job    system-backup-recurring-job
     Given Create volume 0 with    size=2Gi    numberOfReplicas=1    dataEngine=${DATA_ENGINE}
+    And Attach volume 0
+    And Wait for volume 0 healthy
     When Create system-backup recurringjob 0    parameters={"volume-backup-policy":"disabled"}
 
     Then Assert recurringjob 0 not created backup for volume 0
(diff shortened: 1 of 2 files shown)


e2e/tests/regression/test_recurringjob.robot
```diff
@@ -16,6 +16,8 @@ Test Teardown    Cleanup test resources
 Test System Backup Recurring Job When volume-backup-policy is disabled
     [Tags]    recurring-job    system-backup-recurring-job
     Given Create volume 0 with    size=2Gi    numberOfReplicas=1    dataEngine=${DATA_ENGINE}
+    And Attach volume 0
+    And Wait for volume 0 healthy
```
roger-ryao, line 20: The current change focuses on volume in an `attached` state. I suggest expanding the test case to include both an `attached` and a `detached` volume.

```suggestion
Test System Backup Recurring Job When volume-backup-policy is disabled
    [Tags]    recurring-job    system-backup-recurring-job
    Given Create volume 0 with    size=2Gi    numberOfReplicas=1    dataEngine=${DATA_ENGINE}
    And Create volume 1 with    size=2Gi    numberOfReplicas=1    dataEngine=${DATA_ENGINE}
    And Attach volume 1
    And Wait for volume 1 healthy

    When Create system-backup recurringjob 0    parameters={"volume-backup-policy":"disabled"}

    Then Assert recurringjob 0 not created backup for volume 0
    And Assert recurringjob 0 not created backup for volume 1
    And Wait for recurringjob 0 created systembackup to reach Ready state

Test System Backup Recurring Job When volume-backup-policy is if-not-present
    [Tags]    recurring-job    system-backup-recurring-job
    Given Create volume 0 with    size=2Gi    numberOfReplicas=1    dataEngine=${DATA_ENGINE}
    And Create volume 1 with    size=2Gi    numberOfReplicas=1    dataEngine=${DATA_ENGINE}
    And Attach volume 1
    And Wait for volume 1 healthy
    When Create system-backup recurringjob 0    parameters={"volume-backup-policy":"if-not-present"}

    Then Assert recurringjob 0 created backup for volume 0
    And Assert recurringjob 0 created backup for volume 1
    And Wait for recurringjob 0 created systembackup to reach Ready state
```

roger-ryao: Regarding my previous suggestion, it may need to be postponed because I found that the test case `Test System Backup Recurring Job When volume-backup-policy is if-not-present` failed during execution.

roger-ryao (approved):
LGTM

PR #2720, opened 2025-08-25
test(robot): test system backup recurring job with both attached and detached volumes

yangchiu:

Issue https://github.com/longhorn/longhorn-tests/pull/2712#discussion_r2290032380

test system backup recurring job with both attached and detached volumes


roger-ryao (approved):
LGTM

PR #2781, opened 2025-09-19
test(robot): fix not all v2 volumes are detached before upgrading in prerelease check test case

yangchiu:

Issue N/A

fix not all v2 volumes are detached before upgrading in prerelease check test case


yangchiu:
@mergify backport v1.10.x

PR #2786, opened 2025-09-22
test(robot): fix Test system upgrade with a new storage class being default on v2 volumes by detaching v2 volumes before upgrading

yangchiu:

Issue N/A

fix Test system upgrade with a new storage class being default on v2 volumes by detaching v2 volumes before upgrading


chriscchien (approved):
LGTM

PR #2804, opened 2025-09-26
ci: fix issues in rancher chart test pipeline

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11853

fix issues in rancher chart test pipeline

(1) expose longhorn-ui node-port to download support bundles
(2) set max retry for downloading support bundles
(3) adjust workflow to publish test report before downloading support bundle so it won't be blocked by stucking downloading
(4) add wait_for_rancher_webhook in installing longhorn with `terraform apply` because `terraform apply` could accidentally fail due to rancher webhook not found
(5) wait 1 minute before uninstalling longhorn with `terraform destroy` because the dependent terraform.lock.hcl file may need some time to sync
(6) refine check_longhorn_crd_removed codeflow by adding wait mechanism and print more information
(7) manually uninstall longhorn-crd after uninstalling longhorn with `terraform destroy`

### DIFF
--- a/e2e/libs/longhorn_deploy/base.py
+++ b/e2e/libs/longhorn_deploy/base.py
@@ -24,9 +24,19 @@ class Base(ABC):
         return NotImplemented
 
     def check_longhorn_crd_removed(self):
-        all_crd = k8s.get_all_custom_resources()
-        for crd in all_crd.items:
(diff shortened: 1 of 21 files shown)


roger-ryao (approved):
LGTM

PR #2915, opened 2025-11-18
test(robot): add test case Test Delete Instance Manager Of Single Replica Volume

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11525 https://github.com/longhorn/longhorn/issues/13302

add test case Test Delete Instance Manager Of Single Replica Volume


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.10.x

PR #3294, opened 2026-04-07
test(robot): fix RWX strict-local incompatible volume test

roger-ryao:
Observed during verification of https://github.com/longhorn/longhorn/issues/12316:
robot regression test case `Test RWX Volume Without Migratable Should Be Incompatible With Strict-Local` failed unexpectedly.
Screenshot_20260407_193930

Use `Run Keyword And Expect Error` with `retry=False` so the webhook rejection is treated as expected instead of retrying for ~20 minutes.

Also add a None guard in `wait_for_volume_state` to produce a clear AssertionError instead of TypeError when the volume does not exist.


yangchiu (approved):

yangchiu:
@mergify backport v1.11.x

PR #3316, opened 2026-04-20
test(robot): add test case Test Block Volume Should Reject Filesystem Trim

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/12071

add test case Test Block Volume Should Reject Filesystem Trim


chriscchien (approved):

yangchiu:
@mergify backport v1.11.x

PR #3325, opened 2026-04-23
test(robot): add test case Test RWX Failover And Auto Salvage When Volume Is Faulted

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9091

add test case Test RWX Failover And Auto Salvage When Volume Is Faulted

### DIFF
--- a/e2e/libs/metrics/metrics.py
+++ b/e2e/libs/metrics/metrics.py
@@ -217,47 +217,53 @@ def check_longhorn_components_memory_cpu_usage():
             usage["backing-image-manager"]["memory"],
             cold_usage["backing-image-manager"]["memory"])
 
-    is_high_resource_consumption("csi-attacher", "cpu",
-        usage["csi-attacher"]["cpu"],
-        old_usage["csi-attacher"]["cpu"])
-    is_high_resource_consumption("csi-attacher", "memory",
-        usage["csi-attacher"]["memory"],
-        old_usage["csi-attacher"]["memory"])
+    if "csi-attacher" in usage and "csi-attacher" in old_usage:
+        is_high_resource_consumption("csi-attacher", "cpu",
+            usage["csi-attacher"]["cpu"],
+            old_usage["csi-attacher"]["cpu"])
+        is_high_resource_consumption("csi-attacher", "memory",
+            usage["csi-attacher"]["memory"],
+            old_usage["csi-attacher"]["memory"])
 
-    is_high_resource_consumption("csi-provisioner", "cpu",
-        usage["csi-provisioner"]["cpu"],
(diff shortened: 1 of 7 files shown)


e2e/libs/metrics/metrics.py
```diff
@@ -207,47 +207,53 @@ def check_longhorn_components_memory_cpu_usage():
             usage["backing-image-manager"]["memory"],
             cold_usage["backing-image-manager"]["memory"])
```
roger-ryao:
```suggestion
            old_usage["backing-image-manager"]["memory"])
```
Typo

roger-ryao (changes requested):

e2e/tests/regression/test_rwx_volume.robot
```diff
@@ -137,35 +136,67 @@ Test ShareManager Status Current Image After Upgrade
...
+    And Setting auto-salvage is set to true
+    And Create storageclass longhorn-test with    numberOfReplicas=3    dataEngine=${DATA_ENGINE}
+    And Create statefulset 0     volume_type=RWX    sc_name=longhorn-test
+    And Wait for volume of statefulset 0 healthy
+    And Update volume of statefulset 0 replica count to 1
+    And Delete replica of statefulset 0 volume on all replica node
+    And Write 1 MB data to file data.bin in statefulset 0
+    And Get Longhorn components resource usage
+
+    When Power off volume node of statefulset 0
```
roger-ryao, line 194: If I remember correctly, test cases involving actions like `stopping kubelet` or `Power off` should be categorized under negative tests. Does this classification need to be adjusted?

e2e/libs/metrics/metrics.py
yangchiu: Updated. Thank you!

e2e/tests/regression/test_rwx_volume.robot
yangchiu: I think it depends on whether we are` testing a corner case or intentionally injecting negative factors to test resiliency` or `verifying a behavior`. What do you think this test case should be categorized as? I am fine with either.

roger-ryao: Let's add the `node-down` tag to the case for now and leave the case location as is. Since this issue originated from an internal ticket, we can reorganize it based on the case tags.

yangchiu: Updated. Thank you.

roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.11.x

PR #3327, opened 2026-04-23
test: revert aws instance type to the lower-cost t2.xlarge

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/10949

t3.xlarge is too expensive, revert to using t2.xlarge instead.

To test nvme type driver, use arm64.


derekbit (approved):

yangchiu:
@mergify backport v1.11.x

PR #3328, opened 2026-04-24
test(robot): add io error test cases

chriscchien:

Add test cases
- Replica Fails When Disk Has I/O Errors

Issue longhorn/longhorn#12288

Known issue for v2 data engine https://github.com/longhorn/longhorn/issues/13354

### DIFF
--- a/e2e/libs/keywords/instancemanager_keywords.py
+++ b/e2e/libs/keywords/instancemanager_keywords.py
@@ -2,7 +2,6 @@ from instancemanager import InstanceManager
 
 from utility.utility import logging
 
-
 class instancemanager_keywords:
 
     def __init__(self):
--- a/e2e/libs/keywords/io_keywords.py
+++ b/e2e/libs/keywords/io_keywords.py
@@ -1,0 +1,141 @@
+from utility.utility import logging
+from node_exec import NodeExec
+
+import time
+
+class io_keywords:
+
+    def __init__(self):
+        pass
+
+    def setup_dm_linear_device_from_block_disk(self, block_disk_path, dm_device_name, node_name):
+        """
+        Setup a device mapper linear device from a raw block device.
+        """
+        logging(f"Setting up dm-linear device {dm_device_name} from block disk {block_disk_path} on node {node_name}")
+
+        cmd = f"test -b {block_disk_path}"
+        result = NodeExec(node_name).issue_cmd(cmd)
+        if result and ("No such" in result or "not a block" in result):
+            raise Exception(f"Block device {block_disk_path} does not exist or is not a block device")
+
+        cmd = f"blockdev --getsz {block_disk_path}"
+        sectors = NodeExec(node_name).issue_cmd(cmd).strip()
+        if not sectors or not sectors.isdigit():
+            raise Exception(f"Failed to get device size in sectors. Result: {sectors}")
+
+        cmd = f'echo "0 {sectors} linear {block_disk_path} 0" | dmsetup create --noudevsync {dm_device_name}'
+        create_result = NodeExec(node_name).issue_cmd(cmd)
+        if create_result and ("failed" in create_result.lower() or "error" in create_result.lower()):
+            raise Exception(f"Failed to create dm device. Result: {create_result}")
+
+        cmd = f"dmsetup table --noudevsync {dm_device_name}"
+        table = NodeExec(node_name).issue_cmd(cmd)
+        if "failed" in table.lower() or "No such device" in table:
+            raise Exception(f"Device verification failed. Table result: {table}")
+
+        result = {
+            "real_dev": block_disk_path,
+            "sectors": sectors,
+            "dm_device": f"/dev/mapper/{dm_device_name}"
+        }
+        logging(f"Successfully created dm device {dm_device_name}: {result}")
+        return result
+
+    def format_and_mount_dm_device(self, dm_device_name, mount_point, node_name):
+        """Format dm device as ext4 and mount it."""
+        logging(f"Formatting and mounting dm device {dm_device_name} to {mount_point} on node {node_name}")
+
+        dm_path = f"/dev/mapper/{dm_device_name}"
+
+        # Format device as ext4
+        cmd = f"mkfs.ext4 -F {dm_path}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        # Create mount point
+        cmd = f"mkdir -p {mount_point}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        # Mount device with errors=continue option
+        cmd = f"mount -t ext4 -o errors=continue {dm_path} {mount_point}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        logging(f"Successfully formatted and mounted {dm_device_name} to {mount_point}")
+
+    def force_unmount_dm_device(self, mount_point, node_name):
+        """Force unmount a dm device mount point by killing processes and using lazy unmount."""
+        logging(f"Force unmounting {mount_point} on node {node_name}")
+
+        # Kill all processes using the mount point
+        cmd = f"fuser -km {mount_point} || true"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        # Wait for processes to terminate
+        timeout = 10
+        start_time = time.time()
+        while time.time() - start_time < timeout:
+            cmd = f"fuser {mount_point} 2>/dev/null"
+            result = NodeExec(node_name).issue_cmd(cmd)
+            if not result or not result.strip():
+                logging(f"All processes using {mount_point} have terminated")
+                break
+            time.sleep(0.5)
+        else:
+            logging(f"Warning: Timeout waiting for processes to terminate on {mount_point}")
+
+        # Try normal unmount first, fallback to lazy unmount
+        cmd = f"umount {mount_point} || umount -l {mount_point}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        logging(f"Successfully force unmounted {mount_point}")
+
+    def switch_dm_device_to_error(self, dm_device_name, sectors, node_name):
+        """Switch device mapper device from linear to error target. This will make all I/O operations fail."""
+        logging(f"Switching dm device {dm_device_name} to error mode on node {node_name}")
+
+        # Load the error table (inactive)
+        cmd = f'echo "0 {sectors} error" | dmsetup load --noudevsync {dm_device_name}'
+        NodeExec(node_name).issue_cmd(cmd)
+
+        # Suspend the device without locking filesystem
+        cmd = f"dmsetup suspend --noudevsync --nolockfs {dm_device_name}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        # Resume device to activate the loaded table
+        cmd = f"dmsetup resume --noudevsync {dm_device_name}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        logging(f"Successfully switched {dm_device_name} to error mode")
+
+    def switch_dm_device_to_linear(self, dm_device_name, block_device_path, sectors, node_name):
+        logging(f"Switching dm device {dm_device_name} to linear mode on node {node_name}")
+
+        # Suspend the device
+        cmd = f"dmsetup suspend --noudevsync {dm_device_name}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        # Reload with linear mapping
+        cmd = f'echo "0 {sectors} linear {block_device_path} 0" | dmsetup reload {dm_device_name}'
+        NodeExec(node_name).issue_cmd(cmd)
+
+        # Resume without flushing
+        cmd = f"dmsetup resume --noudevsync --noflush {dm_device_name}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        cmd = f"dmsetup status {dm_device_name}"
+        status = NodeExec(node_name).issue_cmd(cmd)
+        logging(f"dm device status after switching to linear mode: {status}")
+
+    def cleanup_dm_device(self, dm_device_name, node_name):
+        logging(f"Cleaning up dm device {dm_device_name} on node {node_name}")
+        cmd = f"dmsetup remove --noudevsync --deferred {dm_device_name}"
+        NodeExec(node_name).issue_cmd(cmd)
+
+        logging(f"Successfully cleaned up dm device {dm_device_name}")
+
+    def wipe_block_device_signatures(self, block_device_path, node_name):
+        logging(f"Wiping filesystem signatures from {block_device_path} on node {node_name}")
+
+        cmd = f"wipefs -a {block_device_path}"
+        NodeExec(node_name).issue_cmd(cmd)
+        logging(f"Successfully wiped signatures from {block_device_path}")
--- a/e2e/tests/negative/io_error.robot
+++ b/e2e/tests/negative/io_error.robot
@@ -1,0 +1,112 @@
+*** Settings ***
+Documentation    Disk I/O Error Test Cases
+
+Test Tags    negative    disk    io-error
+
+Resource    ../keywords/variables.resource
+Resource    ../keywords/common.resource
+Resource    ../keywords/volume.resource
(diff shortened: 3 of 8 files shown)


e2e/libs/keywords/io_keywords.py
```diff
@@ -0,0 +1,166 @@
+from utility.utility import logging
+from node_exec import NodeExec
+
+
+class io_keywords:
+
+    def __init__(self):
+        pass
+
+    def _get_real_device_path(self, volume_name, node_name):
```
yangchiu: It may not be necessary to create a sub-function for each command. Try merge them into one.

e2e/libs/keywords/io_keywords.py
```diff
@@ -0,0 +1,166 @@
...
+        """Get the real device path from /dev/longhorn/{volume_name}."""
+        cmd = f"readlink -f /dev/longhorn/{volume_name}"
```
yangchiu: When would the real device path be different from `/dev/longhorn/{volume_name}`?

e2e/libs/keywords/io_keywords.py
```diff
@@ -0,0 +1,166 @@
...
+        """Get device size in sectors."""
+        cmd = f"blockdev --getsz {device_path}"
+        sectors = NodeExec(node_name).issue_cmd(cmd).strip()
+        if not sectors or not sectors.isdigit():
+            raise Exception(f"Failed to get device size in sectors. Result: {sectors}")
+        return sectors
+
+    def _create_dm_linear_device(self, dm_device_name, real_dev, sectors, node_name):
+        """Create dm-linear device mapping."""
+        cmd = f'echo "0 {sectors} linear {real_dev} 0" | dmsetup create --noudevsync {dm_device_name}'
```
yangchiu: Do you know why we need `linear` and `--noudevsync` options?

e2e/libs/keywords/io_keywords.py
```diff
@@ -0,0 +1,166 @@
...
+        """Force unmount a dm device mount point by killing processes and using lazy unmount."""
+        logging(f"Force unmounting {mount_point} on node {node_name}")
+
+        self._kill_processes_using_mount(mount_point, node_name)
+        self._wait_for_processes_to_terminate()
+        self._unmount_with_fallback(mount_point, node_name)
+
+        logging(f"Successfully force unmounted {mount_point}")
+
+    def _load_error_table(self, dm_device_name, sectors, node_name):
```
yangchiu: Same as above. Maybe these sub-functions can be merged into `switch_dm_device_to_error`.

e2e/libs/keywords/io_keywords.py
```diff
@@ -0,0 +1,166 @@
...
+        self._mount_device(dm_path, mount_point, node_name)
+
+        logging(f"Successfully formatted and mounted {dm_device_name} to {mount_point}")
+
+    def _kill_processes_using_mount(self, mount_point, node_name):
+        """Kill all processes using the mount point."""
+        cmd = f"fuser -km {mount_point} 2>/dev/null || true"
+        NodeExec(node_name).issue_cmd(cmd)
+
+    def _wait_for_processes_to_terminate(self):
```
yangchiu: This function doesn't wait for any process to terminate. It simply sleeps.

e2e/tests/negative/io_error.robot
```diff
@@ -0,0 +1,120 @@
...
+    ...                4. Disable default disk on node 0, add dm device as Longhorn disk.
+    ...                5. Create 1GB test volume with 3 replicas, attach to node 0.
+    ...                6. Switch dm device from linear to error mode (dmsetup) to inject I/O errors.
+    ...                7. Write data to test volume to trigger I/O errors on the dm disk replica.
+    ...                8. Verify test volume becomes degraded due to failed replica.
+    ...
+    [Setup]    Setup device mapper environment
+    [Teardown]    Cleanup device mapper environment
+    IF    '${DATA_ENGINE}' == 'v2'
+        Skip    Test only validate on v1 data engine
```
yangchiu: Do you know why this device mapper approach doesn't work for v2 data engine? We even have an existing block device e.g., `/dev/xvdh` for `dmsetup create` and don't need to create a dedicated disk/volume ourselves. Does the error target part not work for v2 data engine?

e2e/tests/negative/io_error.robot
```diff
@@ -0,0 +1,120 @@
...
+    END
+    
+    Given Create volume 2 with    size=1Gi    numberOfReplicas=3    dataEngine=v1
+    And Attach volume 2 to node 0
+    And Wait for volume 2 healthy
+
+    When Switch dm device to error mode    dm_device_name=${dm_device_name}    node_id=0
+    And Write data 0 500 MB to volume 2
+    Then Wait for volume 2 degraded
+
```
yangchiu: Verify the replica is failed and the data is intact.

e2e/tests/negative/io_error.robot
```diff
@@ -0,0 +1,120 @@
...
+    ${dm_info} =    Create and mount dm disk from volume ${base_volume_name} as ${dm_device_name} on node ${node_id} to ${mount_path}
+    Log    Device mapper info: ${dm_info}
+
+    When Disable node ${node_id} default disk
+    And Add filesystem disk ${disk_name} to node ${node_id} with path ${mount_path}
+
+Cleanup dm disk environment on node ${node_id} for ${data_engine}
+    [Documentation]    Cleanup device mapper disk and restore default disk.
+    ...                Handles disk scheduling disable, volume deletion, unmounting (v1 only), and cleanup.
+    When Disable disk ${disk_name} scheduling on node ${node_id}
```
yangchiu: Will it work if we leave the disk/volume cleanup to the existing cleanup functions in `Cleanup test resources` and only perform  `dmsetup remove` at the end?

e2e/tests/negative/io_error.robot
```diff
@@ -0,0 +1,120 @@
...
+*** Keywords ***
+Setup device mapper environment
+    Set up test environment
+    Setup dm disk environment on node 0
+
+Cleanup device mapper environment
+    Cleanup dm disk environment on node 0 for v1
+    Cleanup test resources
+
+Setup dm disk environment on node ${node_id}
```
yangchiu, line 26: Would you consider moving this function to `io.resource`? Also, maybe `Add dm device as Longhorn disk on node` is clearer?

e2e/tests/negative/io_error.robot
yangchiu: I mean, what if we omit the following steps:
```
And Delete volume 2    wait=False
Then Force unmount dm disk at ${mount_path} on node ${node_id}
And Wait for volume 2 deleted
And Delete volume ${disk_name}
And Cleanup mounted dm disk ${dm_device_name} at ${mount_path} on node ${node_id}
And Delete disk ${disk_name} on node ${node_id}    
```
and only keep `_remove_dm_device`?

yangchiu (changes requested):

e2e/tests/negative/io_error.robot
chriscchien: After test case complete, volume 2 still have failed replica running on dm disk, it is necessary to perform the steps to clean dm environment

1. Delete volume 2 first, so Longhorn stops using the dm-backed disk.
2. Unmount the dm disk.
3. Remove the dm device.
4. Delete the backing Longhorn volume.
5. Delete the Longhorn disk entry.

e2e/tests/negative/io_error.robot
chriscchien: The device mapper method not work on SPDK because SPDK use user space, I/O requests travel directly from application to the hardware not through kernel https://spdk.io/doc/about.html.

e2e/libs/keywords/io_keywords.py
chriscchien: The `linear` target is needed because the test first needs a normal working dm device backed by the Longhorn volume. Longhorn then uses the mounted filesystem disk normally. Later, we switch the same dm device from the `linear` target to the `error` target to simulate a disk I/O errors at runtime.
```
linear target = transfer I/O to /dev/longhorn/volume
error target  = return EIO to all I/O
```
Because the `dmsetup` commands all executed in pod, without arg `--noudevsync`, the command will hang. (If we run `dmsetup` directly on node, the arg can skip).

e2e/tests/negative/io_error.robot
chriscchien: Updated!
v1 https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7458/
v2 https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7459/

yangchiu:
Revisit this after the release.

yangchiu:
Revisit this after `v1.12.0` release.

e2e/tests/negative/io_error.robot
yangchiu: One question, we still have the IO latency negative factor to implement to simulate slow IO: https://github.com/longhorn/longhorn/issues/12895. Do you have a plan in mind for handling this on both v1 and v2 volumes?

chriscchien: Currently not, due to the known Chaos Mesh issue with Longhorn usage: https://github.com/longhorn/longhorn/issues/12288#issuecomment-4311504669 

I will investigate alternative methods after the recent release tasks are completed.

yangchiu:
More investigation is needed. I'll need some time to clarify the questions raised in the discussion.

chriscchien:
For v1 volume, the device map path is like below in the test code(in the QA meeting, the device mapper is mount directly from `/dev/xvdh`, however below method sould also equivalent for this test because since test not guarantee the extra block disk is always exist, using a  v1 volume as the backing disk source is a workaround)
```
/dev/mapper/${dm_device_name}
  -> linear
  -> /dev/longhorn/${disk_name}  (v1 Longhorn base volume)
  -> ext4
  -> mounted at ${mount_path}
  -> added to Longhorn as filesystem disk ${disk_name}
```

image

For v2 volume, the path should be like below, however there is an issue https://github.com/longhorn/longhorn/issues/13354 need to clarify first
```
/dev/mapper/${dm_device_name}
  -> linear
  -> /dev/xvdh   (actual disk)
```

e2e/tests/negative/io_error.robot
```diff
@@ -0,0 +1,97 @@
...
+    And Attach volume 2 to node 0
+    And Wait for volume 2 healthy
+
+    When Switch dm device to error mode    dm_device_name=${dm_device_name}    node_id=0
+    And Write data 0 500 MB to volume 2
+    Then Wait for volume 2 degraded
+    And Volume 2 should have no running replica on node 0
+    And Check volume 2 data is intact
+
+Replica Fails When Block Disk Has I/O Errors
```
yangchiu: Does this IO error test case still have different steps or implementation for v1 and v2 volume? Can they merge into a single test case?

Even though a extra disk needs to be created for v1 volumes, after this extra setup step, the remaining steps should be the same for both v1 and v2?

Also, although there is an issue, you can still add the test case for v2 volumes so that developers to easily reproduce, investigate and clarify the issue.

yangchiu (changes requested):

chriscchien:
Update

- v1 : Replica Fails When Disk Has I/O Errors + Test Volume Basic)
   - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8287/
- v2 : Replica Fails When Disk Has I/O Errors + Test V2 Volume Basic)
   - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8288/

e2e/libs/keywords/instancemanager_keywords.py
```diff
@@ -1,6 +1,9 @@
 from instancemanager import InstanceManager
 
 from utility.utility import logging
+from utility.utility import get_longhorn_client
```
yangchiu: If the imports are not used, they can be removed.

e2e/tests/negative/io_error.robot
```diff
@@ -0,0 +1,111 @@
...
+    And Disable node ${node_id} default disk
+    And Add filesystem disk ${disk_name} to node ${node_id} with path ${mount_path}
+
+Setup dm disk for v2 on node ${node_id}
+    [Documentation]    Setup dm device as block disk for V2 data engine.
+    ...                Removes default BLOCK disk (which uses /dev/xvdh), creates dm-linear on /dev/xvdh, adds dm device as block disk.
+    When Disable default block disk on node ${node_id}
+    And Delete default block disk on node ${node_id}
+    And Create dm linear device from block device ${DISK_PATH} as ${dm_device_name} on node ${node_id}
+    #${dm_device_path} =    Set Variable    /dev/mapper/${dm_device_name}
```
yangchiu: Unused comment.

e2e/tests/negative/io_error.robot
```diff
@@ -0,0 +1,111 @@
...
+    When Create and mount dm disk from block device ${DISK_PATH} as ${dm_device_name} on node ${node_id} to ${mount_path}
+    And Disable node ${node_id} default disk
+    And Add filesystem disk ${disk_name} to node ${node_id} with path ${mount_path}
+
+Setup dm disk for v2 on node ${node_id}
+    [Documentation]    Setup dm device as block disk for V2 data engine.
+    ...                Removes default BLOCK disk (which uses /dev/xvdh), creates dm-linear on /dev/xvdh, adds dm device as block disk.
+    When Disable default block disk on node ${node_id}
+    And Delete default block disk on node ${node_id}
+    And Create dm linear device from block device ${DISK_PATH} as ${dm_device_name} on node ${node_id}
```
yangchiu, line 55: If the `${DISK_PATH}` is a BDF, e.g., `0000:00:1f.0`, will the dm linear device still work?

yangchiu (changes requested):

e2e/tests/negative/io_error.robot
chriscchien: No, need to find the undetlying disks first for BDF. Currently we are using `/dev/xvdh` `/dev/nvme1n1` so the dm linear device works as expected.

yangchiu: Only using block device path (`/dev/xvdh` or `/dev/nvme1n1`) is just a temporary work around in the test code for the unresolved issue. It makes us skip the `nvme` driver test and potentially miss any related issues. Eventually we need to to enable using `BDF` to test `nvme` driver. 

We should figure out a way to handle this, maybe running a dedicated test during the release, or running the test on another cloud provider, rather than assuming it won't happen.

yangchiu (changes requested):

e2e/tests/negative/io_error.robot
chriscchien: Currently there is a BDF path hard coded in robot test suite, when provision arm64, test framework uses `/dev/nvme1n1` as disk PATH. Therefore I  dont have a environmet to test is. 

Created https://github.com/longhorn/longhorn/issues/13395, I will work on it once test framework support provision with BDF path.

yangchiu: You modified the hard-coded path [before](https://github.com/longhorn/longhorn-tests/pull/3471). I believe you can do the same here to modify it again and create a custom test image to test the BDF path in this test case.

yangchiu (changes requested):

e2e/tests/negative/io_error.robot
chriscchien: The modify is because https://github.com/longhorn/longhorn/issues/13243. Do you plan to let amd64 CI supports BDF when running v2?

yangchiu: > The modify is because [longhorn/longhorn#13243](https://github.com/longhorn/longhorn/issues/13243).

Yes, that's the unresolved issue I was referring to.

> Do you plan to let amd64 CI supports BDF when running v2?

Feel free to work on https://github.com/longhorn/longhorn/issues/13395. There's no need to wait for CI.

chriscchien: If there are no other concerns, could we proceed with this PR and move it forward? THank you.

yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3334, opened 2026-04-27
test(robot): add test cases Test Auto Balance In Unstable Cluster

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/12927 https://github.com/longhorn/longhorn/issues/12508

add test cases Test Auto Balance In Unstable Cluster

### DIFF
--- a/e2e/tests/regression/test_scheduling.robot
+++ b/e2e/tests/regression/test_scheduling.robot
@@ -1,7 +1,7 @@
 *** Settings ***
 Documentation    Scheduling Test Cases
 
-Test Tags    regression    replica
+Test Tags    regression    replica    scheduling
 
 Resource    ../keywords/variables.resource
 Resource    ../keywords/common.resource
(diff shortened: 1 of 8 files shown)


e2e/tests/regression/test_scheduling.robot
```diff
@@ -460,3 +460,112 @@ Test Scheduling Replicas To Different Disks On The Same Node
...
+    ...    9. Wait for a replica to be auto balanced to node 2.
+    ...    10. Verify replicas are stable: one in zone-a and one in zone-b for a sufficient time.
+
+    # Step 1: Set up zones
+    Given Set k8s node 0 zone zone-a
+    And Set k8s node 1 zone zone-a
+    And Set k8s node 2 zone zone-b
+
+    # Step 2: Create unstable node condition on node 2
+    And Stop kubelet on node 2 for 2100 seconds
```
roger-ryao, line 490: This test takes at least 30 minutes to complete. We can add a `long-running` tag so it can be excluded from quick CI runs with `--exclude long-running`.

e2e/tests/regression/test_scheduling.robot
```diff
@@ -460,3 +460,112 @@ Test Scheduling Replicas To Different Disks On The Same Node
...
+    ...    8. Enable scheduling for node 2.
+    ...    9. Wait for a replica to be auto balanced to node 2.
+    ...    10. Verify replicas are stable: one in zone-a and one in zone-b for a sufficient time.
+
+    # Step 1: Set up zones
+    Given Set k8s node 0 zone zone-a
+    And Set k8s node 1 zone zone-a
+    And Set k8s node 2 zone zone-b
+
+    # Step 2: Create unstable node condition on node 2
```
roger-ryao, line 489: ~~If I remember correctly, test cases involving actions like `stopping kubelet` should be categorized under `negative tests`. Does this classification need to be adjusted?~~

Let's add the `kubelet-restart` tag to the case.

e2e/tests/regression/test_scheduling.robot
```diff
@@ -460,3 +460,112 @@ Test Scheduling Replicas To Different Disks On The Same Node
...
+    # Step 2: Create and attach a 2-replica volume with replicas on node 0 and node 1
+    And Disable node 2 scheduling
+    And Create volume 0 with    numberOfReplicas=2    replicaAutoBalance=least-effort    dataEngine=${DATA_ENGINE}
+    And Attach volume 0 to node 0
+    And Wait for volume 0 attached
+    And Wait for volume 0 healthy
+    And Enable node 2 scheduling
+
+    # Step 3: Stop kubelet on node 1 for 35 minutes
+    And Stop kubelet on node 1 for 2100 seconds
```
roger-ryao, line 551: This test takes at least 30 minutes to complete. We can add a `long-running` tag so it can be excluded from quick CI runs with `--exclude long-running`.

e2e/tests/regression/test_scheduling.robot
yangchiu: Is it a good idea to skip a test case just because it takes time? Maybe we can have a `coretest` test set, like in pytest, if we only need a basic test.

e2e/tests/regression/test_scheduling.robot
yangchiu: Added. Thank you!

e2e/tests/regression/test_scheduling.robot
yangchiu: Add the tag for now. Try not to rely on it.

roger-ryao: I understand your concern, but I don't have a great idea at the moment.
 
The only benefit I can think of right now is that it would help us more quickly identify which test cases are time consuming when we need to gather statistics in the future. 

From my personal experience, 20 minutes is probably the limit for a test case, and anything over 30 minutes is really too long. 

As you mentioned, using a tag like `coretest` could address the need for users who want a quick and simple way to test basic functionality.

roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.11.x

PR #3335, opened 2026-04-27
fix(uninstall): use manifests in the specified repo and branch

derekbit:

Issue #


yangchiu (approved):

yangchiu:
@mergify backport v1.11.x

yangchiu:
@derekbit It would be better to create a ticket to illustrate the issue so that all members can easily understand the context.

derekbit:
> @derekbit It would be better to create a ticket to illustrate the issue so that all members can easily understand the context.

Thank you. Created a ticket https://github.com/longhorn/longhorn/issues/13002 for the fix.

PR #3348, opened 2026-04-29
test(robot): fail the cpu and memory usage monitoring only if it's stably high

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13000 https://github.com/longhorn/longhorn/issues/10932

fail the cpu and memory usage monitoring only if it's stably high

### DIFF
--- a/e2e/tests/continuous_io.robot
+++ b/e2e/tests/continuous_io.robot
@@ -79,6 +79,14 @@ Continuous IO Test
         And Check deployment rwx-nfs-soft-mount-deploy data in file data.txt is intact
         And Check deployment bi-deploy data in file data.txt is intact
         ${current}=    Get Time    epoch
-
-        And Check Longhorn components resource usage
     END
+
+    And Stop writing data to pod of deployment rwo-deploy
+    And Stop writing data to pod of deployment rwx-deploy
+    And Stop writing data to pod of deployment strict-local-deploy
+    And Stop writing data to pod of deployment rwx-nfs-4-2rwx-nfs-4-2-deploy
+    And Stop writing data to pod of deployment rwx-nfs-hard-mount-deploy
+    And Stop writing data to pod of deployment rwx-nfs-soft-mount-deploy
+    And Stop writing data to pod of deployment bi-deploy
+
+    And Check Longhorn components resource usage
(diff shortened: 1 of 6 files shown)


e2e/tests/continuous_io.robot
```diff
@@ -79,6 +79,14 @@ Continuous IO Test
...
+
+    And Stop writing data to pod of deployment rwo-deploy
+    And Stop writing data to pod of deployment rwx-deploy
+    And Stop writing data to pod of deployment strict-local-deploy
+    And Stop writing data to pod of deployment rwx-nfs-4-2rwx-nfs-4-2-deploy
+    And Stop writing data to pod of deployment rwx-nfs-hard-mount-deploy
+    And Stop writing data to pod of deployment rwx-nfs-soft-mount-deploy
+    And Stop writing data to pod of deployment bi-deploy
+
+    And Check Longhorn components resource usage
```
roger-ryao, line 92: Since we don't have baseline numbers yet, we could  measure component resource usage at three points:

1. Before stopping data writes
2. Immediately after stopping data writes
3. After stopping data writes 30 ~ 60 seconds

```robot
And Check Longhorn components resource usage    # Before stopping (peak under I/O load)

And Stop writing data to pod of deployment rwo-deploy
And Stop writing data to pod of deployment rwx-deploy
...

And Check Longhorn components resource usage    # Immediately after stopping

And Sleep    30s
And Check Longhorn components resource usage    # ~30-60s after stopping (recovery check)
```



```robot
Check Longhorn components resource usage
    [Arguments]    ${fail_on_high}=${False}
    ${status}    ${msg} =    Run Keyword And Ignore Error    check_longhorn_components_memory_cpu_usage
    IF    '${status}' == 'FAIL'
        IF    ${fail_on_high}
            Fail    ${msg}
        ELSE
            Log    ${msg}    WARN
        END
    END
```

All three use `fail_on_high=${False}` to just log the values without failing the test. 
Once we have enough data, we can decide the right threshold and when to actually fail.

yangchiu: Based on https://github.com/longhorn/longhorn/issues/12988#issuecomment-4332086213, `the values depends on the CPU model and IO pattern`. Even if you collect data through a large number of test runs to obtain statistically significant values, it may be still not persuasive enough for devs to regard it as an issue.

At this moment, the furthest we can go is to focus only on `if the values can go back to normal after completing the IO testing, it should be fine` according to https://github.com/longhorn/longhorn/issues/12988#issuecomment-4332086213.

roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.12.x v1.11.x

PR #3355, opened 2026-04-30
fix(dr,ha): make flaky test stable

davidcheng0922:

https://github.com/longhorn/longhorn/issues/12606
https://github.com/longhorn/longhorn/issues/9277

- For the DR volume test, it occasionally gets stuck during “Attach volume.” To avoid failures caused by attaching to an unhealthy node, I updated it to “And Attach volume 1 to a healthy node”.

- For ha regression test, https://github.com/longhorn/longhorn/issues/12606#issuecomment-4332181283


yangchiu (approved):

yangchiu:
@mergify backport v1.11.x

PR #3393, opened 2026-05-12
test(robot): add test case Test Encrypted Volume After Node Reboot And Instance Manager Crash

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11558

add test case Test Encrypted Volume After Node Reboot And Instance Manager Crash

### DIFF
--- a/e2e/tests/negative/instance_manager_crash.robot
+++ b/e2e/tests/negative/instance_manager_crash.robot
@@ -10,6 +10,7 @@ Resource    ../keywords/storageclass.resource
 Resource    ../keywords/persistentvolumeclaim.resource
 Resource    ../keywords/deployment.resource
 Resource    ../keywords/workload.resource
+Resource    ../keywords/host.resource
 Resource    ../keywords/longhorn.resource
 Resource    ../keywords/volume.resource
 Resource    ../keywords/backup.resource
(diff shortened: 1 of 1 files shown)


e2e/tests/negative/instance_manager_crash.robot
```diff
@@ -116,3 +117,11 @@ Crash Single Instance Manager While RWX Volume Backup Is Restoring
 Crash Single Instance Manager While RWX Encrypted Volume Backup Is Restoring
     [Tags]    encrypted    rwx    volume    backup-restore
     Test Instance Manager Crash During Backup Restoration    access_mode=RWX    encrypted=${True}
+
+Test Encrypted Volume After Node Reboot And Instance Manager Crash
+    [Tags]    encrypted    rwo
```
roger-ryao, line 122: It would be great to include 

```robot
 [Documentation]     Issue: https://github.com/longhorn/longhorn/issues/11510
```

This will help us quickly understand the background and origin of the test case.

yangchiu: Yes. Thanks for the reminder!

roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.12.x

PR #3396, opened 2026-05-13
test(robot): fix flaky test case DR Volume Node Reboot During Initial Restoration

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13047

fix flaky test case DR Volume Node Reboot During Initial Restoration


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3397, opened 2026-05-13
test(robot): check dr volume is expanded after source volume is expanded

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/9278#issuecomment-4386640574

check dr volume is expanded after source volume is expanded


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3423, opened 2026-05-21
test(robot): add case Test V2 Orphan Replica Cleanup After Node Power Cycle

chriscchien:

Issue longhorn/longhorn#6046
- [master](https://10.115.5.5/job/private/job/longhorn-e2e-test/416/):
- [v1.12.x](https://10.115.5.5/job/private/job/longhorn-e2e-test/417/):

### DIFF
--- a/e2e/keywords/host.resource
+++ b/e2e/keywords/host.resource
@@ -73,6 +73,11 @@ Power on off nodes
         Remove Values From List    ${powered_off_nodes}    ${powered_off_node}
     END
 
+Power on node ${node_id}
+    ${powered_on_node} =    get_node_by_index    ${node_id}
+    power_on_node_by_name    ${powered_on_node}
+    Remove Values From List    ${powered_off_nodes}    ${powered_on_node}
+
 Power off node ${node_id}
     ${powered_off_node} =     get_node_by_index    ${node_id}
     Append to list      ${powered_off_nodes}     ${powered_off_node}
--- a/e2e/keywords/replica.resource
+++ b/e2e/keywords/replica.resource
@@ -31,3 +31,18 @@ There should be no replica of volume ${volume_id} on node ${node_id} disk ${disk
     ${node_name} =    get_node_by_index    ${node_id}
     ${disk_uuid} =    get_disk_uuid    ${node_name}    ${disk_name}
     wait_for_disk_replica_count    volume_name=${volume_name}    node_name=${node_name}    disk_uuid=${disk_uuid}    count=0
+
+Delete replica
+    [Arguments]    ${replica_name}
+    [Documentation]    Delete a replica custom resource by its exact name using kubectl.
+    delete_replica_by_name    ${replica_name}
+
+Verify replica lvol is deleted from SPDK lvol list on node ${node_id}
+    [Arguments]    ${replica_name}
+    ${node_name} =    get_node_by_index    ${node_id}
+    verify_replica_lvol_deleted_from_spdk_lvol    ${node_name}    ${replica_name}
+
+Verify replica lvol exists in SPDK lvol list on node ${node_id}
+    [Arguments]    ${replica_name}
+    ${node_name} =    get_node_by_index    ${node_id}
+    verify_replica_lvol_exists_in_spdk_lvol    ${node_name}    ${replica_name}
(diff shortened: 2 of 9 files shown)


e2e/keywords/host.resource
```diff
@@ -73,6 +73,10 @@ Power on off nodes
         Remove Values From List    ${powered_off_nodes}    ${powered_off_node}
     END
 
+Power on node ${node_id}
+    ${powered_on_node} =    get_node_by_index    ${node_id}
+    Run keyword And Ignore Error    power_on_node_by_name    ${powered_on_node}
+
```
yangchiu, line 80: `Remove Values From List    ${powered_off_nodes}`

e2e/keywords/host.resource
```diff
@@ -73,6 +73,10 @@ Power on off nodes
         Remove Values From List    ${powered_off_nodes}    ${powered_off_node}
     END
 
+Power on node ${node_id}
+    ${powered_on_node} =    get_node_by_index    ${node_id}
+    Run keyword And Ignore Error    power_on_node_by_name    ${powered_on_node}
```
yangchiu: Since this is not a general cleanup function that powers on all previously powered-off nodes, but instead explicitly targets a specific node, the `Run Keyword And Ignore Error` should be omitted.

e2e/libs/keywords/replica_keywords.py
```diff
@@ -17,3 +19,9 @@ def wait_for_disk_replica_count(self, volume_name=None, node_name=None, disk_uui
 
     def get_replica_names(self, volume_name, numberOfReplicas=3):
         return self.replica.get_replica_names(volume_name, numberOfReplicas)
+
+    def delete_replica_cr_by_command(self, replica_name, namespace="longhorn-system"):
```
yangchiu: There is an existing function `delete_replica_by_name`. You can refine it to use `kubectl`.

e2e/libs/instancemanager/instancemanager.py
```diff
@@ -168,3 +168,47 @@ def get_instance_manager_pod_on_node(self, node_name, engine_type):
         label_selector = f"longhorn.io/component=instance-manager,longhorn.io/data-engine={engine_type},longhorn.io/node={node_name}"
         ims = list_pods(constant.LONGHORN_NAMESPACE, label_selector)
         return ims[0].metadata.name
+
+    def get_spdk_lvols(self, node_name):
+        logging(f"Getting SPDK lvols on node {node_name}")
+        im_pod_name = self.get_instance_manager_pod_on_node(node_name, "v2")
+        exec_cmd = [
```
yangchiu: We can use the string format like:
```
cmd = f"kubectl -n {namespace} delete replica {replica_name}"
```
for consistency.

e2e/keywords/replica.resource
```diff
@@ -31,3 +31,18 @@ There should be no replica of volume ${volume_id} on node ${node_id} disk ${disk
     ${node_name} =    get_node_by_index    ${node_id}
     ${disk_uuid} =    get_disk_uuid    ${node_name}    ${disk_name}
     wait_for_disk_replica_count    volume_name=${volume_name}    node_name=${node_name}    disk_uuid=${disk_uuid}    count=0
+
+Delete replica CR
```
yangchiu: From my understanding, when we try to delete a replica in robot test, it always delete the replica `CR`. Maybe the `CR` suffix can be removed from the keyword, and existing functions can be reused.

yangchiu (changes requested):

chriscchien:
Updated

- `Test V2 Orphan Replica Cleanup After Node Power Cycle`
   - [master-head](https://10.115.5.5/job/private/job/longhorn-e2e-test/464/)
   - [v1.12.x](https://10.115.5.5/job/private/job/longhorn-e2e-test/463/)

- `Test Replica Rebuilding Per Volume Limit`(Double checked because of the change to `When Delete 4 replicas of volume 0`, which uses `delete_replica_by_name`.)
   - [master-head](https://10.115.5.5/job/private/job/longhorn-e2e-test/465/)

e2e/libs/keywords/replica_keywords.py
```diff
@@ -1,4 +1,6 @@
 from replica import Replica
+from utility.utility import logging
```
yangchiu: Remove these imports of they're not used.

yangchiu (changes requested):

e2e/libs/keywords/replica_keywords.py
chriscchien: Updated!

yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3432, opened 2026-05-25
test: fix upgrade engine image test when installing via helm chart

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13192

fix upgrade engine image test when installing via helm chart

### DIFF
--- a/e2e/libs/volume/crd.py
+++ b/e2e/libs/volume/crd.py
@@ -4,15 +4,11 @@ import yaml
 from kubernetes import client
 from kubernetes.client.rest import ApiException
 
-from engine import Engine
-from enginefrontend import EngineFrontend
-
 from node_exec import NodeExec
 
 from utility.constant import LABEL_TEST
 from utility.constant import LABEL_TEST_VALUE
 import utility.constant as constant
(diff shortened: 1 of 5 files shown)


e2e/libs/volume/crd.py
```diff
@@ -30,9 +26,6 @@ def __init__(self):
         super().__init__()
         self.core_api = client.CoreV1Api()
         self.obj_api = client.CustomObjectsApi()
-        self.retry_count, self.retry_interval = get_retry_count_and_interval()
```
chriscchien, line 33: There are some `time.sleep(self.retry_interval)` exist in this file, is remove`self.retry_count, self.retry_interval = get_retry_count_and_interval()` safe?

chriscchien (approved):

e2e/libs/volume/crd.py
yangchiu: It's in base.py

yangchiu:
@mergify backport v1.12.x

roger-ryao:
@mergify backport v1.11.x

PR #3434, opened 2026-05-25
test(robot): refine wait_for_workload_pod_(not)_on_node function to check all pods of a workload instead of only the first one

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13142

refine wait_for_workload_pod_(not)_on_node function to check all pods of a workload instead of only the first one


chriscchien:
Hi @yangchiu , could you check if `wait_for_workload_pod_on_node`  also need to include the same update? thank you.

chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3456, opened 2026-05-28
test(robot): skip backing image tests by data engine

chriscchien:

Issue N/A

Some test cases involve v2 volumes while still using v1 backing images, use `--exclude backing-image` will miss the coverage. Skip unsupported cases directly in the test case instead.


derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #3469, opened 2026-06-01
fix: use v1 block-mode volumes as host disks

derekbit:

Issue #

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -6529,10 +6529,11 @@ def create_rwx_volume_with_storageclass(client,
     return volume_name
 
 
-def create_volume(client, vol_name, size, node_id, r_num):
+def create_volume(client, vol_name, size, node_id, r_num,
+                  data_engine=DATA_ENGINE):
     volume = client.create_volume(name=vol_name, size=size,
                                   numberOfReplicas=r_num,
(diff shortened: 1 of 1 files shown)


manager/integration/tests/common.py
```diff
@@ -6563,10 +6564,12 @@ def cleanup_volume_by_name(client, vol_name):
 
 
 def create_host_disk(client, vol_name, size, node_id):
-    # create a single replica volume and attach it to node
-    volume = create_volume(client, vol_name, size, node_id, 1)
+    # create a single replica v1 volume and attach it to node
+    volume = create_volume(client, vol_name, size, node_id, 1,
+                           data_engine="v1")
```
chriscchien, line 6570: If `data_engine` is explicitly set to `v1`, will this change affect test cases when they are run with the v2 data engine, such as `test_node_disk_update`?

derekbit: Why does it matter?

derekbit: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/11040/parameters/a
All test cases in test_node.py are passed. I think they are not impacted.

chriscchien (approved):

derekbit:
@mergify backport v1.12.x

PR #3471, opened 2026-06-02
test(pytest): use AIO disk on arm64 machine

chriscchien:

Issue longhorn/longhorn#13243

Use AIO disk on arm64 machine instead of BDF path disk when running test on arm64 machine

arm64 v2 core test https://10.115.5.5/job/private/job/longhorn-tests-regression/476/


derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #3473, opened 2026-06-03
fix: wait for replica creation in best-effort data locality test

carterli0407-cell:
Issue [#13222](https://github.com/longhorn/longhorn/issues/13222)

This PR fixes the flaky failure in `test_best_effort_data_locality`.

The test creates a volume with `dataLocality=best-effort` and `numberOfReplicas=1`. The expected behavior is that the replica object is created first while the volume is detached, remains unscheduled, and then gets scheduled to the local node after the volume is attached.

The previous test logic asserted the replica count inside the retry loop. As a result, the test could fail on the first check if the Longhorn volume controller had not finished creating the replica CR yet.

This PR separates the replica creation wait from the data-locality validation. The test now waits until the expected replica count is observed before checking that the replica remains unscheduled before attachment.

The support bundle showed that the replica CR was created shortly after volume creation, so this is a test-side timing issue rather than a Longhorn scheduling failure.

Verified with Jenkins:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/11083/
```text
-k test_best_effort_data_locality --count 10

### DIFF
--- a/manager/integration/tests/test_scheduling.py
+++ b/manager/integration/tests/test_scheduling.py
@@ -2489,13 +2489,17 @@ def best_effort_data_locality_test(client, volume_name):  # NOQA
                                   numberOfReplicas=number_of_replicas,
                                   dataLocality="best-effort")
 
-    # replica should stay unscheduled until volume is attached
-    for _ in range(10):
-        time.sleep(RETRY_INTERVAL)
+    # Replica should stay unscheduled until volume is attached.
+    wait_for_volume_replica_count(client,
+                                  volume_name,
+                                  number_of_replicas)
+
+    for i in range(10):
         volume = client.by_id_volume(volume_name)
         assert len(volume.replicas) == number_of_replicas
-        assert volume.replicas[0]['hostId'] == ""
+        assert volume.replicas[0]["hostId"] == ""
         assert not volume.replicas[0].running
+        time.sleep(RETRY_INTERVAL)
 
     self_node = get_self_host_id()
     # attach volume to the self_node

### CODE REVIEW

File: manager/integration/tests/test_scheduling.py [hunk 1]
```diff
@@ -2489,13 +2490,13 @@ def best_effort_data_locality_test(client, volume_name):  # NOQA
...
-        time.sleep(RETRY_INTERVAL)
-        volume = client.by_id_volume(volume_name)
-        assert len(volume.replicas) == number_of_replicas
-        assert volume.replicas[0]['hostId'] == ""
-        assert not volume.replicas[0].running
+    # Wait until the replica object is created.
+    wait_for_replica_count(client, volume_name, number_of_replicas)
+
+    # Replica should stay unscheduled until volume is attached.
+    volume = client.by_id_volume(volume_name)
```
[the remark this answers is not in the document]
carterli0407-cell: Updated.

manager/integration/tests/test_scheduling.py
```diff
@@ -64,6 +64,7 @@
 from common import wait_scheduling_failure
 from common import wait_for_rebuild_complete
 from common import wait_for_rebuild_start
+from common import wait_for_replica_count
```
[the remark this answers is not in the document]
carterli0407-cell: Updated.

manager/integration/tests/test_scheduling.py
```diff
@@ -2489,13 +2489,14 @@ def best_effort_data_locality_test(client, volume_name):  # NOQA
...
-        assert volume.replicas[0]['hostId'] == ""
-        assert not volume.replicas[0].running
+    # Wait until the replica object is created.
+    volume = wait_for_volume_replica_count(client,
+                                           volume_name,
+                                           number_of_replicas)
+
+    # Replica should stay unscheduled until volume is attached.
+    assert volume.replicas[0]['hostId'] == ""
+    assert not volume.replicas[0].running
```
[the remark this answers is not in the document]
carterli0407-cell: Updated.

manager/integration/tests/test_scheduling.py
```diff
@@ -2489,13 +2489,17 @@ def best_effort_data_locality_test(client, volume_name):  # NOQA
...
+                                           number_of_replicas)
+
+    # Replica should stay unscheduled until volume is attached.
     for _ in range(10):
-        time.sleep(RETRY_INTERVAL)
         volume = client.by_id_volume(volume_name)
-        assert len(volume.replicas) == number_of_replicas
         assert volume.replicas[0]['hostId'] == ""
         assert not volume.replicas[0].running
+        time.sleep(RETRY_INTERVAL)
```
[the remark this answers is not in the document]
carterli0407-cell: Resolved.

manager/integration/tests/test_scheduling.py
```diff
@@ -2489,14 +2489,20 @@ def best_effort_data_locality_test(client, volume_name):  # NOQA
...
+
+    for i in range(10):
         assert len(volume.replicas) == number_of_replicas
-        assert volume.replicas[0]['hostId'] == ""
+        assert volume.replicas[0]["hostId"] == ""
         assert not volume.replicas[0].running
 
+        if i != 9:
+            time.sleep(RETRY_INTERVAL)
+            volume = client.by_id_volume(volume_name)
```
derekbit: The original logic looks solid, and the initial `by_id_volume` in the for loop won't cause significant negative impact. The new modification adds unnecessary complexity to the implementation. I recommend reverting to the original approach.

carterli0407-cell: cc @derekbit ,
Yeah, I agree the original validation logic was good. I mainly changed this part while trying to address the GitHub Copilot comments.

derekbit: I see. If you think the comment from Copilot is not good enough, you can simply add an explanation in the comment. Reviewers will check them.

carterli0407-cell: OK. Then I will revert to the original approach.

carterli0407-cell: @derekbit , I just updated to the original logic.
Please let me know if you have any questions. 
Thank you.

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.9.x v1.10.x v1.11.x v1.12.x

PR #3482, opened 2026-06-07
test(robot): add test case Test Restoring Volume From Backup With Missing Blocks

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/6689

add test case Test Restoring Volume From Backup With Missing Blocks

### DIFF
--- a/e2e/libs/backupstore/s3.py
+++ b/e2e/libs/backupstore/s3.py
@@ -157,7 +157,7 @@ class S3(Base):
         try:
             minio_api.remove_object(bucket_name, file_path)
         except ResponseError as err:
-            print(err)
+            logging(err)
         logging(f"Deleted file {file_path} in backupstore")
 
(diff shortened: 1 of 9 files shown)


e2e/libs/backupstore/s3.py
```diff
@@ -206,8 +209,11 @@ def delete_random_backup_block(self, volume_name):
 
         try:
             minio_api.remove_object(bucket_name, object_file)
+            logging(f"Removed backup block file {object_file} in backupstore")
```
chriscchien, line 212: Could we add the same logging to the other backup stores as well?

yangchiu: Updated.

chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3483, opened 2026-06-08
test(robot): add case Test V2 Volume Upgrade Data Plane

chriscchien:

Issue longhorn/longhorn#8904

Add case Test V2 Volume Upgrade Data Plane

#### Special notes for reviewers:

The original test [steps](https://github.com/longhorn/longhorn/issues/6001#issuecomment-2239541696) were written based on Longhorn v1.7.0. Some behaviors in the original steps are different with current product behavior:

- `volume.spec.targetNodeID` is no longer available, therefore this test uses `engineNodeID` instead.
- Volumes no longer become degraded during the engine upgrade.
- The transient engine state changes are difficult to observe reliably, therefore this test focus on validates the RAID bdev location to confirm whether the engine target has moved to the expected node.

https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8191/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8192/

### DIFF
--- a/e2e/libs/volume/crd.py
+++ b/e2e/libs/volume/crd.py
@@ -850,8 +850,15 @@ class CRD(Base):
     def activate(self, volume_name):
         return Rest().activate(volume_name)
 
-    def upgrade_engine_image(self, volume_name, engine_image_name):
-        return Rest().upgrade_engine_image(volume_name, engine_image_name)
+    def upgrade_engine_image(self, volume_name, engine_image_name, use_patch=False):
+        if use_patch:
+            logging(f"Upgrading volume {volume_name} engine image to {engine_image_name} using kubectl patch")
+            exec_cmd = f"""
+            kubectl -n {constant.LONGHORN_NAMESPACE} patch volume.longhorn.io {volume_name} --type='merge' -p '{{"spec":{{"image": "{engine_image_name}"}}}}'
+            """
+            subprocess_exec_cmd(exec_cmd)
+        else:
+            return Rest().upgrade_engine_image(volume_name, engine_image_name)
 
     def wait_for_engine_image_upgrade_completed(self, volume_name, engine_image_name):
         return Rest().wait_for_engine_image_upgrade_completed(volume_name, engine_image_name)
--- a/e2e/tests/regression/test_v2.robot
+++ b/e2e/tests/regression/test_v2.robot
@@ -20,6 +20,7 @@ Resource    ../keywords/k8s.resource
 Resource    ../keywords/orphan.resource
 Resource    ../keywords/replica.resource
 Resource    ../keywords/engine_frontend.resource
+Resource    ../keywords/engine_image.resource
 
 Test Setup    Set up v2 test environment
 Test Teardown    Cleanup test resources
@@ -298,7 +299,6 @@ Test Default Block Disks Delete And Re-add
     Then Check volume 0 data is intact
 
 Test V2 Volume Engine Live Switchover
-    [Tags]    v2
     [Documentation]    Test v2 volume cross-node initiator and target support with live switchover.
     ...                Verify data integrity is maintained when engine moves across nodes
     ...                while volume and enginefront remain on the original node.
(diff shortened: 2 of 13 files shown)


e2e/libs/instancemanager/instancemanager.py
```diff
@@ -206,3 +254,53 @@ def verify_replica_lvol_exists_in_spdk_lvol(self, node_name, replica_name):
...
+        cmd = f"kubectl exec -n {constant.LONGHORN_NAMESPACE} {im_pod_name} -- go-spdk-helper raid get"
+        output = subprocess_exec_cmd(cmd)
+        logging(f"SPDK raid bdevs output on node {node_name}:\n{output}")
+        return output
+
+    def verify_raid_bdev_exists_on_node(self, node_name):
+        logging(f"Verifying raid bdev exists on node {node_name}")
+        for i in range(self.retry_count):
+            try:
+                raid_output = self.get_spdk_raid_bdevs(node_name)
```
yangchiu: Do we know what the expected output should look like?

e2e/libs/volume/crd.py
```diff
@@ -839,6 +839,16 @@ def activate(self, volume_name):
     def upgrade_engine_image(self, volume_name, engine_image_name):
         return Rest().upgrade_engine_image(volume_name, engine_image_name)
 
+    def upgrade_engine_image_with_patch(self, volume_name, engine_image_name):
```
yangchiu: You can overwrite the `upgrade_engine_image` function above.

e2e/tests/regression/test_v2.robot
```diff
@@ -373,3 +373,102 @@ Test V2 Volume Engine Live Switchover
...
+    ...                https://github.com/longhorn/longhorn/issues/6001#issuecomment-2239541696
+    IF    '${DATA_ENGINE}' == 'v1'
+        Skip    Test only validate on v2 data engine
+    END
+    ${CUSTOM_LONGHORN_ENGINE_IMAGE}=    Get Environment Variable    CUSTOM_LONGHORN_ENGINE_IMAGE    default='undefined'
+
+    Given Setting deleting-confirmation-flag is set to true
+    And Uninstall Longhorn
+    And Check all Longhorn CRD removed
+    And Install Longhorn stable version
```
yangchiu, line 443: Should we also check whether the Longhorn stable version is defined at the beginning of this test case?

e2e/tests/regression/test_v2.robot
```diff
@@ -373,3 +373,102 @@ Test V2 Volume Engine Live Switchover
...
+    And Wait for longhorn ready
+    And Enable v2 data engine and add block disks
+
+    When Create storageclass longhorn-test with    dataEngine=v2    numberOfReplicas=2
+    And Create persistentvolumeclaim 0    volume_type=RWO    sc_name=longhorn-test
+    And Create persistentvolumeclaim 1    volume_type=RWO    sc_name=longhorn-test
+    And Create deployment 0 with persistentvolumeclaim 0
+    ...    node_selector={"kubernetes.io/hostname":"${NODE_0}"}
+    And Create deployment 1 with persistentvolumeclaim 1
```
yangchiu, line 452: Are these 2 deployments identical? Should we create different types of workloads e.g., RWO and RWX instead?

e2e/tests/regression/test_v2.robot
```diff
@@ -373,3 +373,102 @@ Test V2 Volume Engine Live Switchover
...
+    ...    node_selector={"kubernetes.io/hostname":"${NODE_0}"}
+    And Wait for volume of deployment 0 healthy
+    And Wait for volume of deployment 1 healthy
+    And Record v2 instance manager UIDs
+
+    When Upgrade Longhorn to custom version
+    And Wait for longhorn ready
```
yangchiu, line 459: Does this mean we are already able to upgrade Longhorn with attached v2 volumes now? If so, should be update the related upgrade test cases?

e2e/tests/regression/test_v2.robot
```diff
@@ -373,3 +373,102 @@ Test V2 Volume Engine Live Switchover
...
+    Then Check v2 instance manager did not restart
+    And Wait for volume of deployment 0 healthy
+    And Wait for volume of deployment 1 healthy
+
+    # Disable node 2 scheduling to ensure the RAID bdev can only move to node 1 later
+    When Disable node 2 scheduling
+    And Set node 0 upgrade requested to true
```
yangchiu, line 466: Do we know the effect of setting upgrade requested to true? Why don't we need to set this during a normal Longhorn upgrade?

e2e/tests/regression/test_v2.robot
```diff
@@ -373,3 +373,102 @@ Test V2 Volume Engine Live Switchover
...
+    And Verify raid bdev does not exist on node 1
+    And Verify raid bdev does not exist on node 2
+
+    When Update volume of deployment 0 engineNodeID to node 1
+    And Update volume of deployment 1 engineNodeID to node 1
+    Then Volume of deployment 0 engine CR should be on node 1
+    And Volume of deployment 1 engine CR should be on node 1
+    And Volume of deployment 0 should be attached to node 0
+    And Volume of deployment 1 should be attached to node 0
+    When Upgrade v2 volumes engine to ${CUSTOM_LONGHORN_ENGINE_IMAGE}
```
yangchiu, line 477: Do we know the criteria for upgrading the v2 engine image? Let's clarify the series of operations before and after the engine image upgrade to get a clear picture of what we should do to upgrade a v2 engine image.

e2e/tests/regression/test_v2.robot
```diff
@@ -373,3 +373,102 @@ Test V2 Volume Engine Live Switchover
...
+    Then Volume of deployment 0 engine CR should be on node 0
+    And Volume of deployment 1 engine CR should be on node 0
+    And Volume of deployment 0 should be attached to node 0
+    And Volume of deployment 1 should be attached to node 0
+
+    And Verify raid bdev exists on node 0
+    And Verify raid bdev does not exist on node 1
+    And Verify raid bdev does not exist on node 2
+
+    # Reinstall to avoid side effects of upgrade with attached v2 volumes
```
yangchiu, line 495: What side effects do we observe here?

yangchiu (changes requested):

e2e/libs/instancemanager/instancemanager.py
chriscchien: Below output will print in logs
```
SPDK raid bdevs output on node ip-10-0-1-55:
[
	{
		"name": "pvc-cf7282a7-61d1-4788-9de1-625679d075df-e-0",
		"aliases": [
			"e2336213-01f3-4089-acaa-3e7087bb0fca"
		],
		"product_name": "Raid Volume",
		"block_size": 512,
		"num_blocks": 6291456,
		"uuid": "e2336213-01f3-4089-acaa-3e7087bb0fca",
		"assigned_rate_limits": {
			"rw_ios_per_sec": 0,
			"rw_mbytes_per_sec": 0,
			"r_mbytes_per_sec": 0,
			"w_mbytes_per_sec": 0
		},
		"claimed": true,
		"claim_type": "exclusive_write",
		"zoned": false,
		"supported_io_types": {
			"read": true,
			"write": true,
			"unmap": true,
			"write_zeroes": true,
			"flush": true,
			"reset": true,
			"compare": false,
			"compare_and_write": false,
			"abort": false,
			"nvme_admin": false,
			"nvme_io": false
		},
		"memory_domains": [
			{
				"dma_device_id": "system",
				"dma_device_type": 1
			},
			{
				"dma_device_id": "system",
				"dma_device_type": 1
			}
		],
		"driver_specific": {
			"raid": {
				"strip_size_kb": 0,
				"state": "online",
				"raid_level": "raid1",
				"num_base_bdevs": 2,
				"num_base_bdevs_discovered": 2,
				"num_base_bdevs_operational": 2,
				"base_bdevs_list": [
					{
						"name": "pvc-cf7282a7-61d1-4788-9de1-625679d075df-r-81040788n1",
						"uuid": "ecebe285-f4e7-5c37-94fe-50543a28f6ff",
						"is_configured": true,
						"data_offset": 0,
						"data_size": 6291456
					},
					{
						"name": "pvc-cf7282a7-61d1-4788-9de1-625679d075df-r-fe04fb80n1",
						"uuid": "e8b4e618-d671-541f-a25d-07f2582da3e8",
						"is_configured": true,
						"data_offset": 0,
						"data_size": 6291456
					}
				],
				"superblock": false
			}
		}
	}
]
```

e2e/libs/volume/crd.py
chriscchien: Updated!

---

~I prefer seperate them. it is because `upgrade_engine_image`  use REST API to change volume but `upgrade_engine_image_with_patch` use CR patch.~

~Using patch is because currently change engine image on v2 volume via REST API is blocked.~

~This is also explained in the comment:~
```
        """
        This is required for v2 volumes as the API may be blocked.
        """
```

e2e/tests/regression/test_v2.robot
chriscchien: From the test result, the RWX volume RAID bdev did not move to the expected node during the engine upgrade, which caused the test case to fail.

But it is not mentioned in the test [steps](https://github.com/longhorn/longhorn/issues/6001#issuecomment-2239541696), we can wait  https://github.com/longhorn/longhorn/issues/9104 then verify the v2 RWX upgrade.

e2e/tests/regression/test_v2.robot
chriscchien: No, https://github.com/longhorn/longhorn/issues/9104 is targeting v1.13.0

e2e/tests/regression/test_v2.robot
chriscchien: We can wait https://github.com/longhorn/longhorn/issues/9104.

This test is focus on raid bdev will switch between nodes when engine upgrade.

e2e/tests/regression/test_v2.robot
chriscchien: Updated!

e2e/tests/regression/test_v2.robot
chriscchien: We do not observe any specific side effects in this test.

The reinstall is a precautionary cleanup because upgrading Longhorn with attached V2 volumes is not a supported scenario. The official documentation states that V2 volumes should be detached before upgrading since V2 Data Engine does not support live upgrades.

e2e/tests/regression/test_v2.robot
chriscchien: I believe we don't need to manually set this during a normal Longhorn upgrade because the Longhorn upgrade flow/controllers handle this internally as part of the manager/instance-manager image upgrade. 

cc @derekbit

e2e/tests/regression/test_v2.robot
yangchiu: Then do we know why the test case needs to create 2 identical deployments?

e2e/tests/regression/test_v2.robot
yangchiu: The test case you implement in this PR is Test V2 Volume Upgrade _**Control**_ Plane. Is it the correct name?

e2e/tests/regression/test_v2.robot
yangchiu: Does that mean this test case is testing a special workaround for upgrading v2 engine image that won't occur in a real world upgrade scenario? 

@derekbit Could you help elaborate on [the test steps](https://github.com/longhorn/longhorn/issues/6001#issuecomment-2239541696)?

We'd like to better understand the upgrade flow in the test steps and how it relates to an actual v2 engine image upgrade.

e2e/tests/regression/test_v2.robot
yangchiu: @derekbit Could you help provide more background on this test case? Is it intended to test the upgrade flow or the raid bdev switch?

e2e/tests/regression/test_v2.robot
yangchiu: @derekbit Could we have more information about why the special workaround in the test steps works for upgrading the v2 engine image? It would be appreciated if we could understand the details under the hood.

chriscchien:
Test result
- Test V2 Volume Upgrade Control Plane
   - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8298/
- Pre-release Checks (involine engine upgrade)
   - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8301/

e2e/tests/regression/test_v2.robot
chriscchien: Updated!

e2e/tests/regression/test_v2.robot
chriscchien: Test [steps](https://github.com/longhorn/longhorn/issues/6001#issuecomment-2239541696) designed by developer were using two workloads.

I think using two workloads helps avoid testing only the single-volume happy path.

cc @derekbit

yangchiu:
Leaving this PR to @derekbit for review since I'm having difficulty understanding this test case. @derekbit should have more context on it.

derekbit:
Haven't read the ticket. @chriscchien Could you briefly introduce it?

chriscchien:
> Haven't read the ticket. @chriscchien Could you briefly introduce it?

Test case is come from https://github.com/longhorn/longhorn/issues/6001#issuecomment-2239541696, the main scenario is:
- Deploy workloads with v2 volumes.
- Upgrade Longhorn while the v2 volumes are still attached.
- Upgrade the v2 engine image. by setting upgradeRequested
- Verify that the SPDK RAID bdev is moved to the expected node when the engine is in upgrade flow.
- Reinstall Longhorn at the end to avoid any side effects from the unsupported/unsafe attached-v2-volume Longhorn upgrade scenario for the remaining tests.

My understanding is that this case focuses on validating whether the SPDK RAID bdev can switch between nodes correctly when a v2 volume goes through the engine upgrade flow.

The issue was created around v1.7.0 and this case is included in the v2 sanity test scope, so it needs to be performed during release testing.

Some question during review and may need test case designer to clarify:
- [Why use two workloads](https://github.com/longhorn/longhorn-tests/pull/3483#discussion_r3458971094)
- [Did the test scenario included RWX volume](https://github.com/longhorn/longhorn-tests/pull/3483#discussion_r3449439212)
- [What is the side effect when upgrade with v2 volume attached, because test case reinstalled longhorn in the end](https://github.com/longhorn/longhorn-tests/pull/3483#discussion_r3449524496)
- [What is the purpose of set node `spec.upgradeRequested` to true](https://github.com/longhorn/longhorn-tests/pull/3483#discussion_r3449476608)

chriscchien:
Close this PR.

The test steps are outdated, and we will have v2 upgrade support in v1.13.0. A new test case will be needed for that.

I will also remove the related test from the `v2 volume sanity check` test case in Qase.

cc @yangchiu , @derekbit

PR #3485, opened 2026-06-08
test(robot): refine test case Pre-release Checks to restore backups to volumes with different data engines

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/6689

refine test case Pre-release Checks to restore backups to volumes with different data engines


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.12.x

PR #3493, opened 2026-06-09
test(robot): add test case Test Reusing Failed Replica After Node Back

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11657

add test case Test Reusing Failed Replica After Node Back


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3504, opened 2026-06-16
test(robot): add test case Test Node Drain After Node Deletion

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/11595

add test case Test Node Drain After Node Deletion


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3505, opened 2026-06-17
ci: move user-volumes-patch to initial machine config for Talos `1.12.x` support

roger-ryao:
longhorn/longhorn#12965

Apply `user-volumes-patch.yaml` at initial deployment via `config_patches` instead of `post-deployment` patching.
Also update default instance type to `t3.xlarge`.

According to the official Talos `v1.13` documentation for AWS deployment,
there is a known issue that prevents Talos from running on t2 instance types.
Therefore, t3.xlarge is used instead of t2 for all control plane and worker nodes.

Reference: https://docs.siderolabs.com/talos/v1.13/platform-specific-installations/cloud-platforms/aws#create-the-ec2-instances

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -89,7 +89,10 @@ ISCSI_DEV_PATH = "/dev/disk/by-path"
 ISCSI_PROCESS = "iscsid"
 
 if os.environ.get("CLOUDPROVIDER") == "aws":
-    if os.uname().machine == "x86_64":
+    if os.environ.get("DISTRO") == "talos":
+        # Talos: nvme2n1 for Longhorn user volume, nvme1n1 for v2 block tests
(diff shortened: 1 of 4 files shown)


yangchiu:
@roger-ryao Make sure it works with both v1 and v2 volumes.

roger-ryao:
> @roger-ryao Make sure it works with both v1 and v2 volumes.

OK. As I currently have other higher-priority tasks to address, the verification will be pushed to next week.

roger-ryao:
> @roger-ryao Make sure it works with both v1 and v2 volumes.

For v2 create an ISSUE https://github.com/longhorn/longhorn/issues/13440  to track this. This needs more time to determine the proper solution.

yangchiu (approved):

PR #3532, opened 2026-06-30
test(robot): add test case Node Down Between Migration And Confirmation

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13136

add test case Node Down Between Migration And Confirmation


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.12.x

PR #3533, opened 2026-06-30
test(robot): skip 30s sleep unless a block disk is deleted

roger-ryao:
longhorn/longhorn#12999

The 30s sleep in `reset_disks_for_all_nodes` was added to handle delays
in block device detection after removing disks (longhorn/longhorn#11860).

Previously, the sleep ran unconditionally on every call, even when no
block disks were removed, causing unnecessary slowdowns in test execution.

This PR optimizes the behavior so that the sleep only triggers when a
block disk was actually deleted.

Changes:
- manager/integration/tests/common.py:  remove sleep 30s  due to https://github.com/longhorn/longhorn/issues/12999#issuecomment-4860550740

- e2e/libs/node/node.py: track block disk deletion and sleep only when a block disk is deleted

longhorn/longhorn#12999

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -3836,7 +3836,6 @@ def reset_disks_for_all_nodes(client, add_block_disks=False):  # NOQA
             }
             default_disks["v1"]["default"] = False
 
-    block_disk_deleted = False
     nodes = client.list_node()
     for n in nodes:
         node = n  # Captures the correct value of n in the closure.
@@ -3865,10 +3864,6 @@ def reset_disks_for_all_nodes(client, add_block_disks=False):  # NOQA
                     break
 
         if cleanup_required:
-            for _, disk in iter(node.disks.items()):
-                if disk.diskType == "block":
-                    block_disk_deleted = True
-                    break
             update_disks = get_update_disks(node.disks)
             for disk_name, disk in iter(update_disks.items()):
                 disk.allowScheduling = False
@@ -3925,13 +3920,6 @@ def reset_disks_for_all_nodes(client, add_block_disks=False):  # NOQA
             wait_for_disk_status(client, node.name, name,
                                  "storageReserved",
                                  expected_reserved_storage)
-
-    # for block type disks added by bdf (nvme disk driver)
-    # disk deletion takes some time, wait the device show up on the host
-    # normally less than 30 seconds
-    # ref: https://github.com/longhorn/longhorn/issues/11860
-    if block_disk_deleted:
-        time.sleep(30)
 
 
 def reset_settings(client):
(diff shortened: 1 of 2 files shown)


manager/integration/tests/common.py
yangchiu, line 1: Is this function called during every test case's teardown function? If so, does that mean we're spending extra time unnecessarily resetting the disks?

Does robot test do the same thing? If so, it should need the same refinement. If not, we can follow the approach used in robot test to avoid unnecessarily resetting the disks and save time.

manager/integration/tests/common.py
```diff
@@ -3925,7 +3930,8 @@ def reset_disks_for_all_nodes(client, add_block_disks=False):  # NOQA
     # disk deletion takes some time, wait the device show up on the host
     # normally less than 30 seconds
     # ref: https://github.com/longhorn/longhorn/issues/11860
-    time.sleep(30)
+    if block_disk_deleted and BLOCK_DEV_PATH.startswith("/dev/nvme"):
```
yangchiu: It might not be correct. The block device path name depends on how the operating system exposes the storage device.

For a physical NVMe device, it's like `/dev/nvme1n1`.
For a Xen Virtual Block Device, it's like `/dev/xvdh`.
For a VirtIO Block Device, it's like `/dev/vdc`.

All of them are valid block device paths.

Also you need to take the BDF into consideration.

yangchiu (changes requested):

manager/integration/tests/common.py
roger-ryao: ~~I discussed this issue with @davidcheng0922  last week. He mentioned that this problem only occurs with NVMe, and AIO and VirtIO should not require this kind of sleep.~~

~~As for the BDF part, I missed it and will add the necessary condition check.~~

roger-ryao: > I discussed this issue with @davidcheng0922 last week. He mentioned that this problem only occurs with NVMe, and AIO and VirtIO should not require this kind of sleep.
> 
> As for the BDF part, I missed it and will add the necessary condition check.

I just checked with @davidcheng0922  again. He mentioned that `AIO` and `VirtIO` complete faster than `NVMe`, but that does not mean they do not require the 30 second sleep. I will update the condition check.

davidcheng0922: If I remember correctly, the issue happens when disk deletion is not fully completed before the instance manager is torn down. `aio` may be faster than `nvme`, but the problem might not be limited to the engine side.

The flow is roughly:

1. Longhorn Manager detects that the disk has been removed.
2. The controller syncs the state.
3. Longhorn Manager sends a request to the instance manager.
4. SPDK starts removing the disk driver.

If the instance manager is deleted too quickly, the deletion process may not have completed yet, which can cause the issue to happen again.

Waiting until the device disappears is probably the safest approach.

manager/integration/tests/common.py
roger-ryao: Updated

manager/integration/tests/common.py
roger-ryao: Updated

yangchiu (approved):

yangchiu:
@mergify backport v1.12.x v1.11.x

PR #3534, opened 2026-06-30
test(robot): Add case V2 Replica Migration Should Not Cause IO Stall

chriscchien:

Issue longhorn/longhorn#13311

Add case `V2 Replica Migration Should Not Cause IO Stall`

Test result

- master: [passed](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8338/)
- v1.12.0: [failed](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8339/), confirms this test case covered issue [13309](https://github.com/longhorn/longhorn/issues/13309) 

image

### DIFF
--- a/e2e/keywords/io.resource
+++ b/e2e/keywords/io.resource
@@ -3,6 +3,7 @@ Library             ../libs/keywords/io_keywords.py
 Library             ../libs/keywords/instancemanager_keywords.py
 Library             ../libs/keywords/common_keywords.py
 Library             ../libs/keywords/node_keywords.py
+Library             ../libs/keywords/workload_keywords.py
 
 *** Keywords ***
 Create dm linear device from block device ${block_disk_path} as ${dm_device_name} on node ${node_id}
@@ -44,3 +45,12 @@ Wipe block device ${block_device_path} on node ${node_id}
     [Documentation]    Wipe filesystem and partition signatures from a block device.
     ${node_name} =    get_node_by_index    ${node_id}
     wipe_block_device_signatures    ${block_device_path}    ${node_name}
+
+Start fsync writer on ${workload_kind} ${workload_id}
+    [Documentation]    Start a background fsync writer that tracks max per-write latency in memory.
+    ${workload_name} =    generate_name_with_suffix    ${workload_kind}    ${workload_id}
+    ${pod_name} =    get_workload_pod_name    ${workload_name}
+    start_fsync_writer    ${pod_name}
+
+Assert no IO stall greater than ${threshold_sec} seconds
+    assert_no_io_stall    ${threshold_sec}
--- a/e2e/libs/keywords/io_keywords.py
+++ b/e2e/libs/keywords/io_keywords.py
@@ -1,12 +1,70 @@
-from utility.utility import logging
+from utility.utility import logging, get_retry_count_and_interval
 from node_exec import NodeExec
 
+import subprocess
+import threading
 import time
 
 class io_keywords:
 
     def __init__(self):
-        pass
+        self._fsync_writer_process = None
+        self._fsync_writer_thread = None
+        self._max_elapsed = 0.0
+        self._max_elapsed_lock = threading.Lock()
+
+    def start_fsync_writer(self, pod_name, namespace="default"):
+        logging(f"Starting fsync writer on pod {pod_name}")
+
+        cmd = [
+            "kubectl", "exec", pod_name, "-n", namespace, "--",
+            "sh", "-c",
+            "mkdir -p /data && while true; do"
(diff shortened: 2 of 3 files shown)


e2e/keywords/io.resource
```diff
@@ -44,3 +44,39 @@ Wipe block device ${block_device_path} on node ${node_id}
     [Documentation]    Wipe filesystem and partition signatures from a block device.
     ${node_name} =    get_node_by_index    ${node_id}
     wipe_block_device_signatures    ${block_device_path}    ${node_name}
+
+Start fsync writer on ${workload_kind} ${workload_id} and log per-write latency to ${file}
+    [Documentation]    Start a background fsync writer that logs per-write timing to /data/${file}
+    ${command}=    Catenate    SEPARATOR=${SPACE}
+    ...    nohup sh -c 'while true; do
+    ...    dd if=/dev/zero of=/data/f bs=64k count=1 conv=fsync 2>&1
```
yangchiu: `f` doesn't seem like a meaningful file name.

e2e/keywords/io.resource
```diff
@@ -44,3 +44,39 @@ Wipe block device ${block_device_path} on node ${node_id}
...
+    ...    | awk -F" copied, | s," "/copied/ {print \\$2}"
```
yangchiu: If we're only interested in the elapsed time (the numeric value `0.089056`), we can append just that value to the log file and omit the unit (`seconds`) as well. That would also eliminate the need for the following step:

```
Evaluate
    ...    [float(x) for x in re.findall(r'(?m)^\\s*([0-9]+(?:\\.[0-9]+)?)\\s+seconds', '''${latency_data}''')]
```

since the log would already contain plain numeric values that can be consumed directly.

e2e/keywords/io.resource
```diff
@@ -44,3 +44,39 @@ Wipe block device ${block_device_path} on node ${node_id}
...
+    ...    0.029434 seconds, 2.1MB/s
+    ${workload_name} =    generate_name_with_suffix    ${workload_kind}    ${workload_id}
+    ${latency_data} =    Run Command
+    ...    kubectl exec deploy/${workload_name} -- cat /data/${file}
+
+    ${latencies} =    Evaluate
+    ...    [float(x) for x in re.findall(r'(?m)^\\s*([0-9]+(?:\\.[0-9]+)?)\\s+seconds', '''${latency_data}''')]
+    ...    modules=re
+
+    ${max_latency} =    Evaluate    max(${latencies})
```
yangchiu: If the log file contains only plain numeric values, we can obtain the max value with a single shell command using `Run commands in`.

e2e/keywords/io.resource
```diff
@@ -44,3 +44,39 @@ Wipe block device ${block_device_path} on node ${node_id}
...
+    Set Test Variable    ${max_latency}
+
+Assert no IO stall from ${workload_kind} ${workload_id} latency log ${file}
+    [Arguments]    ${threshold_sec}=3.0
+    Get max latency from ${workload_kind} ${workload_id} latency log ${file}
+    Log    Maximum latency: ${max_latency}s
+
+    Should Be True    ${max_latency} < ${threshold_sec}
```
yangchiu: This keyword fails the test case immediately when the condition is not met, leaving no opportunity to investigate the issue.

yangchiu (changes requested):

e2e/keywords/io.resource
chriscchien: Full I/O log will copied to /tmp/test-report/lat.log now

e2e/keywords/io.resource
```diff
@@ -44,3 +44,32 @@ Wipe block device ${block_device_path} on node ${node_id}
     [Documentation]    Wipe filesystem and partition signatures from a block device.
     ${node_name} =    get_node_by_index    ${node_id}
     wipe_block_device_signatures    ${block_device_path}    ${node_name}
+
+Start fsync writer on ${workload_kind} ${workload_id} and log per-write latency to ${file}
+    [Documentation]    Start a background fsync writer that logs per-write timing to /data/${file}
+    ${command}=    Catenate    SEPARATOR=${SPACE}
```
yangchiu: It would be even better if we can keep the elapsed time in a variable instead of appending it to a log file, and update the variable only when the new elapsed time is greater than the current value. It would completely eliminate the need of a separate file.

e2e/keywords/io.resource
yangchiu: Once the the test case fails and the test build completes, the environment is destroyed, and the copied log file is lost as well, so the effort to copy the log file is to no avail.

yangchiu (changes requested):

yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3550, opened 2026-07-07
test(robot): add nodePublishSecretRef support for encrypted volume PVs

roger-ryao:
longhorn/longhorn#13364

Add `node_publish_secret_name`/`node_publish_secret_namespace` parameters through the volume_keywords -> Volume -> Base -> PersistentVolume call chain so that PVs created for restored/cloned encrypted volumes set nodePublishSecretRef.

Update `test_encrypted_volume.robot` to pass `node_publish_secret_name` for all `Create deployment ... with volume ...` steps that mount restored/cloned encrypted volumes, and remove the v2 data engine skip for backup restore to encrypted volume now that it is supported.


e2e/tests/regression/test_encrypted_volume.robot
```diff
@@ -339,8 +337,10 @@ Test Encrypted Volume Backup Restore To Encrypted Volume
     And Wait for volume 3 detached
     # Mount the restored volumes via deployments so that CSI opens the LUKS container.
     # Must use longhorn-crypto SC (with node-stage-secret-ref) so luksOpen is triggered.
-    And Create deployment 2 with volume 2    sc_name=longhorn-crypto    node_stage_secret_name=longhorn-crypto
-    And Create deployment 3 with volume 3    sc_name=longhorn-crypto    node_stage_secret_name=longhorn-crypto
+    And Create deployment 2 with volume 2    sc_name=longhorn-crypto    node_stage_secret_name=longhorn-crypto    node_publish_secret_name=longhorn-crypto
```
yangchiu, line 340: If the persistent volume created here does not inherit the `node-stage-secret-name` and `node-publish-secret-name` referenced by the `longhorn-crypto` storage class, then there doesn't seem to be a need to use `longhorn-crypto` in this case.

Based on the [documentation](https://longhorn.io/docs/1.13.0/snapshots-and-backups/backup-and-restore/restore-statefulset/), it appears that it has its own mechanism and does not rely on the storage class that is passed to it (in the example, it uses the default `longhorn` storage to restore an encrypted volume).

If that's true, is the statement still correct?
```
Must use longhorn-crypto SC (with node-stage-secret-ref) so luksOpen is triggered.
```

@mantissahz, could you help confirm the behavior?

Is the `longhorn-crypto` storage class (which defines `node-stage-secret-name` and `node-publish-secret-name`) required here?
Or are the explicit `node-stage-secret-name` and `node-publish-secret-name` fields sufficient?
If the explicit `node-stage-secret-name` and `node-publish-secret-name` are required, are `provisioner-secret-name` and `node-expand-secret-name` required as well?

yangchiu (changes requested):

e2e/tests/regression/test_encrypted_volume.robot
mantissahz: > Must use longhorn-crypto SC (with node-stage-secret-ref) so luksOpen is triggered.

The `luksOpen` operation is triggered at the `NodeStageVolume` step. It depends on the `node-stage-secret-name` and `node-publish-secret-name` (The `NodePublishVolume` might restage the volumes).
The storage class is not used at the `NodeStageVolume` and `NodePublishVolume` steps.

> Is the longhorn-crypto storage class (which defines node-stage-secret-name and node-publish-secret-name) required here?

Yes, when creating a PV, it will use the storage class.
After the PV is created, the operations such as staging, publishing, and expanding will depend on the `PersistentVolume.spec.csi.nodeStageSecretRef` or `PersistentVolume.spec.csi.nodePublishSecretRef` fields.

yangchiu: @mantissahz Does that mean we don't need the `longhorn-crypto` storage class, and that the default `longhorn` storage class is sufficient in this case?

mantissahz: For creating the PV of the restored encrypted volume, in theory, yes.
But I still recommend consistency for create, restore, and attach operations because I think it still needs the original storage class used to create the Lognhorn volume when creating the PV.

What's your concern for using the default `longhorn` or the created `longhorn-crypto` storage class?

yangchiu: @mantissahz The [documentation](https://longhorn.io/docs/1.13.0/snapshots-and-backups/backup-and-restore/restore-statefulset/) uses default `longhorn` storage class to restore an encrypted volume. Do you think the documentation should be refined?

Also, could you help provide an example of a scenario where `node-stage-secret-name` and `node-publish-secret-name` are required? Thanks!

mantissahz: > @mantissahz The [documentation](https://longhorn.io/docs/1.13.0/snapshots-and-backups/backup-and-restore/restore-statefulset/) uses default longhorn storage class to restore an encrypted volume. Do you think the documentation should be refined?

Yes, we should point out it should be the same to the storage class of the  original volume or an encryption enabled storage class.

> Also, could you help provide an example of a scenario where node-stage-secret-name and node-publish-secret-name are required? Thanks!

When creating a PV for an encrypted volume, these parameters should be provided to attaching the volume to the assigned node like this test case.

Every CSI attach operation will use parameters `node-stage-secret-name` and `node-publish-secret-name` for `NodeStageVolume` and `NodePublishVolume` steps.

yangchiu: Sorry for the typo.. I mean `provisioner-secret-name` and `node-expand-secret-name`.

yangchiu: `provisioner-secret-name` is used with the storage class when creating the volume. Therefore, it should not be used after the volume is created. For only restoring an encrypted volume, no, because creating/provisioning is provided by the storage class.

`node-expand-secret-name` is used when a CSI expand request is received. If we will expand the restored encrypted volume, the node-expand-secret-name is needed.

yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3551, opened 2026-07-08
test(pytest): remove Default Data Path setting from disk labeled test

roger-ryao:
longhorn/longhorn#9083

The `Default Data Path` setting (SETTING_DEFAULT_DATA_PATH) is no longer used to control where the default disk is created.

Update `test_node_default_disk_labeled` to always expect the default disk at `DEFAULT_DISK_PATH`, instead of setting a random path. Also clean up the reset_disk_settings fixture that reset this setting.


manager/integration/tests/test_node.py
```diff
@@ -1595,18 +1588,18 @@ def test_node_default_disk_labeled(client, core_api, random_disk_path,  reset_de
     node = client.by_id_node(cases["disk_exists"])
     if DATA_ENGINE == "v1":
         assert len(node.disks) == 1
-        assert node.disks[list(node.disks)[0]].path == \
-            DEFAULT_DISK_PATH
+        assert node.disks[list(node.disks)[0]].path.rstrip('/') == \
```
yangchiu: If the default data path has been refined from /var/lib/longhorn/ to /var/lib/longhorn and we agree on the convention of omitting the trailing /, we could also consider changing DEFAULT_DISK_PATH to /var/lib/longhorn, so  we wouldn’t need to add rstrip() everywhere.

yangchiu (changes requested):

manager/integration/tests/test_node.py
roger-ryao: Update

davidcheng0922 (approved):
LGTM

yangchiu (approved):

PR #3563, opened 2026-07-14
fix(infra): support calico cni ip-in-ip mode

COLDTURNIP:

Issue longhorn/longhorn#13438 https://github.com/longhorn/longhorn/issues/13521

The Calico CNI plugin works in IP-in-IP mode, i.e. the IP packet is packed into another IP packet and forward to other nodes. However, AWS EC2 checks the source & destination, and drops such packets due to security concern. This PR allows such traffic for such CNI setup.


yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3572, opened 2026-07-16
test(robot): fix failed upgrade test cases caused by default data path change

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13377 https://github.com/longhorn/longhorn/issues/13528

fix failed upgrade test cases caused by default data path change


roger-ryao (approved):
LGTM

PR #3575, opened 2026-07-17
test: fix test_upgrade by waiting for replica deletion instead of replica count reduction

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13470

fix test_upgrade by waiting for replica deletion instead of replica count reduction


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3578, opened 2026-07-17
test(robot): fix test case Test Node ID Change During Backing Image Creation failure caused by execute_command exception

yangchiu:

Issue N/A

fix test case Test Node ID Change During Backing Image Creation failure caused by execute_command exception


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3585, opened 2026-07-18
test(robot): refine test case Power Off Replica Node Should Not Rebuild New Replica On Same Node with retry mechanism

yangchiu:

Issue N/A

refine test case Power Off Replica Node Should Not Rebuild New Replica On Same Node with retry mechanism


roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.12.x

PR #3586, opened 2026-07-19
test(robot): refactor instance manager disk size assertion to support  v1/v2 raw size check

roger-ryao:
https://github.com/longhorn/longhorn/issues/13163

- Add `wait_for_disk_size_in_instance_manager_pod` retry method in instancemanager_keywords.py
- Rework `Assert disk size in instance manager pod for ... is` keyword into `Assert disk size in instance manager for ${resource_kind} ${resource_id}` with named args `expected_disk_size` and optional `raw_size`
- Add `Assert v2 raw disk size in instance manager` keyword to verify v2 RAW device size independently of the encrypted device size
- Handle encrypted device naming difference for v2 (`${volume_name}-encrypted`)
- Update case in `test_encrypted_volume.robot` to use new keyword

### DIFF
--- a/e2e/keywords/longhorn.resource
+++ b/e2e/keywords/longhorn.resource
@@ -181,23 +181,34 @@ Check longhorn manager pods not restarted after test start
 Check longhorn backing image manager image is ${expected_image}
     check_backing_image_manager_images    ${expected_image}
 
-Assert disk size in instance manager pod for ${resource_kind} ${resource_id} is ${size}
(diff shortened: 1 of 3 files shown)


chriscchien (approved):
LGTM

Test Encrypted Volume Basic 
- https://10.115.5.5/job/private/job/longhorn-e2e-test/571/  v1
- https://10.115.5.5/job/private/job/longhorn-e2e-test/572/ v2

yangchiu:
@mergify backport v1.12.x

PR #3599, opened 2026-07-23
test(robot): use v1 volume as source of extra block disk

chriscchien:

longhorn/longhorn#13563

test ruelst with v2 data engine: https://10.115.5.5/job/private/job/longhorn-e2e-test/574/
- Test Volume Expansion When Node Disk Is Full
- Test Replica Auto Balance Disk In Pressure
- Test Replica Auto Balance Disk In Pressure With Stopped Volume Should Not Block
- Test No Transient Error In Engine Status During Eviction
- Test Scheduling Replicas To Different Disks On The Same Node


yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3611, opened 2026-07-26
test(robot): fix name 'time' is not defined

yangchiu:

Issue N/A

fix name 'time' is not defined

### DIFF
--- a/e2e/libs/keywords/host_keywords.py
+++ b/e2e/libs/keywords/host_keywords.py
@@ -1,6 +1,7 @@
 from robot.libraries.BuiltIn import BuiltIn
 
 import os
+import time
 
 from host import Harvester, Aws, Vagrant
 from host.constant import NODE_REBOOT_DOWN_TIME_SECOND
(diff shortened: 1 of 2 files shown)


e2e/libs/keywords/host_keywords.py
```diff
@@ -1,6 +1,7 @@
 from robot.libraries.BuiltIn import BuiltIn
 
 import os
+import time
```
roger-ryao, line 4: `import time` is already imported at the top of the file, so the local imports in [execute_command_on_node_and_not_expect_output](https://github.com/yangchiu/longhorn-tests/blob/756853ff5c3a61b4b16e1e7e6b029733637dc20b/e2e/libs/keywords/host_keywords.py#L80)  and [execute_command_on_node_and_wait_for_output](https://github.com/yangchiu/longhorn-tests/blob/756853ff5c3a61b4b16e1e7e6b029733637dc20b/e2e/libs/keywords/host_keywords.py#L91) are unnecessary.

roger-ryao (changes requested):

e2e/libs/keywords/host_keywords.py
yangchiu: Removed. Thanks!

roger-ryao (approved):
LGTM

yangchiu:
@mergify backport v1.12.x

PR #3612, opened 2026-07-27
test(robot): refine reset_disks to handle local-disk

yangchiu:

Issue N/A

refine reset_disks to handle local-disk

local-disk uses /var/lib/longhorn/* as its path, which breaks the reset_disks logic


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3616, opened 2026-07-27
test(robot): improve encrypted volume tests and add v2 upgrade/DR cases

roger-ryao:
-  longhorn/longhorn#13603
-  longhorn/longhorn#13163

- Increase volume sizes from `100 Mi` to `512 Mi` and writes from `10 MB` to `256 MB`

- Fix encrypted volume stuck in attached bug by adding `cleanup_node_exec_on_node` immediately after `Check volume endpoint on node` (longhorn/longhorn#13603)

- Add scale down before/scale up after logic around Longhorn upgrade for v2 data engine in Test Encrypted Volume Upgrade

- Add `Test Encrypted Volume Upgrade For v2` Full v1.11.3/v1.12.0 > v1.12.1/master upgrade scenario for v2 encrypted volumes, covering DR volumes, replica rebuild, expansion, workload reattach, and backup/restore

- Add `Test Encrypted DR Volume Activation` Verifies that an encrypted standby (DR) volume can be created from a backup, restored, activated, and attached with correct data and device size (longhorn/longhorn#13163)

https://github.com/longhorn/longhorn/issues/13163#issuecomment-4620507665

### DIFF
--- a/e2e/keywords/host.resource
+++ b/e2e/keywords/host.resource
@@ -171,6 +171,11 @@ Check volume endpoint on node of ${resource_kind} ${resource_id}
     [Documentation]    Find which node ${resource_kind} ${resource_id}'s volume is attached to,
     ...                then run "ls -l /dev/longhorn/${volume_name}" on that node and log the output.
     ...                Fails the test if the endpoint is empty or doesn't exist.
+    ...                The node-exec pod must be cleaned up immediately after use.
+    ...                Leaving a privileged pod with host /var/run bind-mounted on the
+    ...                same node as an encrypted volume will cause the volume to become
+    ...                stuck in the attached state and fail to detach.
+    ...                See: https://github.com/longhorn/longhorn/issues/13603
     ${resource_name} =    generate_name_with_suffix    ${resource_kind}    ${resource_id}
     ${volume_name} =    Run Keyword If    '${resource_kind}' == 'volume'
     ...        Set Variable    ${resource_name}
--- a/e2e/tests/regression/test_encrypted_volume.robot
+++ b/e2e/tests/regression/test_encrypted_volume.robot
@@ -31,29 +31,29 @@ Test Encrypted Volume Basic
     [Documentation]    Test basic encrypted volume operations for both RWO and RWX volumes.
     ...                Deployment 0 = RWO, Deployment 1 = RWX.
     ...
-    ...                Expected sizes for a 100 Mi volume:
-    ...                    - Replica backend file = 116 Mi (requested + 16 Mi pre-allocated).
+    ...                Expected sizes for a 512 Mi volume:
+    ...                    - Replica backend file = 528 Mi (requested + 16 Mi pre-allocated).
     ...                    - v2 data engine: Replica backend N/A (format differs, only device size is verified).
-    ...                    - dm-crypt device (RWO instance manager) = 100 Mi (full requested size).
-    ...                    - dm-crypt device (RWX sharemanager pod) = 100 Mi (full requested size).
-    ...                    - Mounted filesystem (workload pod, RWO/RWX) = ~100 Mi (accounting for filesystem overhead).
+    ...                    - dm-crypt device (RWO instance manager) = 512 Mi (full requested size).
+    ...                    - dm-crypt device (RWX sharemanager pod) = 512 Mi (full requested size).
+    ...                    - Mounted filesystem (workload pod, RWO/RWX) = ~512 Mi (accounting for filesystem overhead).
     Given Create crypto secret
     When Create storageclass longhorn-crypto with    encrypted=true    dataEngine=${DATA_ENGINE}
-    And Create persistentvolumeclaim 0    volume_type=RWO    sc_name=longhorn-crypto    storage_size=100Mi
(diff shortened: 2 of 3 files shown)


e2e/keywords/host.resource
```diff
@@ -171,13 +171,19 @@ Check volume endpoint on node of ${resource_kind} ${resource_id}
...
+    ...                stuck in the attached state and fail to detach.
+    ...                See: https://github.com/longhorn/longhorn/issues/13603
     ${resource_name} =    generate_name_with_suffix    ${resource_kind}    ${resource_id}
     ${volume_name} =    Run Keyword If    '${resource_kind}' == 'volume'
     ...        Set Variable    ${resource_name}
     ...    ELSE
     ...        get_workload_volume_name    ${resource_name}
     ${node_name} =    get_volume_node    ${volume_name}
     execute_command_on_node_and_not_expect_output    ls -l /dev/longhorn/${volume_name}    ${node_name}    No such file or directory
+    cleanup_node_exec_on_node    ${node_name}
```
yangchiu: We can clean up the pod directly at the end of `NodeExec.issue_cmd()`

e2e/tests/regression/test_encrypted_volume.robot
```diff
@@ -728,88 +744,341 @@ Test Encrypted Volume Upgrade
...
+    IF    '${DATA_ENGINE}' == 'v1'
+        Then Assert disk size in instance manager for deployment 8    expected_disk_size=512Mi
+        And Assert replica file size of deployment 8 is 528Mi
+    ELSE
+        Then Assert disk size in instance manager for deployment 8    expected_disk_size=496Mi    raw_size=512Mi
+    END
     And Check deployment 8 file data.txt checksum matches checksum 1
 
+Test Encrypted Volume Upgrade For v2
+    [Tags]    rwo    expansion    replica-rebuild    engine-upgrade    dr-volume    backup    restore    v2
```
yangchiu: Use `upgrade` tag instead of `engine-upgrade`.

`engine-upgrade` tag should refer to upgrading a volume's engine image.

yangchiu (changes requested):

e2e/keywords/host.resource
roger-ryao: Hi @yangchiu 
     I've looked into this, and I think if we clean up the pod directly at the end of `NodeExec.issue_cmd()`, it would likely affect  [`keep_writing_data(self, volume_name, size)`](https://github.com/longhorn/longhorn-tests/blob/88f53b50d3627f7ba2059bd66f16b7e1c7fe4b25/e2e/libs/volume/crd.py#L737) ,since that function relies on the pod.  So I think a better way might be to add `NodeExec(node_name).cleanup()` at the end of [`execute_command_on_node_and_not_expect_output(self, cmd, node_name, output)`](https://github.com/longhorn/longhorn-tests/blob/88f53b50d3627f7ba2059bd66f16b7e1c7fe4b25/e2e/libs/keywords/host_keywords.py#L79) instead.
     
WDYT?

yangchiu: Is `NodeExec.cleanup()` added to `execute_command_on_node_and_not_expect_output` only because `execute_command_on_node_and_not_expect_output` is called right before the workloads are scaled down in the test case? If we modify this test case or add a new test case in the future that calls other `execute_command_on_node` functions before detaching a volume, could we still run into the same issue?

yangchiu: > it would likely affect [keep_writing_data(self, volume_name, size)](https://github.com/longhorn/longhorn-tests/blob/88f53b50d3627f7ba2059bd66f16b7e1c7fe4b25/e2e/libs/volume/crd.py#L737), since that function relies on the pod.

Need further investigation since it's not the expected behavior (the write process is expected to stay alive on the host). If you don't have time to look into it now, could you create a ticket to track it?

roger-ryao: > Is `NodeExec.cleanup()` added to `execute_command_on_node_and_not_expect_output` only because `execute_command_on_node_and_not_expect_output` is called right before the workloads are scaled down in the test case? 

Yes

> If we modify this test case or add a new test case in the future that calls other `execute_command_on_node` functions before detaching a volume, could we still run into the same issue?

You're right, we could still run into the same issue in that scenario. 
I tested different volume combinations to narrow down the scope, and found that this issue specifically occurs when scaling down a workload with an encrypted RWO volume (both v1 and v2 data engine). Other volume combinations don't run into this issue.

I've documented the detailed investigation here: https://github.com/longhorn/longhorn/issues/13603#issuecomment-5086565217

> Need further investigation since it's not the expected behavior (the write process is expected to stay alive on the host). If you don't have time to look into it now, could you create a ticket to track it?

Create ticket https://github.com/longhorn/longhorn/issues/13625

yangchiu: > You're right, we could still run into the same issue in that scenario.

So this is more of a temporary workaround for this particular scenario. We should consider a more general solution that fixes the issue once and for all.

yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3618, opened 2026-07-28
test(linked-clone):  add robot tests for refactored linked clone feature

shuo-wu:

Issue longhorn/longhorn#12552


e2e/libs/keywords/linked_clone_keywords.py
```diff
@@ -0,0 +1,578 @@
...
+    # Volume creation
+    # -------------------------------------------------------------------------
+
+    def get_snapshot_name(self, volume_name, snapshot_id):
+        """Return the actual Longhorn snapshot CR name for the given snapshot_id."""
+        snapshot = self.snapshot.get(volume_name, snapshot_id)
+        assert snapshot is not None, \
+            f"Snapshot with id={snapshot_id} not found for volume {volume_name}"
+        logging(f"Snapshot id={snapshot_id} of volume {volume_name} has name: {snapshot.name}")
+        return snapshot.name
```
roger-ryao, line 38: This can be simplified by reusing `snapshot_keywords.get_snapshot_name`

e2e/libs/keywords/linked_clone_keywords.py
```diff
@@ -0,0 +1,578 @@
...
+                created = True
+                break
+            except Exception as e:
+                logging(f"Failed to create linked-clone volume {clone_volume_name} "
+                        f"(attempt {i}): {e}")
+                time.sleep(self.retry_interval)
+
+        assert created, f"Failed to create linked-clone volume {clone_volume_name}"
+        self.volume.wait_for_volume_to_be_created(clone_volume_name)
+        logging(f"Linked-clone volume {clone_volume_name} created successfully")
```
roger-ryao, line 105: This function can reference [this PR](https://github.com/longhorn/longhorn-tests/commit/eb6d8c52bb726a5385599a3f78e1f784a36b1fed#diff-3d96e807641b35598c48e9b899f85db2783d46f945f7f5d0c227326f879ead46) to refactor `create_linked_clone_volume` so that it directly calls `create_volume` for better reusability.

roger-ryao (changes requested):

e2e/libs/keywords/linked_clone_keywords.py
```diff
@@ -0,0 +1,578 @@
...
+        for i in range(self.retry_count):
+            snap = self.snapshot.get_snapshot_by_name(volume_name, snapshot_name)
+            if snap is None:
+                logging(f"Snapshot CR {snapshot_name} of volume {volume_name} "
+                        f"has been deleted")
+                return
+            time.sleep(self.retry_interval)
+        assert False, \
+            f"Timed out waiting for snapshot CR {snapshot_name} of volume " \
+            f"{volume_name} to be deleted"
```
roger-ryao, line 546: The retry loop in `wait_for_named_snapshot_cr_deleted` can be replaced by `snapshot.wait_for_snapshot_to_be_deleted`, which already handles it.


```suggestion
     def wait_for_named_snapshot_cr_deleted(self, volume_name, snapshot_name):
         """Wait until the named snapshot CR no longer exists."""
         logging(f"Waiting for snapshot CR {snapshot_name} of volume {volume_name} to be deleted")
         self.snapshot.wait_for_snapshot_to_be_deleted(volume_name, snapshot_name)
```

roger-ryao (dismissed):

e2e/libs/keywords/common_keywords.py
```diff
@@ -58,8 +58,10 @@ def execute_command(self, cmd):
         return subprocess_exec_cmd(cmd)
 
     def execute_command_and_expect_output(self, cmd, expected_output):
-        res = subprocess_exec_cmd(cmd)
-        retry_count, _ = get_retry_count_and_interval()
+        try:
```
yangchiu, line 61: `subprocess_exec_cmd` is already wrapped in a `try`/`except` block under the hood.

e2e/libs/keywords/common_keywords.py
```diff
@@ -68,7 +70,8 @@ def execute_command_and_expect_output(self, cmd, expected_output):
 
         if not expected_pattern.search(res):
             logging(f"Failed to find {expected_output} in {cmd} result: {res}")
-            time.sleep(retry_count)
+            _, retry_interval = get_retry_count_and_interval()
+            time.sleep(retry_interval)
```
yangchiu, line 74: This breaks the convention of using `retry_count` to control the debugging wait time.

e2e/libs/snapshot/rest.py
```diff
@@ -24,7 +24,14 @@ def create(self, volume_name, snapshot_id, waiting):
 
         snapshot_created = False
         for i in range(self.retry_count):
-            snapshots = self.list(volume_name)
+            try:
+                snapshots = volume.snapshotList().data
+            except Exception as e:
+                if i < 10:
```
yangchiu, line 30: Why do we need to raise an exception here after 10 retries? What error would we encounter if we kept the original logic instead?

e2e/libs/volume/base.py
```diff
@@ -88,6 +92,81 @@ def get_last_data_checksum(self, volume_name):
...
+    def get_expanded_data_size_mb(self, volume_name):
+        v = get_annotation_value(
+            group="longhorn.io",
+            version="v1beta2",
+            namespace=constant.LONGHORN_NAMESPACE,
+            plural="volumes",
+            name=volume_name,
+            annotation_key=self.ANNOT_EXPANDED_DATA_SIZE_MB,
+        )
+        return int(v) if v else 200
```
yangchiu, line 168: Hard-coding the value `200` in multiple places could lead to maintenance issues. Consider defining it as a constant and using it as the default value for the corresponding function parameter.

e2e/keywords/volume.resource
```diff
@@ -174,7 +180,7 @@ Delete volume ${volume_id}
 
 Write data to volume ${volume_id}
     ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
-    ${checksum} =    write_volume_random_data   ${volume_name}    2048
+    ${checksum} =    write_volume_random_data   ${volume_name}    200
```
yangchiu, line 183: Hard-coding the value `200` in multiple places could lead to maintenance issues. Consider defining it as a constant and using it as the default value for the corresponding function parameter.

e2e/libs/keywords/volume_keywords.py
```diff
@@ -156,6 +160,14 @@ def write_volume_random_data(self, volume_name, size_in_mb, data_id=None):
         logging(f'Writing {size_in_mb} MB random data to volume {volume_name}')
         return self.volume.write_random_data(volume_name, size_in_mb, data_id)
 
+    def write_volume_data_at_offset(self, volume_name, offset_mb, size_mb=200):
```
yangchiu, line 163: Hard-coding the value `200` in multiple places could lead to maintenance issues. Consider defining it as a constant and using it as the default value for the corresponding function parameter.

yangchiu (changes requested):

yangchiu:
If there aren't updates to the PR in the near term, I might merge it to keep things moving forward and address the review comments in follow-up improvements.

e2e/libs/volume/crd.py
```diff
@@ -606,21 +609,75 @@ def write_random_data(self, volume_name, size, data_id):
 
         endpoint = self.get_endpoint(volume_name)
 
+        # Limit the checksum to the written region (not the full device) so
```
yangchiu: I suspect the checksum change is causing some test cases to fail:

The following test cases fail on this PR due to checksum mismatches:

```
Crash Single Instance Manager While RWO Encrypted Volume Backup Is Restoring
Crash Single Instance Manager While RWX Encrypted Volume Backup Is Restoring
Test Volume Attached Node Is Down During Encrypted Volume Restoration
Test Corrupting Source Replica While Backup Creation
Test Restoring Volume From Backup With Missing Blocks
```

https://ci.longhorn.io/job/longhorn-e2e-test-pull-request-check/378/

All of them pass on master-head:

https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8915/

yangchiu (changes requested):

e2e/libs/volume/crd.py
shuo-wu: It's probably due to the above annotation usage. I am verifying it

 Write data to volume 0  (annotation set so the test will operatespecific data size) → restore into volume 2 (no annotation)

yangchiu: Storing test information in CR annotations is error-prone. We should consider removing these kinds of usage altogether eventually. cc @longhorn/qa

yangchiu: @shuo-wu You could consider store the necessary information in a robot variable, similar to what's done: https://github.com/longhorn/longhorn-tests/blob/master/e2e/keywords/variables.resource#L17.

yangchiu: > Storing test information in CR annotations is error-prone. We should consider removing these kinds of usage altogether eventually. cc @longhorn/qa

https://github.com/longhorn/longhorn/issues/13795 is created for this.

yangchiu: > It's probably due to the above annotation usage. I am verifying it
> 
>  Write data to volume 0  (annotation set so the test will operatespecific data size) → restore into volume 2 (no annotation)

The test cases failed due to checksum mismatches rather than missing annotations, so the checksum calculation logic may also be worth investigating.

shuo-wu: The root cause analysis:
1. `write_random_data` was changed to use  `dd count=N | md5sum`  plus a `write-size-mb` annotation, and made `get_checksum` fit it. A restored volume never goes through `write_random_data`, so it had no annotation then calculated the whole-device md5 and compared it against a partial-region checksum. This leads to mismatching.
2. Shared keyword silently downsized. The code had changed `Write data to volume ${volume_id}`  from `2048 MB` → `200 MB`, affecting 84 call sites across the whole suite. 


Now, we have reverted both changes and ask the linked-clone tests to use the explicit new keyword `Write ${WRITE_SIZE_MB} MB data to volume X at offset 0`  instead. Besides, there is no annotation used in the linked-clone tests now.

yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3624, opened 2026-08-04
test(robot): refine test case Test Volume Expansion After Iscsid Restart

yangchiu:

Issue N/A

refine test case Test Volume Expansion After Iscsid Restart

1. aws auto-assign public ips can change after an instance reboots, which currently breaks ssh access. further improvements are needed.
2. as a temporary workaround, separate the iscsid restart test case from other test cases that may reboot a node
3. since the ssh failure is caused by the ip change, retrying is ineffective. remove the retry mechanism accordingly


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3632, opened 2026-08-06
test(robot): refine test cases Test Best/Least Effort Auto Balance In Unstable Cluster

yangchiu:

Issue N/A

refine test cases Test Best/Least Effort Auto Balance In Unstable Cluster

1. kubelet restart has been reimplelented using ssh, but aws auto-assign public ips can change after an instance reboots, which currently breaks ssh access. further improvements are needed.
2. as a temporary workaround, separate the kubelet restart related test cases from regression test cases that may reboot a node
3. improve error handling


chriscchien:
> 2. as a temporary workaround, separate the kubelet restart related test cases from regression test cases that may reboot a node

Do we still need to keep this workaround because if https://github.com/longhorn/longhorn-tests/pull/3636 is ready, the IP change issue should be fixed.

yangchiu:
> Do we still need to keep this workaround because if #3636 is ready, the IP change issue should be fixed.

Yes, that's the preferred categorization.

chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3633, opened 2026-08-06
test(robot): refine test case SnapshotBack Proxy Request Should Be Sent To Correct Instance-Manager Pod teardown

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13657

refine test case SnapshotBack Proxy Request Should Be Sent To Correct Instance-Manager Pod teardown


chriscchien (approved):

yangchiu:
@mergify backport v1.12.x

PR #3635, opened 2026-08-06
test(pytest): fix flaky case `test_volume_iscsi_basic_with_backing_image` by enhancing `wait_for_device_login`

roger-ryao:
https://github.com/longhorn/longhorn/issues/13662#issuecomment-5199977624

- Increased the retry limit for checking the device login directory
- Added an error print statement to capture `os.listdir` exceptions rather than failing silently.


yangchiu (approved):

yangchiu:
@mergify backport v1.12.x

PR #3639, opened 2026-08-12
test(robot): add replica rebuild TCP keepalive test

dnesting:

Issue longhorn/longhorn#13703 https://github.com/longhorn/longhorn/issues/13767

Reproduce a V1 replica rebuild interruption by rate-limiting the rebuild data transfer and withholding replies on the idle FileSend connection to ensure TCP keepalives get dropped. Verify that the original control connection remains established and that the rebuild completes with intact data.

I relied heavily on Codex to understand your testing framework, and while I was able to get the test running in a local Docker-based cluster that I think approximated your test environment, I'm not entirely confident it will run correctly in the real one.  Even so, the test appears to fail currently and passes with the fix at https://github.com/longhorn/longhorn-engine/pull/1970.

### DIFF
--- a/e2e/keywords/network.resource
+++ b/e2e/keywords/network.resource
@@ -83,3 +83,31 @@ Disconnect network port ${port_number} of ${engine_type} instance manager networ
     ${node_name} =    get_node_by_index    ${node_id}
     ${pod_name} =    get_instance_manager_pod_on_node    ${node_name}    ${engine_type}
     disconnect_pod_network    ${pod_name}    ${duration}    ${port_number}
+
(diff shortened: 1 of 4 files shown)


yangchiu:
@dnesting Could you please include `Signed-off-by: Author Name <authoremail@example.com>` in the commit message to pass the DCO check? Thank you!

yangchiu:
@chriscchien Since you've verified the original ticket https://github.com/longhorn/longhorn/issues/13703, you can move this test PR forward.

yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

PR #3641, opened 2026-08-18
test(robot): add v2 live upgrade new instance manager e2e coverage

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9104


e2e/keywords/host.resource
```diff
@@ -93,6 +83,14 @@ Power off node ${node_id}
     Append to list      ${powered_off_nodes}     ${powered_off_node}
     power_off_node_by_name    ${powered_off_node}
 
+Power off node named ${node_name}
```
yangchiu: This can be replaced with the existing keyword `Power off node ${node_id}`:
```
When Power off node ${source_engine_node}
```

bin/latest_image
```diff
@@ -0,0 +1,2 @@
+longhornio/longhorn-manager-test:179a2a7dd-dirty
```
yangchiu: Accidentally submitted unnecessary files.

e2e/keywords/host.resource
```diff
@@ -93,6 +83,14 @@ Power off node ${node_id}
...
+    Append to list    ${powered_off_nodes}    ${node_name}
+    power_off_node_by_name    ${node_name}
+
+Power off node named ${node_name} without waiting
```
yangchiu: This can be achieved by improving the existing keyword with `&{config}`:
```
Power off node ${node_id}
    [Arguments]    &{config}
    ${powered_off_node} =     get_node_by_index    ${node_id}
    Append to list      ${powered_off_nodes}     ${powered_off_node}
    power_off_node_by_name    ${powered_off_node}    &{config}
```

e2e/keywords/longhorn.resource
```diff
@@ -27,6 +27,12 @@ ${LONGHORN_REPO_DIR}    /tmp/longhorn/longhorn
 Wait for longhorn ready
     wait_for_all_instance_manager_running
 
+Wait for all instance managers running with default image
```
yangchiu: Can we simplify this by extending the existing keyword `Upgrade v1 volumes engine to ${custom_engine_image}` to v2 data engine?

e2e/keywords/longhorn.resource
```diff
@@ -27,6 +27,12 @@ ${LONGHORN_REPO_DIR}    /tmp/longhorn/longhorn
...
+    wait_for_all_instance_manager_running_with_default_image
+
+Wait for all instance manager upgrades completed
```
yangchiu: Can we simplify this by extending the existing keyword `Upgrade v1 volumes engine to ${custom_engine_image}` to v2 data engine?

e2e/keywords/longhorn.resource
```diff
@@ -163,6 +166,19 @@ Upgrade Longhorn to custom version
         Fail    Upgrading Longhorn failed
     END
 
+Start Longhorn upgrade to custom version
```
yangchiu: This can be achieved by the existing keyword:
```
Upgrade Longhorn to custom version    wait=${False}
```

e2e/keywords/longhorn.resource
```diff
@@ -163,6 +166,19 @@ Upgrade Longhorn to custom version
...
+    [Arguments]    &{config}
+    ${upgrade_process} =    upgrade_longhorn    wait=False    &{config}
+    Set Test Variable    ${upgrade_process}
+
+Wait for Longhorn custom upgrade to complete
```
yangchiu: `Wait for Longhorn upgrade to complete` might be a better name ,since this keyword may not be limited to upgrade to Longhorn `custom` version.

e2e/keywords/longhorn.resource
```diff
@@ -163,6 +166,19 @@ Upgrade Longhorn to custom version
...
+    [Arguments]    ${timeout}=600
+    ${upgraded} =    wait_for_longhorn_upgrade_process    ${upgrade_process}    timeout=${timeout}
```
yangchiu: It would be better to use the existing variable `RETRY_COUNT` is the timeout.

e2e/keywords/volume.resource
```diff
@@ -332,6 +332,11 @@ Wait for volume ${volume_id} migration to be ready
     ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
     wait_for_volume_migration_to_be_ready    ${volume_name}
 
+Get volume ${volume_id} current engine node
```
yangchiu: Is the `current engine node` the same as `the volume attached node`? If so, the existing function `get_volume_node` can be reused.

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
```diff
@@ -0,0 +1,190 @@
...
+    ...    Test steps
+    ...    1. Uninstall the existing Longhorn system and install the stable version for upgrade testing.
+    ...    2. Enable the v2 data engine, add block disks, and wait for Longhorn to become ready.
+    ...    3. Create 3 attached v2 volumes, wait for each volume to become healthy, and write data to each volume.
+    ...    4. Upgrade Longhorn to the custom version under test and wait for Longhorn to become ready again.
+    ...    5. Verify all instance managers are running with the default target image.
+    ...    6. Verify all upgraded attached v2 volumes remain healthy and their data stays intact.
+    Given Prepare v2 upgrade environment
+
+    FOR    ${i}    IN RANGE    3
```
yangchiu: If all 3 volumes have the same settings, why do we need to upgrade 3 identical volumes? Should we try volumes with different settings instead? For example, RWX, single replica, migratable, sharding etc.

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
```diff
@@ -0,0 +1,190 @@
...
+        And Create volume vol${i} with    dataEngine=v2
+        And Attach volume vol${i}
+        And Wait for volume vol${i} healthy
+        And Write data to volume vol${i}
+    END
+
+    When Upgrade Longhorn to custom version
+    Then Wait for longhorn ready
+    And Setting allow-v2-instance-manager-automatic-upgrade is set to true
+    And Wait for all instance managers running with default image
```
yangchiu: Do `Wait for all instance managers running with default image` and `Wait for all instance manager upgrades completed` validate both `v1` and `v2` instance managers?

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
```diff
@@ -0,0 +1,190 @@
...
+    When Create storageclass longhorn-test-rwx with    dataEngine=v2    numberOfReplicas=3
+    And Create persistentvolumeclaim 0    volume_type=RWX    sc_name=longhorn-test-rwx
+    And Create deployment 0 with persistentvolumeclaim 0
+    And Wait for volume of persistentvolumeclaim 0 healthy
+    And Write 100 MB data to file data.txt in deployment 0
+
+    When Upgrade Longhorn to custom version
+    Then Wait for longhorn ready
+    And Setting allow-v2-instance-manager-automatic-upgrade is set to true
+    And Setting replica-disk-soft-anti-affinity is set to true
```
yangchiu: Could we know the purpose of setting these affinities?

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
```diff
@@ -0,0 +1,190 @@
...
+
+Test System V2 Upgrade With Volume Node Down Before Upgrade Eventually Recovers
+    [Documentation]    Verify that the v2 system upgrade can still converge when the
+    ...    attached volume node is already down before the upgrade starts, as long as
+    ...    the node eventually comes back and the volume can recover to healthy.
+    ...    Test steps
+    ...    1. Uninstall the existing Longhorn system and install the stable version for upgrade testing.
+    ...    2. Enable the v2 data engine, add block disks, and wait for Longhorn to become ready.
+    ...    3. Create 1 attached v2 volume with 3 replicas, wait for it to become healthy, and write data to it.
+    ...    4. Power off the current engine node of the attached volume and wait for the Longhorn node to be reported down.
```
yangchiu: Since the requirement is power off the node _before_ the upgrade, do we still need the `upgrade_longhorn    wait=False` operation here? Will the down node block the upgrade?

e2e/keywords/longhorn.resource
yangchiu, line 31: Does this mean `v1` and `v2` volumes are upgraded using different mechanisms? For v1 volumes, we upgrade the engine image of an individual v1 volume, and it doesn't affect other v1 volumes. For v2 volumes, we directly upgrade the instance manager, which automatically affects all v2 volumes. Am I understanding this correctly?

yangchiu (changes requested):

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
davidcheng0922: You are right. For this test, I want to verify the case where volumes are distributed across all worker nodes. Previously, I did not attach each volume to a specific node. Each volume should be attached to a different node.

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
```diff
@@ -0,0 +1,190 @@
...
+    And Wait for longhorn ready
+    And Set default backupstore
+
+Prepare attached v2 volume ${volume_id} with ${replica_count} replicas
+    Create volume ${volume_id} with    dataEngine=v2    numberOfReplicas=${replica_count}
+    Attach volume ${volume_id} to node 0
+    Wait for volume ${volume_id} healthy
+    Write data to volume ${volume_id}
+
+Wait for instance manager upgrade relocation node for volume ${volume_id} on node ${node_name}
```
yangchiu: I couldn't to find where `Wait for instance manager upgrade relocation node for volume ${volume_id} on node ${node_name}` is used. Could you help point me to it?

Does that mean a v2 volume will be reattached to a different node during the upgrade?

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
```diff
@@ -0,0 +1,190 @@
...
+    ${relocation_node}=    wait_for_instance_manager_upgrade_relocation_node_for_volume_on_node
+    ...    ${volume_id}    ${node_name}
+    [Return]    ${relocation_node}
+
+Wait for Longhorn node named ${node_name} down
```
yangchiu: It'd be better to move this keyword to *.resource layer.

e2e/libs/keywords/longhorn_deploy_keywords.py
```diff
@@ -18,5 +25,53 @@ def install_longhorn_system(self, custom_cmd="", install_stable_version=False, l
...
     def enable_storage_network_setting(self):
         self.longhorn.enable_storage_network_setting()
+
+    def wait_for_instance_manager_upgrade_relocation_node_for_volume_on_node(self, volume_id, node_name, timeout=600, poll_interval=5):
+        volume_name = generate_name_with_suffix("volume", volume_id)
+        namespace = get_longhorn_namespace()
+        deadline = time.time() + int(timeout)
+        last_state = ""
+
+        while time.time() < deadline:
```
yangchiu: It can use the same wait mechanism as `wait_for_all_instance_manager_upgrades_completed` for consistency.

yangchiu (changes requested):

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
davidcheng0922: ~Yes. The Longhorn upgrade is considered complete only after the volume becomes healthy.~

~If we wait for the upgrade synchronously while the volume node is still down, the upgrade may be~ ~blocked because the volume can remain `degraded` until the node comes back. Therefore, this test~ ~starts the upgrade with wait=False, powers the node back on, and then waits for the upgrade process~ ~to complete.~

The test is no more needed, it is belonged to offline upgrade

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
davidcheng0922: No more needed, remove it. Thank you

e2e/keywords/longhorn.resource
davidcheng0922: Yes, that is correct. The v1 live upgrade is performed volume by volume, so upgrading one v1 volume does not affect other v1 volumes.

For v2, due to the current engine/instance-manager architecture limitation, the upgrade is performed node by node by upgrading the v2 instance manager. As a result, all v2 volumes whose engines are managed by the instance manager on that node can be affected during the upgrade.

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
davidcheng0922: During the v2 instance manager upgrade, the volume engine can be temporarily moved to another node, but this should not change the original volume attachment from the workload's perspective.

Before upgrade: Volume A, Engine A -> Node A
During upgrade: Volume A -> Node A, Engine A -> Node B
After upgrade: Volume A, Engine A -> Node A

e2e/tests/negative/upgrade_with_v2_new_instance_manager.robot
davidcheng0922: Thank you

e2e/libs/keywords/longhorn_deploy_keywords.py
davidcheng0922: Thank you

bin/latest_image
```diff
@@ -0,0 +1,2 @@
+longhornio/longhorn-manager-test:f14509db2-dirty
```
yangchiu: Is this file required to be included in the PR?

yangchiu (changes requested):

yangchiu:
@davidcheng0922 Please mark the PR as ready for review when it's ready. Thank you!

bin/latest_image
davidcheng0922: Thank you, I remove the files in the /bin

davidcheng0922:
> @davidcheng0922 Please mark the PR as ready for review when it's ready. Thank you!

I've shared this test case for your reference. Going forward, there are a few things that could be improved, for example, the loop count.

Thank you

yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

PR #3653, opened 2026-08-19
test(robot): add test case Test NVMe TCP IO Queue Count

chriscchien:

Issue longhorn/longhorn#13783

This PR has some minor differences from the original [manual test plan](https://github.com/longhorn/longhorn/issues/13706#issuecomment-5300291168):

- Step 1 – Default behavior: Skip comparing the default nvme io queue count with the number of CPU cores. because the kernel default may vary by environment. Instead, record the initial default queue count and verify in end of test case.
- Step 8 – Resource effect: Skip the memory usage comparison because this step is optional and the result may vary depending on the test environment.

N/A

### DIFF
--- a/e2e/keywords/volume.resource
+++ b/e2e/keywords/volume.resource
@@ -110,6 +110,35 @@ Update volume ${volume_id} ${key} to ${value}
     ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
     update_volume_spec   ${volume_name}    ${key}    ${value}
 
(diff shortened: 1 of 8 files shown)


derekbit:
@chriscchien Could you resolve the conflict? Thanks.

yangchiu (dismissed):

yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

PR #3658, opened 2026-08-24
test: add volume expansion test cases

derekbit:

Issue longhorn/longhorn#13386

### DIFF
--- a/e2e/keywords/snapshot.resource
+++ b/e2e/keywords/snapshot.resource
@@ -152,3 +152,12 @@ Longhorn snapshot associated with csi volume snapshot ${csi_volume_snapshot_id} 
     ${workload_name} =   generate_name_with_suffix    ${workload_kind}    ${workload_id}
     ${volume_name} =    get_workload_volume_name    ${workload_name}
     wait_for_snapshot_to_be_created    ${volume_name}    ${longhorn_snapshot_name}
+
+Volume ${volume_id} should have expansion snapshot containing volume name
+    ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
+    assert_expansion_snapshot_name_contains_volume_name    ${volume_name}
+
+Volume ${volume_id_1} and volume ${volume_id_2} should have distinct expansion snapshots
+    ${volume_name_1} =    generate_name_with_suffix    volume    ${volume_id_1}
+    ${volume_name_2} =    generate_name_with_suffix    volume    ${volume_id_2}
+    assert_volumes_have_distinct_expansion_snapshots    ${volume_name_1}    ${volume_name_2}
--- a/e2e/libs/keywords/snapshot_keywords.py
+++ b/e2e/libs/keywords/snapshot_keywords.py
@@ -138,3 +138,25 @@ class snapshot_keywords:
         assert snapshot is not None, f"Snapshot {snapshot_id} not found in volume {volume_name}"
         logging(f"Got snapshot name {snapshot.name} for snapshot {snapshot_id} of volume {volume_name}")
         return snapshot.name
+
+    def assert_expansion_snapshot_name_contains_volume_name(self, volume_name):
+        snapshot_name = self.snapshot.get_expansion_snapshot_name(volume_name)
+        assert volume_name in snapshot_name, \
+            f"Expect expansion snapshot {snapshot_name} to contain volume " \
+            f"name {volume_name}"
+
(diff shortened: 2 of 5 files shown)


e2e/libs/keywords/volume_keywords.py
```diff
@@ -72,6 +73,13 @@ def wait_for_volume_expand_to_size(self, volume_name, size):
         logging(f'Waiting for volume {volume_name} to expand to {size}')
         return self.volume.wait_for_volume_expand_to_size(volume_name, size)
 
+    def expand_volume_to_size(self, volume_name, size):
```
yangchiu: Duplicates the existing `expand_volume` function.

yangchiu: We should not implement logic in keyword layer.

e2e/keywords/volume.resource
```diff
@@ -110,6 +110,10 @@ Update volume ${volume_id} ${key} to ${value}
     ${volume_name} =    generate_name_with_suffix    volume    ${volume_id}
     update_volume_spec   ${volume_name}    ${key}    ${value}
 
+Expand volume ${volume_id} to size ${size}
```
yangchiu: Duplicates the existing keyword `Expand volume ${volume_id} to ${size}`.

yangchiu (changes requested):

e2e/keywords/snapshot.resource
```diff
@@ -152,3 +152,12 @@ Longhorn snapshot associated with csi volume snapshot ${csi_volume_snapshot_id}
     ${workload_name} =   generate_name_with_suffix    ${workload_kind}    ${workload_id}
     ${volume_name} =    get_workload_volume_name    ${workload_name}
     wait_for_snapshot_to_be_created    ${volume_name}    ${longhorn_snapshot_name}
+
+Volume ${volume_id} should have expansion snapshot containing volume name
```
yangchiu, line 156: Both `Volume ${volume_id} should have expansion snapshot containing volume name` and `Volume ${volume_id_1} and volume ${volume_id_2} should have distinct expansion snapshots` don't exactly test the changed introduced by https://github.com/longhorn/longhorn/issues/13386.

After https://github.com/longhorn/longhorn/issues/13386, an expansion snapshot follows the naming `expand-<size>-<volume_name>`, we could directly validate this explicitly in the test steps like:
```
Run command and wait for output
...    kubectl get snapshot ...
...    expand-<size>-<volume_name>
```

By validating `expand-<size>-<volume_name>`, the test case would not only cover `Test Volume Expansion Snapshot Name Contains Volume Name`, but also make `Test Volume Expansion Snapshot Name Not Colliding Across Volumes` redundant.

Moreover, I think the test case being implemented in this PR is more suitable as a unit test rather than an e2e test?

e2e/libs/keywords/snapshot_keywords.py
```diff
@@ -138,3 +138,25 @@ def get_snapshot_name(self, volume_name, snapshot_id):
...
         logging(f"Got snapshot name {snapshot.name} for snapshot {snapshot_id} of volume {volume_name}")
         return snapshot.name
+
+    def assert_expansion_snapshot_name_contains_volume_name(self, volume_name):
+        snapshot_name = self.snapshot.get_expansion_snapshot_name(volume_name)
+        assert volume_name in snapshot_name, \
+            f"Expect expansion snapshot {snapshot_name} to contain volume " \
+            f"name {volume_name}"
+
+    def assert_volumes_have_distinct_expansion_snapshots(self, volume_name_1,
```
yangchiu, line 148: If an expansion snapshot follows the naming `expand-<size>-<volume_name>`, 2 volumes cannot have non-distinct expansion snapshot names.

Makefile
```diff
@@ -1,6 +1,6 @@
 MACHINE := longhorn
 
-.PHONY: validate ci
+.PHONY: validate ci package
```
yangchiu, line 3: Does this PR accidentally include the changes from https://github.com/longhorn/longhorn-tests/pull/3657/changes?

yangchiu (changes requested):

Makefile
derekbit: yes

derekbit:
The test cases might be not required.
https://github.com/longhorn/longhorn-tests/pull/3658#discussion_r3867370767
I'll leave it to QA to implement the test case instead.

PR #3663, opened 2026-08-25
test(negative): reproduce stuck attach from stale stopped v1 engine record

derekbit:

Issue Longhorn/longhorn#13687 https://github.com/longhorn/longhorn/issues/13688

Add an e2e case for longhorn/longhorn#13687. It injects a leaked `stopped` v1 engine process record into the target node's instance manager (via `longhorn-instance-manager process create` with a binary that exits 0, using `--port-count 1` so the record settles in `stopped` rather than `running`), then attaches the volume to that node and asserts it becomes healthy.

On v1.12.0/v1.12.1 the volume stays wedged in `attaching` and the case fails; with the longhorn-manager fix that reaps the stale record, the attach proceeds and the case passes.

Adds injection/query helpers to the v1 instance manager lib, exposes them through instancemanager_keywords, and wires up the engine.resource keywords.

### DIFF
--- a/e2e/keywords/engine.resource
+++ b/e2e/keywords/engine.resource
@@ -3,6 +3,7 @@ Documentation       Longhorn engine related keywords
 
 Library             ../libs/keywords/common_keywords.py
 Library             ../libs/keywords/engine_keywords.py
+Library             ../libs/keywords/instancemanager_keywords.py
 
 *** Keywords ***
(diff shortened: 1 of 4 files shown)


yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

PR #3677, opened 2026-08-28
test(robot): add test case Test Create Default Disk On Labeled Nodes With V2 Data Engine

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13880

add test case Test Create Default Disk On Labeled Nodes With V2 Data Engine


derekbit (approved):

derekbit:
@mergify backport v1.13.x

PR #3678, opened 2026-08-30
test(v2): add e2e cases for v2 block disk device recovery

derekbit:

Issue Longhorn/longhorn#13893

Add negative cases for a disk whose device is left detached from the kernel, for the device release on disk removal, and for the release after an instance manager restart, plus a regression case for the remove and re-add cycle. Add node helpers to detect a PCI BDF path, query the device driver, bind and unbind it, and wait for a disk to become schedulable. A device is treated as detached when it is bound to a userspace PCI driver or to no driver at all, since an interrupted bind can leave either state. The cases skip on a non-BDF disk path, where the scenario cannot be reproduced.


yangchiu (dismissed):

yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

PR #3679, opened 2026-08-30
chore(robot): add tags to e2e test cases and document them

derekbit:

Issue #

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -145,6 +145,145 @@ pip install -r requirements.txt
 
 Once the test completed, the test result can be found at /tmp/test-report folder.
 
+### Test tags
+
+Every test suite declares `Test Tags` in its `*** Settings ***` section, and individual test cases may add
+more specific tags with `[Tags]`. Tags are how the CI pipelines and `./run.sh -i <tag>` / `-e <tag>` select
+which test cases to run.
+
+Conventions:
+
+- Use lower-case and hyphens for new tags (`backing-image`, `node-reboot`), never underscores or spaces.
+  A few existing tags (`engine_image`, `system_backup`, `support bundle`) do not follow this, but they are
+  kept as they are so the CI job configurations referring to them keep working.
+- Every suite must carry at least one *suite* tag (`regression` or `negative`) plus one *feature area* tag.
+- Put `[Tags]` **before** `[Documentation]`, otherwise the `...` continuation lines of the documentation are
+  parsed as tags by Robot Framework.
+- Only add an *issue* tag when the test exists to cover one specific upstream issue.
+
+#### Suite tags
+
+| Tag | Description |
+| --- | --- |
+| `regression` | Regression test cases under `tests/regression/`. |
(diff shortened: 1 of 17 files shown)


e2e/README.md
```diff
@@ -145,6 +145,145 @@ pip install -r requirements.txt
...
+### Test tags
+
+Every test suite declares `Test Tags` in its `*** Settings ***` section, and individual test cases may add
+more specific tags with `[Tags]`. Tags are how the CI pipelines and `./run.sh -i <tag>` / `-e <tag>` select
+which test cases to run.
+
+Conventions:
+
+- Use lower-case and hyphens for new tags (`backing-image`, `node-reboot`), never underscores or spaces.
+  A few existing tags (`engine_image`, `system_backup`, `support bundle`) do not follow this, but they are
```
yangchiu, line 157: There is no CI job will be affected by these tags. They are free to be renamed to align wit h the convention.

e2e/README.md
```diff
@@ -145,6 +145,145 @@ pip install -r requirements.txt
...
+| `node-delete` | Node deletion. |
+| `node-drain` | Node drain. |
+| `service-restart` | Longhorn / Kubernetes service restart. |
+| `kubelet-restart` | Kubelet restart or stop. |
+| `network` / `network-disconnect` | Network disruption. |
+| `reattach` | Volume reattachment. |
+
+#### Issue tags
+
+Tags named `longhorn-<issue number>` (`longhorn-8355`, `longhorn-9865`, `longhorn-10210`) mark test cases
```
yangchiu, line 284: These kinds of tags are not useful. We should remove them and directly reference the issue in the documentation section instead.

cc @longhorn/qa

e2e/README.md
```diff
@@ -145,6 +145,145 @@ pip install -r requirements.txt
...
+- Put `[Tags]` **before** `[Documentation]`, otherwise the `...` continuation lines of the documentation are
+  parsed as tags by Robot Framework.
+- Only add an *issue* tag when the test exists to cover one specific upstream issue.
+
+#### Suite tags
+
+| Tag | Description |
+| --- | --- |
+| `regression` | Regression test cases under `tests/regression/`. |
+| `negative` | Negative / resilience test cases under `tests/negative/`. |
```
yangchiu, line 169: Just for the information, directly specifying `-i negative` to run all `negative` test cases is not practical because it could take several days. Consider using more specific tags like `manual`, `node-reboot`, `network-disconnect`, etc. to run the negative tests parallel, similar to what we've done for release testing.

cc @longhorn/qa

yangchiu (changes requested):

e2e/README.md
yangchiu: We should also include this in the README to suggest that users run the tests by specifying tags to limit the test scope, rather than running all test cases in a single test run, which could take days to complete.

yangchiu (changes requested):

PR #3681, opened 2026-08-31
test(robot): fix failed test cases due to inconsistent longhorn default path /var/lib/longhorn

yangchiu:

Issue N/A

fix failed test cases due to inconsistent longhorn default path /var/lib/longhorn

related failed test cases:

```
Test Volume Expansion When Node Disk Is Full
Test Scheduling Replicas To Different Disks On The Same Node
Test No Transient Error In Engine Status During Eviction
Test Replica Auto Balance Disk In Pressure With Stopped Volume Should Not Block
Test backing image handle node disk deleting events
Test Replica Auto Balance Disk In Pressure
```

### DIFF
--- a/e2e/libs/node/node.py
+++ b/e2e/libs/node/node.py
@@ -8,7 +8,6 @@ from robot.libraries.BuiltIn import BuiltIn
 from utility.constant import DISK_BEING_SYNCING
 from utility.constant import NODE_UPDATE_RETRY_INTERVAL
 from utility.constant import DEFAULT_BLOCK_DISK_NAME
-from utility.constant import DEFAULT_FILESYSTEM_DISK_NAME_PREFIX
 import utility.constant as constant
 from utility.utility import get_longhorn_client
 from utility.utility import get_retry_count_and_interval
@@ -18,7 +17,7 @@ from node_exec import NodeExec
 
 class Node:
 
-    DEFAULT_DISK_PATH = "/var/lib/longhorn"
+    DEFAULT_DISK_PATH = "/var/lib/longhorn/"
     DEFAULT_VOLUME_PATH = "/dev/longhorn/"
 
     all_nodes = None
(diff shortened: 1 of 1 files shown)


chriscchien (approved):

yangchiu:
@mergify backport v1.13.x

mantissahz:
Hi @yangchiu,
I'm not sure this ticket https://github.com/longhorn/longhorn/issues/13910 is still valid. If modifying the configuration to fix test cases is OK, we can also modify the regression configuration to fix the test case.

yangchiu:
> Hi @yangchiu, I'm not sure this ticket [longhorn/longhorn#13910](https://github.com/longhorn/longhorn/issues/13910) is still valid. If modifying the configuration to fix test cases is OK, we can also modify the regression configuration to fix the test case.

@mantissahz @derekbit I'm not sure either. The default data path on the Longhorn manager side has been switching between `/var/lib/longhorn` and `/var/lib/longhorn/` recently. It's quite confusing, and I can't tell which one we actually want.

derekbit:
> > Hi @yangchiu, I'm not sure this ticket [longhorn/longhorn#13910](https://github.com/longhorn/longhorn/issues/13910) is still valid. If modifying the configuration to fix test cases is OK, we can also modify the regression configuration to fix the test case.
> 
> @mantissahz @derekbit I'm not sure either. The default data path on the Longhorn manager side has been switching between `/var/lib/longhorn` and `/var/lib/longhorn/` recently. It's quite confusing, and I can't tell which one we actually want.

No. `/var/lib/longhorn/` exists for very long while. @mantissahz Please clarify the root cause of the failure.

mantissahz:
>No. /var/lib/longhorn/ exists for very long while. @mantissahz Please clarify the root cause of the failure.

The ticket longhorn/longhorn#13910 is a pytest test case, not an e2e test case.
Therefore, it is not included in this fix.
The root cause of longhorn/longhorn#13910 is still the default data path `/var/lib/longhorn/`, and the [pytest deploy manifest](https://github.com/longhorn/longhorn-tests/blob/3b01e5d4d499297b983ac8a5f16a113f45ae63b7/manager/integration/deploy/test.yaml#L60-L61) and [default common setting](https://github.com/longhorn/longhorn-tests/blob/3b01e5d4d499297b983ac8a5f16a113f45ae63b7/manager/integration/tests/common.py#L150) are still using `/var/lib/longhorn`.

yangchiu:
Just make sure https://github.com/longhorn/longhorn/issues/9083 won't change `/var/lib/longhorn/` to `/var/lib/longhorn` again in the future. Otherwise, pytest will fail again.

PR #3690, opened 2026-09-02
test(negative): cover longhorn-global-manager component

hookak:

Issue longhorn/longhorn#13059

Registers the `longhorn-global-manager` Deployment introduced by longhorn/longhorn#13100 in the component inventories (`longhorn_workloads`, resource-usage metrics, pytest `check_longhorn`) and includes it in the `Test Longhorn Components Recovery` deletion list.

Adds a `Test Longhorn Global Manager Leader Failover` case: delete the lease-holding pod, verify a standby takes over the `longhorn-global-manager` lease, then verify the controllers hosted by the new leader — the Kubernetes PV controller by a RWX volume created after the failover becoming ready, and the Kubernetes pod controller by workload remount recovery after instance-manager deletion.

The added cases need the `longhorn-global-manager` Deployment introduced by longhorn/longhorn#13100, so please merge this after that chart PR. Until then they fail on a master-manifest installation.

Verified on a 4-node k3s v1.35 cluster with the PR builds: the leader-failover case, `Test Longhorn Components Recovery` (including the added deletion), and resource-usage recording all pass.

### DIFF
--- a/e2e/tests/negative/component_resilience.robot
+++ b/e2e/tests/negative/component_resilience.robot
@@ -80,6 +80,7 @@ Test Longhorn Components Recovery
     And Delete Longhorn component instance-manager pod on node 1
     And Delete Longhorn Deployment longhorn-ui pod
     And Delete Longhorn Deployment longhorn-driver-deployer pod
+    And Delete Longhorn Deployment longhorn-global-manager pod
 
     Then Wait for Longhorn components all running
     And Wait for volume 0 healthy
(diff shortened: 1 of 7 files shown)


derekbit:
Hello @longhorn/qa Let's prioritize reviewing the PR first.

manager/integration/tests/common.py
```diff
@@ -4313,6 +4313,7 @@ def check_longhorn(core_api):
     ready = False
     has_engine_image = False
     has_driver_deployer = False
+    has_global_manager = False
```
yangchiu: This breaks the test_upgrade test case, which upgrade Longhorn from a previous version to the latest version.

`check_longhorn()` won't return because  `longhorn-global-manager` doesn't exist in the previous version of Longhorn.

e2e/tests/negative/component_resilience.robot
```diff
@@ -195,3 +196,37 @@ Test Longhorn Dynamic Provisioned RWO Volume Recovery
     When Delete replica of deployment 0 volume on replica node
     And Wait until volume of deployment 0 replica rebuilding started on replica node
     Then Delete ${DATA_ENGINE} instance manager of deployment 0 volume and wait for recover
+
+Test Longhorn Global Manager Leader Failover
```
yangchiu, line 200: This test case is flaky:

https://ci.longhorn.io/job/longhorn-e2e-test-pull-request-check/576/

```

00:00:00.000KEYWORD Collections . Then List Should Contain Value ${global_manager_pods}, ${new_leader_pod}
Documentation:	
Fails if the value is not found from list.

Start / End / Elapsed:	20260908 11:51:17.060 / 20260908 11:51:17.060 / 00:00:00.000
11:51:17.060	FAIL	[ longhorn-global-manager-76c9868988-7gkh8 | longhorn-global-manager-76c9868988-848wn | longhorn-global-manager-76c9868988-gdfq6 ] does not contain value 'longhorn-global-manager-76c9868988-wtx52'.
```

yangchiu (changes requested):

e2e/tests/negative/component_resilience.robot
derekbit: Hello @hookak 
Could you help check it? Thanks.

manager/integration/tests/common.py
hookak: Fixed in https://github.com/longhorn/longhorn-tests/pull/3690/commits/f52ef667a: the check now probes the Deployment first and requires a running longhorn-global-manager pod only when the Deployment exists.

e2e/tests/negative/component_resilience.robot
hookak: Fixed in https://github.com/longhorn/longhorn-tests/pull/3690/commits/07e043d3f: the test now waits for the Deployment to be fully running again after the failover and checks that the new holder is one of the live longhorn-global-manager pods.

hookak:
@yangchiu Thanks for the review! I've rebased on master and addressed both comments. Could you re-run the test when you have a moment?

yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

PR #3698, opened 2026-09-03
test(pytest): fix flaky test_setting_concurrent_rebuild_limit on v2

derekbit:

Issue Longhorn/longhorn#13189

The test case crashes volume 1's in-flight rebuild and then asserts that the volumes rebuild one at a time. On v2 the crash and the assertions race.

crash_replica_processes() is not equivalent between the data engines:
- For v1 it kills the replica process, so the failure is instantaneous and wait_for_replica_failed() blocks until it is registered.
- For v2 it only runs `go-spdk-helper expose stop`, which removes the NVMf subsystem while the lvol and spdk_tgt keep running. Nothing dies, so the engine only notices through the NVMe-oF replicaFastIOFailTimeoutSec/ replicaCtrlrLossTimeoutSec timeouts. The replica also never satisfies wait_for_replica_failed(): its instance stays registered in the instance manager and it is replaced by a new replica rather than being marked failed, which is why the v2 path passed wait_to_fail=False and did not synchronise on the crash at all. The test therefore entered the verification loop while volume 1's rebuild was still reported as in_progress, and later observed the ~4s window where it reports "error" before volume 2 picks up the rebuild slot. The loop asserted rebuildStatus[0].state == "in_progress" whenever the other volume was idle, so it failed on that transient state.

Add wait_for_replica_crashed() as the v2 counterpart of wait_for_replica_failed(), waiting for the engine to register the crash (replica gone, failedAt set, not running, or mode ERR), and use it in crash_replica_processes() so wait_to_fail works for both data engines. The v1 path is unchanged. Drop the v2 special case in the test so the crash is synchronised the same way it is on v1.

### DIFF
--- a/manager/integration/tests/common.py
+++ b/manager/integration/tests/common.py
@@ -2595,8 +2595,11 @@ def crash_replica_processes(client, api, volname, replicas=None,
             exec_instance_manager(api, r.instanceManagerName, kill_command)
 
         if wait_to_fail is True:
+            wait_crashed = wait_for_replica_failed
+            if DATA_ENGINE == "v2":
+                wait_crashed = wait_for_replica_crashed
             thread = create_assert_error_check_thread(
-                wait_for_replica_failed,
+                wait_crashed,
                 client, volname, r['name'], RETRY_COUNTS, RETRY_INTERVAL_SHORT
             )
             threads.append(thread)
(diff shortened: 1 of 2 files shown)


derekbit:
Testing
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12040/ Passed
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12041/ Passed

derekbit:
cc @chriscchien for review

chriscchien (approved):
LGTM https://github.com/longhorn/longhorn/issues/13189

derekbit:
@mergify backport v1.13.x

PR #3709, opened 2026-09-05
test(robot): add test case Test V2 Continuous Write

roger-ryao:
https://github.com/longhorn/longhorn/issues/13937
https://github.com/longhorn/longhorn/issues/13940

- Add regression test `Test V2 Continuous Write` to verify that continuous write operations to V2 volumes do not get stuck and data remains intact (#13937).
- Add `Delete statefulset` keyword in `statefulset.resource` to support workload cleanup in test cases.

### DIFF
--- a/e2e/keywords/common.resource
+++ b/e2e/keywords/common.resource
@@ -112,68 +112,21 @@ Set disk path based on host provider and architecture
     END
 
 Enable v2 data engine and add block disks
-    [Arguments]    ${wait}=${True}    ${interrupt_mode_enabled}=${False}
+    [Arguments]    ${wait}=${True}
     set_disk_path_based_on_host_provider_and_architecture
+    ${RUN_V2_INTERRUPT_MODE}=    Get Environment Variable    RUN_V2_INTERRUPT_MODE    false
+    IF    "${RUN_V2_INTERRUPT_MODE}" == "true"
+        update_setting    data-engine-interrupt-mode-enabled    {"v2":"true"}
+    END
     ${V2_DATA_ENGINE_ENABLED}=    get_setting    v2-data-engine
     # if it's already configured, no need to do it again
     Return From Keyword If    "${V2_DATA_ENGINE_ENABLED}" == "true"
     update_setting    data-engine-cpu-mask    {"v2": "0x1"}
-    ${v2_data_engine_enabled}=    get_setting    v2-data-engine
-
-    IF    ${interrupt_mode_enabled}
-        Enable interrupt mode for v2 data engine    ${v2_data_engine_enabled}
-    END
-
-    # v2 data engine and its block disks are already configured, nothing to do
-    IF    "${v2_data_engine_enabled}" == "true"
-        RETURN
-    END
-
     update_setting    v2-data-engine    true
     ${worker_nodes}=    get_worker_nodes
     FOR    ${worker_node}    IN    @{worker_nodes}
         add_disk    ${DEFAULT_BLOCK_DISK_NAME}    ${worker_node}    block    ${DISK_PATH}    wait=${wait}
     END
-
-    Wait for block disks schedulable on nodes    @{worker_nodes}
-
-Enable interrupt mode for v2 data engine
-    [Arguments]    ${v2_data_engine_enabled}
-    ${is_interrupt_mode_enabled} =    Data engine interrupt mode is enabled for v2
-    IF    ${is_interrupt_mode_enabled}
-        RETURN
-    END
-
-    Setting data-engine-interrupt-mode-enabled is set to {"v2": "true"}
-
-    # only if v2 data engine was already enabled do v2 instance manager pods
-    # (and their block disks) already exist for this update to disrupt; a
-    # fresh environment gets everything created with interrupt mode already
-    # applied below, so there is nothing to wait for here
-    IF    "${v2_data_engine_enabled}" == "true"
-        ${worker_nodes} =    get_worker_nodes
-        Check v2 instance manager pods recreated
-        Wait for block disks unschedulable then schedulable on nodes    @{worker_nodes}
-    END
-
-Wait for block disks ${state:schedulable|unschedulable} on nodes
-    [Arguments]    @{node_names}
-    # add_disk only waits for the disk to report Ready; the Schedulable
-    # condition can lag behind that (or briefly flips during IM pod
-    # recreation), so confirm the expected state separately per node
-    ${expected_schedulable} =    Set Variable If    '${state}' == 'schedulable'    ${True}    ${False}
-    FOR    ${node_name}    IN    @{node_names}
-        wait_for_node_disks_schedulable    ${node_name}    disk_type=block    expected_schedulable=${expected_schedulable}
-    END
-
-Wait for block disks unschedulable then schedulable on nodes
-    [Arguments]    @{node_names}
-    # recreating the v2 instance manager pods (e.g. after toggling
-    # data-engine-interrupt-mode-enabled) briefly makes already-added block
-    # disks unschedulable while the new pod re-attaches them; confirm that
-    # full transition instead of assuming the disks are still ready
-    Wait for block disks unschedulable on nodes    @{node_names}
-    Wait for block disks schedulable on nodes    @{node_names}
 
 Init worker node variables
     ${worker_nodes}=    get_worker_nodes
--- a/e2e/keywords/common.resource
+++ b/e2e/keywords/common.resource
@@ -112,17 +112,68 @@ Set disk path based on host provider and architecture
     END
 
 Enable v2 data engine and add block disks
-    [Arguments]    ${wait}=${True}
+    [Arguments]    ${wait}=${True}    ${interrupt_mode_enabled}=${False}
     set_disk_path_based_on_host_provider_and_architecture
     ${V2_DATA_ENGINE_ENABLED}=    get_setting    v2-data-engine
     # if it's already configured, no need to do it again
     Return From Keyword If    "${V2_DATA_ENGINE_ENABLED}" == "true"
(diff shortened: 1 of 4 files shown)


derekbit:
@roger-ryao Could you resolve the conflicts? Thanks.

e2e/tests/regression/test_v2.robot
```diff
@@ -643,72 +645,22 @@ Test CPU Manager Policy And Data Engine Number Of CPU Cores
...
-        Run command
-        ...    kubectl label node ${node_name} node.longhorn.io/create-default-disk=config --overwrite
-        Run command
-        ...    kubectl annotate node ${node_name} node.longhorn.io/default-disks-config='[{"name":"block-disk","path":"${DISK_PATH}","diskType":"block","allowScheduling":true}]' --overwrite
+Test V2 Continuous Write In Interrupt Mode
+    [Tags]    regression    v2    interrupt-mode
+    [Documentation]    Continuously write data to v2 volume in interrupt mode and 
+    ...                verify write does not get stuck and data is intact.
+    ...                - Issue: https://github.com/longhorn/longhorn/issues/13937
+    [Setup]    Set up v2 test environment    interrupt_mode_enabled=${True}
```
yangchiu: This PR only adds a single test case for v2 interrupt mode, which doesn't quite match the PR title `support v2 interrupt mode setup and add regression test`. The title suggests that an extra `interrupt mode` parameter is added to the pipeline and sets up the interrupt mode environment throughout the entire test.

e2e/tests/regression/test_v2.robot
```diff
@@ -643,72 +645,22 @@ Test CPU Manager Policy And Data Engine Number Of CPU Cores
...
-        Run command and wait for output
-        ...    kubectl get nodes.longhorn.io ${node_name} -n ${LONGHORN_NAMESPACE} -o yaml
-        ...    block-disk:
-        Run command and wait for output
-        ...    kubectl get nodes.longhorn.io ${node_name} -n ${LONGHORN_NAMESPACE} -o yaml
-        ...    path: ${DISK_PATH}
-        Run command and wait for output
-        ...    kubectl get nodes.longhorn.io ${node_name} -n ${LONGHORN_NAMESPACE} -o yaml
-        ...    diskType: block
+    FOR    ${i}    IN RANGE    ${LOOP_COUNT}
```
yangchiu: Currently we don't actively set `LOOP_COUNT` during release testing. Suggesting replace it with a fixed value that can reproduce the original issue first.

e2e/tests/regression/test_v2.robot
```diff
@@ -643,72 +645,22 @@ Test CPU Manager Policy And Data Engine Number Of CPU Cores
...
+        When Create statefulset ${i} using RWO volume with longhorn-test-v2-interrupt storageclass and size 3 Gi
+        And Wait for volume of statefulset ${i} healthy
+        Then Write 2560 MB data to file data in statefulset ${i}
+        And Check statefulset ${i} data in file data is intact
+        And Delete statefulset ${i}
```
yangchiu, line 1068: Will deleting the statefulset also delete the corresponding PV/PVC/Volume?

e2e/keywords/common.resource
```diff
@@ -91,37 +91,82 @@ Get test start time
...
+        wait_for_node_disks_schedulable    ${node_name}    disk_type=block    expected_schedulable=${expected_schedulable}
+    END
+
+Wait for block disks unschedulable then schedulable on nodes
+    [Arguments]    @{node_names}
+    # recreating the v2 instance manager pods (e.g. after toggling
+    # data-engine-interrupt-mode-enabled) briefly makes already-added block
+    # disks unschedulable while the new pod re-attaches them; confirm that
+    # full transition instead of assuming the disks are still ready
+    Wait for block disks unschedulable on nodes    @{node_names}
```
yangchiu: I'm not sure whether we need this fine-grained check. What if the 3rd disk transitions from `unschedulable` to `schedulable` while you're checking the 1st and 2nd disks?

e2e/keywords/common.resource
```diff
@@ -91,37 +91,82 @@ Get test start time
...
     FOR    ${worker_node}    IN    @{worker_nodes}
         add_disk    ${DEFAULT_BLOCK_DISK_NAME}    ${worker_node}    block    ${DISK_PATH}
     END
 
+    Wait for block disks schedulable on nodes    @{worker_nodes}
+
+Enable interrupt mode for v2 data engine
+    [Arguments]    ${v2_data_engine_enabled}
+    ${is_interrupt_mode_enabled} =    Data engine interrupt mode is enabled for v2
+    IF    ${is_interrupt_mode_enabled}
```
yangchiu: It seems like this won't happen because `data-engine-interrupt-mode-enabled` will be reset in the test case's teardown function?

e2e/keywords/common.resource
```diff
@@ -91,37 +91,82 @@ Get test start time
...
+
+    Setting data-engine-interrupt-mode-enabled is set to {"v2": "true"}
+
+    # only if v2 data engine was already enabled do v2 instance manager pods
+    # (and their block disks) already exist for this update to disrupt; a
+    # fresh environment gets everything created with interrupt mode already
+    # applied below, so there is nothing to wait for here
+    IF    "${v2_data_engine_enabled}" == "true"
+        ${worker_nodes} =    get_worker_nodes
+        Check v2 instance manager pods recreated
```
yangchiu: It would be clearer to make this logic explicitly test steps if what we are checking is the behavior during the transition from `v2 normal mode` to `v2 interrupt mode`.

e2e/keywords/setting.resource
```diff
@@ -26,3 +26,17 @@ Update Setting To Valid Value And Verify
 Setting ${setting_key} should be ${setting_value}
     ${current_value} =    get_setting    ${setting_key}
     Should Be Equal    ${current_value}    ${setting_value}
+
+Data engine interrupt mode is enabled for ${data_engine}
+    [Documentation]    data-engine-interrupt-mode-enabled may be stored either as a legacy
```
yangchiu: Does that mean the value of `data-engine-interrupt-mode-enabled` is sometime `true` and sometime `{"v2": "true"}`? Do we know how this happens? Should it be considered a bug?

yangchiu (changes requested):

e2e/tests/regression/test_v2.robot
roger-ryao: yes, it will also delete  the corresponding PV/PVC/Volume.

e2e/tests/regression/test_v2.robot
roger-ryao: I will update  the PR title later.

e2e/keywords/setting.resource
roger-ryao: ~~As shown in the screenshot, based on my experience, I expected `data-engine-interrupt` to be specific to v2 only, so a simple `true`/`false` value should be sufficient.~~

~~However, the actual value of `data-engine-interrupt-mode-enabled` is `{"v2": "true"}`.~~

~~To avoid potential issues if this changes in the future, I made this modification accordingly.~~

~~If you think this change is not necessary, please let me know and I'll revert this part.~~

Screenshot_20260907_101450
Screenshot_20260907_101419

roger-ryao: According to https://longhorn.io/docs/1.12.1/references/settings/#value-format-types-by-supported-data-engines, I will update later.

yangchiu (changes requested):
@chriscchien @roger-ryao Please work together to come up with a desirable approach for supporting v2 interrupt mode in robot test. https://github.com/longhorn/longhorn-tests/pull/3712/ and https://github.com/longhorn/longhorn-tests/pull/3709/ seem to be duplicating each other. Thanks!

roger-ryao:
> @chriscchien @roger-ryao Please work together to come up with a desirable approach for supporting v2 interrupt mode in robot test. #3712 and #3709 seem to be duplicating each other. Thanks!

Hi @yangchiu 
       We could merge  https://github.com/longhorn/longhorn-tests/pull/3712 first, as it already covers the v2 interrupt mode environment setup. 
       I will then update this PR to remove the duplicated setup logic and rename the regression test case from `Test V2 Continuous Write In Interrupt Mode` to `Test V2 Continuous Write`  to also cover `Interrupt Mode` & `polling mode`, and fix the conflicts.

roger-ryao:
/jenkins test -t "Test V2 Continuous Write" -v RETRY_COUNT:300 -v DATA_ENGINE:v2

derekbit (commented):
This is not enough because it only writes data using 1 single thread. 
Please refer to https://github.com/longhorn/longhorn-tests/pull/3772 and refine the all continuous io test cases.

derekbit:
In addtion, the test case should be in continuous_io.robot instead.

roger-ryao:
> In addtion, the test case should be in continuous_io.robot instead.

The original goal of this PR was make sure basic DD write operations do not get stuck on V2 volumes (issue #[13937](https://github.com/longhorn/longhorn/issues/13937)).
So I believe this is a basic regression test, not a negative test. 
My PR description may have caused the confusion. I need to think of a better name for this PR

yangchiu:
> > In addtion, the test case should be in continuous_io.robot instead.
> 
> The original goal of this PR was make sure basic DD write operations do not get stuck on V2 volumes (issue #[13937](https://github.com/longhorn/longhorn/issues/13937)). So I believe this is a basic regression test, not a negative test. My PR description may have caused the confusion. I need to think of a better name for this PR

@roger-ryao If it's just a basic regression test case, refine the existing `Test V2 Volume Basic` might be sufficient.

yangchiu (changes requested):
https://github.com/longhorn/longhorn-tests/pull/3709#issuecomment-5787311539

PR #3733, opened 2026-09-10
fix(pytest): test_engine_crash_for_dr_volume

c3y1huang:

Issue longhorn/longhorn#13869

test_engine_crash_for_dr_volume asserts the DR volume is healthy immediately after the activated volume's pod starts. The engine crash earlier in the test can fail a replica, and its rebuild may still be in progress, so the volume can still be degraded at that moment.

The test previously passed because an engine-level restore error faulted the whole volume and auto-salvage reused all replicas in place, so no rebuild overlapped activation. With v2 volume per-replica restore failure introduced in longhorn/longhorn#13869, only the affected replica fails, and a real rebuild is required, exposing the missing wait.

`None`


derekbit (approved):

derekbit:
@mergify backport v1.13.x

PR #3737, opened 2026-09-11
fix(deploy): add NetworkPolicy for longhorn-manager to accept longhorn-test pod req

shuo-wu:

This updated [longhorn-manager Network Policy](https://github.com/longhorn/longhorn/blob/026b8a2c6863b50e750b7393fd9befc0cd9a58ad/deploy/longhorn.yaml#L106-L139) will block the test pod process without this extra rule

### DIFF
--- a/manager/integration/deploy/test.yaml
+++ b/manager/integration/deploy/test.yaml
@@ -17,6 +17,26 @@ subjects:
   name: longhorn-test-service-account
   namespace: default
 ---
+# The longhorn-manager policy only admits a few in-namespace components by default,
+# so this test pod - which lives in the default namespace - cannot reach the REST API on 9500.
+#
+# Allowing all ingress is deliberate rather than lazy. This relaxes a hardening measure,
+# which is acceptable only because it is scoped to a test manifest that already grants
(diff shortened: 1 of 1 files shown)


yangchiu (changes requested):
If we use the script to create the network policies for longhorn tests on demand: https://github.com/longhorn/longhorn-tests/blob/master/e2e/README.md?plain=1#L39, do we still need to modify `test.yaml`?

shuo-wu:
@yangchiu Last time when I tried to run **the python test**, all tests were blocked by the network policy.

If it's no longer needed, feel free to close this PR.

yangchiu:
Currently, we still have some older Longhorn versions in use that don't enable network policies by default, and longhorn-tests may run against any of these versions. That's why we still need to apply the network policies on demand for longhorn-tests rather than hard-coding them in `test.yaml`.

Currently, we handle this using the script mentioned in the `README.md`. Once all Longhorn versions without network policies enabled reach EOL, we can consider hard-coding the network policies in `test.yaml`. The PR can be reopened at that time.

PR #3740, opened 2026-09-14
test(robot): add test case Test Metrics Scrape Sources NetworkPolicy

yangchiu:

Issue https://github.com/longhorn/longhorn/issues/13996

add test case Test Metrics Scrape Sources NetworkPolicy


roger-ryao (approved):
LGTM

PR #3746, opened 2026-09-16
test(robot): improve CSI clone node and disk selector test with dual disks

roger-ryao:
longhorn/longhorn#12792

Enhance regression coverage for longhorn/longhorn#12792 by validating diskSelector against multiple disks on the target node.

Previously, each node had only a single disk available, meaning replicas had no choice but to be placed on that disk regardless of diskSelector.

Key improvements:
- Create an additional disk on node 1 so multiple disks are available.
- Tag default and extra disks separately (hosting-fast vs hosting-slow).
- Assert source and clone replicas strictly land on the targeted disk.
- Verify source PVC attachment is not placed on untagged nodes during clone.
- Add test coverage for best-effort dataLocality and graceful scheduling failure.
- Add keywords for default disk retrieval, per-disk tagging, clone PVC without waiting for bound, and PVC disk-level replica assertions.
- Unify set_disk_tags and set_node_disks_tags implementation in node library.
- Fix wait_for_disk_replica_count and wait_for_replica_count to reliably support count=0 checks.


roger-ryao:
> This pull request is now in conflict. Could you fix it @roger-ryao? 🙏

Fixed

PR #3749, opened 2026-09-17
fix: improve wait_for_pods_volume_delete

derekbit:

Issue Longhorn/longhorn#14028

The Longhorn API can temporarily return a 404 while listing volumes during wait_for_pods_volume_delete(), because the volume CR may already be gone while the test helper is still polling for all pod-backed volumes to disappear.

Handle this race by retrying the specific "failed to list volume: volume.longhorn.io" 404 instead of failing immediately. Also precompute the pod volume names and return as soon as none of them remain, which keeps the polling logic focused on the volumes the helper is waiting for.


chriscchien (approved):

derekbit:
@mergify backport v1.13.x

PR #3754, opened 2026-09-17
ci: improve HAL sle-micro cluster configuration

roger-ryao:

https://github.com/longhorn/longhorn/issues/13393

- Add vm_affinity targeting hal-node-13 and hal-node-9 to avoid scheduling on problematic HAL nodes
- Increase VM memory size from 8GB to 16GB
- Add dev-longhorn and qa-longhorn cluster role template bindings
- Prefix cluster name with sle-micro for clearer identification in Rancher


yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

PR #3755, opened 2026-09-17
test(robot): fix nested loop issue around update_disks when disk fails to be scheduled

yangchiu:

Issue N/A

fix nested loop issue around update_disks when disk fails to be scheduled


chriscchien (dismissed):

chriscchien (approved):

yangchiu:
@mergify backport v1.13.x

PR #3758, opened 2026-09-18
fix(rbac): remove CSI controller Secret access

COLDTURNIP:

Issue longhorn/longhorn#14020

As we removed the Secret reading permission from the CSI sidecars, we don't expect `csi.storage.k8s.io/provisioner-secret-*` parameters to be set during volume creation.


roger-ryao:
/jenkins test -i "regressionANDencrypted" -v RETRY_COUNT:300 -v DATA_ENGINE:v1

yangchiu (approved):

yangchiu:
@mergify backport v1.13.x

